PR #2, opened 2016-05-21
Initial Cut of the Build

brthor:
Lots of stuff coming in, much is duplicated from the cli repo, with changes mentioned below.

We've copied a number of things here but the long term plan is to reduce the duplication in the following ways:
  - Publish the Debian Package Tool as a CLI Extension Tool from here, Consume this in the CLI
  - Microsoft.Dotnet.Build.Framework will go away when we make the switch to MSBuild, so it didn't make sense to invest in reducing the duplication

A couple changes included as well
 - Create a Standalone host build with packaging/publishing of the shared framework and shared Host
 - Download install scripts from cli repo
 - Create SharedFramework deb package with dependency on fixed version of the host deb @piotrpMSFT 
 - Publish a SharedFramework Version Script

A couple more things left before merge:
 - Finish configuring VSO signed builds
 - Remove stub package builder, it's not needed any longer **[EDIT: It's needed until we lock the host]**

A few Questions:
 - Can we assume the LICENSE of this repo is the same as cli? @blackdwarf 

PTAL @piotrpMSFT @eerhardt @livarcocc @weshaggard


build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -46,6 +46,125 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+                    // Copy the shared framework + host Archives
+                    CopyBlobs($"{Channel}/Binaries/{SharedFrameworkNugetVersion}/", targetContainer);
+
+                    // Copy the shared framework installers
+                    CopyBlobs($"{Channel}/Installers/{SharedFrameworkNugetVersion}/", $"{Channel}/Installers/Latest/");
+
+                    // Copy the shared host installers
+                    CopyBlobs($"{Channel}/Installers/{SharedHostNugetVersion}/", $"{Channel}/Installers/Latest/");
+
+                    // Generate the CLI and SDK Version text files
```
eerhardt: CLI?

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -46,6 +46,125 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+                    List<string> versionFiles = new List<string>() { "win.x86.version", "win.x64.version", "ubuntu.x64.version", "rhel.x64.version", "osx.x64.version", "debian.x64.version", "centos.x64.version" };
+                    string cliVersion = Utils.GetCliVersionFileContent(c);
+                    string sfxVersion = Utils.GetSharedFrameworkVersionFileContent(c);
+                    foreach (string version in versionFiles)
+                    {
+                        AzurePublisherTool.PublishStringToBlob($"{Channel}/dnvm/latest.{version}", cliVersion);
```
eerhardt: CLI? - probably shouldn't do anything with CLI in this whole repo.

dotnet-install.ps1
```diff
@@ -0,0 +1,346 @@
+#
```
eerhardt: Probably don't need this file, right?  since it is getting downloaded above.

build_projects/dotnet-host-build/build.sh
```diff
@@ -85,7 +85,9 @@ done < "$REPOROOT/branchinfo.txt"
 [ -z "$DOTNET_INSTALL_DIR" ] && export DOTNET_INSTALL_DIR=$REPOROOT/.dotnet_stage0/$(uname)
 [ -d "$DOTNET_INSTALL_DIR" ] || mkdir -p $DOTNET_INSTALL_DIR
 
-$REPOROOT/scripts/obtain/dotnet-install.sh --channel $CHANNEL --verbose
+
+DOTNET_INSTALL_SCRIPT_URL="https://raw.githubusercontent.com/dotnet/cli/b939c88e0cf7c2e204f91a5161acdd2f5a732bf9/scripts/obtain/dotnet-install.sh"
```
eerhardt: Maybe instead of a hard-coded commit, should we use the latest install script in `rel/1.0.0`?

dotnet-install.ps1
brthor: deleted this in a later commit, will squash

brthor: ah, never mind, I accidentally checked in one of the ones that got downloaded, will delete

build_projects/dotnet-host-build/PublishTargets.cs
brthor: When I thought I got them all 😄

build_projects/dotnet-host-build/PublishTargets.cs
brthor: what are these dnvm files anyways, do we still use these?

build_projects/dotnet-host-build/build.sh
brthor: My intention here was that changes to the install script in the cli repo wouldn't break this repo, if that ever happened. 

What benefit do you see from taking the latest over a hard coded version?

build_projects/dotnet-host-build/PublishTargets.cs
eerhardt: I think "dnvm" is a hold-over from DNX days.

Searching the repo, the "dnvm" folder only appears to be used just for these "latest.version" files. They are used by the `dotnet-install` scripts to determine the latest version for each platform (i.e. this is how we get Stage0).

Maybe when we refactor/restructure our published artifacts (https://github.com/dotnet/cli/issues/3120), "dnvm" should be removed completely.  @sokket

build_projects/dotnet-host-build/build.sh
eerhardt: But we are taking the actual latest CLI right?  It doesn't really make sense to use a hard-coded install script version to install the latest CLI.  I would think they should be either both hard-coded versions or both "latest".

Another thought:  the BuildTools stuff needs a CLI as well.  Currently in the CLI repo, we download 2 versions: the BuildTools version, and the "latest".  Maybe we should change this to only require 1 version of the CLI.  i.e. use the BuildTools version for both BuildTools and for running the build.

brthor:
Commits not showing up for now...

VSO sigining is configured. 
PR Feedback is addressed
Channel Switched
Setup a testing myget feed
Disabled deb publishing

All this is for a test setup until the CLI is ready and we bring over whatever latest changes we need.

build_projects/dotnet-host-build/build.sh
brthor: It's unlikely that we would make such a drastic install script change, and if we do I hope we maintain some level of backcompat so I'll go ahead and take latest. 

While I agree that there should be only one version of the CLI in use, I don't think it makes sense to figure out how to integrate these currently since pretty soon we'll only be using buildtools.

src/compilers/project.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": "1.0.0-*",
+  "buildOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0-rc3-*"
```
eerhardt: is this going to require "stub" packages?

brthor: Until we lock the host yes

brthor: Because we'll still be consuming what we just built

brthor:
@dotnet-bot test RHEL7.2 x64 Release Build please

PR #3, opened 2016-05-23
Corehost history

brthor:
Bring over Corehost directory history

cc @schellap


schellap:
Nice, good job! I can see the change and the diffs per commit.

PR #10, opened 2016-05-25
Add update-dependencies to core-setup.

eerhardt:
Also - add the ability to specify environment variables on the command line during our build.  This allows us to use VSO secret variables.

#### Note
This was copied from the CLI repo - however I made the following change to it:

1. In UpdateFilesTargets.cs, instead of reading the "Latest.txt" file and getting just the pre-release version, it now expects the "Latest_Packages.txt" file to be specified.  This has every package ID and full version in it, which allows us to update versions even if the major, minor, or patch has changed.  It also allows us to remove the Regexs we were using to tell if this was a "CoreFX" package. Now we only update packages that are specified in the "Latest_Packages.txt" file.

@brthor

### DIFF
--- a/build_projects/Microsoft.DotNet.Cli.Build.Framework/BuildSetup.cs
+++ b/build_projects/Microsoft.DotNet.Cli.Build.Framework/BuildSetup.cs
@@ -52,10 +52,17 @@ namespace Microsoft.DotNet.Cli.Build.Framework
 
         public int Run(string[] args)
         {
-            var targets = new[] { BuildContext.DefaultTarget };
-            if(args.Length > 0)
+            string[] targets;
+            string[] environmentVariables;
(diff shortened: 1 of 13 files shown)


build_projects/Microsoft.DotNet.Cli.Build.Framework/BuildSetup.cs
```diff
@@ -52,10 +52,17 @@ public BuildSetup UseTargetsFrom<T>()
 
         public int Run(string[] args)
         {
-            var targets = new[] { BuildContext.DefaultTarget };
-            if(args.Length > 0)
+            string[] targets;
```
brthor: :+1:

brthor:
LGTM thanks for adding hte Environment variables!

:shipit:

PR #13, opened 2016-05-25
stop nuking latest in core-setup

brthor:
Right now we're nuking all of the cli binaries in latest in every core-setup build.
We shouldn't need to delete these files, we can just overwrite.

cc @eerhardt @schellap @sokket

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -71,12 +71,6 @@ namespace Microsoft.DotNet.Host.Build
                 }
                 else
                 {
-                    // This is an old drop of latest so remove all old files to ensure a clean state
-                    AzurePublisherTool.ListBlobs($"{targetContainer}")
-                        .Select(s => s.Replace("/dotnet/", ""))
-                        .ToList()
-                        .ForEach(f => AzurePublisherTool.TryDeleteBlob(f));
-
                     // Drop the version file signaling such for any race-condition builds (see above comment).
                     AzurePublisherTool.DropLatestSpecifiedVersion(targetVersionFile);
                 }
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -71,6 +71,15 @@ namespace Microsoft.DotNet.Host.Build
                 }
                 else
                 {
+                    Regex versionFileRegex = new Regex(@"(?<version>\d\.\d\.\d)-(?<release>.*)?");
+
+                    // Delete old version files
+                    AzurePublisherTool.ListBlobs($"{targetContainer}")
+                        .Select(s => s.Replace("/dotnet/", ""))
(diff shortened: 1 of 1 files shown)


build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -71,12 +71,6 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
                 }
                 else
                 {
-                    // This is an old drop of latest so remove all old files to ensure a clean state
-                    AzurePublisherTool.ListBlobs($"{targetContainer}")
```
sokket: One thing I forgot about - we drop a version-specific file that was deleted here. These will add up unless we delete them here

brthor: Can we let the add up until your changes come in? We could search for files that match a regex but it's a little ugly.

eerhardt: Maybe skip "latest" here, but still delete everything that isn't "latest".

brthor: The issue is we're deleting all the binaries here as well, so we need to match against file patterns that match a version, I think I'll use a regex from updatedependencies.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -71,9 +71,12 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
                 }
                 else
                 {
-                    // This is an old drop of latest so remove all old files to ensure a clean state
+                    Regex versionFileRegex = new Regex(@"(?<version>\d\.\d\.\d)-(?<release>.*)?");
```
brthor: @eerhardt Took this from update dependencies, looks good?

eerhardt: what does the 'latest' version blob names look like?  they don't match this regex at all I assume...

eerhardt: We should really consider splitting the "channels" between CLI and core-setup. But maybe that is what @sokket is doing right now, right?

eerhardt:
:shipit:

build_projects/dotnet-host-build/PublishTargets.cs
brthor: Separate channels would eliminate problems like this, but create greater complexity for install scripts. since they'd need to reason about two channels and how they map to one another

eerhardt: We will need that eventually though.  The Shared Framework should be publishing to its "rtm" channel while CLI will be publishing to its "preview" channel.

eerhardt:
Is this still broken?  We didn't get a "FinalizeBuild" run for our last build.

brthor:
taking a look

brthor:
FinalizeBuild appears to be fine. Is there a specific artifact which you're not seeing?

PR #14, opened 2016-05-26
[Do not Merge]Add other meta packages

weshaggard:
Adds the following meta-packages:
Microsoft.NETCore
Microsoft.NETCore.Console
Microsoft.NETCore.UniversalWindowsPlatform

Also updates the tools used for the pkg projects.

@ericstj  @jhendrixMSFT @eerhardt


pkg/projects/Microsoft.NETCore.Console/Microsoft.NETCore.Console.pkgproj
```diff
@@ -0,0 +1,32 @@
...
+        <PackageVersion>$([System.Text.RegularExpressions.Regex]::Match('%(Identity)', '(\d+\.\d+.\d+[^&quot;]*)'))</PackageVersion>
+      </PackageMatch>
+
+      <Dependency Include="@(PackageMatch -> '%(PackageId)')" Condition="'%(PackageVersion)' != ''">
+        <Version>%(PackageVersion)</Version>
+      </Dependency>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <VersionSuffix>-rc3-$(CLIBuildVersion)</VersionSuffix>
```
eerhardt: Where does "CLIBuildVersion" come from?  The core-setup repo shouldn't have properties named "CLIBuildVersion".  This is repeated below for the other packages.

pkg/projects/Microsoft.NETCore.Console/project.json
```diff
@@ -0,0 +1,51 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore" :"5.0.1-rc3-24123-01",
```
eerhardt: These versions are now out of date, as I updated to 24126 this morning. If you want to update them you can run:

`powershell build_projects\update-dependencies\update-dependencies.ps1 -Targets UpdateFiles` from the root of the repo.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
```diff
@@ -0,0 +1,32 @@
...
+        <PackageVersion>$([System.Text.RegularExpressions.Regex]::Match('%(Identity)', '(\d+\.\d+.\d+[^&quot;]*)'))</PackageVersion>
+      </PackageMatch>
+
+      <Dependency Include="@(PackageMatch -> '%(PackageId)')" Condition="'%(PackageVersion)' != ''">
+        <Version>%(PackageVersion)</Version>
+      </Dependency>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <VersionSuffix>-rc3-$(CLIBuildVersion)</VersionSuffix>
```
eerhardt: Should this property be up in the general `PropertyGroup` next to <Version>?  It doesn't really make sense that it is inside this Target, since it has nothing to do with reading dependencies from project.json.

pkg/init-tools.cmd
```diff
@@ -5,13 +5,13 @@ set INIT_TOOLS_LOG=%~dp0init-tools.log
 if [%PACKAGES_DIR%]==[] set PACKAGES_DIR=%~dp0packages\
 if [%TOOLRUNTIME_DIR%]==[] set TOOLRUNTIME_DIR=%~dp0Tools
 set DOTNET_PATH=%TOOLRUNTIME_DIR%\dotnetcli\
-if [%DOTNET_CMD%]==[] set DOTNET_CMD=%DOTNET_PATH%bin\dotnet.exe
+if [%DOTNET_CMD%]==[] set DOTNET_CMD=%DOTNET_PATH%dotnet.exe
```
eerhardt: Does the init-tools.sh bash script need to be updated as well?

pkg/pack.cmd
```diff
@@ -55,7 +55,7 @@ call "%__ProjectDir%\init-tools.cmd"
 
 :: Restore dependencies mainly to obtain runtime.json
 pushd "%__ProjectDir%\deps"
-"%__ProjectDir%\Tools\dotnetcli\bin\dotnet.exe" restore --source "https://dotnet.myget.org/F/dotnet-core" --packages "%__ProjectDir%\packages"
+"%__ProjectDir%\Tools\dotnetcli\dotnet.exe" restore --source "https://dotnet.myget.org/F/dotnet-core" --packages "%__ProjectDir%\packages"
```
eerhardt: Need to update pack.sh?

pkg/init-tools.cmd
weshaggard: Yes I suspect that is why the non-windows CI builds failed.

pkg/pack.cmd
weshaggard: I will get it.

pkg/projects/Microsoft.NETCore.Console/Microsoft.NETCore.Console.pkgproj
weshaggard: This is being passed by pack.cmd/sh but I will update it for these and the other NETCore.App.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
weshaggard: Your right this is better to go at the top property group now. The reason it was here to begin with was because we were including some additional version numbers from the dependencies but that is not needed here nor is it interesting for these packages.

weshaggard:
Updated PR based on feedback. I still think there is further clean-up that can and should happen to the overall way we are passing properties from the CLI build scripts to the BuildTools scripts but we can do that at a later point.

pkg/projects/Microsoft.NETCore.Console/project.json
```diff
@@ -0,0 +1,51 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore": "5.0.1-rc3-24126-00",
+    "Microsoft.NETCore.ConsoleHost": "1.0.0-rc3-24126-00",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.2-rc3-24126-00",
```
gkhanna79: IIRC, the Microsoft.NetCore.App metapackage was referencing a "NetCore.Runtimes" metapackage that, in turn, pulled in CoreCLR. Here, we are pulling in CoreCLR package directly. Is this difference intentional or should we strive for consistency?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
```diff
@@ -0,0 +1,38 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore": "5.0.1-rc3-24126-00",
+    "Microsoft.NETCore.Portable.Compatibility": "1.0.1-rc3-24126-00",
+    "Microsoft.NETCore.Runtime": "1.0.2-rc3-24126-00",
```
gkhanna79: NETCore.Runtime pulls in CoreCLR package? If so, will this package come from closed build for for UWP since that is what should happen?

eerhardt:
> I still think there is further clean-up that can and should happen to the overall way we are passing properties from the CLI build scripts to the BuildTools scripts but we can do that at a later point.

Can you log an issue for this?  So we can track it.

pkg/projects/Microsoft.NETCore.Console/project.json
weshaggard: This is a different meta-package, primarily used for some ad-hoc testing scenarios but no shipping scenarios, for my initial port I just ported it as is. For this case though I do expect it to directly reference CoreCLR package.

pkg/projects/descriptions.json
```diff
@@ -20,8 +25,23 @@
...
+        "Description": "Provides a set of packages that can be used when building portable libraries on .NETCore based platforms.",
+        "CommonTypes": [ ]
+    },
+    {
         "Name": "Microsoft.NETCore.App",
         "Description": "A set of .NET API's that are included in the default .NET Core application model.",
         "CommonTypes": [ ]
     },
+    {
+        "Name": "Microsoft.NETCore.Console",
```
gkhanna79: Is there any relationship (from composition standpoint) between this and NETCore.App? Likewise for .NETCore/.NETCoreApp/.NETCore.Console?

pkg/projects/Microsoft.NETCore.Console/project.json
eerhardt: Do we still need this package if it doesn't ship?

pkg/init-tools.sh
```diff
@@ -47,15 +47,15 @@ if [ ! -e $__PROJECT_JSON_FILE ]; then
     echo "Running: $__scriptpath/init-tools.sh" > $__init_tools_log
     if [ ! -e $__DOTNET_PATH ]; then
         echo "Installing dotnet cli..."
-        __DOTNET_LOCATION="https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/${__DOTNET_TOOLS_VERSION}/${__DOTNET_PKG}.${__DOTNET_TOOLS_VERSION}.tar.gz"
+        __DOTNET_LOCATION="https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/${__DOTNET_TOOLS_VERSION}/${__DOTNET_PKG}.${__DOTNET_TOOLS_VERSION}.tar.gz"
```
eerhardt: Do you know why this doesn't use the `dotnet-install` scripts?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -40,10 +40,6 @@
         <Exclude>Compile</Exclude>
       </Dependency>
     </ItemGroup>
-
-    <PropertyGroup>
-    <VersionSuffix>-rc3-$(CLIBuildVersion)</VersionSuffix>
```
eerhardt: If this is gone now, how are we getting the commit count to put in our version number?

Note - that NETCore.App has to be higher than 004312, since that is the current version.

eerhardt:
:shipit:  with some questions for my knowledge.

pkg/projects/Microsoft.NETCore.Console/project.json
weshaggard: Honestly I don't know who all is using it right now so for now I'm blinding bringing to get us out of TFS but we can consider removing later.

weshaggard:
>Can you log an issue for this? So we can track it.

https://github.com/dotnet/core-setup/issues/20

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
weshaggard: This is not new the UWP  meta-package has always pulled in the CoreCLR package. It will pull whichever version of the package we are actually publishing, whether that comes from open or closed doesn't matter it just gets it from the feed.

pkg/projects/descriptions.json
weshaggard: Currently there isn't a direct release ship between these. .NETCore.Console was a precursor to .NETCore.App. Microsoft.NETCore package is used by PCL projects but might be superseded by NETStandard.Library. For now my main intent is to get these building somewhere in the open so we don't start publishing bogus versions of these when we switch to shipping from the open. I will not be surprised at all if we end up deleting some of these later.

pkg/init-tools.sh
weshaggard: This is copied from corefx and we haven't fully looked into using dotnet-install scripts yet, it is on my list to follow-up on though, as I really don't like having all this logic in these scripts.

pkg/projects/descriptions.json
ericstj: No, this was a meta-package I made like a year or more ago when @KrzysztofCwalina was first doing the console-app-model prototype.  It's sense been used in various demos / test scenarios.  Its purpose was to include every package that was supported x-plat.  I think we can phase it out once we clean up consumers: https://github.com/search?utf8=%E2%9C%93&q=%22Microsoft.NETCore.Console%22&type=Code&ref=searchresults

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
weshaggard: The version number gets passed via https://github.com/dotnet/core-setup/blob/master/pkg/pack.cmd#L68 as BuildNumberMajor and BuildNumberMinor. With those we end up building a version that looks like `-rc3-004315-00`.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
gkhanna79: However, the feed will soon have the CoreCLR package being published from open, which requires reverse forwarders to be inplace. Since that will not work on Win10, by simply moving this definition out, we are breaking that support.

Prior to the feed changes, the package came from closed build, which would have just worked due to api-set linkage. Should there be a plan for defining a RID for Win10 as we discussed in the past?

Also, given that CoreCLR gets installed via the MSI for UWP, what does it mean to pulling in this package?

pkg/projects/descriptions.json
eerhardt: Can we log an issue to remove this, if we feel it is the right thing to do?

ericstj:
LGTM
Here's a thought on how to change/improve the build process.  Baseline versions are defined in the Microsoft.Private.PackageBaseline.  We could reference the package IDs without version and let baselining lift everything up to the correct version during build.  That'd get rid of your project.json hack.  The PJ is needed for NETCore.App because we actually produce an MSI/zip that contains it.  For these meta-packages it isn't needed.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
eerhardt: :( so you mean NETCore.App will now get that `-00` in its version number too?

Let's avoid that because we are using commits.  There will never be a need to change `-00`.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
weshaggard: I think we definitely need to think about this some more to think through the consequences. We may need to have a win10 version of CoreCLR or something.

For UWP in particular as you hint at we don't actually pull CoreCLR from the packages as we use the AppX package that contains in which we ship with the CoreRuntimeSDK.

gkhanna79: > we use the AppX package that contains in which we ship with the CoreRuntimeSDK.

Given this, what does it even mean to have CoreCLR nuget package in this manifest package?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
weshaggard: Yes it means it will start getting a `-00` as well. The host packages were already getting that as well. Today we are using commits but it isn't clear to me yet if that will remain the way we do it. If we switch this to full BuildTools for example that might change. Even with commit versioning there may be cases were you need to spin multiple builds from the same commit.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
weshaggard: It is there for completeness because Microsoft.NETCore.Runtime pulls in both the CoreCLR and NETNative packages.

weshaggard:
Please do not merge as I need to figure out how transition these from the closed builds to the open builds. I may have to change the versioning scheme because most of these have already published with higher version numbers. See https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.UniversalWindowsPlatform `5.2.0-rc3-24126-00`

weshaggard:
I'm likely going to need to do the same versioning trick I did when we moved NETCore.App and change the versions to be "rc3-3<version>" in order to ensure these always have a higher version then what is already on our myget feeds.

weshaggard:
This PR is outdated and I'm going to rebuild it overtop of my recent PR https://github.com/dotnet/core-setup/pull/111. So closing this.

ericstj:
Where's your new PR?  #111 doesn't add these back and current builds of TFS stopped producing them.

weshaggard:
@ericstj https://github.com/dotnet/core-setup/pull/126

PR #19, opened 2016-05-26
Overwrite path_helper file on install.

vcsjones:
If the package's installer is run multiple times, the installation
path would be appended for every time the package had been run. This
change overwrites the file initially, and if subsequent paths need
to be added, then those can be appended. This ensures the "dotnet"
path helper file is clean for each install.

Fixes #18.

PR #21, opened 2016-05-26
Remove appending to dotnet paths

### LINKED ISSUE #18: Installing packages appends many times to /etc/paths.d/dotnet
## Steps to reproduce

Run the .pkg for the SDK installer multiple times. Can be the same version, can be different ones.

## Expected  behavior

The `/etc/paths.d/dotnet` file contains a single line to the path of where the CLI tools are run.

## Actual behavior

Every time the package installer is run, another path line is added:

```
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
..etc
```

I only discovered this because it looks like at some point, the path was once `/usr/local/share/dotnet/bin`, which was getting added to my path and causing problems.

## Environment data
`dotnet --info` output:

>.NET Command Line Tools (1.0.0-preview1-002700)
>
>Product Information:
> Version:     1.0.0-preview1-002700
> Commit Sha:  0336f6bb34

>Runtime Environment:
> OS Name:     Mac OS X
> OS Version:  10.11
> OS Platform: Darwin
> RID:         osx.10.11-x64

blackdwarf:
This fixes appending multiple times to the `/etc/paths.d/dotnet` file for the shared host. 

Fixes #18

/cc @brthor @eerhardt

### DIFF
--- a/packaging/osx/sharedhost/scripts/postinstall
+++ b/packaging/osx/sharedhost/scripts/postinstall
@@ -11,6 +11,6 @@ INSTALL_DESTINATION=$2
 chmod 755 $INSTALL_DESTINATION/dotnet
 
 # Add the installation directory to the system-wide paths
-echo $INSTALL_DESTINATION | tee -a /etc/paths.d/dotnet
+echo $INSTALL_DESTINATION | tee /etc/paths.d/dotnet
 
 exit 0


blackdwarf:
@brthor @eerhardt this looks good? Should I *gasp* merge/

brthor:
are you sure we should overwrite that file every time? 

It seems fine to me.

eerhardt:
:shipit: 

@vcsjones - Your code in #19 was different. Do you see advantages to doing it your way over this way?

vcsjones:
@eerhardt No, `tee` is already known to work and there were some suspected problems with using redirection as discussed in [in the comments of #18](https://github.com/dotnet/core-setup/issues/18#issuecomment-221922605).

PR #24, opened 2016-05-26
Update NETCore.App meta-package dependencies to .NETcoreApp1.0

weshaggard:
Fixes https://github.com/dotnet/cli/issues/2924

This limits the dependnecies of Microsoft.NETcore.App to only
be appicable to netcoreapp1.0 and so if people try to restore
it with other tfm's like netstandard they will not get any
of the dependencies and the meta-package will essentially be empty.

cc @ericstj @davidfowl

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -39,6 +39,11 @@
       <Dependency Condition="'%(Dependency.Identity)' == 'System.Runtime.Loader'">
         <Exclude>Compile</Exclude>
       </Dependency>
+
+      <File Include="$(PlaceholderFile)" >
+        <SkipPackageFileCheck>true</SkipPackageFileCheck>
+        <TargetPath>lib/netcoreapp1.0</TargetPath>
+      </File>
     </ItemGroup>
 
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


ericstj:
I'd actually expect it to not install since it won't have any applicable assets.  Can you confirm that's the case?

weshaggard:
I didn't test installing via VS. However I did have a reference to it with a tfm of netstandard1.0 and there were no errors but just an empty package in the lock file.

```
{
  "locked": false,
  "version": 2,
  "targets": {
    ".NETStandard,Version=v1.0": {
      "Microsoft.NETCore.App/1.0.0-rc3-004315": {
        "type": "package"
      }
    }
  },
  "libraries": {
    "Microsoft.NETCore.App/1.0.0-rc3-004315": {
      "sha512": "gC7ikPLzvhlnhuHyh5PFAKxwzOljvaGMrCdr7mIEbxbV72/ZFmGHWXh9BY4S23Z0015m2Xssghjcolulwsyd1Q==",
      "type": "package",
      "path": "Microsoft.NETCore.App/1.0.0-rc3-004315",
      "files": [
        "Microsoft.NETCore.App.1.0.0-rc3-004315.nupkg.sha512",
        "Microsoft.NETCore.App.nuspec",
        "ThirdPartyNotices.txt",
        "dotnet_library_license.txt"
      ]
    }
  },
  "projectFileDependencyGroups": {
    "": [
      "Microsoft.NETCore.App >= 1.0.0-rc3-004315"
    ],
    ".NETStandard,Version=v1.0": []
  },
  "tools": {},
  "projectFileToolGroups": {}
}
```

ericstj:
What happens if you put a placeholder in `lib\netcoreapp1.0\_._`

weshaggard:
I just pushed an update to add the placeholder with that in place. People will get the following restore error if they try to use this from anything other then netcoreapp1.0.

```
Errors in C:\temp\netcoreapp\project.json
    Package Microsoft.NETCore.App 1.0.0-rc3-004316 is not compatible with netstandard1.0 (.NETStandard,Version=v1.0). Package Microsoft.NETCore.App 1.0.0-rc3-004316 supports: netcoreapp1.0 (.NETCoreApp,Version=v1.0)
    One or more packages are incompatible with .NETStandard,Version=v1.0.
```

ericstj:
LGTM

PR #30, opened 2016-05-27
dotnet-deb-tool: ship the DebianPackageTool as a CLI tool

brthor:
This PR changes the DebianPackageTool to ship as a CLI tool so we can reduce code duplication.

It also makes the necessary build changes so it can be used inside the build, removing the old duplicated scripts we had. 

PTAL @eerhardt


build_projects/dotnet-host-build/DebTargets.cs
```diff
@@ -57,6 +73,17 @@ public static BuildTargetResult GenerateSharedFrameworkDeb(BuildTargetContext c)
             var inputRoot = c.BuildContext.Get<string>("SharedFrameworkPublishRoot");
             var debFile = c.BuildContext.Get<string>("SharedFrameworkInstallerFile");
             var objRoot = Path.Combine(Dirs.Output, "obj", "debian", "sharedframework");
+            var debianConfigFile = Path.Combine(Dirs.RepoRoot,
+                "packaging", "deb-package", "sharedframework", "dotnet-sharedframework-debian_config.json");
```
eerhardt: Can we make a Dirs property for this folder? `packaging\deb-package'  Or at least some sort of common place?

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -244,6 +273,11 @@ public static BuildTargetResult RestorePackages(BuildTargetContext c)
                 .Execute()
                 .EnsureSuccessful();
 
+            dotnet.Restore("--verbosity", "verbose", "--disable-parallel")
+                .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "tools", "dotnet-deb-tool"))
```
eerhardt: How about just restore from the "tools" directory, so we don't need 2 calls here?

Also, `RepoRoot` instead of `BuildDirectory`.  I think that reads cleaner.

tools/dotnet-deb-tool/project.json
```diff
@@ -0,0 +1,25 @@
...
+    }
+  },
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0-rc3-004308"
+    }
+  },
+
+  "packInclude": {
```
eerhardt: Does this give a warning?  I thought this was `packOptions` now.

build_projects/dotnet-host-build/PrepareTargets.cs
brthor: I don't think this will work since one of the calls utilizes a global.json inside the `independent` folder but I'll try it out. 

:+1: for `RepoRoot` Been wanting to get rid of `BuildDirectory` everywhere for some time

tools/dotnet-deb-tool/project.json
brthor: Yes it gives a warning, do you happen to know if there exists some documentation on the `packOptions`?

tools/dotnet-deb-tool/Program.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
eerhardt: copyright?  on all new code files.

tools/dotnet-deb-tool/project.json
brthor: found it
https://github.com/dotnet/core-proposals/blob/ajbaaska/projectmodel-update/DotNetCore-ProjectModel.md#pack-options

tools/dotnet-deb-tool/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+
+namespace Dotnet.Deb.Tool
```
eerhardt: `Microsoft.DotNet`, to match other namespaces.

tools/dotnet-deb-tool/ArgumentEscaper.cs
```diff
@@ -0,0 +1,206 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Dotnet.Deb.Tool
+{
+    public static class ArgumentEscaper
```
eerhardt: Is this an exact dupe of https://github.com/dotnet/core-setup/blob/master/build_projects/Microsoft.DotNet.Cli.Build.Framework/ArgumentEscaper.cs?  If so, maybe we can just compile it into the deb-tool using shared sources?  That way it isn't duplicated for a third time (in case it ever needs to change).

build_projects/shared-build-targets-utils/Utils/DebPackageCreator.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        private void InitializeDotnetDebTool()
+        {
+            CreateAndRestoreToolConsumingProject();
+        }
+
+        private void CreateAndRestoreToolConsumingProject()
+        {
+            FS.Mkdirp(_consumingProjectDirectory);
```
eerhardt: Should we delete the folder first?  To ensure it is clean between runs.

tools/dotnet-deb-tool/Program.cs
brthor: fixed.

tools/dotnet-deb-tool/Program.cs
brthor: fixed.

build_projects/shared-build-targets-utils/Utils/DebPackageCreator.cs
eerhardt: Looks like that code is happening in the 2 places where this is called.  Maybe we can move them both in here.

tools/dotnet-deb-tool/ArgumentEscaper.cs
brthor: I wish we just had the ability to pass a string[] args to process.Start...

Anyways, It's already duplicated 3 times, between utils and the two build_projects, do you think it'd be preferable to just take a dependency on CLI.Utils nuget package instead?

build_projects/shared-build-targets-utils/Utils/DebPackageCreator.cs
brthor: :+1:

build_projects/dotnet-host-build/PrepareTargets.cs
eerhardt: seems like it should work... `dotnet restore` should go through all the dirs, right? and process each one as if it was called directly on it.

tools/dotnet-deb-tool/ArgumentEscaper.cs
eerhardt: If I were to do it, I would just shared-source-compile it from the build_project.  That way you don't have an external dependency to deal with.  You only have an internal dependency on the code living in the build_project, which could easily be duped at a later date.

eerhardt:
Should the nupkg be published somewhere so the CLI repo can consume it?

eerhardt:
LGTM.  This is gonna be pretty cool when it is done.

build_projects/dotnet-host-build/PrepareTargets.cs
brthor: It is working. fixed.

brthor:
> Should the nupkg be published somewhere so the CLI repo can consume it?

Indeed it is. It is placed in Dirs.Packages which gets picked up by the Windows x64 VSO build.

tools/dotnet-deb-tool/ArgumentEscaper.cs
brthor: Yeah I agree that is better, and it is actually copied from there.

eerhardt:
> Indeed it is. It is placed in Dirs.Packages which gets picked up by the Windows x64 VSO build.

Oh right - I forgot about magic.

tools/dotnet-deb-tool/project.json
brthor: fixed.

build_projects/dotnet-host-build/DebTargets.cs
brthor: fixed.

brthor:
@dotnet-bot test this please

brthor:

eerhardt:
@brthor - is the build flaky?

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -240,10 +269,15 @@ public static BuildTargetResult RestorePackages(BuildTargetContext c)
             var dotnet = DotNetCli.Stage0;
 
             dotnet.Restore("--verbosity", "verbose", "--disable-parallel")
-                .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "tools", "independent", "RuntimeGraphGenerator"))
+                .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "tools"))//, "independent", "RuntimeGraphGenerator"))
                 .Execute()
                 .EnsureSuccessful();
 
+            //dotnet.Restore("--verbosity", "verbose", "--disable-parallel")
```
eerhardt: Can these be removed please?

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -240,10 +269,15 @@ public static BuildTargetResult RestorePackages(BuildTargetContext c)
             var dotnet = DotNetCli.Stage0;
 
             dotnet.Restore("--verbosity", "verbose", "--disable-parallel")
-                .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "tools", "independent", "RuntimeGraphGenerator"))
+                .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "tools"))//, "independent", "RuntimeGraphGenerator"))
```
eerhardt: Please remove the commented out section.

brthor:
The OS X build was failing in an unrelated way I hadn't seen, we'll see if it's persistent.

Ubuntu builds had issues with the directory deletion refactoring.

PR #31, opened 2016-05-27
Sync shared-build-targets-utils with CLI

brthor:
Accompanying CLI Pr incoming tomorrow.

PTAL @eerhardt

### DIFF
--- a/build_projects/dotnet-host-build/DebTargets.cs
+++ b/build_projects/dotnet-host-build/DebTargets.cs
@@ -51,7 +51,8 @@ namespace Microsoft.DotNet.Host.Build
         [BuildPlatforms(BuildPlatform.Ubuntu)]
         public static BuildTargetResult GenerateSharedFrameworkDeb(BuildTargetContext c)
         {
-            var packageName = Monikers.GetDebianSharedFrameworkPackageName(c);
+            var sharedFrameworkNugetVersion = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
+            var packageName = Monikers.GetDebianSharedFrameworkPackageName(sharedFrameworkNugetVersion);
             var sharedHostVersion = c.BuildContext.Get<HostVersion>("HostVersion").LockedHostVersion;
             var version = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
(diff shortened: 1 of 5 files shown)


build_projects/shared-build-targets-utils/DependencyVersions.cs
```diff
@@ -8,5 +8,11 @@ namespace Microsoft.DotNet.Cli.Build
     public class DependencyVersions
     {
         public static readonly string CoreCLRVersion = "1.0.2-rc3-24126-01";
+
+        public static readonly string SharedFrameworkVersion = "1.0.0-rc3-004312";
```
eerhardt: These versions are always going to be updated, and get out of date.  I don't think these should be here.

build_projects/shared-build-targets-utils/Utils/Crossgen.cs
```diff
@@ -154,15 +154,9 @@ public void CrossgenDirectory(string sharedFxPath, string pathToAssemblies)
 
                 IList<string> crossgenArgs = new List<string> {
                     "-readytorun", "-in", file, "-out", tempPathName,
+                    "-JITPath", GetLibCLRJitPathForVersion(),
                     "-platform_assemblies_paths", platformAssembliesPaths
                 };
-
```
eerhardt: Your PR is failing because you are removing this "good" change in core-setup.  The existing code needs to be ported into CLI, which is being done here:  https://github.com/dotnet/cli/pull/3223

build_projects/shared-build-targets-utils/DependencyVersions.cs
brthor: 👍

build_projects/shared-build-targets-utils/Utils/Crossgen.cs
brthor: Makes sense, I wasn't sure which way it had gone.

brthor: fixed.

build_projects/shared-build-targets-utils/DependencyVersions.cs
brthor: fixed.

eerhardt:
:shipit:

PR #35, opened 2016-05-27
publish dotnet-deb-tool on ubuntu

brthor:
We can't publish this on windows because it doesn't retain file permissions correctly.
@eerhardt


brthor:
merging to unblock publishing the tool.

PR #43, opened 2016-05-29
Add support for Ubuntu 16.04, Fedora 23 and openSUSE 13.2

ellismg:
While doing the PR, I did a few small things as well:

 - I simplified up how we compute the Dockerfile to use when dockerrun is not passed one.  Doing thing required that I rename the directories of the current files, but I can't rename them atomically because the VSO builds have the paths hard coded.  Once this lands, I'll update VSO and remove the old copies.
 -  I needed to add a temporary copy of `dotnet-install.sh` so I could be able to download the new stage0's.  Once we can upstream the change to dotnet/cli, we can revert 5caaef7

Once this lands, I will add VSO jobs for the new distros.

@brthor @livarcocc @eerhardt


build_projects/shared-build-targets-utils/Utils/Monikers.cs
```diff
@@ -15,9 +16,33 @@ public class Monikers
 
         public static string GetProductMoniker(BuildTargetContext c, string artifactPrefix, string version)
         {
-            string osname = GetOSShortName();
-            var arch = CurrentArchitecture.Current.ToString();
-            return $"{artifactPrefix}-{osname}-{arch}.{version}";
+            string rid = RuntimeEnvironment.GetRuntimeIdentifier();
+
+            if (rid == "ubuntu.16.04-x64" || rid == "fedora.23-x64" || rid == "opensuse.13.2-x64")
```
joshfree: consider moving these literals to static readonly values on a RID class instead of having future modifiers copy/paste these throughout the cs files

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -32,10 +32,13 @@ public class CompileTargets
             { "osx.10.10-x64", "osx.10.10-x64" },
             { "osx.10.11-x64", "osx.10.10-x64" },
             { "ubuntu.14.04-x64", "ubuntu.14.04-x64" },
+            { "ubuntu.16.04-x64", "ubuntu.16.04-x64" },
             { "centos.7-x64", "rhel.7-x64" },
```
joshfree: do we need an entry for ol.7?

scripts/docker/centos.7/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# Install CoreCLR and CoreFx dependencies
+RUN yum -q -y install unzip libunwind gettext libcurl-devel openssl-devel zlib libicu-devel
+
+# RUN apt-get update && \
+#     apt-get -qqy install unzip curl libicu-dev libunwind8 gettext libssl-dev libcurl3-gnutls zlib1g liblttng-ust-dev lldb-3.6-dev lldb-3.6 
+
+# Install Build Prereqs
+# CMake 3.3.2 from GhettoForge; LLVM 3.6.2 built from source ourselves;
+RUN yum install -y http://mirror.symnds.com/distributions/gf/el/7/plus/x86_64/cmake-3.3.2-1.gf.el7.x86_64.rpm \
+        https://matell.blob.core.windows.net/rpms/clang-3.6.2-1.el7.centos.x86_64.rpm \
```
joshfree: can you move these to a more official location or else create a tracking issue to do so?

scripts/obtain/dotnet-install.sh
```diff
@@ -0,0 +1,659 @@
...
+
+    local uname=$(uname)
+    if [ "$uname" = "Darwin" ]; then
+        echo "osx"
+        return 0
+    else
+        if [ -e /etc/os-release ]; then
+            . /etc/os-release
+
+            case "$ID.$VERSION_ID" in
```
joshfree: same question here: do we need entries for ol7, mint17, etc

joshfree:
over LGTM besides comments above

scripts/obtain/dotnet-install.sh
ellismg: We should do that in the real installer in cli.  I'll
file a bug over there.

scripts/docker/centos.7/Dockerfile
ellismg: I will create an issue.  I have some ideas on a better way to do this.

build_projects/dotnet-host-build/CompileTargets.cs
ellismg: I do not think it's needed here as we do not do OL specific builds.

joshfree:
@ellismg do you plan to merge this tonight?

ellismg:
@joshfree I figured it was a good practice to have someone else from the CLI actually give it a once over, but I'm happy to merge it now and just deal with the feedback later if you'd prefer.

joshfree:
Let's get these distros enabled tonight and create a tracking issue for any feedback that comes up from CLI (hopefully @livarcocc can help finish up those :smile))

ellismg:
@joshfree SGTM.  I'll merge this and then figure out how to get the VSO jobs set up.

joshfree:
Thanks Matt.

ellismg:
It would appear is if I don't have access to VSO to add the additional jobs so I will need @eerhardt @brthor or @piotrpMSFT to do this for me (or give me access).

I will manually do the builds that I can't add and publish them.

ellismg:
*grumble grumble* this broke some publishing stuff (which is not covered in CI), I am working on fixing stuff up, so there will be a follow up PR.

scripts/docker/ubuntu.14.04/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+FROM ubuntu:14.04
+
+# Misc Dependencies for build
+RUN apt-get update && apt-get -qqy install curl unzip gettext sudo
+
+# This could become a "microsoft/coreclr" image, since it just installs the dependencies for CoreCLR (and stdlib)
+RUN echo "deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.6 main"  | tee /etc/apt/sources.list.d/llvm.list && \
+    curl http://llvm.org/apt/llvm-snapshot.gpg.key | apt-key add - && \
+    apt-get update && apt-get -qqy install\
+	libc6 \
```
eerhardt: Why is this list so much bigger than the Ubuntu16.04 list?

scripts/docker/centos.7/Dockerfile
```diff
@@ -0,0 +1,54 @@
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# Dockerfile that creates a container suitable to build dotnet-cli
+FROM centos:7.1.1503
+
+RUN yum -q -y install deltarpm
```
eerhardt: We already supported centos.7, why do we need this docker file now?

pkg/projects/descriptions.json
```diff
@@ -5,6 +5,11 @@
         "CommonTypes": [ ]
     },
     {
+       "Name": "NuGet3MinVersion",
```
eerhardt: Is this ever used?  I don't see it being used anywhere.

scripts/docker/ubuntu.14.04/Dockerfile
brthor: To determine this list @ellismg had an awesome bash one liner that ran `ldd` against the shared libraries distributed by corefx and coreclr to determine hte entire package closure. 

Undoubtedly some of these packages are contained as dependenices of others, but it doesn't seem straightforward to determine which. So I'd say this is similar to the lock file of our package dependencies on ubuntu 14.04.

pkg/projects/descriptions.json
ellismg: The packaging targets from buildtools consume it: https://github.com/dotnet/buildtools/blob/6bc1be1914be31bfa7e11178322cc3f36fe5fb2f/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets

scripts/docker/centos.7/Dockerfile
ellismg: The commit message for the commit that introduced this made this clear, but GitHub makes it hard to see such a thing.  I wanted to have the directories named after their rid to simplify logic in dockerbuild.sh, but since the VSO jobs were using the old names, I had to copy these around in the interm.

scripts/dockerrun.sh
```diff
@@ -55,18 +55,21 @@ if [ -z "$DOCKERFILE" ]; then
...
-        elif [ "$(cat /etc/*-release | grep -cim1 centos)" -eq 1 ]; then
-            echo "Detected current OS as CentOS, using 'centos' image"
-            export DOCKERFILE=scripts/docker/centos
-        elif [ "$(cat /etc/*-release | grep -cim1 debian)" -eq 1 ]; then
-            echo "Detected current OS as Debian, using 'debian' image"
-            export DOCKERFILE=scripts/docker/debian
         else
-            echo "Unknown Linux Distro. Using 'ubuntu' image"
-            export DOCKERFILE=scripts/docker/ubuntu
+            if [ -e /etc/os-release ]; then
```
ellismg: @eerhardt This is the logic I simplified that required me to duplicate the Dockerfiles.

scripts/docker/ubuntu.14.04/Dockerfile
eerhardt: The question really is:  shouldn't the Ubuntu14 and Ubuntu16 docker files look very, very similar?  In fact, I think these two files should share implementation where it makes sense, instead of being:

1. inconsistent
2. duplicated

scripts/docker/centos.7/Dockerfile
eerhardt: Do we have an issue logged to remove the old versions?

scripts/docker/ubuntu.14.04/Dockerfile
brthor: I disagree that these dockerfiles should share implementation because I don't think we should need to reason about anything outside of the dockerfile itself to determine what the environment the docker image created from the dockerfile looks like. 

Having a shared implementation would require adding in a file to the docker container from our repo, and then executing that file with in as a part of the image build. So far we've managed to avoid needing to do that. This would add additional complexity and room for error between the different environments where the docker container is built since it's no longer self-contained.

What benefit do you see from the shared implementation?

eerhardt: A few benefits:

1. When a new dependency is added, we only need to update it in 1 place.
2. When we go to add Ubuntu17, we won't need to duplicate this list again.
3. When trying to figure out all our dependencies, we won't have to compare these lists across X versions of Ubuntu.

Basically it is the standard [DRY principle] (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

brthor: 1. The dependency is not guaranteed to be of the same version across different versions of ubuntu, and eventually we'll need to fork this logic as older versions of ubuntu are not updated. Having that forked logic inside of a shared script instead of in the appropriate dockerfiles creates conditional logic you need to reason about to determine what exact environment the dockerfile creates when things aren't working.

2. For the same reason as above that is not guaranteed. 

3. Why do we need to figure out our dependencies set across multiple versions of ubuntu? Do we do the same comparison between fedora and ubuntu?

PR #48, opened 2016-05-31
Reverting embedding version resource breaks Linux

schellap:
@eerhardt I am undoing while I investigate the issue. What changed is the cmake defines, not sure if that is what broke it.

In any case reverting to debug what is causing the problem and also going to add tests this week.

Cc @joshfree


schellap:
Closing in favor of https://github.com/dotnet/core-setup/pull/49

PR #52, opened 2016-05-31
Update branchinfo for master branch

brthor:
Changing to map to corefx branches. release/1.0.0 will stay as this file was previously.

cc @ellismg @eerhardt


ellismg:
LGTM

brthor:
@dotnet-bot test Ubuntu x64 Release Build please

branchinfo.txt
```diff
@@ -4,6 +4,6 @@
 MAJOR_VERSION=1
 MINOR_VERSION=0
 PATCH_VERSION=0
-RELEASE_SUFFIX=rc3
-CHANNEL=preview
+RELEASE_SUFFIX=rc4
+CHANNEL=master
 BRANCH_NAME=master
```
eerhardt: > release/1.0.0 will stay as this file was previously.

Doesn't this line need to be updated in release/1.0.0?

brthor: yes :) Thanks for pointing that out.

eerhardt: There will need to be a reckoning here someday.  "CHANNEL" and "BRANCH_NAME" should really be combined.  Can you do it here and now?

eerhardt:
@brthor - we should drop the "4000" offset in the 'master' branch, now that we are building "rc4" packages.

https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PrepareTargets.cs#L22

brthor:
Good idea @eerhardt, we can delete the packages which have already published,

Is the version here expected to be `1.1.0`?

eerhardt:
It will eventually be 1.1.0, when corefx moves to 1.1.0. core-setup should stay in step with corefx.

brthor:
Thanks @eerhardt, I had seen a comment somewhere mentioning this and wanted to ensure we had it right.

PR #56, opened 2016-06-01
Add Libuv to Microsoft.NETCore.App

weshaggard:
Fixes https://github.com/dotnet/cli/issues/2277

Adds Libuv to Microsoft.NETCore.App (aka Shared framework).

cc @muratg @davidfowl @moozzyk @Petermarcu @piotrpMSFT

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/project.json
+++ b/pkg/projects/Microsoft.NETCore.App/project.json
@@ -38,7 +38,8 @@
     "System.Threading.Tasks.Extensions": "4.0.0-rc3-24131-00",
     "System.Threading.Tasks.Parallel": "4.0.1-rc3-24131-00",
     "System.Threading.Thread": "4.0.0-rc3-24131-00",
-    "System.Threading.ThreadPool": "4.0.10-rc3-24131-00"
+    "System.Threading.ThreadPool": "4.0.10-rc3-24131-00",
+    "Libuv": "1.9.0-rc2-21014"
   },
   "frameworks": {
     "netstandard1.5": {}


eerhardt:
In general, I don't like the path we are going further down.  Because ASP.NET needs something, we think that **all** .NET Core projects should reference it.  I think this design is broken.

pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -38,7 +38,8 @@
     "System.Threading.Tasks.Extensions": "4.0.0-rc3-24131-00",
     "System.Threading.Tasks.Parallel": "4.0.1-rc3-24131-00",
     "System.Threading.Thread": "4.0.0-rc3-24131-00",
-    "System.Threading.ThreadPool": "4.0.10-rc3-24131-00"
+    "System.Threading.ThreadPool": "4.0.10-rc3-24131-00",
+    "Libuv": "1.9.0-rc2-21014"
```
moozzyk: do you have the aspnetrelease feed? I did not see it in the NuGet.config...

eerhardt: This dependency comes from ASP.NET?

@weshaggard - do we have a circular reference again?

joshfree:
I thought this was already discussed and shot down.

schellap:
Would there be any dependency stomping now or in the future, due to this package? By adding this to the NETCore.App package we are also increasing the servicing surface area.

eerhardt:
For some background info,

When publishing a "self-contained" app with the CLI, ALL dependencies get published to the output folder.  So if we add this dependency, every self-contained .NET Core app will be getting this native assembly, whether it is used or not.

pkg/projects/Microsoft.NETCore.App/project.json
moozzyk: Libuv does not depend on anything. It only has native binaries in it.

weshaggard: I moved it to dotnet-core feed as I didn't want to depend on aspnet feed.

moozzyk: :+1:

ellismg: @moozzyk Where does this package build from and how can I produce it myself?  I would like to continue to understand how to build all the stuff that goes into Microsoft.NETCore.App from source.

moozzyk: @ellismg - this is how how/where we build it https://github.com/aspnet/Libuv-build. From this repo we build three separate libuv-build packages (Microsoft.AspNetCore.Internal.libuv-Darwin, Microsoft.AspNetCore.Internal.libuv-Linux and Microsoft.AspNetCore.Internal.libuv-Windows) which we then fold into the Libuv package in https://github.com/aspnet/libuv-package repo.

ellismg: @moozzyk Thanks!

joshfree:
@weshaggard can you send this to shiproom if the plan is to add this to /release/1.0.0?

mellinoe:
> When publishing a "self-contained" app with the CLI, ALL dependencies get published to the output folder. So if we add this dependency, every self-contained .NET Core app will be getting this native assembly, whether it is used or not.

While true (and definitely not good), this is the status quo when using a meta-package. You will pretty much always publish unnecessary files if reference a meta package, because nothing is trimmed. For example, 90% of the assemblies published from the "dotnet new" template project are dead weight; they can be deleted without issue. Right now, you have to manually trim dependencies (aka don't use a meta package) to reduce file size. [Related discussion](https://github.com/dotnet/cli/issues/3297).

joshfree:
Circling back, ASP.NET is going to maintain and support libuv.  The dependency is small and if people go standalone, we should be able to trim this out of the published output if people don't use it.

joshfree:
@weshaggard you'll need to port this from master to the release branch

PR #64, opened 2016-06-01
[WIP] Lowercase ID and version in NuGet global packages directory

joelverhagen:
DO NOT MERGE.

Porting over changes previously in https://github.com/dotnet/cli/pull/2722.


src/corehost/cli/deps_entry.cpp
```diff
@@ -167,19 +169,21 @@ bool deps_entry_t::to_hash_matched_path(const pal::string_t& base, pal::string_t
...
     // Build the hash file path str.
     pal::string_t hash_file;
-    hash_file.reserve(base.length() + library_name.length() + library_version.length() + nupkg_filename.length() + 3);
+    hash_file.reserve(base.length() + lower_name.length() + lower_version.length() + nupkg_filename.length() + 3);
     hash_file.assign(base);
-    append_path(&hash_file, library_name.c_str());
-    append_path(&hash_file, library_version.c_str());
+    append_path(&hash_file, lower_name.c_str());
+    append_path(&hash_file, lower_version.c_str());
     append_path(&hash_file, nupkg_filename.c_str());
```
schellap: I assume that the deps file preserves casing and only the File System doesn't?

This needs to be ported to release branches as well.

Cc @gkhanna79.

joelverhagen: Yes, that's correct. Could you explain more about the release branches bit? I am unfamiliar with the branching scheme.

gkhanna79:
We have master and release branches for core-setup. @joelverhagen why are we not fixing this in master first and then port the change to release?

CC @piotrpMSFT @joshfree @weshaggard

src/corehost/cli/deps_entry.cpp
schellap: The host as part of Microsoft.NETCore.App is RTMing this June. We are pretty much entering Escrow this week. The release branch is: `release/1.0.0` on the main fork similar to `master`.

gkhanna79: > We are pretty much entering Escrow this week

We have entered Escrow!

schellap: Sorry, just saw this is on the releas branch. The mobile UI didn't show the branch. @gkhanna79 this is tested with CLI's CI quite a while back. But did not make the RC2 checkin.

schellap:
LGTM.

joelverhagen:
>  @joelverhagen why are we not fixing this in master first and then port the change to release?

If that's the process, I'll create this PR hanging off of master.

gkhanna79:
@joelverhagen Per @eerhardt in https://github.com/dotnet/core-setup/pull/92#issuecomment-223996169, he will be doing a mop-up RI from release/1.0.0 of core-setup to master. So you are okay for now to commit into release/1.0.0 only.

joelverhagen:
@dotnet-bot test RHEL7.2 x64 Release Build please

schellap:
What is the plan for this? Do we want this in RTM? @piotrpMSFT.

joelverhagen:
There is an active discussion about whether to pull this in for RTM. 
/cc @rrelyea, @yishaigalatzer who have more context.

schellap:
1.0.0 netcore runtime needs to be serviced, when nuget is upgraded.

rrelyea:
yes, we are aware that corhost will need a change in conjunction with this.

eerhardt:
I assume that this is NOT going into release/1.0.0.  We are building stable versions now and are locking down the Shared Framework very quickly.

piotrpMSFT:
@joelverhagen any updates here?

schellap:
@piotrpMSFT, offline cache layout on Unix uses lowcase and that can be upgraded, right? Because otherwise the host might have to do two probes one for lower and the 1.0.0 style for back compat.

piotrpMSFT:
offline install in CLI uses `dotnet restore` to build out the `packages` directory so it will be consistent with whatever NuGet does during that operation.

piotrpMSFT:
This issue is starting to hinder CLI development since it prevents us from taking an updated NuGet dependencies. What work remains here that prevents the PR from being merged?

gkhanna79:
@joelverhagen has setup time to discuss this prior to merge later this week. We can determine next steps based upon that.

joelverhagen:
You're faster at responding than I, @gkhanna79 :smile:. Essentially I have some questions about how the [corresponding changes in core-setup](https://github.com/dotnet/core-setup/pull/64) will work with projects already build for the Corehost that RTM'd.

piotrpMSFT:
How soon is the meeting? @eerhardt is running out of runway...

joelverhagen:
Tomorrow, 11 AM PDT.

piotrpMSFT:
great :D

eerhardt:
I'm working on a hack now in the CLI that may work around the issue:

https://github.com/dotnet/cli/pull/3937/commits/e9bab82a796ae422d75b0485816b4df6df75c6eb

Basically, the "deps.json" writer writes all the package names and versions in lower case.  The host resolver uses the casing as-is from the deps.json file.

eerhardt:
I was able to work around this by writing the deps.json file using lower cased package names and versions.  See https://github.com/dotnet/cli/pull/3937/commits/efc0efe4f95020bbc67323c3979498ac1d90ef32

joelverhagen:
Closing since this change should be based off of `master` (not `rel/1.0.0`).

PR #66, opened 2016-06-01
remove commit count offset and remove BRANCH_NAME

brthor:
cc @eerhardt


brthor:
@eerhardt made the changes to use BranchInfo in update-dependencies.

build_projects/update-dependencies/Config.cs
```diff
@@ -37,7 +40,7 @@ public class Config
         private Lazy<string> _gitHubOriginOwner;
         private Lazy<string> _gitHubUpstreamOwner = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_UPSTREAM_OWNER", "dotnet"));
         private Lazy<string> _gitHubProject = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_PROJECT", "core-setup"));
-        private Lazy<string> _gitHubUpstreamBranch = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_UPSTREAM_BRANCH", "master"));
+        private Lazy<string> _gitHubUpstreamBranch = new Lazy<string>(() => s_branchInfo.Entries["BRANCH_NAME"]);
```
eerhardt: This should still use `GetEnvironmentVariable`.  The branch name from branchinfo.txt should just be the default (if no one set the variable).  It is safer, that way we can always modify these inputs without changing the code.

brthor: alright sounds good.

brthor:
fixed to use the environment variable

eerhardt:
:shipit:

PR #68, opened 2016-06-02
Fix version.txt file contents

schellap:

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -258,7 +258,7 @@ namespace Microsoft.DotNet.Host.Build
             var arch = IsWinx86 ? "x86" : "x64";
             
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
-            var hostNugetversion = hostVersion.LatestHostVersion;
+            var hostNugetversion = hostVersion.LatestHostVersion.ToString();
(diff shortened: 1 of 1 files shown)


schellap:
@eerhardt PTAL.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -258,7 +258,7 @@ public static BuildTargetResult PackagePkgProjects(BuildTargetContext c)
             var arch = IsWinx86 ? "x86" : "x64";
             
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
-            var hostNugetversion = hostVersion.LatestHostVersion;
+            var hostNugetversion = hostVersion.LatestHostVersion.ToString();
```
eerhardt: If this bug doesn't convince people to not use `var`, I don't know what will.

eerhardt:
:shipit:

eerhardt:
@schellap - it appears that there is other issues with this VerInfo struct.  We have been getting the following build warning/error:

```
F:\Agent\_work\12\s\build_projects\shared-build-targets-utils\Utils\HostVersion.cs(34,27): warning CS0114: 'HostVersion.VerInfo.ToString()' hides inherited member 'ValueType.ToString()'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword.
```

Fixing that would also solve this issue.

eerhardt:
Fixing warning with #69

PR #86, opened 2016-06-04
Don't delete mscorlib.dll from Shared FX

schellap:
@gkhanna79 @eerhardt PTAL. Because the meaning of mscorlib.dll has changed with the renaming and facade, this needs to be present in the Shared FX. Once you guys agree, I will port this to release/1.0.0 -- this will be a problem with rc4 packages we would eventually ship out of.


eerhardt:
@ellismg was already doing this in https://github.com/dotnet/core-setup/pull/85/commits/e1d659dec8e9f90e508503029f5812da3bdffa63.

Given the conversation over there, I think it is the right approach.

ellismg:
LGTM.

PR #87, opened 2016-06-04
Add correct code to replace Runtime.CoreCLR and NETCore.Platforms hard-coded versions.

eerhardt:
Fixes the TODOs in the core-setup update-dependencies script to find the dependencies correctly.

If the dependencies aren't found, the script will inject a "NOT FOUND" string in place of the version, which will tell us really quickly that something is wrong. These packages should always be found, even if we are already on the latest version.

@ellismg @dagood


ellismg:
LGTM

PR #91, opened 2016-06-06
Fix x86 build by setting ExecutionPolicy for packaging scripts.

AustinWise:
Error message was: "File packaging\windows\host\generatemsi.ps1 cannot be loaded because running scripts is disabled on this system."
This error only happened when I did the x86 build.

### DIFF
--- a/build_projects/dotnet-host-build/MsiTargets.cs
+++ b/build_projects/dotnet-host-build/MsiTargets.cs
@@ -122,7 +122,7 @@ namespace Microsoft.DotNet.Host.Build
             }
             Directory.CreateDirectory(wixObjRoot);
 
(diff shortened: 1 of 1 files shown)


brthor:
@AustinWise Why not just set this on your machine? this is generally what we do on our dev machines.

eerhardt:
@weshaggard - can you weigh in here if this change is OK? I'm no powershell expert in any way and I am concerned about having this kind of setting in our scripts.

AustinWise:
I figured out why I was running into this problem. I had set the execution policy for 64-bit Powershell, but not 32-bit Powershell.

eerhardt:
Ok good.  Is it safe to close this PR then?

AustinWise:
Sure.

PR #117, opened 2016-06-08
Publish host pdbs to Azure

schellap:
@eerhardt PTAL. This seems to have been ignored, just remembered as a fit and finish before shipping!

Cc @joshfree

### DIFF
--- a/pkg/dir.props
+++ b/pkg/dir.props
@@ -6,6 +6,7 @@
     <BinDir Condition="'$(BinDir)'==''">$(ProjectDir)bin/</BinDir>
     <TestWorkingDir Condition="'$(TestWorkingDir)'==''">$(BinDir)tests/</TestWorkingDir>
     <PackagesOutDir Condition="'$(PackagesOutDir)'==''">$(BinDir)packages/</PackagesOutDir>
+    <SymbolPackagesOutDir Condition="'$(SymbolPackagesOutDir)'==''">$(PackagesOutDir)</SymbolPackagesOutDir>
 
     <!-- Input Directories -->
     <PackagesDir Condition="'$(PackagesDir)'==''">$(ProjectDir)packages/</PackagesDir>
@@ -25,6 +26,7 @@
 
   <PropertyGroup>
     <PackageOutputPath>$(PackagesOutDir)</PackageOutputPath>
+    <SymbolPackageOutputPath>$(SymbolPackagesOutDir)</SymbolPackageOutputPath>
     <RuntimeIdGraphDefinitionFile>$(ProjectDir)packages\Microsoft.NETCore.Platforms\1.0.1-rc4-24203-06\runtime.json</RuntimeIdGraphDefinitionFile>
     <PackageLicenseFile>$(ProjectDir)projects/dotnet_library_license.txt</PackageLicenseFile>
     <PackageThirdPartyNoticesFile>$(ProjectDir)projects/ThirdPartyNotices.txt</PackageThirdPartyNoticesFile>
--- a/pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetHost.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetHost.pkgproj
@@ -13,12 +13,14 @@
   
   <ItemGroup>
     <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/dotnet.exe" />
+    <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/dotnet.pdb">
+      <IsSymbolFile>true</IsSymbolFile>
+    </ArchitectureSpecificNativeFile>
 
     <File Include="@(ArchitectureSpecificNativeFile)">
       <TargetPath>runtimes/$(PackageTargetRuntime)/native</TargetPath>
     </File>
     <File Include="$(ProjectDir)/version.txt" />
-    <File Include="$(DotNetHostBinDir)/dotnet.pdb" />
   </ItemGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
--- a/pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetHostPolicy.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetHostPolicy.pkgproj
@@ -17,12 +17,14 @@
     </Dependency>
 
     <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/hostpolicy.dll"/>
+    <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/hostpolicy.pdb">
+      <IsSymbolFile>true</IsSymbolFile>
+    </ArchitectureSpecificNativeFile>
 
     <File Include="@(ArchitectureSpecificNativeFile)">
       <TargetPath>runtimes/$(PackageTargetRuntime)/native</TargetPath>
     </File>
     <File Include="$(ProjectDir)/version.txt" />
-    <File Include="$(DotNetHostBinDir)/hostpolicy.pdb" />
   </ItemGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
--- a/pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetHostResolver.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetHostResolver.pkgproj
@@ -17,12 +17,14 @@
     </Dependency>
 
     <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/hostfxr.dll"/>
+    <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/hostfxr.pdb">
+      <IsSymbolFile>true</IsSymbolFile>
+    </ArchitectureSpecificNativeFile>
 
     <File Include="@(ArchitectureSpecificNativeFile)">
       <TargetPath>runtimes/$(PackageTargetRuntime)/native</TargetPath>
     </File>
     <File Include="$(ProjectDir)/version.txt" />
-    <File Include="$(DotNetHostBinDir)/hostfxr.pdb" />
   </ItemGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />


eerhardt:
I'll review this when it is green.

schellap:
@eerhardt sorry it was a missing semi colon, but I added few checks to the build.

Cc @gkhanna79 symbols for host wasn't getting published to Azure.

build_projects/dotnet-host-build/PackageTargets.cs
```diff
@@ -49,6 +50,41 @@ public static BuildTargetResult GenerateVersionBadge(BuildTargetContext c)
...
+
+            foreach (var file in Directory.GetFiles(Dirs.CorehostLatest, "*.pdb", SearchOption.TopDirectoryOnly))
+            {
+                var destFile = file.Replace(Dirs.CorehostLatest, nativeSymbolsRoot);
+                File.Copy(file, destFile, true);
+                c.Warn(destFile);
+            }
+            FixPermissions(nativeSymbolsRoot);
+            
+            var expectedFiles = new string[] { "dotnet.pdb", "hostfxr.pdb", "hostpolicy.pdb" };
```
gkhanna79: What about symbols for non-Windows?

build_projects/dotnet-host-build/PackageTargets.cs
```diff
@@ -49,6 +50,41 @@ public static BuildTargetResult GenerateVersionBadge(BuildTargetContext c)
...
+            var nativeSymbolsRoot = Path.Combine(Dirs.Output, "obj", "nativeSymbols");
+            var nativeSymbolsZipFileDir = Path.Combine(Dirs.Intermediate, "nativeSymbols");
+
+            Utils.DeleteDirectory(nativeSymbolsRoot);
+            Directory.CreateDirectory(nativeSymbolsRoot);
+
+            Utils.DeleteDirectory(nativeSymbolsZipFileDir);
+            Directory.CreateDirectory(nativeSymbolsZipFileDir);
+
+            foreach (var file in Directory.GetFiles(Dirs.CorehostLatest, "*.pdb", SearchOption.TopDirectoryOnly))
```
brthor: Should we be publishing CoreHostLatest here or CoreHostLocked?

As we up the build number and the host is locked on some previous build, the pdbs in this package won't be representative of the binaries we shipped if we don't use CoreHostLocked.

eerhardt:
I think we should be following the same thing that CoreFX and CoreCLR are doing.  Here is a link:

[Symbol indexing guidance](onenote:https://microsoft.sharepoint.com/teams/netfx/corefx/Documents/CoreFx Notes/Engineering/OSS.one#Symbol%20indexing%20guidance&section-id={3B2F124D-C418-47B2-AF63-0A43940A6DAC}&page-id={FC7D1B53-8F80-48D5-A1B4-DA862EB7E43E}&end) [(Web View)](https://microsoft.sharepoint.com/teams/netfx/corefx/_layouts/OneNote.aspx?id=%2Fteams%2Fnetfx%2Fcorefx%2FDocuments%2FCoreFx%20Notes&wd=target%28Engineering%2FOSS.one%7C3B2F124D-C418-47B2-AF63-0A43940A6DAC%2FSymbol%20indexing%20guidance%7CFC7D1B53-8F80-48D5-A1B4-DA862EB7E43E%2F%29)

Note that there are sub-pages to that OneNote page.

eerhardt:
So speaking with @dagood, we should be making `symbols.nupkg` files for each of our packages that contain binaries and pushing them to MyGet.  This is what the CoreFX and CoreCLR are doing.

/cc @weshaggard @ericstj

dagood:
You can check out https://github.com/dotnet/corefx/blob/master/src/Native/pkg/dir.targets to see how CoreFX creates native symbol packages with platform-based auto-detection. https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/PackageLibs.targets#L164 from buildtools does some pdb auto-detection.

I haven't looked into core-setup enough to tell how to integrate those ideas. I'm not against publishing symbols to multiple places it it helps them get out sooner (or help people find them) though.

build_projects/dotnet-host-build/PackageTargets.cs
schellap: The pdbs are always for the latest. The locked will have its pdbs uploaded when it was built.

brthor: The binaries we package are not the latest, why should the pdbs be?

schellap: What gets packaged with netcore.app is a fixed version. But what gets produced each time is a new binary and is pushed with the build. Its version will be latest and pdbs will be latest.

build_projects/dotnet-host-build/PackageTargets.cs
schellap: Sorry, are there some other files that need to be included?

dagood: In CoreFX and CoreCLR we strip xplat symbols into `.so.dbg` and `.dylib.dwarf` files. By default, symbols might still exist in the `.so` and `.dylib` files, but I don't know for sure about core-setup.

schellap:
> I think we should be following the same thing that CoreFX and CoreCLR are doing. 

Is this for consistency or there is nice VS integration for symbols.nupkg for native binaries? This is the CLI guidance for RC2, why change this now? We should anyway move to msbuild based system for the repo post preview2.

https://github.com/dotnet/cli/blob/cd0f847cd73cf8831de3b2c5a244f34730fb0bb1/build_projects/dotnet-cli-build/PublishTargets.cs#L266

I would prefer to have this solution for Escrow in the short term.

build_projects/dotnet-host-build/PackageTargets.cs
brthor: @schellap The host binaries getting pushed to azure are consumed in the cli. It is the locked version of the corehost, and if it is latest that is a bug.

schellap:
@weshaggard, @ericstj is there some switch to turn on auto symbol packaging in the pkgproj system? We already use buildtools for packaging.

build_projects/dotnet-host-build/PackageTargets.cs
gkhanna79: Correct!

gkhanna79:
> this is the CLI guidance for RC2
@schellap CLI is all managed but host is native. Need to ensure that symbols for native components are available as well.

build_projects/dotnet-host-build/PackageTargets.cs
brthor: https://github.com/dotnet/core-setup/blob/master/build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs#L171 

https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PackageTargets.cs#L101 

Locked bits are published in the zips.

eerhardt:
> Is this for consistency or there is nice VS integration for symbols.nupkg for native binaries?

It is for less work all around.  We don't need to document all the different places/ways we publish symbols. (i.e. if you're looking for XX look here, if you are looking for YY look here, if you need these kinds of symbols for this binary, look in this third spot).  

We can "stand on the shoulders of giants" here and not "reinvent the wheel". Let's do things the way the rest of the product does them, for everyone's sanity.

dagood:
Agreed on making symbol packages. A baseline I think we should go for is every library package on our MyGet feed having symbols available in a symbol package. (Unless symbols don't exist.) They are extremely discoverable, and MyGet indexes their contents when possible on a symbols feed with its URL on the gallery page.

There's one thing missing to get symbol packages building (it shows up in pack.cmd logs):

```
Nuspec C:\git\core-setup\pkg\bin\packages\specs\Microsoft.NETCore.App.nuspec does not contain symbol or source files. Not creating symbol package.
Nuspec C:\git\core-setup\pkg\bin\packages\specs\Microsoft.NETCore.DotNetHost.nuspec does not contain symbol or source files. Not creating symbol package.
Nuspec C:\git\core-setup\pkg\bin\packages\specs\runtime.win7-x64.Microsoft.NETCore.DotNetHost.nuspec does not contain any files matching src\**\*. Not creating symbol package.
Nuspec C:\git\core-setup\pkg\bin\packages\specs\Microsoft.NETCore.DotNetHostPolicy.nuspec does not contain symbol or source files. Not creating symbol package.
Nuspec C:\git\core-setup\pkg\bin\packages\specs\runtime.win7-x64.Microsoft.NETCore.DotNetHostPolicy.nuspec does not contain any files matching src\**\*. Not creating symbol package.
Nuspec C:\git\core-setup\pkg\bin\packages\specs\Microsoft.NETCore.DotNetHostResolver.nuspec does not contain symbol or source files. Not creating symbol package.
Nuspec C:\git\core-setup\pkg\bin\packages\specs\runtime.win7-x64.Microsoft.NETCore.DotNetHostResolver.nuspec does not contain any files matching src\**\*. Not creating symbol package.
```

It's also a good idea to specify the output location. I'm going to try this locally to see if I spot any more config that needs to happen.

dagood:
It looks like the auto-detection doesn't do much in this case... I think it's because the pkgproj isn't referencing csproj's. (I wouldn't expect it to, but I didn't realize that was how the autodetection worked.)

Anyway, these changes make Windows native symbol packages: https://github.com/dotnet/core-setup/compare/master...dagood:symbol-packages.

I also took a look on OSX and there are no `.dwarf` files being generated, so I think stripping xplat native symbols and packaging them could be done in another PR since there's significant changes to get that going.

schellap:
@dagood, your commit just worked.

@weshaggard how does the publishing work for the symbol packages? Is this something your Azure Pull script would do once the symbol nupkgs are on Azure?

schellap:
We are good even if it is on the Azure servers and not on the feed in the short term.

eerhardt:
> how does the publishing work for the symbol packages? Is this something your Azure Pull script would do once the symbol nupkgs are on Azure?

@schellap - see PublishCoreHostPackagesToFeed()

https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/dotnet-host-build/PublishTargets.cs#L154

```C#
        private static void PublishCoreHostPackagesToFeed()
        {
            var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}";

            Directory.CreateDirectory(Dirs.PackagesNoRID);
            AzurePublisherTool.DownloadFilesWithExtension(hostBlob, ".nupkg", Dirs.PackagesNoRID);

            string nugetFeedUrl = EnvVars.EnsureVariable("NUGET_FEED_URL");
            string apiKey = EnvVars.EnsureVariable("NUGET_API_KEY");
            NuGetUtil.PushPackages(Dirs.PackagesNoRID, nugetFeedUrl, apiKey);

            string githubAuthToken = EnvVars.EnsureVariable("GITHUB_PASSWORD");
            VersionRepoUpdater repoUpdater = new VersionRepoUpdater(githubAuthToken);
            repoUpdater.UpdatePublishedVersions(Dirs.PackagesNoRID, $"build-info/dotnet/core-setup/{Channel}/Latest").Wait();
        }
```

`NuGetUtil.PushPackages(Dirs.PackagesNoRID` should push any nupkgs that are in the `Dirs.PackagesNoRID` directory.

eerhardt:
And one step before that is each platform needs to publish its `.symbols.nupkg` files along with its regular `.nupkg` files to Azure.  Each platform publishes its packages to Azure, and then "FinalizeBuild" comes through and pushes them all to myget.org at the same time.

See https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/dotnet-host-build/PublishTargets.cs#L254-L265 for how the packages get pushed to Azure on each platform.

schellap:
@dagood I just used your commit and that pretty much nailed it. Appreciate the work you have put into this -- it is so easy to consume for us.

eerhardt:
:shipit:

schellap:
@dagood or @eerhardt is there a concept of `nuget restore` for symbol packages? In this case, we only publish symbols for `win7-x64` and `win7-x86`. Would this be a problem for other platforms as the full closure is not there? or is it simply just a name match with `symbols` appended?

pkg/dir.props
```diff
@@ -20,11 +21,12 @@
   <PropertyGroup>
     <BaseOutputPath Condition="'$(BaseOutputPath)'==''">$(BinDir)</BaseOutputPath>
   </PropertyGroup>
-
-  <Import Project="version.props" Condition="Exists('version.props')" />
+
```
weshaggard: nit: added spaces at the end of lines.

gkhanna79:
@schellap Is there anything outstanding here?

dagood:
@schellap No problem, glad it worked. :smile: The story isn't great right now for xplat native, though... CoreCLR and CoreFX do most of that manually, and differently from each other. But to be clear, I don't this PR needs to be concerned with that.

Symbol packages aren't a part of `nuget restore`--they're just a delivery mechanism for symbols that MyGet lets you download again manually. So no need to worry about that either.

schellap:
> Symbol packages aren't a part of nuget restore--they're just a delivery mechanism for symbols that MyGet lets you download again manually. So no need to worry about that either.

Nothing outstanding based on this

gkhanna79:
@schellap This PR is in master. Does this need to be merged in release/1.0.0? If so, please file an issue for it and follow the usual process to get it in for rtm.

PR #126, opened 2016-06-09
[master]Port other meta-packages

weshaggard:
Adds the following meta-packages:
Microsoft.NETCore
Microsoft.NETCore.UniversalWindowsPlatform

We moved these packages from another repo with a different versioning scheme
so to ensure we produce a higher version we are shipping one which starts with
3 as a temporary measure until next milestone ships.

cc @ericstj


ericstj:
LGTM, did you diff the nuspecs with those we produced in TFS?

weshaggard:
>did you diff the nuspecs with those we produced in TFS?

Yes the only differences I see in the specs are, the exact versions, the servicability/schema, and the open builds don't have the SHA information in the description.

ericstj:
OK, those diffs are acceptable.  Thanks for checking.  Feel free to merge.

PR #127, opened 2016-06-09
[release]Port other meta-packages 

weshaggard:
Port https://github.com/dotnet/core-setup/pull/111
Port https://github.com/dotnet/core-setup/pull/126

cc @ericstj


build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -252,54 +252,43 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
...
+
+            msbuildProps.AppendLine(@"<Project xmlns=""http://schemas.microsoft.com/developer/msbuild/2003"">");
+            msbuildProps.AppendLine("  <PropertyGroup>");
+            msbuildProps.AppendLine($"    <Platform>{arch}</Platform>");
+            msbuildProps.AppendLine($"    <DotNetHostBinDir>{Dirs.CorehostLatest}</DotNetHostBinDir>");
+            msbuildProps.AppendLine($"    <HostVersion>{hostVersion.LatestHostPolicyVersion.WithoutSuffix}</HostVersion>");
+            msbuildProps.AppendLine($"    <HostResolverVersion>{hostVersion.LatestHostFxrVersion.WithoutSuffix}</HostResolverVersion>");
+            msbuildProps.AppendLine($"    <HostPolicyVersion>{hostVersion.LatestHostVersion.WithoutSuffix}</HostPolicyVersion>");
+            msbuildProps.AppendLine($"    <BuildNumberMajor>{hostVersion.LatestHostBuildMajor}</BuildNumberMajor>");
+            msbuildProps.AppendLine($"    <BuildNumberMinor>{hostVersion.LatestHostBuildMinor}</BuildNumberMinor>");
```
brthor: @weshaggard Can you help me understand how this translates to the version of Microsoft.NETCore.App?

We have an inconsistency here between the version of Microsoft.NETCore.App generated after this change and the version persisted in the shared framework redist causing portable apps to break.

There seems to be an additional `-00` appended. Is that expected and we should account for it in other places, or should we remove it here?

weshaggard: It gets generated by https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L16. 

We can update the Microsoft.NETCore.App.pkgproj to set that and exclude the "-00". I will preemptively create a PR with that change if that helps others in consuming this.

eerhardt: > We can update the Microsoft.NETCore.App.pkgproj to set that and exclude the "-00". I will preemptively create a PR with that change if that helps others in consuming this.

We are down the path of `Microsoft.NETCore.App` containing the "-00" suffix already. Let's stick with what we have.

PR #128, opened 2016-06-09
Mark host and NETCore.App versions stable

schellap:
@ericstj this is the PR marking packages stable in the core-setup repo.

Currently failing because buildtools appends rc3 etc., automatically... looking.


schellap:
This is a branch on the main fork.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -18,19 +18,24 @@ public struct VerInfo
...
                 Release = release;
                 BuildMajor = buildMajor;
                 BuildMinor = buildMinor;
+                CommitCountString = commitCountString;
             }
 
             public string WithoutSuffix => $"{Major}.{Minor}.{Patch}";
 
+            public string MsiBuildMajor => string.IsNullOrEmpty(BuildMajor) ? BuildMajor : CommitCountString;
+            public string MsiBuildMinor => string.IsNullOrEmpty(BuildMinor) ? BuildMinor : "00";
```
gkhanna79: Is the convention "0" or "00" for package that come from open @jhendrixMSFT ?

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -72,7 +78,7 @@ public override string ToString()
         // Locked muxer for consumption in CLI.
         //
         public bool IsLocked = false; // Set this variable to toggle muxer locking.
-        public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00") : LatestHostFxrVersion;
-        public VerInfo LockedHostVersion    => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00") : LatestHostVersion;
+        public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00", "002468") : LatestHostFxrVersion;
```
gkhanna79: What are the Locked versions about?

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
jhendrixMSFT: For VSO builds they start at "01".

schellap: We dont have these concepts in core-setup. This is for an explicit number so we control it than leave it to buildtools.

pkg/stable.packages.props
```diff
@@ -0,0 +1,950 @@
...
+    </StablePackage>
+    <StablePackage Include="System.ServiceModel.Security">
+      <Version>4.0.1</Version>
+    </StablePackage>
+    
+    <!-- NETCore 1.0 CoreCLR -->
+    <StablePackage Include="Microsoft.NETCore.Runtime.CoreCLR">
+      <Version>1.0.2</Version>
+    </StablePackage>
+
```
ericstj: We may be able to remove everything above here because @weshaggard's logic for building the meta-package pulls the full version from the project.json.  In that case we'd get stable versions just by updating that project.json when we ingest a new build.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -49,12 +54,13 @@ public override string ToString()
...
 
         // Full versions and package information.
+        public bool EnsureStableVersion => true;
         public string LatestHostPrerelease => "rc3";
-        public string LatestHostBuildMajor => CommitCountString;
-        public string LatestHostBuildMinor => "00";
-        public VerInfo LatestHostVersion => new VerInfo(1, 0, 1, LatestHostPrerelease, LatestHostBuildMajor, LatestHostBuildMinor);
-        public VerInfo LatestHostFxrVersion => new VerInfo(1, 0, 1, LatestHostPrerelease, LatestHostBuildMajor, LatestHostBuildMinor);
-        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 0, 1, LatestHostPrerelease, LatestHostBuildMajor, LatestHostBuildMinor);
+        public string LatestHostBuildMajor => "";
```
eerhardt: Can we get rid of these empty strings?  Are they necessary?

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -72,7 +78,7 @@ public override string ToString()
         // Locked muxer for consumption in CLI.
         //
         public bool IsLocked = false; // Set this variable to toggle muxer locking.
-        public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00") : LatestHostFxrVersion;
-        public VerInfo LockedHostVersion    => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00") : LatestHostVersion;
+        public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00", "002468") : LatestHostFxrVersion;
+        public VerInfo LockedHostVersion    => IsLocked ? new VerInfo(1, 0, 1, "rc2", "002468", "00", "002468") : LatestHostVersion;
```
eerhardt: `rc2`??

pkg/stable.packages.props
eerhardt: :+1: we just need to write the stable package versions to `dotnet/versions` and the auto-PR functionality will pick them up.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -49,12 +54,13 @@ public override string ToString()
         //
 
         // Full versions and package information.
+        public bool EnsureStableVersion => true;
```
eerhardt: I don't think we should turn this on until we are 99.99% sure we aren't going to spin a new build of the DotNetHost

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
eerhardt: I'm not sure I understand the name here: `MsiBuildMajor`?  This is the version of the Msi?

schellap:
```
/opt/code/pkg/Tools/Packaging.targets(1041,5): error : Error when creating nuget lib package from /opt/code/pkg/bin/packages/specs/Microsoft.NETCore.App.nuspec. System.IO.InvalidDataException: A stable release of a package should not have on a prerelease dependency. Either modify the version spec of dependency "Libuv [1.9.0-rc2-21014, )" or update the version field.
```

In this build host packages are producing stable nupkgs.

LibUV and CSC are not stable so there is an expected error Cc @ericstj 

Cc @weshaggard

pkg/stable.packages.props
ericstj: Neat.  So once corefx/coreflc pushes these out you'll get them.

weshaggard: @eerhardt are you sure the update will work without a prerelease versions? I didn't believe it work with the package list but just the version number which is only the prerelease version.

eerhardt: For `dotnet/core-setup` and `dotnet/cli`, I changed over to read the Latest_Packages.txt file, parse it into "DependencyInfo" objects, and we do the full replacement of the version for any packages that match the ID.  It works really well. When NETStandard.Library switched from `1.5` to `1.6` we didn't even notice.  The tool took care of it.

See https://github.com/dotnet/core-setup/blob/master/build_projects/update-dependencies/UpdateFilesTargets.cs#L41-L59

and 

https://github.com/dotnet/core-setup/blob/master/build_projects/update-dependencies/UpdateFilesTargets.cs#L136-L164

eerhardt: I should say though, `dotnet/corefx` and `dotnet/coreclr` do not do "update dependencies" like `core-setup` and `cli` do.  They still are based on just the single prerelease version.

My comment above `we just need to write the stable package versions to dotnet/versions and the auto-PR functionality will pick them up.` was strictly meant for the current repo - `core-setup`.

ericstj:
This PR cannot go in until we have stable dependency packages.  It needs to be serialized after the CoreFx, CoreCLR, WCF, LibUV, and Roslyn builds.

schellap:
Blocked on #141

schellap:
@dotnet-bot test this please

schellap:
```
13:33:38 [39;49m[31m/opt/code/pkg/Tools/Packaging.targets(1041,5): error : Error when creating nuget lib package from /opt/code/pkg/bin/packages/specs/Microsoft.NETCore.UniversalWindowsPlatform.nuspec. System.IO.InvalidDataException: A stable release of a package should not have on a prerelease dependency. Either modify the version spec of dependency "Microsoft.NETCore.Portable.Compatibility [1.0.1-rc3-24208-00, )" or update the version field. [/opt/code/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj]
```

gkhanna79:
CC @weshaggard

weshaggard:
As called out by @ericstj in an earlier comment we are going to need to update the project.json files for the meta-packages to point to stable versions in order to build them stable.

schellap:
> As called out by @ericstj in an earlier comment we are going to need to update the project.json files for the meta-packages to point to stable versions in order to build them stable.

The auto-updater would pick this up right? I hope it is not a manual hack to just mark stable and expecting NuGet restore to pick up recent version of the packages -- asking because I don't see `Microsoft.CSharp` stable version on our dotnet-core feed.

```
A stable release of a package should not have on a prerelease dependency. Either modify the version spec of dependency "Microsoft.CSharp [4.0.1-rc3-24210-10, )" 
```

schellap:
Sorry just saw `Microsoft.CSharp` stable is there! I'll update this.

schellap:
> Either modify the version spec of dependency "Microsoft.CodeAnalysis.CSharp [1.3.0-beta1-20160610-04, )"

A stable version for the above is not available on the feed.

image

schellap:
@joshfree who can provide these stable packages, see the `[HACK]` commit above.

```
13:16:45 Errors in /mnt/resource/j/workspace/dotnet_core-setup/release_1.0.0/debug_centos7.1_x64_prtest/TestAssets/TestProjects/StandaloneTestApp/project.json
13:16:45     Unable to resolve 'Microsoft.CodeAnalysis.VisualBasic (>= 1.3.0)' for '.NETCoreApp,Version=v1.0'.
13:16:45     Unable to resolve 'Microsoft.CodeAnalysis.CSharp (>= 1.3.0)' for '.NETCoreApp,Version=v1.0'.
```

ericstj:
Those will be coming from the roslyn team.  I believe either @cston or @tmat is working on that.

cston:
@schellap, @ericstj Roslyn packages are being updated in https://github.com/dotnet/roslyn/pull/11945.

ericstj:
@dotnet-bot test this please

ericstj:
Looks like there are some more places where the build infra thinks NETCore.App will have pre-release versions.

ericstj:
I think the culprit is 'SharedFrameworkNugetVersion'.  Trying to find the source of this property.  /cc @weshaggard @eerhardt

ericstj:
Found it: https://github.com/dotnet/core-setup/blob/52ebe205182e9e475e87e03a5caef92de1d56231/build_projects/shared-build-targets-utils/Utils/BuildVersion.cs#L10

ericstj:
Lots of green here :smile:

ericstj:
Ok, now time for a proper review.  @weshaggard @eerhardt please take a look ASAP.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -19,24 +19,29 @@ public struct VerInfo
...
 
             public string GenerateMsiVersion()
             {
-                return Version.GenerateMsiVersion(Major, Minor, Patch, Int32.Parse(BuildMajor));
+                return Version.GenerateMsiVersion(Major, Minor, Patch, Int32.Parse(VerRsrcBuildMajor));
             }
 
             public string WithoutSuffix => $"{Major}.{Minor}.{Patch}";
 
+            public string VerRsrcBuildMajor => !string.IsNullOrEmpty(BuildMajor) ? BuildMajor : CommitCountString;
```
joshfree: should VerInfo ctor require buildMajor to be not null or empty?

joshfree: otherwise LGTM

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -267,9 +267,10 @@ public static BuildTargetResult GenerateMSbuildPropsFile(BuildTargetContext c)
             msbuildProps.AppendLine($"    <HostVersion>{hostVersion.LatestHostPolicyVersion.WithoutSuffix}</HostVersion>");
             msbuildProps.AppendLine($"    <HostResolverVersion>{hostVersion.LatestHostFxrVersion.WithoutSuffix}</HostResolverVersion>");
             msbuildProps.AppendLine($"    <HostPolicyVersion>{hostVersion.LatestHostVersion.WithoutSuffix}</HostPolicyVersion>");
-            msbuildProps.AppendLine($"    <BuildNumberMajor>{hostVersion.LatestHostBuildMajor}</BuildNumberMajor>");
-            msbuildProps.AppendLine($"    <BuildNumberMinor>{hostVersion.LatestHostBuildMinor}</BuildNumberMinor>");
-            msbuildProps.AppendLine($"    <PreReleaseLabel>{hostVersion.LatestHostPrerelease}</PreReleaseLabel>");
+            msbuildProps.AppendLine($"    <BuildNumberMajor></BuildNumberMajor>");
```
weshaggard: Do these properties actually need to be cleared? I believe they are also used to put version numbers into the binaries which we still want to happen for the hosting assemblies correct?

weshaggard:
LGTM - Just please verify that the file versions in the binaries still contain the build numbers.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -19,24 +19,29 @@ public struct VerInfo
...
+            public string VerRsrcBuildMinor => !string.IsNullOrEmpty(BuildMinor) ? BuildMinor : "00";
```
eerhardt: Can we rename these properties to follow normal .NET property names?  What does "Rsrc" mean?

If we can't rename these properties, comments are really needed in order to make this code readable.

build_projects/shared-build-targets-utils/Utils/BuildVersion.cs
```diff
@@ -7,7 +7,7 @@ public class BuildVersion : Version
         public string SimpleVersion => $"{Major}.{Minor}.{Patch}.{CommitCountString}";
         public string VersionSuffix => $"{ReleaseSuffix}-{CommitCountString}";
         public string NuGetVersion => $"{Major}.{Minor}.{Patch}-{VersionSuffix}";
```
eerhardt: Does this version need to be updated too?

pkg/stable.packages.props
```diff
@@ -0,0 +1,105 @@
...
+      <Version>1.0.0</Version>
+    </StablePackage>
+    <StablePackage Include="Microsoft.NETCore">
+      <Version>5.0.1</Version>
+    </StablePackage>
+    <StablePackage Include="Microsoft.NETCore.UniversalWindowsPlatform">
+      <Version>5.2.0</Version>
+    </StablePackage>
+    <StablePackage Include="Microsoft.NETCore.DotNetHost">
+      <Version>1.0.1</Version>
```
eerhardt: 1.0.1 seems to be duplicated a lot.  Would it make sense to put that in 1 place?

pkg/stable.packages.props
```diff
@@ -0,0 +1,105 @@
...
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <!-- NETCore 1.0 core-setup -->
+    <StablePackage Include="Microsoft.NETCore.App">
+      <Version>1.0.0</Version>
+    </StablePackage>
+    <StablePackage Include="Microsoft.NETCore">
+      <Version>5.0.1</Version>
+    </StablePackage>
+    <StablePackage Include="Microsoft.NETCore.UniversalWindowsPlatform">
```
eerhardt: Are we really shipping this as a "stable" package?  Is it part of .NET Core 1.0?

eerhardt:
>     "netstandard1.5": {}

I'm not sure if this matters (and I know it is outside the scope of this change), but is this supposed to be netstandard1.6?

---
Refers to: pkg/projects/Microsoft.NETCore.App/project.json:45 in d99ac6e. [](commit_id = d99ac6e182eaa2267b1146f0f58467c5279533b2, deletion_comment = False)

pkg/stable.packages.props
weshaggard: Yes we are shipping a stable UWP package with .NET Core RTM.

weshaggard:
The netstandard1.5 TFM in the meta-package project.json files doesn't really matter. It does get used for restoring to ensure the package actually exists but it doesn't get used at all in the making of the meta-package itself.

build_projects/shared-build-targets-utils/Utils/BuildVersion.cs
ericstj: What uses it?  I had the impression this was used to represent the nuget version picked up which isn't yet stable, but that assumption may be incorrect.

eerhardt: If nothing uses it, we should remove it.

pkg/stable.packages.props
ericstj: Think of this as a database.  It's autogenerated.  I'd prefer to keep things independent rather than introduce some accidental coupling between IDs.

build_projects/shared-build-targets-utils/Utils/BuildVersion.cs
eerhardt: It looks like there are 2 "dead" code places that use it:

https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/shared-build-targets-utils/Utils/Monikers.cs#L54-L57

https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/shared-build-targets-utils/Utils/Utils.cs#L150-L154

Those methods were copied over from CLI, but aren't invoked, so they can be removed along with this property.

ericstj: Appears to be used here: https://github.com/dotnet/core-setup/blob/da800066269351dddae310dc262e2184c674d790/build_projects/shared-build-targets-utils/Utils/Utils.cs#L150 which I seems to never be called.

pkg/stable.packages.props
eerhardt: Oh?  I didn't know this was generated. What generates it?

build_projects/shared-build-targets-utils/Utils/BuildVersion.cs
ericstj: >  2 "dead" code places 

Yep, found those and cleaned them up

build_projects/dotnet-host-build/CompileTargets.cs
schellap: This is used only for packaging. The build.sh takes care of baking in version info and rid info into the binary.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
schellap: NuGet versioning doesn't take a build major because of its notion of stable. But MSI version will be 1.0.1.3456. The reason is two different MSI binaries should not be 1.0.1.

pkg/stable.packages.props
ericstj: It isn't part of the build.  It's an dumb tool that I wrote a year or so ago around the first time we did this for UWP.

build_projects/dotnet-host-build/CompileTargets.cs
weshaggard: OK this is something to keep in mind if we start building any managed code as part of this BuildTools infrastructure in core-setup.

schellap:
Waiting for :shipit: from @ericstj, once green will merge it

ericstj:
:shipit:

eerhardt:
Can we ensure we remove things like "[WIP]" from PR titles when we think a PR is ready?  It gets really confusing when PRs are ready or not.

Also - this has a "NO MERGE" Label on it.

PR #134, opened 2016-06-09
fix portable apps

brthor:
@weshaggard

This is accounting for version changes in #127 

Please comment if this is the right fix here. After tests were merged they caught the inconsistency and our official builds are currently down. We'll have to flow this change upstack to cli as well.

cc @eerhardt


eerhardt:
Merging since this is blocking getting a new CLI.

@Petermarcu

Petermarcu:
Go!

weshaggard:
I'm actually OK removing the "-xx" from the version number for this repo as well if it makes consuming it upstack easier.

PR #139, opened 2016-06-10
Fix the VersionRepoUpdater to handle symbols.nupkg files correctly.

eerhardt:
When the VersionRepoUpdater pushes a change to `dotnet/versions`, it is putting a bunch of blank lines at the beginning of the Packages.txt file. This can break anyone trying to parse the file.

The problem is that our Regex to parse package Ids and versions based on the file name doesn't account for .symbols.nupkg files.  Fixing this generates the correct file again.

See https://github.com/dotnet/versions/commit/9ab28562290e732c3d09ca2fae4a84a50d22cc26 for an example of a bad Packages.txt file.

@dagood @weshaggard

### DIFF
--- a/build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
+++ b/build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
@@ -12,7 +12,7 @@ namespace Microsoft.DotNet.Cli.Build
 {
     public class VersionRepoUpdater
     {
-        private static Regex s_nugetFileRegex = new Regex("^(.*?)\\.(([0-9]+\\.)?[0-9]+\\.[0-9]+(-([A-z0-9-]+))?)\\.nupkg$");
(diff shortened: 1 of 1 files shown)


build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.DotNet.Cli.Build
 {
     public class VersionRepoUpdater
     {
-        private static Regex s_nugetFileRegex = new Regex("^(.*?)\\.(([0-9]+\\.)?[0-9]+\\.[0-9]+(-([A-z0-9-]+))?)\\.nupkg$");
+        private static Regex s_nugetFileRegex = new Regex("^(.*?)\\.(([0-9]+\\.)?[0-9]+\\.[0-9]+(-([A-z0-9-]+))?)(\\.symbols)?\\.nupkg$");
```
weshaggard: I would suggest turning these into named groups (`(<symbols>?\\.symbols)`) instead of remember and figuring out the correct index.

eerhardt: That's a good idea. Fixed.

PR #143, opened 2016-06-10
[master] Tests + Portable App Fix

brthor:
Bringing over tests to master, going to merge when green.

cc @eerhardt


brthor:
Tests are failing here:

```
Error: assembly specified in the dependencies manifest was not found -- package: 'Microsoft.CSharp', version: '4.0.1-rc3-24210-06', path: 'lib/netstandard1.3/Microsoft.CSharp.dll'
```

cc @schellap @eerhardt any ideas?

schellap:
See: "problem with strict deps email" thread. NuGet is dependent on assemblies not part of CLI and FX.

I'll file an issue once I am at work.

brthor:
@schellap This doesn't seem to have anything to do with NuGet as this is a muxer invocation of a published test app which is failing. 

I'll take a look at the thread though.

schellap:
How are you generating the deps file? With an old NETCore.App?

eerhardt:
@brthor - the 'master' branch is using `rc4` versions of corefx, coreclr, etc.

You'll probably need to upgrade your tests to using `rc4`.

brthor:
You're right @eerhardt that's the crux of the issue but after chatting with @schellap I'm convinced we have a conceptual issue here.

My tests are relying on a lower version of the sharedfx, so the host will automatically roll it forward to the one we just built. However the deps file written out by restore contains version X of Microsoft.CSharp.dll while the deps file of the built sharedfx contains version Y of Microsoft.CSharp.dll

After the rollforward, the host goes looking for version X of Microsoft.CSharp.dll in the sharedfx, and can't find it because it contains version Y. The issue here is that publish should not write out version X of Microsoft.CSharp.dll to the deps file of the app, should publish version X next to the app, or the host needs to forcefully upgrade the version to versionY, which is extremely dangerous without reasoning about the full graph of dependencies in the way nuget would.

brthor:
What I'm not clear on though is why we don't see this issue in built apps

PR #155, opened 2016-06-13
re-enable deb publishing

brthor:
@eerhardt @joshfree @piotrpMSFT


eerhardt:
Is this really all that is needed?  Why was it shut off in the first place?

eerhardt:
Is it because both repos would have been publishing the packages to debian?

brthor:
Let's talk offline. Our dev feed is published to Microsoft.com/net so this has potential impact for preview1 bits

brthor:
see here:
https://github.com/dotnet/cli/issues/2973

PR #166, opened 2016-06-14
Revert "use docker for ubuntu 16 jenkins builds"

brthor:
Reverts dotnet/core-setup#164

docker is not available on the Jenkins CI machines. 

netci.groovy changes are unfortunately only testable post-merge.

@piotrpMSFT


piotrpMSFT:
Chatted with @brthor offline. since we cannot test netci.groovy offline we only learned that there is a CI client configuration issue after merging. Reverting the PR to unblock builds. We are working on an alternative approach now.

PR #171, opened 2016-06-14
Include LICENSE and ThirdPartyNotices in artifacts

ellismg:
For each artifact, include the relvent LICENSE file in the root and a
ThirdPartyNotice.txt file if there's third party code that ships as part
of the product.


ellismg:
@mellinoe Can you take this the rest of the way.  There are a few open issues in my mind:

1. Understand if we need ThirdPartyNotices.txt for either the muxer or the hostfxr stuff.  We include Apache licensed code from Casablanca (see https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/json/casablanca/LICENSE.txt for the license there).
2. Understand if we need to include the host LICENSE file as an artifact that is dropped by the host MSI.  I believe that they should be included in the hostfxr and sharedframework MSIs because of how we run heat.

joshfree:
@richlander

mellinoe:
Let's continue this in #173 .

PR #173, opened 2016-06-15
[continued from #171] Include LICENSE and ThirdPartyNotices in artifacts

mellinoe:
This builds on #171.

@ellismg 's changes:

Include LICENSE and ThirdPartyNotices in artifacts
For each artifact, include the relvent LICENSE file in the root and a ThirdPartyNotice.txt file if there's third party code that ships as part of the product.

Additional Changes (@mellinoe)
Include ThirdPartyNotices for corehost components
* Include the license text for the C++ REST SDK, used by corehost
* Include a ThirdPartyNotices.txt file next to hostfxr.dll which includes the license text for the C++ REST SDK.
* Include the C++ REST SDK license text in the shared framework's ThirdPartyNotices.txt, because hostfxr and hostpolicy both use json parsing code from it, and they are included in the shared framework.

Still need to consider this:

> Understand if we need to include the host LICENSE file as an artifact that is dropped by the host MSI. I believe that they should be included in the hostfxr and sharedframework MSIs because of how we run heat.

@ellismg Could you clarify what you meant by that?

### DIFF
--- a/src/corehost/cli/json/casablanca/LICENSE.txt
+++ b/src/corehost/cli/json/casablanca/LICENSE.txt
@@ -1,1 +1,0 @@
-https://github.com/Microsoft/cpprestsdk
--- a/build_projects/dotnet-host-build/PackageTargets.cs
+++ b/build_projects/dotnet-host-build/PackageTargets.cs
@@ -69,6 +69,14 @@ namespace Microsoft.DotNet.Host.Build
             }
             FixPermissions(sharedHostRoot);
 
+            File.Copy(
+                Path.Combine(Dirs.RepoRoot, "resources", "ThirdPartyNotices.txt"),
+                Path.Combine(sharedHostRoot, "ThirdPartyNotices.txt"));
+
+            File.Copy(
+                Path.Combine(Dirs.RepoRoot, "resources", "LICENSE.txt"),
+                Path.Combine(sharedHostRoot, "LICENSE.txt"));
+
             c.BuildContext["SharedHostPublishRoot"] = sharedHostRoot;
             return c.Success();
         }
--- a/packaging/windows/host/host.wxs
+++ b/packaging/windows/host/host.wxs
@@ -39,6 +39,10 @@
             <Component Id="cmpCoreHost" Directory="DOTNETHOME" Guid="{45399BBB-DDA5-4386-A2E9-618FB3C54A18}">
                 <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe" />
             </Component>
+            <Component Id="cmpLicenseFiles" Directory="DOTNETHOME" Guid="{A61CBE5B-1282-4F29-90AD-63597AA2372E}">
+                <File Id="fileLicenseTxt" KeyPath="yes" Source="$(var.HostSrc)\LICENSE.txt" />
+                <File Id="fileThirdPartyNoticesTxt" Source="$(var.HostSrc)\ThirdPartyNotices.txt" />
+            </Component>
         </ComponentGroup>
     </Fragment>
 </Wix>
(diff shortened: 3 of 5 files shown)


ellismg:
> @ellismg Could you clarify what you meant by that?

When you install the MSIs, should LICENSE end up in the installed directory (e.g C:\Program Files\dontet) as well.  I think for everything except the host this will happen naturally (because we use heat.exe to construct the list of files to install and where to install them based on the layout we produce earlier) but the muxer MSI has an [explicit list of files to install](https://github.com/dotnet/core-setup/blob/master/packaging/windows/host/host.wxs#L38).

mellinoe:
Yes, good question. We do want the CLI EULA to be in the root directory (https://github.com/dotnet/cli/issues/3037), but folks could have the shared framework installed without the SDK.

@leecow What do we want to do here?

resources/third-party-notices/sharedframework/ThirdPartyNotices.txt
```diff
@@ -0,0 +1,184 @@
...
+zlib
+1995-2013 Jean-loup Gailly and Mark Adler
+zlib/libpng License
+This software is provided 'as-is', without any express or implied warranty. In no event will the authors be held liable for any damages arising from the use of this software.
+Permission is granted to anyone to use this software for any purpose, including commercial applications, and to alter it and redistribute it freely, subject to the following restrictions:
+1. The origin of this software must not be misrepresented; you must not claim that you wrote the original software. If you use this software in a product, an acknowledgment in the product documentation would be appreciated but is not required.
+2. Altered source versions must be plainly marked as such, and must not be misrepresented as being the original software.
+3. This notice may not be removed or altered from any source distribution.
+
+License notice for Libuv
```
joshfree: @Eilon please review

leecow:
@mellinoe - Can a shared - only install be to any arbitrary directory or will it always include /Microsoft.NETCore.App at some point in the path? If the latter then including the license at this level could cover the scenario.

resources/third-party-notices/sharedframework/ThirdPartyNotices.txt
Eilon: Looks like what we were told to put here, so this part is :shipit:

mellinoe:
I think we are eventually going to support installing shared framework's in arbitrary locations, but it doesn't currently work.

resources/third-party-notices/sharedframework/ThirdPartyNotices.txt
mellinoe: Thanks @Eilon . We did indeed get the text from here: https://github.com/aspnet/libuv-build/blob/dev/build/License.txt

Eilon: Actually @moozzyk to confirm that this is the guidance he received. He did all the *actual* work I think 😄

moozzyk: This is correct. I was told to replace thirdpartynotices.txt we had with this License.txt in our packages. Since packages produced by libuv-build and libuv-package contain only libuv I put the same license in all the package. Note that packages built in the libuv-build repo do not ship on NuGet. Only the package built by libuv-package end on NuGet so technically it is safer to take the License from there but as I said, I updated the License in both repos so it is the same.

src/corehost/cli/json/casablanca/LICENSE.txt
```diff
@@ -1 +1,20 @@
 https://github.com/Microsoft/cpprestsdk
+
```
joshfree: remove this file

joshfree:
as discussed in person, @mellinoe is going to revise this PR and then send an update for review

@richlander @leecow

mellinoe:
@jfree, @richlander , @leecow, @ellismg 

I've updated the PR with the new plan. LICENSE.txt and ThirdPartyLicenses.txt are now bundled in with the host, which installs to the root directory (`Program Files\dotnet`, etc.). It is therefore included in the shared framework installer and in the shared framework archives (zip/tar).

mellinoe:
@brthor Could you give the changes a look as well?

joshfree:
LGTM.  Reviewed in person with @mellinoe.

joshfree:
:shipit: when CI is green

brthor:
LGTM :shipit:

PR #178, opened 2016-06-15
Remove unused code and fix error message

schellap:
@brthor PTAL. Testing VSO definition as well with this change in master.

### DIFF
--- a/src/corehost/common/pal.h
+++ b/src/corehost/common/pal.h
@@ -184,8 +184,6 @@ namespace pal
     bool load_library(const char_t* path, dll_t* dll);
     proc_t get_symbol(dll_t library, const char* name);
     void unload_library(dll_t library);
-
-    bool find_coreclr(pal::string_t* recv);
 }
 
 #endif // PAL_H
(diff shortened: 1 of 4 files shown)


brthor:
Nice! We've got an issue here, @mellinoe made a fix in release/1.0.0 you can quickly make it in master as well

projects under `test` need to target Microsoft.NETCore.App `1.0.0` since that's what the sdk brings

projects under `TestAssets` need to target `1.1.0-*` since that's what we're building

mellinoe:
There is also a project under tools/dotnet-deb-tool which needs updating.

brthor:
good point @mellinoe , `dotnet-deb-tool` can reference `1.0.0` it has no need for the latest here.

schellap:
>  @mellinoe made a fix in release/1.0.0 
@mellinoe is there a commit you have?

brthor:
> @mellinoe is there a commit you have?

Even if you take the commit ensure TestAssets are targeted differently than they are on release/1.0.0

mellinoe:
It's this commit: https://github.com/dotnet/core-setup/commit/d7b37f34cbe759821db61cb724037187886094d6

schellap:
@dotnet-bot test this please

PR #179, opened 2016-06-16
Bump minor versions and update dependencies to stable

weshaggard:
Bump minor versions of things shipping in core-setup.

This updates dependencies to stable versions of corefx/coreclr until we start publishing higher versions of them.

It update any dependencies on NETCore.App to stable 1.0.0.

Update test projects to pull 1.0.1-beta-* to match the patch version
bump we have done in this branch.

cc @ericstj @eerhardt @brthor @schellap


brthor:
LGTM

eerhardt:
Lgtm

ericstj:
LGTM

test/HostActivationTests/project.json
```diff
@@ -1,16 +1,15 @@
 {
   "version": "1.0.0-*",
-
   "dependencies": {
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "1.0.0-rc3-004338"
+      "version": "1.0.0"
     },
```
schellap: Should be 1.0.1-beta-*

tools/dotnet-deb-tool/project.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "1.0.0-rc3-004308"
+      "version": "1.0.0"
     }
   },
```
schellap: Same as above, cc @brthor

schellap: Qn for @brthor

test/TestUtils/project.json
```diff
@@ -17,20 +17,17 @@
       ]
     }
   },
-
   "dependencies": {
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "1.0.0-rc3-4338"
+      "version": "1.0.0"
     },
```
schellap: 1.0.1-beta-*

test/HostActivationTests/project.json
weshaggard: This seems to be pulling from stage0 when it runs which is currently pulling the 1.0.0 stable version which is why I changed it to that. As far as I can tell this isn't running on the shared framework we are producing.

tools/dotnet-deb-tool/project.json
weshaggard: I believe this is also running on stage0 which is why it wasn't star versioned before as well.

test/TestUtils/project.json
weshaggard: I put this to stable 1.0.0 because of the others but I can attempt to make it consume the one being built but I suspect it also runs on stage0.

test/HostActivationTests/project.json
brthor: The test projects need to run against the sharedfx in the sdk, so `1.0.0` is right here. The `TestAssets` will run against the sharedfx produced, and there is xunit scenario coverage.

tools/dotnet-deb-tool/project.json
brthor: No need to run against latest here. Again this needs to run against the shared fx in the sdk.

test/TestUtils/project.json
brthor: The test projects need to run against the sharedfx in the sdk, so 1.0.0 is right here. The TestAssets will run against the sharedfx produced, and there is xunit scenario coverage.

There are multiple reasons here and we had a lot of back and forth when introducing the tests. This prevents the need to create a patched stage0 with the new sharedfx, leaving us in a chicken/egg problem if we introduce a change the sdk isn't yet compatible with. Furthermore, all tests are runnable via the Visual Studio test runner. If you target this to the latest built version, the Visual Studio test runner won't work because it will be running an older dotnet. 

There's no loss in test coverage as we have explicit TestApps in `TestAssets` that run against latest.

PR #180, opened 2016-06-16
Fix publish location for hostfxr in Azure

schellap:
@brthor PTAL.

Cc @joshfree Fix publish location for a host component in Azure.

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -343,7 +343,7 @@ namespace Microsoft.DotNet.Host.Build
         [Target]
         public static BuildTargetResult PublishHostFxrArchiveToAzure(BuildTargetContext c)
         {
-            var version = SharedFrameworkNugetVersion;
+            var version = HostFxrNugetVersion;
             var archiveFile = c.BuildContext.Get<string>("HostFxrCompressedFile");
 
             AzurePublisherTool.PublishArchive(archiveFile, Channel, version);


gkhanna79:
@schellap What is the bug for this that details the problem?

gkhanna79:
And should this be fixed in master as well?

schellap:
@brthor @eerhardt  what is the general policy on the URL format: https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/1.0.0/dotnet-hostfxr-win-x86.1.0.1.zip

Specifically,
preview/Binaries/1.0.0 --> Here does the 1.0.0 imply component version or the repo (release) version. It would be confusing to have such a download link like: 

https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/1.0.0/dotnet-hostfxr-win-x86.1.0.1.zip

schellap:
@gkhanna79 the bug and the master PR is linked in this PR.

gkhanna79:
Fixes https://github.com/dotnet/core-setup/issues/181

gkhanna79:
@schellap Why do we need to take this fix right now?

gkhanna79:
Does this break anything? @eerhardt Thoughts?

schellap:
@gkhanna79, we can fix this in the blob storage as well for RTM, so no this is not a blocker. It should be fixed in the servicing branch definitely for future packages.

@eerhardt, If we aligned on repo/release version for the container key then, we need to fix few other places to use the right version in the container name.

brthor:
@schellap Nobody is consuming this that I know of, so this should be okay (not breaking) as is.

gkhanna79:
Closing per discussion in https://github.com/dotnet/core-setup/issues/181

PR #188, opened 2016-06-18
Publish zip only for Windows used by Antares

schellap:
Be a good citizen about Azure blob storage. Antares only needs Windows.

Cc @brthor


schellap:
@brthor don't bother with this PR. It is better to have these archived.

PR #192, opened 2016-06-23
Ensure native artifacts have a build-id in them

ellismg:
These build-ids are helpful for collating artifacts back to builds, especially from dumps.

### DIFF
--- a/src/corehost/cli/setup.cmake
+++ b/src/corehost/cli/setup.cmake
@@ -60,6 +60,8 @@ endif()
 # containing the reference instead of using definitions from other modules.
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Xlinker -Bsymbolic -Bsymbolic-functions")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id=sha1")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--build-id=sha1")
     add_compile_options(-fstack-protector-strong)
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     add_compile_options(-fstack-protector)


src/corehost/CMakeLists.txt
```diff
@@ -1,2 +1,8 @@
 cmake_minimum_required (VERSION 2.6)
+
+if (CMAKE_SYSTEM_NAME STREQUAL Linux)
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id=sha1")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--build-id=sha1")
+endif ()
+
```
schellap: NIT: add to `setup.cmake` where these linker flags are defined.

ellismg: Fixed, please take another look.

src/corehost/cli/setup.cmake
```diff
@@ -60,6 +60,8 @@ endif()
 # containing the reference instead of using definitions from other modules.
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Xlinker -Bsymbolic -Bsymbolic-functions")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id=sha1")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--build-id=sha1")
     add_compile_options(-fstack-protector-strong)
```
schellap: @ellismg LGTM. For my education, the BuildID[SHA1] will be stored somewhere for us to know which one it is?

Actually, if you crack open the ELF binary and look at the strings, search for "commit" (if you use :%!xxd in vim), you can get the SHA1 etc.

ellismg: `file` will print the build id.  It's is also in a note section with the name .note.gnu.build-id.  `readelf -n` can pull it up.

```

matell@matell-ub1604:~/git/core-setup/artifacts/ubuntu.16.04-x64/corehost$ file dotnet
dotnet: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=21196e6acf7429e72b193aac8a9cba8b98761ab1, not stripped
matell@matell-ub1604:~/git/core-setup/artifacts/ubuntu.16.04-x64/corehost$ readelf -n dotnet

Displaying notes found at file offset 0x0000021c with length 0x00000020:
  Owner                 Data size       Description
  GNU                  0x00000010       NT_GNU_ABI_TAG (ABI version tag)
    OS: Linux, ABI: 2.6.32

Displaying notes found at file offset 0x0000023c with length 0x00000024:
  Owner                 Data size       Description
  GNU                  0x00000014       NT_GNU_BUILD_ID (unique build ID bitstring)
    Build ID: 21196e6acf7429e72b193aac8a9cba8b98761ab1
```

We've found this is more likely to be present in coredumps and it also differs per distro, which will be helpful for some other stuff we want to build.

ellismg:
The Windows builds are failing because https://dotnetcli.blob.core.windows.net/dotnet/Sdk/preview/latest.version is not present on blob storage?

eerhardt:
There's two issues here.

1. Windows gets the dotnet-install script from the CLI repo and UNIX gets the dotnet-install script from inside the core-setup repo. They need to be consistent. (The Windows approach is the correct way. Core-setup shouldn't be duplicating the script.)
2. I updated the CLI rel/1.0.0 publish layout and it no longer has the "preview" channel.  When core-setup invokes the dotnet-install script, it should be passing "rel-1.0.0" as the channel. Or maybe not passing a channel and just using the default channel in the script.

schellap:
@dotnet-bot test this please

mellinoe:
@ellismg @schellap Do we still want this to go in? Latest build didn't have any problems.

gkhanna79:
LGTM.

schellap:
Yes, IIRC, the latest build is fine because I believe I fixed the scripts many months ago. Hence the test this please :)

ellismg:
I would still like it to go in.

Issue #194, opened 2016-06-25
Host is not adding "project" dependencies to TPA


schellap:
## Steps to reproduce

`dotnet new`
Edit project.json and remove type: platform to make it standalone
```
dotnet restore --infer-runtimes
dotnet publish
cd bin\Debug\netcoreapp1.0\win7-x64\publish
```

Replace "package" to "project" for the CoreCLR runtime reference in "libraries" section of `app.deps.json`

```
    "runtime.win7-x64.Microsoft.NETCore.Runtime.CoreCLR/1.0.2": {
      "type": "project",
      "serviceable": true,
      ...
    },
```

## Expected  behavior
```
temp\bin\Debug\netcoreapp1.0\win7-x64\publish>temp.exe
Hello World!
```

## Actual behavior
```
temp\bin\Debug\netcoreapp1.0\win7-x64\publish>temp.exe
Could not resolve CoreCLR path. For more details, enable tracing by setting COREHOST_TRACE environment variable to 1
```

## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (1.0.0-preview3-003161)

Product Information:
 Version:            1.0.0-preview3-003161
 Commit SHA-1 hash:  63846c5751

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

## Notes
* This works in 1.0.0 because of #193 
* This bug could cause problems if someone triggered a not yet loaded TPA assembly in a custom load context, but ignored the load override.
* The bug is due to this if-check which should be removed.
  * https://github.com/dotnet/core-setup/blob/b25d39a8521942c70a91ca04210434ab626c292a/src/corehost/cli/deps_format.cpp#L38
  * https://github.com/dotnet/core-setup/blob/b25d39a8521942c70a91ca04210434ab626c292a/src/corehost/cli/deps_format.cpp#L71

Cc @gkhanna79 @brthor @anurse

PR #195, opened 2016-06-25
Allow project to project dependencies in TPA

schellap:
Provides better error message

### DIFF
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -35,11 +35,6 @@ void deps_json_t::reconcile_libraries_with_targets(
     {
         trace::info(_X("Reconciling library %s"), library.first.c_str());
 
-        if (pal::to_lower(library.second.at(_X("type")).as_string()) != _X("package"))
-        {
-            trace::info(_X("Library %s is not a package"), library.first.c_str());
-            continue;
-        }
         if (!library_exists_fn(library.first))
         {
             trace::info(_X("Library %s does not exist"), library.first.c_str());
@@ -68,7 +63,7 @@ void deps_json_t::reconcile_libraries_with_targets(
                 size_t pos = library.first.find(_X("/"));
                 entry.library_name = library.first.substr(0, pos);
                 entry.library_version = library.first.substr(pos + 1);
-                entry.library_type = _X("package");
+                entry.library_type = pal::to_lower(library.second.at(_X("type")).as_string());
                 entry.library_hash = hash;
                 entry.asset_name = asset_name;
                 entry.asset_type = (deps_entry_t::asset_types) i;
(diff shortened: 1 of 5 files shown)


schellap:
Cc @anurse @gkhanna79 @brthor

schellap:
Fixes #194

src/corehost/cli/deps_format.cpp
```diff
@@ -68,7 +63,7 @@ void deps_json_t::reconcile_libraries_with_targets(
                 size_t pos = library.first.find(_X("/"));
                 entry.library_name = library.first.substr(0, pos);
                 entry.library_version = library.first.substr(pos + 1);
-                entry.library_type = _X("package");
+                entry.library_type = pal::to_lower(library.second.at(_X("type")).as_string());
```
gkhanna79: What kind of "types" exist aside from "package" and "project"?

schellap: @anurse, can you or @brthor help answer this to assess impact?

@gkhanna79, the host doesn't have to do a full probe unless it is type package. That's an optimization.

gkhanna79: > the host doesn't have to do a full probe unless it is type package

I am not sure I follow - can you please elaborate?

schellap: For example, "project" assemblies will not be present in "packages" cache or in the FX base, so these probes are useless. We can directly go to app base. The host doesn't do this optimization today. I'm trying to say the host shouldn't otherwise care about the type in the libraries section.

gkhanna79: > We can directly go to app base

This should be done by design, implying that project assemblies should figure out in app.deps.json, similar to explicitly pulled in nuget dependencies and thus, from activation standpoint, there is no difference in resolving these vs explicit dependencies.

schellap: Yes, the "by design" piece is what is happening today and with the fix. Except, in the bug case, the host ignores "project" dependencies in the app.deps.json when adding to TPA. It never broke anyone because of the APP_PATHS setting as the runtime found these in the app base.

I don't agree that there is no difference -- The type "project" tells that they are supposed to be found in the app base, not in the "packages" caches or FX for example, like I said this is an optimization and may even be correctness for the degenerate case of someone putting a project package in the cache.

anurse: As far as I know there is only `project` and `package` for this type field. The only other "type" we have for this kind of entry is a framework reference assembly which only applies to Desktop CLR

src/corehost/corehost.cpp
```diff
@@ -80,7 +80,9 @@ int run(const int argc, const pal::char_t* argv[])
     }
     if (!pal::load_library(fxr_path.c_str(), &fxr))
     {
-        trace::info(_X("Load library of %s failed"), fxr_path.c_str());
+        trace::error(_X("The library %s was found, but loading it from %s failed"), LIBFXR_NAME, fxr_path.c_str());
+        trace::error(_X(" - Installing .NET Core prerequisites might help resolve this problem."));
+        trace::error(_X("     %s"), DOTNET_CORE_URL);
```
anurse: :+1: to more descriptive error messages!

anurse:
LGTM. If I read right all this does is properly read the type field, it doesn't yet change behavior based on that field, right?

schellap:
@eerhardt PTAL the last commit.

eerhardt:
> @eerhardt PTAL the last commit.

I explained the issues and real fixes here:  https://github.com/dotnet/core-setup/pull/192#issuecomment-228390467

Issue #193, opened 2016-06-24
Host should not pass App_Paths and App_Ni_Paths


gkhanna79:
In RC2, the host used to load everything from the app-local unconditionally, which we fixed to load only the assemblies specified in app.deps.json.

However, the host continues to pass app_paths and app_ni_paths to the runtime (https://github.com/dotnet/core-setup/blob/22c97b59b6442eb3454c9a1407b6de91ca11eaa2/src/corehost/cli/hostpolicy.cpp#L70). This means if the application carried additional assemblies, which are not in app.deps.json, then they can get loaded in DefaultContext since the paths are specified while, infact, managed code should be using various Load* methods of AssemblyLoadContext to load such assemblies since only the TPA list corresponds to the "probe" locations for DefaultContext.

CC @jkotas

schellap:
**Note** when fixing this issue, provide adequate compat switches...

PR #197, opened 2016-06-30
Do not set APP_PATHS and APP_NI_PATHS by default

schellap:


schellap:
Instead allow a property to maintain the compatible behavior.

```
"runtimeOptions": {
    "configProperties": {
        "Microsoft.NETCore.DotNetHostPolicy.SetAppPaths": true
    }
}
```

Once this passes PR update the document here: https://github.com/borgdylan/dotnet-cli/blob/master/Documentation/specs/runtime-configuration-file.md#sections

schellap:
Fixes #193

schellap:
Cc @gkhanna79

schellap:
PTAL

src/corehost/cli/hostpolicy.cpp
```diff
@@ -116,12 +110,30 @@ int run(const arguments_t& args)
         property_values.push_back(clrjit_path_cstr.data());
     }
 
+    bool set_app_paths = false;
+
+    // Runtime options config properties.
     for (int i = 0; i < g_init.cfg_keys.size(); ++i)
     {
+        // Provide opt-in compatible behavior by using the switch to set APP_PATHS
```
gkhanna79: It will be good to run scenarios like MusicStore to ensure that nothing is broken. What we want to confirm if that no one relies on this (if they do, they have a workaround), so that we do not need to have this opt-in.

I would expect that the default scenarios enabled by .NET Core do not enable this broken behaviour - is that correct?

schellap: I have tested against MusicStore. The app_paths won't be set by default per this fix. The opt-in provided is `$.runtimeOptions.configProperties.Microsoft.NETCore.DotNetHostPolicy.SetAppPaths`

Correct, AFAIK,

1. We shouldn't have an issue here -- simply because `dotnet build` and `dotnet run` cycle doesn't rely on the `APP_PATHS` and fetches everything from NuGet cache.
2. Native assets already are not on `APP_PATHS` for portable apps, so the bug case -- setting this doesn't help.

PR #199, opened 2016-07-01
[release/1.0.0] Uwp 5.2.1

ericstj:
Updating meta-packages for servicing.

I'm struggling with getting core-setup back on pre-release versioning and making the build process happy with that.  I'd appreciate some help & review @schellap @brthor 

/cc @eerhardt @weshaggard


ericstj:
Tests are still failing because they didn't like my `*` version.  Not sure what the right thing is to do here.

test/HostActivationTests/project.json
```diff
@@ -3,7 +3,7 @@
   "dependencies": {
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "1.0.0"
+      "version": "1.0.0-*"
```
eerhardt: It is a bit tricky to know this without having the background, but the way this works is:

The actual tests themselves use the same version as the "stage0" CLI that gets installed from the build.ps1.  This version needs to stay in sync with the CLI that gets installed.  Thus, this should still be `"version": "1.0.0"`.

Same goes for the `test/TestUtils/project.json` below.  That should be "1.0.0".

The "TestAssets\TestProjects" projects/apps are testing the newly built Shared Framework from within the repo.  So up above in those projects, you have it correct, those should all be "1.0.0-*".

TestAssets/TestProjects/PortableApp/project.json
```diff
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.0.0"
+          "version": "1.0.0-*"
```
eerhardt: Try putting `"version": "1.0.0-servicing-*"` for the version.  The test failure makes it look like it is binding against the "1.0.0" version, and you want to break that.

schellap: @eerhardt is right, but you also need the `1.0.0-servicing-*` to end up in the `project.lock.json`. But because you are disabling the building of those packages by setting the `BuildAllPackages` to `false` the NuGet/MyGet feed wins instead of the local just built package.

I would suggest that we produce all the packages, but prevent publishing them, rather than stop producing them. Or it might be better to create a staging feed/blob storage for testing purposes and then move them to the main feeds.

eerhardt: I agree, we should still be doing full builds, and just publishing to a separate location.  We can even use the same storage account, and just switch the blob container.

https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs#L21

ericstj:
I've been hacking a bit more and still can't get this to work.  In addition to the change I just committed I tried to renable build of packages that we weren't shipping and build those with the servicing moniker but it still didn't fix it.  @schellap do you think you can help out here?

schellap:
@ericstj, schellap/UWP-5.2.1 takes it further. The NETCore.App package is produced correctly, but the project.lock.json still ends up picking `1.0.0`.

schellap:
https://github.com/schellap/core-setup/commit/d4e9a495c4b27251f23477eee44a510fc4a268a8

schellap:
Can someone help me understand if NuGet is doing the expected thing here? Why does it choose `1.0.0` when we specify `1.0.0-servicing-*` and it is available. If I removed NuGet.org from source, it picks up the 1.0.0-servicing.

Command used to restore:
```
E:\clr\git\temp\core-setup\.dotnet_stage0\Windows\x64\dotnet.exe "restore" "--source" "E:\clr\git\temp\core-setup\artifacts\win10-x64\corehost" "--source" "E:\clr\git\temp\core-setup\artifacts\win10-x64\corehostdummypackages" --source https://api.nuget.org/v3/index.json
```

project.json
```
{
 
  "buildOptions": {
    "emitEntryPoint": true
  },
  "dependencies": {},
  "frameworks": {
    "netcoreapp1.0": {
      "dependencies": {
        "Microsoft.NETCore.App": {
          "type": "platform",
          "version": "1.0.0-servicing-*"
        },
        "Newtonsoft.Json": "9.0.1-beta1"
      }
    }
  }
}
```

project.lock.json (with NuGet.org feed)
```
 122       "Microsoft.NETCore.App/1.0.0": {
 123         "type": "package",
 124         "dependencies": {
 125           "Libuv": "1.9.0",
```

project.lock.json (without NuGet.org feed)
```
 122       "Microsoft.NETCore.App/1.0.0-servicing-004504-00": {
 123         "type": "package",
 124         "dependencies": {
 125           "Libuv": "1.9.0",
```

eerhardt:
@emgarten @rrelyea - can you guys answer @schellap's question above on why "1.0.0-servicing-*" is getting resolved to "1.0.0"?

ericstj:
I’m guessing the NuGet algo is to evaluate * on each feed, then take the max of all.  When NuGet evaluates -servicing-* for nuget.org it doesn’t fine anything, then nuget’s “closest” match kicks in and chooses the stable version.  If that is indeed what’s happening I would imagine that “max” algo would need to first prefer the max of pattern matches, then fall back to match of non-pattern matches.

@schellap, can you see if changing the order has any effect? (eg: ensuring nuget.org is last)

ericstj:
@schellap do we need Nuget.org in the source list?  If not let's leave it out for the tests.  That seems like the right thing to do anyhow.

schellap:
@ericstj, please cherry-pick https://github.com/schellap/core-setup/commit/85d9e9285ac9ae0b3976493e683817dce473129d
Made some changes to workaround the issues. Removed a bunch of packages from StableVersion file.

build_projects/dotnet-host-build/TestTargets.cs
```diff
@@ -49,7 +49,9 @@ public static BuildTargetResult RestoreTests(BuildTargetContext c)
             var dotnet = DotNetCli.Stage0;
             CleanBinObj(c, Path.Combine(Dirs.RepoRoot, "test"));
 
-            dotnet.Restore()
+            dotnet.Restore(
+                    "--fallbacksource", Dirs.CorehostLocalPackages,
```
brthor: This should be unnecessary as the test projects themselves are not intended to target the latest generated host.

test/HostActivationTests/project.json
brthor: @eerhardt Has it exactly right here, the projects under `test` target the sharedframework in the downloaded sdk, which is why they target `1.0.0`

The projects under `TestAssets` target the latest built host/sharedfx packages and provide the test coverage. 

This split is necessary to retain the ability to run the projects under `test` outside of the build. For example, you can cd into `test` and use `dotnet test` or even the .net core Visual Studio Test runner, and get the same results as if you were running the tests in the build.

test/TestUtils/project.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "1.0.0"
+      "version": "1.0.0-*"
```
brthor: This should also target `1.0.0`

TestAssets/TestProjects/PortableApp/project.json
```diff
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.0.0"
+          "version": "1.0.0-servicing-004500"
```
brthor: I don't know what you had previously here but I would expect these to target `1.0.0-servicing-*`

ericstj:
@schellap can you have a look at @brthor's comments?

TestAssets/TestProjects/PortableApp/project.json
ericstj: He had `-servicing-*` which was causing NuGet to fetch the 1.0.0 version from nuget.org

schellap: @brthor, from a host testing standpoint, any app should be fine. We will apply roll-forward for portable apps between pre-release to pre-release (but not to production), so this PJ doesn't need to be updated. The reason it would fail if you had `1.0.0` was host roll-forward doesn't go from pre-release to production by design for portable apps. If you had `1.0.0-servicing-*` NuGet always resolves to `1.0.0` hitting the same roll-forward issue.

If you are worried about testing of the just-built `Microsoft.NETCore.App` as well -- it is broken for standalone apps as NuGet resolves it from the feed with `1.0.0` and not `1.0.0-servicing-*`.

schellap: @ericstj, @brthor is right that once this commit goes to the feed, NuGet will always resolve `1.0.0-servicing-004500` against this commit, effectively rendering the tests useless.

brthor: > If you are worried about testing of the just-built Microsoft.NETCore.App as well -- it is broken for standalone apps as NuGet resolves it from the feed with 1.0.0 and not 1.0.0-servicing-*.

I am worried about testing this for the standalone case. Unless the servicing fix goes all the way upstack (is that the plan?), this will be the only coverage. 

Does leaving it at `1.0.0-servicing-004500` resolve to the proper package for standalone?

schellap: @ericstj, I think the only other option to workaround the NuGet issue is to not downgrade the versioning for servicing --- i.e., servicing is `1.0.101`, `1.0.102` and marking master (which is currently `1.0.1`) as `1.1.0` to not conflict with servicing.

brthor: @schellap :+1: 

If we can avoid a version downgrade that will make things simpler, especially if we need to take this change (or further servicing fixes) upstack.

ericstj: @schellap if that is how you'd like to handle servicing for the host packages it is fine by me.  Would you like to submit a separate PR to make host "servicing ready" in release branch then I can just commit after that?

schellap: @ericstj it is not about host servicing. It is NETCore.App version that is the problem. Host has its own versioning scheme. @weshaggard does the above sound reasonable to you? We have master 1.0.1, which causes downgrade of rel to 1.0.0-servicing.

schellap: I'll get a PR out.

ericstj:
Closing in favor of #208

PR #201, opened 2016-07-05
Correct ".NET SDK" to ".NET Core SDK" in help message

hickford:


src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -581,7 +581,7 @@ int muxer_usage()
     trace::println();
     trace::println(_X("If you are debugging the Shared Framework Host, set 'COREHOST_TRACE' to '1' in your environment."));
     trace::println();
-    trace::println(_X("To get started on developing applications for .NET Core, install .NET SDK from:"));
+    trace::println(_X("To get started on developing applications for .NET Core, install .NET Core SDK from:"));
```
hickford: The name ".NET Core" is confusing enough without the app itself getting it mixed up

gkhanna79: Perhaps "..install the SDK from:" ?

hickford: Good idea

PR #202, opened 2016-07-05
Fix muxer help to promote sdk when absent

schellap:


schellap:
Cc @gkhanna79

schellap:
Multiple inputs about the message being confusing to users.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -580,9 +587,13 @@ int muxer_usage()
...
     trace::println();
     trace::println(_X("If you are debugging the Shared Framework Host, set 'COREHOST_TRACE' to '1' in your environment."));
-    trace::println();
-    trace::println(_X("To get started on developing applications for .NET Core, install .NET SDK from:"));
-    trace::println(_X("  %s"), DOTNET_CORE_URL);
+
+    if (include_sdk_link)
+    {
+        trace::println();
+        trace::println(_X("To get started on developing applications for .NET Core, install .NET SDK from:"));
```
gkhanna79: I would suggest to rephrase this as "...install the SDK from: "

gkhanna79:
@piotrpMSFT should chime in on this.

src/corehost/cli/fxr/fx_muxer.cpp
schellap: See https://github.com/dotnet/core-setup/pull/201/files

Cc @hickford

piotrpMSFT:
If the SDK is present we should guide folks towards using it. Perhaps recommending 'dotnet --help'?

/Cc @blackdwarf

blackdwarf:
If we have SDK on the machine, I would say that the message above should not appear, that is, something else should be said. For instance, if the SDK is present:
> Use dotnet --help to get help with the SDK

If SDK is not installed/present:

> To get started on developing applications for .NET Core, install .NET SDK from: https://aka.ms/core-sdk

In this way, the user doesn't get confused by the message since we are aware of the context.

hickford:
Good idea. I was confused why the app nagged me to install the SDK when I'd just installed it.

PR #203, opened 2016-07-07
Increment versions post servicing

ericstj:
This bumps the versions of the packages we intend to ship out of the
servicing branch so that the master will be later than release.

/cc @weshaggard @schellap


schellap:
I think we should put a pre-release tag on master... to indicate bits that are not production releases.

ericstj:
They are all pre-release.  That happens automatically.

ericstj:
Version is just the version base.  We determine if it needs the build's pre-release tag based on the stable version table.
StableVersion means that its the absolute version, and we should bypass lookup in the stable version table.

schellap:
Cool, thanks. LGTM.

PR #205, opened 2016-07-09
Fix build for Alpine Linux

jasonwilliams200OK:
Alpine Linux uses musl-libc and Android uses bionic; what both have in
common is `__GLIBC__` unavailable. This fixes the condition in
Casablanca header where an explicit `__ANDROID__` is used. Instead
checking for `__GLIBC__` fixes the build for both platforms.

Another fix is to include `<sys/time.h>` header which is required for
`timeval` on some Unix platforms.

Progress towards: https://github.com/dotnet/coreclr/issues/917#issuecomment-229827373

/cc @schellap, @ericstj, @jkotas

### DIFF
--- a/src/corehost/cli/json/casablanca/include/cpprest/asyncrt_utils.h
+++ b/src/corehost/cli/json/casablanca/include/cpprest/asyncrt_utils.h
@@ -40,8 +40,9 @@
 
 #ifndef _WIN32
 //#include <boost/algorithm/string.hpp>
-#if !defined(ANDROID) && !defined(__ANDROID__) // CodePlex 269
+#ifdef __GLIBC__
 #include <xlocale.h>
+#include <sys/time.h>
 #endif
 #endif
 


jasonwilliams200OK:
With this fix, I have able to take first step, build muxer binary (with libhostpolicy.so and libhostfxr.so) on Alpine Linux.

```sh
git clone https://github.com/dotnet/core-setup

mkdir build && cd build

../core-setup/src/corehost/build.sh --arch x64 --hostver 1.0.2-beta-000559-00 \
--fxrver 1.0.2-beta-000559-00 --policyver 1.0.2-beta-000559-00 \
--rid alpine.3.4.0-x64 --commithash 3c8ad46cb09d1ac87159b2cc17cb410afb835c9b
```

Now time to fiddle with `dotnet-ubuntu-x64.latest.tar.gz`, to get `Microsoft.NETCore.App` working. :)

PR #206, opened 2016-07-12
Bump master Microsoft.NETCore.App to use 1.1.0 instead of 1.0.1

schellap:
@weshaggard or @ericstj  PTAL.

Cc @brthor @eerhardt

### DIFF
--- a/TestAssets/TestProjects/PortableApp/project.json
+++ b/TestAssets/TestProjects/PortableApp/project.json
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.0.1-beta-*"
+          "version": "1.1.0-beta-*"
         },
         "Newtonsoft.Json": "9.0.1-beta1"
       }
--- a/TestAssets/TestProjects/PortableTestApp/project.json
+++ b/TestAssets/TestProjects/PortableTestApp/project.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.0.1-beta-*"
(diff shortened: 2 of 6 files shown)


schellap:
This is so that `release/1.0.0` can use patch upgrades that don't conflict with `master` branch.

eerhardt:
Do we have a "master plan" here?  When I talked to @weshaggard about this earlier, he said all of the "patch" versions were getting bumped in 'master' of corefx, and so that would be a good plan for the packages in core-setup.  Did this plan change?

ericstj:
That is what happened in core-setup, however @schellap is dealing with issues in the core-setup branch with rebuilding the stable packages in release/1.0.0 for NetCore.App.

Our master plan is summarized as:
1. In master bump patch version of all packages.
2. In release continue to build the release version of packages, disabling any packages which we don't intend to publish/ship.
3. When we determine that we need to publish a new servicing release of a package, bump the version in master to get out of the way of release then bump the package in release.
4. Release can build multiple times with `servicing` prerelease string, then move the serviced package to stable (alternatively if the change is small it can move directly to stable).

Ideally, I'd like to see us follow this same pattern in core-setup but the build and test infrastructure for NetCore.App was getting in the way...

brthor:
@ericstj 

Is the plan you're outlining consistent with what @schellap is currently doing? Is the delta that we're bumping master to `1.1.0` instead of `1.0.2` ?

ericstj:
@brthor yeah, that's the delta.  It's likely a safe thing to do since I expect that any new release of NETCore.App out of master will have new API (since new API already exists in corefx) and thus a minor version bump is appropriate.

ericstj:
LGTM

weshaggard:
LGTM

PR #208, opened 2016-07-12
Get rel ready for servicing

schellap:
@ericstj this PR would take care of #199, can you take a closer look at changes from your side?

@brthor @eerhardt PTAL the rest of the changes.

@weshaggard PTAL the NETCore.App side of things.

Depends on #206


pkg/stable.packages.props
```diff
@@ -2,16 +2,16 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
     <!-- NETCore 1.0 core-setup -->
-    <StablePackage Include="Microsoft.NETCore.App">
+    <!--<StablePackage Include="Microsoft.NETCore.App">
       <Version>1.0.0</Version>
-    </StablePackage>
+    </StablePackage>-->
     <StablePackage Include="Microsoft.NETCore">
```
eerhardt: Do these still need to be here?  Why are these packages in twice?  Since they use the same Include, doesn't the bottom ones just overwrite these?

branchinfo.txt
```diff
@@ -1,9 +1,15 @@
 # This is a file containing environment variables specific to this branch
 # Any line that is not blank and does not start with '#' is interpreted as a variable to set
 # Each line is expected to be in the format "[Name]=[Value]".
+#
+#
+# !!! WARNING !!! WARNING !!! WARNING !!! WARNING !!! WARNING !!! WARNING
+#       "master" branch temporarily produced 1.0.1 builds. The value here
```
eerhardt: What's the plan here going forward?  What if we need to service the Microsoft.NETCore.App package?  What version will we use?

eerhardt:
Looks good to me, other than my two questions.

branchinfo.txt
```diff
@@ -1,9 +1,15 @@
...
+#       should be > 1.0.1
```
brthor: should this be 1.0.2?

brthor:
LGTM

ericstj:
LGTM

branchinfo.txt
schellap: When we service netcore.app I would imagine you'd service from the rel branch rev-ing patch versions.

When you ship master I'd expect it would be 1.1.0 and not a servicing train. Cc @weshaggard.

ericstj:
This looks good to merge.  @schellap do you want to push the button?

PR #212, opened 2016-07-14
Split dotnet and create apphost

schellap:


schellap:
Cc @gkhanna79, I know comments are missing, which I will address. PTAEL (early look)

schellap:
@gkhanna79 It might still be a good idea to review this...

PR #213, opened 2016-07-15
Fix self-process lookup for FreeBSD.

josteink:
Currently dotnet cli stage0 builds for FreeBSD, but does not always work, as it relies on procfs, which is not available on FreeBSD by default.

This patch replaces the current implementation with one based on sysctl(), which ensures it should always resovlve, procfs or no procfs.

### DIFF
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -183,6 +183,24 @@ bool pal::get_own_executable_path(pal::string_t* recv)
     }
     return false;
 }
+#elif defined(__FreeBSD__)
+bool pal::get_own_executable_path(pal::string_t* recv)
+{
+    int mib[4];
+    mib[0] = CTL_KERN;
+    mib[1] = KERN_PROC;
+    mib[2] = KERN_PROC_PATHNAME;
+    mib[3] = -1;
+    char buf[PATH_MAX];
+    size_t cb = sizeof(buf);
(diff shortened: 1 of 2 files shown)


josteink:
cc: @ericstj @schellap

src/corehost/common/pal.unix.cpp
```diff
@@ -183,6 +183,20 @@ bool pal::get_own_executable_path(pal::string_t* recv)
...
 }
+#elif defined(__FreeBSD__)
+bool pal::get_own_executable_path(pal::string_t* recv)
+{
+    int mib[4];
+    mib[0] = CTL_KERN;
+    mib[1] = KERN_PROC;
+    mib[2] = KERN_PROC_PATHNAME;
+    mib[3] = -1;
+    char buf[1024];
```
schellap: We should not ignore the error code from `sysctl`. Handle `ENOMEM` as well?

Some nits:
a `static const int mib[] = { };` 
Use `_countof` or `__arraycount` macros
`PATH_MAX` instead of 1024

josteink:
Sounds reasonable. Just to be absolutely clear about what you consider the right approach...

When you say we shouldn't ignore return codes, you mean only assign and return `true` when we have a good return code, otherwise return `false`.

As for handling `ENOMEM`... Are you thinking about anything *extra* besides aborting and returning `false`?

schellap:
LGTM. Addressing long paths is tracked by: https://github.com/dotnet/cli/issues/1695

We can merge when green.

PR #218, opened 2016-07-17
Allow for backward compatible interface

schellap:
@gkhanna79 PTAL. This doesn't functionally change anything, but sets the right precedence so a future me or a future developer doesn't stumble.


schellap:
Closing because PR #212 is reverted by #219

PR #219, opened 2016-07-18
Revert "Merge pull request #212 from schellap/sig-host"

schellap:
This change reverts PR #212 because:

We need to figure out the following.

1. Unblock rest of the world that is picking core-setup master bits from our apt dev feed (replacing their shipped stable binaries)
2. Figure out why a new dotnet will ever pick up an old hostfxr, i.e., why the update mechanism did not pick up new bits of the hostfxr.dll.
3. Why our master branch beta binaries are pushed to release feeds?
4. The intentional compat break was necessary so a future signed dotnet did not load all old hostfxrs.
  * Figure out if there is a compatible manner to address this.

Cc @gkhanna79 @blackdwarf @joshfree @brthor @piotrpMSFT


schellap:
The following builds of `dotnet` and `hostfxr` have this issue: `1.0.2-beta-000583-00`, `1.0.2-beta-000582-00`, `1.0.2-beta-000581-00`

schellap:
https://github.com/dotnet/core-setup/commits/master

Commit count: 583, 582, 581 since bad commit: cc564bd

gkhanna79:
@blackdwarf @brthor Can you please generate DEB packages to confirm the issue is addressed?

CC @joshfree

brthor:
Deb packages will be generated by CI. I will check out the feed after they have published.

brthor:
Confirmed this issue is addressed with the latest host package: 1.0.2-beta-000585-00

schellap:
@brthor, it would be good to understand 2 and 3 above.

brthor:
@schellap 

For 2: This is a relic of the way relationships between our packages are expressed. It is immediately exacerbated on debian because of the feed structure and apt's natural behavior. 

We express that a hostfxr package is dependent on a ">=" version of the host package present. This relationship is necessary in order to avoid conflicts between different versions of the package chain and support side by side. (i.e. we cannot simply make it an "==" relationship).

Apt, by default, takes the highest version of the host package present, that is compatible with the current package set. So installing preview2 sdk will take the preview2 sdk package, 1.0.0 sharedfx package, 1.0.0 hostfxr package, and the highest available host version. 

As long as we can expect backwards compatibility, this relationship makes sense. If that's not the case we need to merge the host/hostfxr packages again to prevent the issue of them being out of sync, or break Side By Side installation by creating an "==" relationship between the hostfxr and host package. 

For 3:
This has been remedied. The feed available on `dot.net` was changed yesterday.

schellap:
@brthor, for my understanding... Is sharedfx and hostfxr dependency relation == for sxs uninstalls of hostfxr to work? If this relation was >= then apt would pull in a new hostfxr, right, by similar argument? Because we do guarantee that all future hostfxr can load all earlier sharedfx.

brthor:
@schellap Because hostfxr is a SxS component, each hostfxr version has an independent package name, and isn't just bumping a version. The shared framework depends on the package name which includes the version in the title, so it is a "==" relationship. 

Making it a ">=" relationship would remove the ability to install hostfxr components SxS in deb.

PR #224, opened 2016-07-22
Add "restore_tools" that gets a new NuGet version for a workaround

dagood:
To work around https://github.com/NuGet/Home/issues/3190 the recommended way, core-setup needs a new version of NuGet to use `--disable-parallel` to limit HTTP parallelism during `build_projects` restore.

There is a chance that restoring restore_projects will time out just like build_projects does, but there are fewer dependencies so it's less likely.

### DIFF
--- a/build_projects/dotnet-host-build/build.sh
+++ b/build_projects/dotnet-host-build/build.sh
@@ -110,11 +110,19 @@ fi
 # Disable first run since we want to control all package sources
 export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
 
+# Restore the restore tool
+echo "Restoring Restore projects..."
+(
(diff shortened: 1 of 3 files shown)


dagood:
Errors here are due to case sensitivity incompat between restore in the NuGet version I pull in and build in CLI stage_0. Since the [case insensitivity change](https://github.com/NuGet/NuGet.Client/pull/557) was merged before the change that [adds the workaround into NuGet](https://github.com/NuGet/NuGet.Client/pull/690), there's no compatible NuGet version. Barring another layer of workarounds, this isn't going to work with what's out there.

There is a CLI PR out to support the new casing: https://github.com/dotnet/cli/pull/2722. That would make this PR work, but since that PR also upgrades NuGet, it makes this workaround project redundant. Closing.

PR #227, opened 2016-07-26
debian8 deb packages

brthor:
Enables debian packages for our debian 8 builds.

cc @MichaelSimons @naamunds @blackdwarf

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -232,7 +232,7 @@ namespace Microsoft.DotNet.Host.Build
             nameof(PublishTargets.PublishHostFxrInstallerFileToAzure),
             nameof(PublishTargets.PublishSharedFrameworkInstallerFileToAzure),
             nameof(PublishTargets.PublishCombinedMuxerHostFxrFrameworkInstallerFileToAzure))]
-        [BuildPlatforms(BuildPlatform.Ubuntu, BuildPlatform.OSX, BuildPlatform.Windows, , BuildPlatform.Debian)]
+        [BuildPlatforms(BuildPlatform.Ubuntu, BuildPlatform.OSX, BuildPlatform.Windows, BuildPlatform.Debian)]
         public static BuildTargetResult PublishInstallerFilesToAzure(BuildTargetContext c) => c.Success();
 
         [Target(
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -232,7 +232,7 @@ namespace Microsoft.DotNet.Host.Build
             nameof(PublishTargets.PublishHostFxrInstallerFileToAzure),
             nameof(PublishTargets.PublishSharedFrameworkInstallerFileToAzure),
             nameof(PublishTargets.PublishCombinedMuxerHostFxrFrameworkInstallerFileToAzure))]
-        [BuildPlatforms(BuildPlatform.Ubuntu, BuildPlatform.OSX, BuildPlatform.Windows)]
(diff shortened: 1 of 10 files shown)


brthor:
@ellismg liblldb-3.6 was causing issues on debian 8, so I excluded it from the dependency list altogether. 

I recall you stating that it was an unnecessary dependency except for those using libsos, is this still true?

ellismg:
> I recall you stating that it was an unnecessary dependency except for those using libsos, is this still true?

Yes.

brthor:
Cool thanks!

scripts/docker/debian.8/Dockerfile
```diff
@@ -41,6 +41,10 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# liblldb is needed so deb package build does not throw missing library info errors
+RUN wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key|sudo apt-key add -
+RUN echo "deb http://llvm.org/apt/jessie/ llvm-toolchain-jessie-3.6 main" > /etc/apt/sources.list.d/llvm-toolchain.list
+RUN apt-get update && apt-get -qqy install liblldb-3.6
```
MichaelSimons: Generally speaking it is a best practice to clean up the apt cache and remove /var/lib/apt/lists.  This helps keep the image size down.  It would also make this consistent with the previous apt-get logic in the Dockerfile.

brthor: will fix.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -245,7 +245,7 @@ private static bool CheckIfAllBuildsHavePublished()
             nameof(PublishSharedFrameworkDebToDebianRepo),
             nameof(PublishHostFxrDebToDebianRepo),
             nameof(PublishSharedHostDebToDebianRepo))]
-        [BuildPlatforms(BuildPlatform.Ubuntu)]
+        [BuildPlatforms(BuildPlatform.Ubuntu, BuildPlatform.Debian)]
```
naamunds: Should `BuildPlatform.Debian` also be added to the `PublishInstallerFilesToAzure` method in this file?

brthor: yup good catch :)

brthor: fixed.

packaging/deb/dotnet-sharedframework-debian_config.json
```diff
@@ -32,5 +32,9 @@
         "%HOSTFXR_DEBIAN_PACKAGE_NAME%" : {},
         "libssl1.0.0" : {},
         "libcurl3" : {}
-    }
+    },
+
+    "debian_ignored_dependencies" : [
+        "liblldb-3.6"
+    ]
```
naamunds: Why is this change only necessary in the `dotnet-sharedframework-debian_config.json` file, and not also for the other .json files in this directory, for the other hostfxr and sharedhost packages?

brthor: dh_helper which we're using to build the package auto detects package dependencies based libraries linked to the files inside the package. 

libsos....so  is linked to liblldb-3.6 and is only present in the sharedfx debian package.

naamunds:
LGTM

brthor:
I've emailed @chuan to get a new distro added to our repo, going to hold off on this merge until that happens.

scripts/docker/debian.8/Dockerfile
brthor: fixed.

brthor:
We're looking at next week for the distro to get added it looks like.

PR #232, opened 2016-08-05
Adding support for "painting" the native crypto library

blackdwarf:
Recently, Homebrew has changed its policies to disallow the --force
linking of OpenSSL. Because we need OpenSSL in the loader path, we had
several options and the best recommended one was to use
instal_name_tool to add a resolve path to the
System.Security.Cryptography.Native.dylib file.

This change makes that operation automatic when installing from the
PKG as part of its postinstall script. However, the user would still
need to install OpenSSL through brew in order for this to work.

/cc @eerhardt @mellinoe @bartonjs @piotrpMSFT @brthor

### DIFF
--- a/packaging/osx/sharedframework/scripts/postinstall
+++ b/packaging/osx/sharedframework/scripts/postinstall
@@ -6,8 +6,34 @@
 
 PACKAGE=$1
 INSTALL_DESTINATION=$2
+LIBCRYPTO_NAME=libcrypto.1.0.0.dylib
+LIBSSL_NAME=libssl.1.0.0.dylib
+
+# Using this form to allow power users to export both of these
+if [ -z $OPENSSL_PATH ]; then OPENSSL_PATH=/usr/local/opt/openssl/lib; fi
+if [ -z $LINK_DEST ]; then LINK_DEST=/usr/local/lib; fi
+
+# Ugly, but we need this to actually get the version
+#VERSION=${INSTALL_PKG_SESSION_ID##"com.microsoft.dotnet.sharedframework.Microsoft.NETCore.App."}
+#VERSION=${VERSION%%".component.osx.x64"}
+#CORE_LIB_FILE_PATH=$INSTALL_DESTINATION/shared/Microsoft.NETCore.App/$VERSION/System.Security.Cryptography.Native.dylib
+
 
 # A temporary fix for the permissions issue(s)
 chmod -R 755 $INSTALL_DESTINATION/shared
+
+if [ -e "$OPENSSL_PATH/$LIBSSL_NAME" ] && [ -e "$OPENSSL_PATH/$LIBCRYPTO_NAME" ]
+then
+   if [ ! -e "$LINK_DEST/$LIBSSL_NAME" ] && [ ! -e "$LINK_DEST/$LIBCRYPTO_NAME" ]
+   then
+       mkdir -p "$LINK_DEST"
+       ln -s "$OPENSSL_PATH/$LIBSSL_NAME" "$LINK_DEST"
+       ln -s "$OPENSSL_PATH/$LIBCRYPTO_NAME" "$LINK_DEST"
+   fi
+    #sudo install_name_tool \
+        #-add_rpath $OPENSSL_PATH \
+        #$CORE_LIB_FILE_PATH
+fi
+  
 
 exit 0
--- a/packaging/osx/sharedframework/scripts/postinstall
+++ b/packaging/osx/sharedframework/scripts/postinstall
@@ -13,12 +13,6 @@ LIBSSL_NAME=libssl.1.0.0.dylib
 if [ -z $OPENSSL_PATH ]; then OPENSSL_PATH=/usr/local/opt/openssl/lib; fi
 if [ -z $LINK_DEST ]; then LINK_DEST=/usr/local/lib; fi
 
-# Ugly, but we need this to actually get the version
-#VERSION=${INSTALL_PKG_SESSION_ID##"com.microsoft.dotnet.sharedframework.Microsoft.NETCore.App."}
-#VERSION=${VERSION%%".component.osx.x64"}
-#CORE_LIB_FILE_PATH=$INSTALL_DESTINATION/shared/Microsoft.NETCore.App/$VERSION/System.Security.Cryptography.Native.dylib
-
-
 # A temporary fix for the permissions issue(s)
 chmod -R 755 $INSTALL_DESTINATION/shared
 
@@ -30,9 +24,6 @@ then
        ln -s "$OPENSSL_PATH/$LIBSSL_NAME" "$LINK_DEST"
        ln -s "$OPENSSL_PATH/$LIBCRYPTO_NAME" "$LINK_DEST"
    fi
-    #sudo install_name_tool \
-        #-add_rpath $OPENSSL_PATH \
-        #$CORE_LIB_FILE_PATH
 fi
   
 
--- a/packaging/osx/sharedframework/shared-framework-distribution-template.xml
+++ b/packaging/osx/sharedframework/shared-framework-distribution-template.xml
@@ -3,7 +3,7 @@
     <title>{SharedFxBrandName} (x64)</title>
     <license file="eula.rtf" mime-type="application/rtf" />
     <background file="dotnetbackground.png" mime-type="image/png"/>
-    <options customize="never" require-scripts="true" />
+    <options customize="never" require-scripts="false" />
     <volume-check>
         <allowed-os-version>
             <os-version min="10.10" />
@@ -26,19 +26,4 @@
     <pkg-ref id="{SharedFxComponentId}.pkg">{SharedFxComponentId}.pkg</pkg-ref>
     <pkg-ref id="{HostFxrComponentId}.pkg">{HostFxrComponentId}.pkg</pkg-ref>
     <pkg-ref id="{SharedHostComponentId}.pkg">{SharedHostComponentId}.pkg</pkg-ref>
-    <installation-check script="openSSLCheck()" />
-    <script><![CDATA[
-        function openSSLCheck()
-        {
-            var openSSLRoot = "/usr/local/opt/openssl/lib/";
-            if (!system.files.fileExistsAtPath(openSSLRoot + "libcrypto.1.0.0.dylib") || 
-                !system.files.fileExistsAtPath(openSSLRoot + "libssl.1.0.0.dylib")
-                ){
-                    my.result.type = "Fatal";
-                    my.result.message = "OpenSSL was not detected on this machine. Please install OpenSSL to continue. Refer to https://dot.net/core for instructions.";
-                    return false;
-                }
-            return true; 
-        } 
-    ]]></script>
 </installer-gui-script>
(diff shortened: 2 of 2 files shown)


packaging/osx/sharedframework/scripts/postinstall
```diff
@@ -6,8 +6,25 @@
...
+# Ugly, but we need this to actually get the version
+VERSION=${INSTALL_PKG_SESSION_ID##"com.microsoft.dotnet.sharedframework.Microsoft.NETCore.App."}
+VERSION=${VERSION%%".component.osx.x64"}
+CORE_LIB_FILE_PATH=$INSTALL_DESTINATION/shared/Microsoft.NETCore.App/$VERSION/System.Security.Cryptography.Native.dylib
+
 
 # A temporary fix for the permissions issue(s)
 chmod -R 755 $INSTALL_DESTINATION/shared
 
+if [ -e "$OPENSSL_PATH/$LIBSSL_NAME" ] && [ -e "$OPENSSL_PATH/$LIBCRYPTO_NAME" ]
```
brthor: if the user hasn't installed openssl/libcrypto should we be failing silently like this? Shouldn't we at least give a warning so the user can go "brew install openssl"?

blackdwarf: As we discussed, we should actually abort the install if there is no OpenSSL found. Will push another change for that.

packaging/osx/sharedframework/scripts/postinstall
```diff
@@ -6,8 +6,25 @@
 
 PACKAGE=$1
 INSTALL_DESTINATION=$2
+LIBCRYPTO_NAME=libcrypto.1.0.0.dylib
+LIBSSL_NAME=libssl.1.0.0.dylib
+OPENSSL_PATH=/usr/local/opt/openssl/lib
```
bartonjs: You probably shouldn't set this variable if it's already set.  That way non-standard locations and/or macports users can be given the power-user instructions of "set this environment variable to where libcrypto and libssl live, then run the script"

blackdwarf: Good point. Will fix.

blackdwarf: Fixed.

packaging/osx/sharedframework/scripts/postinstall
blackdwarf: Fixed.

blackdwarf:
@dotnet-bot test  RHEL7.2 x64 Release Build please

blackdwarf:
Build on RHEL failed with this:

> error: Unable to resolve 'Microsoft.NETCore.Platforms (>= 1.0.1)' for 'DNXCore,Version=v5.0'.

I'm going to restart it to see if it passes or is it some mixup with myget feeds. 

@dotnet-bot test  RHEL7.2 x64 Release Build please

blackdwarf:
Are we ok with this solution?

packaging/osx/sharedframework/scripts/postinstall
brthor: 👍

brthor:
LGTM :shipit:

packaging/osx/sharedframework/scripts/postinstall
```diff
@@ -6,8 +6,26 @@
 
 PACKAGE=$1
 INSTALL_DESTINATION=$2
+LIBCRYPTO_NAME=libcrypto.1.0.0.dylib
+LIBSSL_NAME=libssl.1.0.0.dylib
```
eerhardt: Are these always `1.0.0`?

bartonjs: 1.0.0 is the only ABI set we have support for, yeah.

packaging/osx/sharedframework/scripts/postinstall
```diff
@@ -6,8 +6,26 @@
...
+VERSION=${VERSION%%".component.osx.x64"}
+CORE_LIB_FILE_PATH=$INSTALL_DESTINATION/shared/Microsoft.NETCore.App/$VERSION/System.Security.Cryptography.Native.dylib
+
 
 # A temporary fix for the permissions issue(s)
 chmod -R 755 $INSTALL_DESTINATION/shared
 
+if [ -e "$OPENSSL_PATH/$LIBSSL_NAME" ] && [ -e "$OPENSSL_PATH/$LIBCRYPTO_NAME" ]
+then
+    sudo install_name_tool \
```
eerhardt: How does this work for tarball installation? Do we tell our customers to run this command themselves?

eerhardt: Or should `dotnet-install.sh` do this as well.

bartonjs:
At the time I thought this approach was good.  I've since learned that OmniSharp has their own copy of the library (independent of the shared runtime); and now I'm concerned that we're optimizing for `dotnet restore` and that anything off the happy path might get harder to diagnose.

Maybe the installer should make the symlinks for you? But since it has to modify the system to do so (well, it could do it in ~/lib) it should be a consented thing.  (Yeah, a terrible "Yes to make it work, No to abort"... but still better than mucking with system state in the dark, or just having the notice scroll by).

At this point I can't see the forest for the trees, so I'm just sharing thoughts.

blackdwarf:
To @bartonjs' point, after talking to Dustin and Gregg, it appears that OmniSharp is a self-contained application. This means that for them to employ this workaround, they would have to do the same `install_name_tool` invocation upon installation of the C# extension. This is not doable, since the installation happens as a child process of VS Code and that cannot escalate privileges; no `sudo`, no `install_name_tool`. 

I would like to see all of the pros and cons of the symlink solution. I understand we are mucking with the system state, but I'm not clear of all the potential pitfalls. If we would have that info, I have the feeling we would have a much crisper picture and would be able to make a final call what the proper solution for automation is, or whether, truly, any automation should happen in the PKG at all.

blackdwarf:
@bartonjs any thoughts on this? :)

blackdwarf:
OK, I will make the necessary changes to this PR for review.

packaging/osx/sharedframework/scripts/postinstall
blackdwarf: It could, yeah.

blackdwarf:
Added it, all reviews are welcome @brthor @bartonjs @eerhardt

packaging/osx/sharedframework/scripts/postinstall
```diff
@@ -6,8 +6,34 @@
...
+LIBSSL_NAME=libssl.1.0.0.dylib
+
+# Using this form to allow power users to export both of these
+if [ -z $OPENSSL_PATH ]; then OPENSSL_PATH=/usr/local/opt/openssl/lib; fi
+if [ -z $LINK_DEST ]; then LINK_DEST=/usr/local/lib; fi
+
+# Ugly, but we need this to actually get the version
+#VERSION=${INSTALL_PKG_SESSION_ID##"com.microsoft.dotnet.sharedframework.Microsoft.NETCore.App."}
+#VERSION=${VERSION%%".component.osx.x64"}
+#CORE_LIB_FILE_PATH=$INSTALL_DESTINATION/shared/Microsoft.NETCore.App/$VERSION/System.Security.Cryptography.Native.dylib
```
bartonjs: Delete the commented out parts?

packaging/osx/sharedframework/scripts/postinstall
```diff
@@ -6,8 +6,34 @@
...
 
+if [ -e "$OPENSSL_PATH/$LIBSSL_NAME" ] && [ -e "$OPENSSL_PATH/$LIBCRYPTO_NAME" ]
+then
+   if [ ! -e "$LINK_DEST/$LIBSSL_NAME" ] && [ ! -e "$LINK_DEST/$LIBCRYPTO_NAME" ]
+   then
+       mkdir -p "$LINK_DEST"
+       ln -s "$OPENSSL_PATH/$LIBSSL_NAME" "$LINK_DEST"
+       ln -s "$OPENSSL_PATH/$LIBCRYPTO_NAME" "$LINK_DEST"
+   fi
+    #sudo install_name_tool \
```
bartonjs: Delete these commented out parts, too? :)

packaging/osx/sharedframework/shared-framework-distribution-template.xml
```diff
@@ -26,4 +26,19 @@
     <pkg-ref id="{SharedFxComponentId}.pkg">{SharedFxComponentId}.pkg</pkg-ref>
     <pkg-ref id="{HostFxrComponentId}.pkg">{HostFxrComponentId}.pkg</pkg-ref>
     <pkg-ref id="{SharedHostComponentId}.pkg">{SharedHostComponentId}.pkg</pkg-ref>
+    <installation-check script="openSSLCheck()" />
+    <script><![CDATA[
+        function openSSLCheck()
+        {
+            var openSSLRoot = "/usr/local/opt/openssl/lib/";
```
bartonjs: The installer script is forgiving on the path, but this isn't...

blackdwarf: Yes. In general, there are two ways to do this:

1. Have a `preinstall` script. This would allow, I believe, a workaround by setting an env variable. However, there is no way to define the error message, so the user would see the generic "installation failed" and would be lost *what* failed. There is a way to see that a preinstall script failed by pressing `Cmd-L` while the installer window is open...but even that doesn't give you a decent message.
2. Do what I did here, which will give you a decent message. 

So the solution is:

1. We remove the env var from postinstall, so we are unforgiving there as well.
2. We remove the check completely; this will get us back to our current state of allowing the install but failing if people try to use .NET Core. 

Thoughts?

packaging/osx/sharedframework/scripts/postinstall
blackdwarf: Oops, will do. :)

packaging/osx/sharedframework/shared-framework-distribution-template.xml
bartonjs: This does seem more user-friendly.  But since it makes life hard for macports and non-standard-path-homebrew users I'd personally cut it over making it be too restrictive.

If can't you read an environment variable into openSSLRoot, then do

```
my.result.message = "OpenSSL was not detected at '" + openSSLRoot + "'. Please install OpenSSL to continue. (etc)";
```
?

blackdwarf: Yeah, the more I think about this, the more I think that maybe this is a bad option...unless we can find out the common location for at least macports and check that as well. @bartonjs do you happen to know this info?

bartonjs: @blackdwarf I don't authoritatively.  https://github.com/dotnet/corefx/issues/9171#issuecomment-226950619 suggests that maybe it puts it in `/opt/local/lib`.

blackdwarf:
OK, I removed the pre-install check since we cannot make it work with the same as with postinstall. This will leave us with the experience we have currently where the user can install regardless if the user has OpenSSL.

blackdwarf:
@brthor @eerhardt @Sridhar-MS @bartonjs if you're ok with this, should we merge?

bartonjs:
LGTM

eerhardt:
:shipit:

PR #236, opened 2016-08-11
Add win10-arm64 and netcoreapp1.1 support

ianhays:
Adds parameterization to the Target framework and Target RID when doing a full build.cmd.  The end result of this is that core-setup packages can now be built for win10-arm64.
- Modifies CMake targets to allow win10-arm64 as a valid RID and arm64 as a valid machine platform.
- build.ps1 now allows passing a TargetArch (e.g. x64, arm64), ToolsDir (used for arm64 native build), and Framework (defaulted to netcoreapp1.0).
- Extracted all instances of RuntimeInformation.CurrentRuntimeEnvironment (and friends) where possible, and added environment variable checks to override the behavior.
- Can now build for a different architecture than the current platform.
- Adds package projects and targets for win10-arm64
- Adds .NETCoreApp1.1 as a targetframework for the Microsoft.NETCore.App meta package.
  - .NETCoreApp1.1 has updated (i.e. beta) package dependency versions as well as a new dependency on Microsoft.NetCore.Targets to ensure native dependencies (clrcompression) are resolved to the correct version.
  - The targetframework definition for .NETCoreApp1.0 is unchanged.

Caveats:
  - The win10-arm64 product zip produced by this is missing libuv but is otherwise at correct parity with the win7-x64 zip.
  - MSI installers for win10-arm64 are being produced but do not yet work correctly.
  - crossgen is disabled for win10-arm64
  - tests are still tethered to the current platform and thus don't work when building for a TargetPlatform other than the current platform.

@ericstj @weshaggard @gkhanna79 

fyi: @livarcocc


build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -180,19 +176,49 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
                 // Create .rc files on Windows.
                 var resourceDir = GenerateVersionResource(c);
 
+                if (configuration.Equals("Release"))
```
gkhanna79: I am surprised you had to add this since it would be required independent of Arm64 change.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -180,19 +176,49 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
...
+                string cmakeHostVer = $"-DCLI_CMAKE_HOST_VER:STRING={hostVersion.LatestHostVersion.ToString()}";
+                string cmakeHostPolicyVer = $"-DCLI_CMAKE_HOST_POLICY_VER:STRING={hostVersion.LatestHostPolicyVersion.ToString()}";
+                string cmakeHostFxrVer = $"-DCLI_CMAKE_HOST_FXR_VER:STRING={hostVersion.LatestHostFxrVersion.ToString()}";
+                string cmakeCommitHash = $"-DCLI_CMAKE_COMMIT_HASH:STRING={commitHash}";
+                string cmakeResourceDir = $"-DCLI_CMAKE_RESOURCE_DIR:STRING={resourceDir}";
+
+                switch (platform.ToLower())
+                {
+                    case "x86":
+                        cmakeBaseRid = "-DCLI_CMAKE_PKG_RID:STRING=win7-x86";
```
gkhanna79: NIT: Extra newline.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -180,19 +176,49 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
...
+                        arch = "x86";
+                        break;
+                    case "arm64":
+                        cmakeBaseRid = "-DCLI_CMAKE_PKG_RID:STRING=win10-arm64";
+                        visualStudio = "Visual Studio 14 2015 Win64";
+                        archMacro = "-DCLI_CMAKE_PLATFORM_ARCH_ARM64=1";
+                        arch = "arm64";
+                        break;
+                    case "x64":
+                    default:
```
gkhanna79: I would suggest that the `default` case should raise a `PlatformNotSupported` exception so that we do not endup compiling x64 for arm32 when that is brought up.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -180,19 +176,49 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
...
+                        cmakeBaseRid = "-DCLI_CMAKE_PKG_RID:STRING=win7-x64";
+                        visualStudio = "Visual Studio 14 2015 Win64";
+                        archMacro = "-DCLI_CMAKE_PLATFORM_ARCH_AMD64=1";
+                        arch = "x64";
+                        break;
+                }
+
+                if (arch == "arm64" && Environment.GetEnvironmentVariable("__ToolsetDir") == null)
```
gkhanna79: Move this check to the `arm64` switch case above.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -211,15 +237,19 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
...
                     Environment.GetEnvironmentVariable("ProgramFiles");
 
-                if (configuration.Equals("Release"))
+                if (arch == "arm64")
                 {
-                    // Cmake calls it "RelWithDebInfo" in the generated MSBuild
-                    configuration = "RelWithDebInfo";
+                    Exec(Path.Combine(pf32, "MSBuild", "14.0", "Bin", "MSBuild.exe"),
+                    Path.Combine(cmakeOut, "ALL_BUILD.vcxproj"),
+                    "/p:useEnv=true",
```
gkhanna79: Refactor this if/else to extract `/p:useEnv=true` for arm64 so that we have a single piece of code to maintain.

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -42,6 +42,17 @@ public class PrepareTargets
         public static BuildTargetResult Init(BuildTargetContext c)
         {
             var configEnv = Environment.GetEnvironmentVariable("CONFIGURATION");
+            string platformEnv = Environment.GetEnvironmentVariable("TARGETPLATFORM") ?? RuntimeEnvironment.OperatingSystemPlatform.ToString();
+            string targetRID = Environment.GetEnvironmentVariable("TARGETRID");
+            if (targetRID == null)
+            {
+                targetRID = RuntimeEnvironment.GetRuntimeIdentifier();
+                if (targetRID.StartsWith("win") && (targetRID.EndsWith("x86") || targetRID.EndsWith("x64")))
```
gkhanna79: Why do we need this check? Isnt targetRID already in the expected form? If not, what form is it in?

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -42,6 +42,17 @@ public class PrepareTargets
...
+                {
+                    targetRID = $"win7-{RuntimeEnvironment.RuntimeArchitecture}";
+                }
+            }
+            string targetFramework = Environment.GetEnvironmentVariable("TARGETFRAMEWORK") ?? "netcoreapp1.0";
```
gkhanna79: Should this be coming from command-line/params instead of env var?

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -42,6 +42,17 @@ public class PrepareTargets
         public static BuildTargetResult Init(BuildTargetContext c)
         {
             var configEnv = Environment.GetEnvironmentVariable("CONFIGURATION");
+            string platformEnv = Environment.GetEnvironmentVariable("TARGETPLATFORM") ?? RuntimeEnvironment.OperatingSystemPlatform.ToString();
+            string targetRID = Environment.GetEnvironmentVariable("TARGETRID");
```
gkhanna79: Who sets up this environment variable?

build_projects/dotnet-host-build/build.ps1
```diff
@@ -17,7 +20,10 @@ if($Help)
     Write-Host ""
     Write-Host "Options:"
     Write-Host "  -Configuration <CONFIGURATION>     Build the specified Configuration (Debug or Release, default: Debug)"
-    Write-Host "  -Architecture  <ARCHITECTURE>      Build the specified architecture (x64 or x86 (supported only on Windows), default: x64)"
+    Write-Host "  -Architecture  <ARCHITECTURE>      Build on the specified architecture (x64 or x86 (supported only on Windows), default: x64)"
```
gkhanna79: Is this script used only within Core-setup build? If not, and there are other users of this, then this maybe a breaking change for them.

build_projects/dotnet-host-build/build.ps1
```diff
@@ -17,7 +20,10 @@ if($Help)
...
+    Write-Host "  -TargetArch  <ARCHITECTURE>        Build for the specified architecture (x64, x86 (supported only on Windows), or arm64, default: x64)"
+    Write-Host "  -ToolsetDir  <TOOLSETDIR>          Path to a toolset to use when building the native host. Used for ARM64. )"
```
gkhanna79: I would call out that `--ToolsetDir` is "temporarily supported for Arm64 only" to let folks know this will go away.

build_projects/dotnet-host-build/build.ps1
```diff
@@ -26,9 +32,39 @@ if($Help)
...
 
+if($TargetArch -eq "arm64")
+{
+    if ($Framework -eq "netcoreapp1.0")
+    {
+        throw "ARM64 is not available on netcoreapp1.0. Pass in '-Framework netcoreapp1.1' to enable ARM64"
+    }
+    $env:__ToolsetDir = $ToolsetDir;
+    $env:TARGETRID = "win10-arm64";
+    $env:DISABLE_CROSSGEN = "1";
```
gkhanna79: Any reason crossgen is being disabled? It is fully supported in CoreCLR repo.

build_projects/dotnet-host-build/build.ps1
gkhanna79: And if this is only used within core-setup, I would suggest renaming this to `HostArchitecture` to make it obvious what it represents.

build_projects/dotnet-host-build/build.ps1
```diff
@@ -26,9 +32,39 @@ if($Help)
...
+    $env:PATH="$ToolsetDir\VC_sdk\bin;$env:PATH";
+    $env:LIB="$ToolsetDir\VC_sdk\lib\arm64;$ToolsetDir\sdpublic\sdk\lib\arm64";
+    $env:INCLUDE="$ToolsetDir\VC_sdk\inc;$ToolsetDir\sdpublic\sdk\inc;$ToolsetDir\sdpublic\shared\inc;$ToolsetDir\sdpublic\shared\inc\minwin;$ToolsetDir\sdpublic\sdk\inc\ucrt;$ToolsetDir\sdpublic\sdk\inc\minwin;$ToolsetDir\sdpublic\sdk\inc\mincore;$ToolsetDir\sdpublic\sdk\inc\abi;$ToolsetDir\sdpublic\sdk\inc\clientcore;$ToolsetDir\diasdk\include";
+}
+
+# No use in specifying a RID if the current and target architecture are equivalent. 
+if($TargetArch -eq "x86" -and $Architecture -ne "x86")
```
gkhanna79: Does the Host architecture required to be checked since the RID is dependent upon the target OS (which will be Windows in this case since we do not support Win->non-Win builds) and target arch?

pkg/pack.sh
```diff
@@ -48,7 +48,7 @@ else
     init_distro_name
 fi
 
-__common_parameters="/p:$__targets_param /p:DistroRid=$__distro_rid /verbosity:minimal"
+__common_parameters="/p:$__targets_param /p:DistroRid=$__distro_rid /verbosity:minimal /p:Platform=x64"
```
gkhanna79: Can you please elaborate on why this change is needed?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -11,10 +11,13 @@
...
     </ProjectReference>
+    <ProjectReference Include="..\Microsoft.NETCore.DotNetHost\Microsoft.NETCore.DotNetHostPolicy.pkgproj">
+      <TargetFramework>.NETCoreApp1.1</TargetFramework>
+    </ProjectReference>
   </ItemGroup>
-
+  
   <Target Name="ReadDependenciesFromProjectJsonFile" BeforeTargets="GetPackageDependencies">
-    <ReadLinesFromFile File="$(MSBuildThisFileDirectory)project.json">
+    <ReadLinesFromFile File="$(MSBuildThisFileDirectory)\netcoreapp1.0\project.json">
```
gkhanna79: Shouldnt this depend upon the target framework as opposed to hardcoded to netcoreapp1.0?

gkhanna79:
CC @schellap even though he is OOF

Modulo comments, change looks reasonable to me.

build_projects/dotnet-host-build/CompileTargets.cs
ianhays: I didn't add it, I just moved it up.

build_projects/dotnet-host-build/CompileTargets.cs
ianhays: Good point.

build_projects/dotnet-host-build/PrepareTargets.cs
ianhays: We do, in build.ps1. Environment variables are how core-setup passes args from build.cmd to the targets.

build_projects/dotnet-host-build/PrepareTargets.cs
ianhays: Configuration already set the precedent by being passed via environment variables so I followed the scheme already in place. It would take a fairly large amount of refactoring to change it to use Main args since there are some static initiatializations (e.g. Dirs.cs) that also check these environment variables. 

Since I imagine this whole repo will at some point be rewritten to use msbuild like CLI is now doing, it's likely not worth changing now.

build_projects/dotnet-host-build/PrepareTargets.cs
ianhays: If the TARGETRID env variable isn't set in build.ps1 (i.e. the TargetArch isn't set), then `RuntimeEnvironment.GetRuntimeIdentifier` returns your version-specific RID e.g. on a win10 x64 machine it returns `win10-x64`. We want it to be `win7-x64`.

build_projects/dotnet-host-build/build.ps1
ianhays: >Is this script used only within Core-setup build? 

I am operating under that assumption, but I made it backwards compatible regardless. If all they set is the `Architecture` then they'll get the same behavior they had before since we set `TargetArch = $Architecture` when `TargetArch` is undefined. 

Which is nice because I didn't have to modify the netci.groovy file since the args to pass to build.cmd are unchanged.

build_projects/dotnet-host-build/build.ps1
ianhays: Good idea.

build_projects/dotnet-host-build/build.ps1
ianhays: I don't set TARGETRID when the target arch is equal to the current arch so that the directory names don't change. This is somewhat convoluted.

Example:

I'm on a `win10-x64` machine and I want to build core-setup for x64. Before this change, the artifacts are put in `core-setup/artifacts/win10-x64` and the tests are hard coded to run using those artifacts.

So now that we have a TargetArchitecture notion, if we're building artifacts for win10-arm64 then it of course makes sense to put them in `core-setup/artifacts/win10-arm64`. But what about when we're on a `win10-x64` machine building for `win7-x64`? Do the artifacts go in `setup/artifacts/win10-x64` like they used to or should they go in `setup/artifacts/win7-x64` since that's the true TargetRID?

My answer was to keep them in their current location so I wouldn't have to modify any tests. To do this, I had to not set the TARGETRID when currentarch == targetarch so that dirs.cs would default directory paths to use `RuntimeEnvironment.GetRuntimeIdentifier()`. If we were to set TARGETRID in build.ps1, then dirs.cs would incorrectly use that to determine the path.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
ianhays: I couldn't find an elegant way to do this in msbuild syntax. This code is duplicated below for netcoreapp1.1.

We want to import both project.jsons (for netcoreapp1.0 and netcoreapp1.1) so that the meta package Microsoft.NETCore.App contains <targetframework> tags for both.

build_projects/dotnet-host-build/build.ps1
gkhanna79: > Do the artifacts go in setup/artifacts/win10-x64 like they used to or should they go in setup/artifacts/win7-x64 since that's the true TargetRID?

I believe the RID's premise is to assume that what you are building upon is what you are going to target as the platform - unless you have been specified the target RID (e.g. in cross compilation scenarios like arm32 or arm64). I believe you should keep the existing behaviour but enable it by specifying the RID to be used:

1) For targeting scenarios, it will be the RID computed by default
2) For cross-targeting scenarios, it will be the RID calculated.
3) When specified RIDs, it will go into the intended RID folder

build_projects/dotnet-host-build/PrepareTargets.cs
weshaggard: We so need a lot of work on core-setup build system.

ianhays: >We so need a lot of work on core-setup build system.

Yes, we really do. Once CLI finishes the transition to msbuild, core-setup is going to be the black sheep with a nonconforming structure and build system. 

IMO it would be worthwhile to put in the time and effort to bring it up to structural parity with the other repos. It was no easy task transitioning from CoreFX to this, and it's only going to get worse as the build system evolves.

pkg/projects/Microsoft.NETCore.App/netcoreapp1.1/project.json
```diff
@@ -0,0 +1,49 @@
+{
```
weshaggard: What are the package differences between netcoreapp1.0 and the netcoreapp1.1 package list?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
weshaggard: I assume the package list is shared? If so then we shouldn't need to do this Regex hack twice. If we do need unique sets then we should just batch on the target frameworks so we don't need to duplicate the logic. 

PS: I wrote this regex hack and I really dislike it, it only exists because we didn't have a good way to have the package versions update with our update scripts. At some point we will fix that in core-setup and we can kill this hack.

ianhays: >I assume the package list is shared? If so then we shouldn't need to do this Regex hack twice. 

What do you mean by shared? The dependencies for netcoreapp1.0 are not equivalent to the dependencies for netcoreapp1.1 if that's what you mean. If they were then you're right, it would be one read and two TargetFramework tags.

>If we do need unique sets then we should just batch on the target frameworks so we don't need to duplicate the logic.

A batch would work. I honestly didn't put too much effort into it because it seemed like a band-aid solution at best. 

>PS: I wrote this regex hack and I really dislike it, it only exists because we didn't have a good way to have the package versions update with our update scripts. At some point we will fix that in core-setup and we can kill this hack.

That would be nice :) I originally tried adding the dependencies for netcoreapp1.1 under the dependencies of the framework, but I noticed they were getting applied to every framework. Then I found the Regex. Ideally we would have some real json parsing to read the dependencies under a framework and associate them with that framework instead of just reading every line and Regex'ing them. That way we would only need one project.json file and we could circumvent the need for duplication or a batch script.

pkg/projects/Microsoft.NETCore.App/netcoreapp1.1/project.json
ianhays: Updated packages:
```
    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.4-beta-24403-04",
    "System.Diagnostics.Process": "4.1.1-beta-24403-05",
    "System.Linq.Expressions": "4.1.1-beta-24403-05",
    "System.Linq.Parallel": "4.0.2-beta-24403-05",
```

new additions:
```
    "Microsoft.NETCore.Targets": "1.0.3-beta-24403-05",
    "System.Runtime": "4.1.1-beta-24403-05",
```

The justification for adding Targets is that the native compression packages use the Targets import to determine which runtime-specific packages to include. When the stable compression package imports the stable Targets package, we end up lacking win10-arm64 assets as the last stable Targets package doesn't acknowledge that RID. Rather than modifying the clrcompression package (and potentially the other native packages for other new RIDS) and forcing updates of that version, I opted to instead have NetCore.App depend on a more recent Targets package.

It's somewhat tricky here because I don't want to update more versions than I need to and I don't want to break compat with netcoreapp1.0. I want netcoreapp1.0 and the associated packages to be as close to they are now as possible while still enabling netcoreapp1.1 to function and pull in the correct, updated assets.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
weshaggard: Ideally we would just put this package list in a msbuild props file and not need to do any special parsing at all. The only reason I didn't do that was we didn't have a great way of refreshing those dependencies where we do for project.json files.

pkg/projects/Microsoft.NETCore.App/netcoreapp1.1/project.json
weshaggard: We shouldn't need to reference Targets or the new System.Runtime. you should instead reference the new NETStandard.Library package which contains both of those. 

Actually I think we can get away with a single package list for both netcoreapp1.0 and netcoreapp1.1 and just update the versions. The asset selection isn't based on the version but based on the contents of the actual packages so everything should just work out.

ianhays: per our in person discussion, I removed the netcoreapp1.1 specific stuff and it looks like it still works. The main thing I needed to change for netcoreapp1.1 to work was that I needed up update the version of NetStandardLibrary being used.

build_projects/dotnet-host-build/build.ps1
ianhays: I hadn't realized it was supported. I pushed an update supporting it.

build_projects/dotnet-host-build/build.ps1
ianhays: Sounds good. That's the behavior that I currently have.

pkg/pack.sh
ianhays: This is a band-aid that shouldn't have been checked-in in its current form.

I added it because the `pack.sh` script on Unix was computing package names as `runtime.ubuntu.14.04-linux.system.io.compression`. It was relying on the `Platform` and `PackagePlatform` variables to be set and was defaulting them oddly when they weren't being set. I'll look into it more and get a better solution or at least figure out why it's failing that way.

ianhays: Got it. Pack.sh imports an msbuild props file written by the build. That props file was using the the wrong variable for its <Platform> tag. I pushed a fix.

ianhays:
Thanks for taking a look @gkhanna79 and @weshaggard! I've responded to all of yalls feedback.

pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -3,22 +3,22 @@
     "Microsoft.CodeAnalysis.CSharp": "1.3.0",
     "Microsoft.CodeAnalysis.VisualBasic": "1.3.0",
```
weshaggard: I'm fine only updating some of these but I do think we should probably update all these at some point.

pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -3,22 +3,22 @@
...
     "Microsoft.CSharp": "4.0.1",
-    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.2",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.4-beta-24403-05",
```
weshaggard: Any reason you are picking build 24403-05? instead of the latest build? 

If you take the latest build you are going to need to update the version to 1.1.0 for the CoreCLR package.

ianhays: No good reason, no.

pkg/projects/Microsoft.NETCore.App/project.json
ianhays: Agreed. I wanted to minimize the already large delta for this PR but eventually these should all be bumped to latest.

pkg/projects/Microsoft.NETCore.App/project.json
weshaggard: I think I would go ahead and update the the latest version.

weshaggard:
LGTM

PR #241, opened 2016-08-25
[WIP - DO NOT MERGE] Enable producing Microsoft.NETCore.App 1.0.1 package

gkhanna79:
@weshaggard @ellismg @eerhardt @mellinoe PTAL

On Windows, post a build, I have confirmed that package name, zips and installation folder contain 1.0.1 as expected.

I have also observed that the host bundled in the SharedFX Bundle MSI is the same as in RTM Bundle MSI and thus, I didn't have to update any host versions.

CC @Petermarcu @leecow @schellap

### DIFF
--- a/branchinfo.txt
+++ b/branchinfo.txt
@@ -3,13 +3,9 @@
 # Each line is expected to be in the format "[Name]=[Value]".
 #
 #
-# !!! WARNING !!! WARNING !!! WARNING !!! WARNING !!! WARNING !!! WARNING
-#       "master" branch temporarily produced 1.0.1 builds. The value here
-#       should be > 1.0.1
-#
 MAJOR_VERSION=1
 MINOR_VERSION=0
-PATCH_VERSION=2
+PATCH_VERSION=1
 RELEASE_SUFFIX=servicing
 CHANNEL=preview
 BRANCH_NAME=release/1.0.0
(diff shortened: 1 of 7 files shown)


branchinfo.txt
```diff
@@ -3,13 +3,9 @@
...
 #
 #
-# !!! WARNING !!! WARNING !!! WARNING !!! WARNING !!! WARNING !!! WARNING
-#       "master" branch temporarily produced 1.0.1 builds. The value here
-#       should be > 1.0.1
-#
 MAJOR_VERSION=1
 MINOR_VERSION=0
-PATCH_VERSION=2
+PATCH_VERSION=1
```
weshaggard: I assume we need to update the master branch as well? Otherwise we might conflict.

weshaggard:
LGTM beyond the question about needing to update master.

branchinfo.txt
gkhanna79: Master is already on 1.1.0, so there should not be any conflict.

eerhardt:
This is producing a Microsoft.NETCore.App package with a prerelease `1.0.1-servicing-004509-00`.  Is there going to be another PR to change it to produce `1.0.1` with no prerelease?

gkhanna79:
Yes - @weshaggard is going to have a seperate PR for stablization of package version.

PR #242, opened 2016-08-26
Mark M.NC.App 1.0.1 stable

weshaggard:
cc @ericstj @eerhardt @gkhanna79

### DIFF
--- a/TestAssets/TestProjects/PortableApp/project.json
+++ b/TestAssets/TestProjects/PortableApp/project.json
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.0.1-servicing-*"
+          "version": "1.0.1"
         },
         "Newtonsoft.Json": "9.0.1-beta1"
       }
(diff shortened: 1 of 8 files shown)


gkhanna79:
Lgtm

gkhanna79:
Why do the legs fail though?

weshaggard:
Do we need new host packages? The host packages are still marked as prerelease in this branch which is why the CI is failing.

weshaggard:
I've updated the PR to mark the Host packages as stable as they already shipped stable and filtered the package built to just Microsoft.NETCore.App which should align with what we are wanting here.

gkhanna79:
> mark the Host packages as stable as they already shipped stable and filtered the package built to just Microsoft.NETCore.App which should align with what we are wanting here.

Sounds good to me. Also lines up with our approach for servicing packages as discussed.

CC @schellap

weshaggard:
unfortunately this still breaks stuff because the core-setup repo relies on all the packages being their or it fails. I'm still working on getting this updated.

PR #246, opened 2016-09-02
Upgrade CLI, buildtools for RTM shared framework

dagood:
See https://github.com/dotnet/corefx/pull/11387.

@weshaggard To make the upgrades work I changed `pack.cmd/sh` to use the `dotnet` host. Does that seem correct? No more corerun in the tools dir.

Using RTM shared framework in buildtools means I should be able to follow up with a change to use the new version of `restore --disable-parallelism` which also disables http parallel. This will help make the core-setup builds not so flaky.

/cc @yizhang82


pkg/pack.cmd
```diff
@@ -2,21 +2,23 @@
...
+set __DotNet=%__ProjectDir%\Tools\dotnetcli\dotnet.exe
+set __MSBuild=%__ProjectDir%\Tools\MSBuild.exe
 
 :: Initialize the MSBuild Tools
 call "%__ProjectDir%\init-tools.cmd"
 
 :: Restore dependencies mainly to obtain runtime.json
 pushd "%__ProjectDir%\deps"
-"%__ProjectDir%\Tools\dotnetcli\dotnet.exe" restore --source "https://dotnet.myget.org/F/dotnet-core" --packages "%__ProjectDir%\packages"
+"%__DotNet%" restore --source "https://dotnet.myget.org/F/dotnet-core" --packages "%__ProjectDir%\packages"
```
weshaggard: Yes this seems reasonable.

weshaggard:
FYI @ramarag 

LGTM but I didn't dig into the CI failures which we should ensure works.

dagood:
CI failures are almost definitely the same as the ones in the corefx PR (https://github.com/dotnet/corefx/pull/11387), caused by a change in the CLI package layout. I'll be updating the other repos' PRs when I get something working in corefx.

dagood:
Closing, these changes were made in https://github.com/dotnet/core-setup/pull/254

PR #247, opened 2016-09-06
Read the path and hashPath properties when reading .deps.json in hostfxr/policy

joelverhagen:
Since a "hashPath" and "path" property are now in the .deps.json file, this PR updates the native components in core-setup to prefer these values over manually constructing the paths.

Addresses the next point in https://github.com/NuGet/Home/issues/2522.

This is a reincarnation of https://github.com/dotnet/core-setup/pull/64.

/cc @gkhanna79 @schellap @eerhardt @livarcocc @piotrpMSFT

### DIFF
--- a/src/corehost/cli/deps_entry.cpp
+++ b/src/corehost/cli/deps_entry.cpp
@@ -118,8 +118,10 @@ bool deps_entry_t::to_full_path(const pal::string_t& base, pal::string_t* str) c
     }
 
     pal::string_t new_base = base;
-    append_path(&new_base, library_name.c_str());
-    append_path(&new_base, library_version.c_str());
+    pal::string_t lower_name = pal::to_lower(library_name);
+    pal::string_t lower_version = pal::to_lower(library_version);
+    append_path(&new_base, lower_name.c_str());
+    append_path(&new_base, lower_version.c_str());
 
     return to_rel_path(new_base, str);
 }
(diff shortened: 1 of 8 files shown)


src/corehost/cli/deps_format.cpp
```diff
@@ -43,6 +43,40 @@ void deps_json_t::reconcile_libraries_with_targets(
...
+        const auto& library_path_iter = properties.find(_X("path"));
+        if (library_path_iter == properties.end())
+        {
+            library_path = _X("");
+        }
+        else
+        {
+            library_path = library_path_iter->second.as_string();
+
+			if (_X('/') != DIR_SEPARATOR)
```
joelverhagen: Seems like I have some inconsistent indentation characters. Will fix.

joelverhagen: Fixed.

src/corehost/cli/deps_entry.cpp
```diff
@@ -164,24 +172,49 @@ bool deps_entry_t::to_hash_matched_path(const pal::string_t& base, pal::string_t
...
-
-    // Build the hash file path str.
+    // Build the directory that contains the hash file.
     pal::string_t hash_file;
-    hash_file.reserve(base.length() + library_name.length() + library_version.length() + nupkg_filename.length() + 3);
-    hash_file.assign(base);
-    append_path(&hash_file, library_name.c_str());
-    append_path(&hash_file, library_version.c_str());
-    append_path(&hash_file, nupkg_filename.c_str());
+    if (library_path.empty())
```
eerhardt: Is it possible to reuse the `deps_entry_t::to_full_path` method above?

src/corehost/cli/deps_format.cpp
```diff
@@ -43,6 +43,40 @@ void deps_json_t::reconcile_libraries_with_targets(
...
+        }
+        else
+        {
+            library_hash_path = library_hash_path_iter->second.as_string();
+
+            if (_X('/') != DIR_SEPARATOR)
+            {
+                replace_char(&library_hash_path, _X('/'), DIR_SEPARATOR);
+            }
+        }
```
eerhardt: Can this chunk of code get extracted into a shared function between this and the library code above?  They look identical except for the string property name to look for.

joelverhagen: Yes, good idea!

src/corehost/cli/deps_entry.cpp
joelverhagen: `to_full_path` checks for file path existence on it's own, via `to_path`. `to_hash_matched_path` does it on it's own (`if (!fstream.good())`) with a specific trace message. I'd rather keep them separate to minimize risk. Does that seem reasonable?

src/corehost/cli/deps_format.cpp
joelverhagen: Fixed.

src/corehost/cli/deps_entry.cpp
eerhardt: Yes, that seems totally fine.  Thanks for explaining.

gkhanna79:
LGTM

gkhanna79:
Please validate the scenarios (with and w/o the data) and confirm they work :)

eerhardt:
:shipit:

PR #249, opened 2016-09-08
Change auto-update to use 1.1.0 build-infos

dagood:
This applies to dependencies on corefx and coreclr. The roslyn dependency might also need to change at some point. (Will there be a 1.1.0 there?)

Change update-dependencies.ps1 to use default CLI channel, like build.ps1 does. CLI was failing to download otherwise.

This will let me set up core-setup Maestro subscriptions.

@gkhanna79 @eerhardt @wtgodbe @weshaggard


gkhanna79:
LGTM

eerhardt:
:shipit:

eerhardt:
test Windows_NT x64 Release Build

PR #250, opened 2016-09-09
Update package versions to be 1.1.0-preview1-*

gkhanna79:
Now that we have preview1 packages published from CoreCLR and CoreFX, update package versions to be 1.1.0-preview1-* in Microsoft.NETCore.App and also update versions of CoreCLR and JIT packages used for precompilation of SharedFW.

These will be updated automatically once Maestro comes online.

@weshaggard @dagood PTAL

CC @schellap @leecow @eerhardt


build_projects/shared-build-targets-utils/DependencyVersions.cs
```diff
@@ -7,9 +7,10 @@ namespace Microsoft.DotNet.Cli.Build
 {
     public class DependencyVersions
     {
-        public static readonly string CoreCLRVersion = "1.0.2";
-        public static readonly string JitVersion = "1.0.2";
-        public static readonly string CoreCLRVersion_Latest = "1.1.0-beta-24415-01";
-        public static readonly string JitVersion_Latest = "1.1.0-beta-24415-01";
+        // TODO: Can these versions be updated by automation?
```
gkhanna79: @dagood Can Maestro be updated to automatically update these versions as well?

eerhardt: Yes, Maestro already knows how to update the CoreCLRVersion and JitVersion properties:

https://github.com/dotnet/core-setup/blob/master/build_projects/update-dependencies/UpdateFilesTargets.cs#L200-L215

I'm not sure why we added these "Latest" properties, or why they are needed now that the property values are both insync.  Can we remove the "_Latest" properties?  @ianhays

pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -3,30 +3,30 @@
...
     "Microsoft.VisualBasic": "10.0.1",
     "NETStandard.Library": "1.6.1-beta-24415-01",
     "System.Buffers": "4.0.0",
     "System.Collections.Immutable": "1.2.0",
     "System.ComponentModel": "4.0.1",
     "System.ComponentModel.Annotations": "4.1.0",
     "System.Diagnostics.DiagnosticSource": "4.0.0",
-    "System.Diagnostics.Process": "4.1.1-beta-24415-01",
+    "System.Diagnostics.Process": "4.3.0-preview1-24508-08",
     "System.Dynamic.Runtime": "4.0.11",
```
eerhardt: How come only some of these versions are being updated, but not all?

build_projects/shared-build-targets-utils/DependencyVersions.cs
eerhardt: You can also remove the TODO.  Automation can update these versions.

gkhanna79: Done.

pkg/projects/Microsoft.NETCore.App/project.json
gkhanna79: I only manually updated the ones that were required to get a clean build. I am relying on Maestro to perform the comprehensive update.

eerhardt: Ok, that's what I assumed, but I wanted to be sure.  Thanks for the explanation.

dagood: These wouldn't be updated in (e.g.) corefx because they're stable, but [core-setup's logic](https://github.com/dotnet/core-setup/blob/c7e704a7bf5f53080b53e409f821bfbf7f4dc225/build_projects/update-dependencies/UpdateFilesTargets.cs#L129) doesn't skip over stable versions so they should all be auto-updated.

build_projects/shared-build-targets-utils/DependencyVersions.cs
```diff
@@ -7,9 +7,9 @@ namespace Microsoft.DotNet.Cli.Build
...
     public class DependencyVersions
     {
-        public static readonly string CoreCLRVersion = "1.0.2";
-        public static readonly string JitVersion = "1.0.2";
-        public static readonly string CoreCLRVersion_Latest = "1.1.0-beta-24415-01";
-        public static readonly string JitVersion_Latest = "1.1.0-beta-24415-01";
+        public static readonly string CoreCLRVersion = "1.1.0-preview1-24508-06";
+        public static readonly string JitVersion = "1.1.0-preview1-24508-06";
+        public static readonly string CoreCLRVersion_Latest = "1.1.0-preview1-24508-06";
+        public static readonly string JitVersion_Latest = "1.1.0-preview1-24508-06";
```
dagood: I'm also curious about the `_Latest` properties and if these can be combined. [It looks like](https://github.com/dotnet/core-setup/blob/05ce78fa94159446cb814e5fac1fa193455e7452/build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs#L45-L47) the `_Latest` is used for crossgen with `win10-arm64`, otherwise non-latest.

(Made a new comment so github doesn't hide the thread as outdated. /cc @ianhays @eerhardt )

ianhays: I added the _Latest properties so we could build win10-arm64 core-setup using the non-stable coreFX and CoreCLR dependencies. There wasn't a clean way to use the existing variables, so I made those new ones.

If `CoreCLRVersion` and `JitVersion` are going to be updated to the non-stable versions, then we can remove `CoreCLRVersion_Latest` and `JitVersion_Latest`  and replace all of their usage.

gkhanna79: Given that we want master of Core-Setup to also use non-stable versions of the JIT and CoreCLR (e.g. daily builds that are published from pipeline), it makes sense to have a single version.

@ianhays Can you please make this fix in master and release/1.1.0 branches (in the latter once I have merged this PR) after validating that Arm64 is not broken?

ianhays: Sure thing.

gkhanna79:
@dotnet-bot test Windows_NT x86 Debug Build  please

PR #254, opened 2016-09-12
Add support for building ubuntu 16.10 and openSUSE 42.1

mellinoe:
This updates buildtools, the CLI version, and a bunch of other project dependencies to enable them to restore and build for new RIDs. I've added all of the packaging projects for openSUSE 42.1 and ubuntu 16.10. I've only added a Dockerfile for openSUSE 42.1 so far. I need to make some modifications to the Ubuntu 16.10 bootstrap CLI on Azure before it will work with the rest of the core-setup build. I've already made those changes to the openSUSE bootstrap CLI.

@brthor , @livarcocc , @Sridhar-MS


scripts/obtain/dotnet-install.sh
```diff
@@ -84,6 +84,10 @@ get_current_os_name() {
                     echo "opensuse.13.2"
                     return 0
                     ;;
+                "opensuse.42.1")
```
brthor: hmm this file shouldn't be checked in here. 

You'll need to make these changes in dotnet/cli in order for downstream consumers to take advantage of it.

brthor:
You'll need to add new rids here as well (For build finalization)

https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PublishTargets.cs#L187

scripts/obtain/dotnet-install.sh
mellinoe: Could you clarify this part? Are we replacing this whole file or something? This was just a part of the pre-existing initialization process. Don't the CLI changes need to follow after this stuff is already merged and built?

brthor: I was confused to see this file checked in here at all, looks like it got added at some point https://github.com/dotnet/core-setup/commit/5caaef7647fee38da701f3772aa3e710ddd10ccc

This is core-setup's special copy of this file and it looks to be only checked in for unix (not windows), my comment was just that in order for everyone else to get these changes they'll need to go into cli as well.

mellinoe:
Errors are because init-tools.sh doesn't properly handle some fallback RIDs (according to @dagood); I'm updating it to the version from corefx. Testing it now...

mellinoe:
@brthor Any other comments on this? CI has passed now.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -122,7 +122,8 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
                         "debian.x64.version",
                         "centos.x64.version",
                         "fedora.23.x64.version",
-                        "opensuse.13.2.x64.version"
+                        "opensuse.13.2.x64.version",
+                        "opensuse.42.1.x64.version"
```
brthor: do you need to add ubuntu 16.10?

pkg/init-tools.sh
```diff
@@ -13,39 +13,84 @@ __BUILD_TOOLS_PATH=$__PACKAGES_DIR/Microsoft.DotNet.BuildTools/$__BUILD_TOOLS_PA
...
 
-OSName=$(uname -s)
-case $OSName in
-    Darwin)
-        OS=OSX
-        __DOTNET_PKG=dotnet-dev-osx-x64
-        ulimit -n 2048
-        ;;
+# Extended version of platform detection logic from dotnet/cli/scripts/obtain/dotnet-install.sh 16692fc
+get_current_linux_name() {
```
brthor: Maintaining this logic in both places is going to suck.

Why not just call the dotnet-install script? Init-tools in cli does this...
https://github.com/dotnet/cli/blob/rel/1.0.0/init-tools.sh#L38

dagood: This way mirrors what we do in corefx because we don't have access to dotnet-install there. I think we want to remove dotnet-install from core-setup in the future, anyway, when it's changed to use buildtools.

mellinoe: My understanding is we're going to consolidate this stuff into buildtools eventually, and in the meantime we should just keep this script in-sync with corefx's version.

brthor: That doesn't solve the problem of needing to maintain this block of logic being copy/pasted all over. 

Since build tools depends on a cli, why doesn't it use dotnet-install like the other consumers of cli?

build_projects/dotnet-host-build/PublishTargets.cs
mellinoe: Yeah, it's going to be in a separate PR. We also need a new Dockerfile and some changes to the debian packaging scripts.

pkg/init-tools.sh
dagood: > Since build tools depends on a cli, why doesn't it use dotnet-install like the other consumers of cli? 

That's (part of) the plan, although we will also need to figure out how to add bootstrap versions. (At least, I'm imagining we don't want to expose those to "normal" users of dotnet-install--maybe I'm wrong there.)

brthor: > That's (part of) the plan, although we will also need to figure out how to add bootstrap versions. (At least, I'm imagining we don't want to expose those to "normal" users of dotnet-install--maybe I'm wrong there.)

Agreed bootstrapping brings in a fair bit of complexity most people don't need to carry about. As long as the plan is eventually to not need to copy this block everywhere, I'm good 👍

brthor:
:shipit:

mellinoe:
Thanks, I should have the rest of the ubuntu.16.10 stuff soon, once I get a clean build of it...

pkg/init-tools.sh
```diff
@@ -66,18 +111,19 @@ if [ ! -e $__PROJECT_JSON_FILE ]; then
 
     if [ ! -e $__BUILD_TOOLS_PATH ]; then
         echo "Restoring BuildTools version $__BUILD_TOOLS_PACKAGE_VERSION..."
-        echo "Running: $__DOTNET_CMD restore \"$__PROJECT_JSON_FILE\" --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE" >> $__init_tools_log
-        $__DOTNET_CMD restore "$__PROJECT_JSON_FILE" --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE >> $__init_tools_log
-        if [ ! -e "$__BUILD_TOOLS_PATH/init-tools.sh" ]; then echo "ERROR: Could not restore build tools correctly. See '$__init_tools_log' for more details."; fi
+        echo "Running: $__DOTNET_CMD restore \"$__PROJECT_JSON_FILE\" --no-cache --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE" >> $__init_tools_log
```
eerhardt: Why `--no-cache` here?

dagood: Parity with corefx: https://github.com/dotnet/corefx/pull/10016

PR #255, opened 2016-09-12
Change 1.1.0 channel to match expected build-info

dagood:
The build-info paths are expected to contain the branch name, so this change is necessary to prevent errors there. There is no reason to stick with `preview` as the channel name, so we might as well make it and the branch name match.


dagood:
I need to confirm that the channel comes from that file in the official builds before I merge this.

dagood:
I see: `build.sh` and `build.ps1` load the branchinfo.txt properties into environment variables that the build fetches.

PR #260, opened 2016-09-13
Add dockerfile for ubuntu 16.10, fix some debian packaging issues

mellinoe:
@brthor


scripts/docker/ubuntu.16.10/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+RUN apt-get update && \
+    apt-get -qqy install \
+        libunwind8 \
+        libkrb5-3 \
+        libicu57 \
+        liblttng-ust0 \
+        libssl1.0.0 \
+        zlib1g \
+        libuuid1 \
+        liblldb-3.5 && \
```
brthor: you should be able to remove this.

tools/dotnet-deb-tool/tool/scripts/config_template_generator.py
```diff
@@ -69,11 +69,13 @@ def generate_rules(config_data, template_dir):
     override_text = ""
 
     if ignored_dependency_packages != None:
-        override_text = "override_dh_shlibdeps:\n\tdh_shlibdeps --dpkg-shlibdeps-params="
+        override_text = "override_dh_shlibdeps:\n\tdh_shlibdeps --dpkg-shlibdeps-params=\""
 
         for package in ignored_dependency_packages:
             override_text += "-x{0} ".format(package)
 
+        override_text += "\""
```
brthor: nice fix, thanks! 👍

scripts/docker/ubuntu.16.10/Dockerfile
mellinoe: There is actually another debhelper command that gives an error if this is not installed. We could (potentially, haven't tested) add some more logic to the python script to pass the "ignore flag" to that command, too, but it didn't seem worth it at this point. The other Ubuntu Dockerfiles install liblldb for the same reason, it looks like.

PR #261, opened 2016-09-13
Use latest packages and fix arm64 build

ianhays:
Fixes the arm64 build by adding a vcvarsall.bat call to set the path variables we need. Also updates the package versions to the latest from coreclr and corefx.

@dagood @gkhanna79

### DIFF
--- a/build_projects/shared-build-targets-utils/DependencyVersions.cs
+++ b/build_projects/shared-build-targets-utils/DependencyVersions.cs
@@ -9,7 +9,5 @@ namespace Microsoft.DotNet.Cli.Build
     {
         public static readonly string CoreCLRVersion = "1.1.0-beta-24509-04";
         public static readonly string JitVersion = "1.1.0-beta-24509-04";
-        public static readonly string CoreCLRVersion_Latest = "1.1.0-beta-24415-01";
-        public static readonly string JitVersion_Latest = "1.1.0-beta-24415-01";
     }
 }
--- a/build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
+++ b/build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
@@ -16,7 +16,7 @@ namespace Microsoft.DotNet.Cli.Build
     public class SharedFrameworkPublisher
     {
         public static string s_sharedFrameworkName = "Microsoft.NETCore.App";
-        
+
         private string _sharedFrameworkTemplateSourceRoot;
         private string _sharedFrameworkNugetVersion;
         private string _sharedFrameworkRid;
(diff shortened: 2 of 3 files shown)


build_projects/shared-build-targets-utils/DependencyVersions.cs
```diff
@@ -7,9 +7,7 @@ namespace Microsoft.DotNet.Cli.Build
 {
     public class DependencyVersions
     {
-        public static readonly string CoreCLRVersion = "1.1.0-beta-24509-04";
-        public static readonly string JitVersion = "1.1.0-beta-24509-04";
-        public static readonly string CoreCLRVersion_Latest = "1.1.0-beta-24415-01";
-        public static readonly string JitVersion_Latest = "1.1.0-beta-24415-01";
+        public static readonly string CoreCLRVersion = "1.1.0-preview1-24512-01";
```
gkhanna79: Master branch should have the "beta" versions that you removed, while 1.1.0 should have the "preview1" version.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -42,9 +42,7 @@ public class SharedFrameworkPublisher
             _corehostLockedDirectory = corehostLockedDirectory;
             _corehostLatestDirectory = corehostLatestDirectory;
             _corehostPackageSource = corehostPackageSource;
-            _crossgenUtil = (sharedFrameworkRid == "win10-arm64") ?
-                new Crossgen(DependencyVersions.CoreCLRVersion_Latest, DependencyVersions.JitVersion_Latest, sharedFrameworkRid) : 
-                new Crossgen(DependencyVersions.CoreCLRVersion, DependencyVersions.JitVersion); 
+            _crossgenUtil = new Crossgen(DependencyVersions.CoreCLRVersion, DependencyVersions.JitVersion, sharedFrameworkRid == "win10-arm64" ? sharedFrameworkRid : null);
```
gkhanna79: Where is this used?

pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -3,7 +3,7 @@
     "Microsoft.CodeAnalysis.CSharp": "1.3.0",
     "Microsoft.CodeAnalysis.VisualBasic": "1.3.0",
     "Microsoft.CSharp": "4.0.1",
-    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.0-beta-24509-04",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.0-preview1-24512-01",
```
gkhanna79: Same as above.

build_projects/shared-build-targets-utils/DependencyVersions.cs
ianhays: fixed.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
ianhays: It's invoked for crossgen later down the line. The RID is so we get the right version since the folders are different for arm64 (e.g. tools/x64_arm64/crossgen.exe instead of just tools/crossgen.exe)

PR #266, opened 2016-09-13
Add rid-moniker mappings for openSUSE 42.1 and Ubuntu 16.10

mellinoe:
I've missed a couple of places that need special handling for the new RIDs. The official builds failed because there wasn't an appropriate mapping for the new RIDs in this code here. Technically I had added it to the list that the build fails and complains about, but there is an additional mapping that is used to go from the actual RID to the key used in the other list (the "badge moniker").

cc: @brthor


brthor:
LGTM :shipit:

PR #273, opened 2016-09-15
Fix host tests by adding global.json, plus partial auto-update of CoreFx, CoreClr

dagood:
My two potential fixes are:
 1. Add a global.json to the host test to avoid referencing the projects under `src`.
 2. Undo `src` project dependency auto-update to allow `Microsoft.Extensions.DependencyModel.Tests` to succeed.

See https://github.com/dotnet/core-setup/pull/267 for details on the errors. I think only one of the two should be necessary, although (1) doesn't fix `Microsoft.Extensions.DependencyModel.Tests`, and (2) might be against the intention of the `src` projects' dependencies. Do the dependencies of `Microsoft.DotNet.PlatformAbstractions` and `Microsoft.Extensions.DependencyModel` need to be updated to latest, or can they stay at 1.0.0 RTM?

/cc @brthor @gkhanna79 @eerhardt @wtgodbe


pkg/deps/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "Microsoft.NETCore.Platforms": "1.0.1"
+    "Microsoft.NETCore.Platforms": "1.0.2-preview1-24513-01"
```
gkhanna79: @weshaggard Didnt this get updated with 1.1.0?

pkg/dir.props
```diff
@@ -27,7 +27,7 @@
   <PropertyGroup>
     <PackageOutputPath>$(PackagesOutDir)</PackageOutputPath>
     <SymbolPackageOutputPath>$(SymbolPackagesOutDir)</SymbolPackageOutputPath>
-    <RuntimeIdGraphDefinitionFile>$(ProjectDir)packages\Microsoft.NETCore.Platforms\1.0.1\runtime.json</RuntimeIdGraphDefinitionFile>
+    <RuntimeIdGraphDefinitionFile>$(ProjectDir)packages\Microsoft.NETCore.Platforms\1.0.2-preview1-24513-01\runtime.json</RuntimeIdGraphDefinitionFile>
```
gkhanna79: Same question here @weshaggard

test/Host/TestUtils/RepoDirectoriesProvider.cs
```diff
@@ -31,7 +31,7 @@ public class RepoDirectoriesProvider
             string corehostPackages = null,
             string corehostDummyPackages = null)
         {
-            _repoRoot = repoRoot ?? Path.Combine(Directory.GetCurrentDirectory(), "..", "..");
+            _repoRoot = repoRoot ?? Path.Combine(Directory.GetCurrentDirectory(), "..", "..", "..");
```
brthor: In hindsight, CWD isn't a fantastic thing to depend on, any ideas on how to improve this?

brthor:
LGTM :shipit:

build_projects/update-dependencies/UpdateFilesTargets.cs
```diff
@@ -87,7 +87,24 @@ public static BuildTargetResult ReplaceProjectJson(BuildTargetContext c)
...
+                "tools"
+            };
+
+            IEnumerable<string> projectJsonFiles = sourceDirsToUpdate
+                .SelectMany(name =>
+                    Directory.GetFiles(
+                        Path.Combine(Dirs.RepoRoot, name),
+                        "project.json",
+                        SearchOption.AllDirectories))
+                .ToArray();
```
dagood: Pointing out new logic I added to avoid upgrading `src` project.jsons. I tested it out, and it produces the exact upgrades that are in this PR.

test/Host/TestUtils/RepoDirectoriesProvider.cs
dagood: I don't know how to get rid of CWD, but to get rid of the `..`s this could be changed to look at parent directories until it hits one that "looks" like the repo root due to some file being present. (Similar to typical `dir.props` or `dir.targets` directory traversal.) `branchinfo.txt` seems like a decent target.

dagood:
@brthor @wtgodbe @gkhanna79 This merge triggered https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=332232

pkg/deps/project.json
weshaggard: Looks like the preview1 versions are still 1.0.2. The ones in master look like they were bumped to 4.3.0 with all the other packages. @ericstj does that sound right?

For the purpose of this repo it is only using this to get the runtimes to validate against so it actually doesn't matter much. I think we could probably even eliminate it completely now as BuildTools has a default version nowadays.

PR #281, opened 2016-09-16
Increase timeout for acquiring lease from 1 to 5 minutes

wtgodbe:
This should solve the 'lease expired' issue in the Windows signing jobs.


wtgodbe:
CC @dagood @gkhanna79 @brthor (we should have this for master as well as release/1.1.0)

gkhanna79:
Thanks for making this change. We should merge this once we have confirmed it works in release/1.1.0.

gkhanna79:
LGTM

brthor:
:shipit:

PR #286, opened 2016-09-16
In 1.1.0, change 1.0.1 to 1.1.0 for PlatformAbstractions and DependencyModel

dagood:
This was colliding with builds in `master`, causing build failures. Plus, these packages are really 1.1.0.

@wtgodbe @gkhanna79 @chcosta @brthor

### DIFF
--- a/src/Microsoft.DotNet.PlatformAbstractions/project.json
+++ b/src/Microsoft.DotNet.PlatformAbstractions/project.json
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.0.1-beta-*",
+  "version": "1.1.0-preview1-*",
   "buildOptions": {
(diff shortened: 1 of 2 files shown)


wtgodbe:
LGTM

gkhanna79:
Will this explain the FileLoadException we saw yesterday?

dagood:
@gkhanna79 Assuming you mean https://github.com/dotnet/core-setup/issues/275, I don't think so, because the issue is that the packages weren't being used and the p2p reference was.

gkhanna79:
Correct. So, was this a case of updates missed out by Maestro?

dagood:
This is an issue with which packages are produced, not consumed, so Maestro isn't involved.

I just noticed that the version here is also `beta`, not `preview1` like the other packages. I'll push a change to fix that too.

brthor:
LGTM :shipit:

PR #299, opened 2016-09-19
Enable auto-lease renewal for Azure blobs

gkhanna79:
@dagood @brthor @eerhardt @wtgodbe PTAL

### DIFF
--- a/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
+++ b/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
@@ -20,6 +20,9 @@ namespace Microsoft.DotNet.Cli.Build
         private static readonly string s_dotnetBlobRootUrl = "https://dotnetcli.blob.core.windows.net/dotnet/";
         private static readonly string s_dotnetBlobContainerName = "dotnet";
 
+        private Task _leaseRenewalTask = null;
+        private CancellationTokenSource _cancellationTokenSource = null;
+
         private string _connectionString { get; set; }
         private CloudBlobContainer _blobContainer { get; set; }
 
@@ -123,9 +126,18 @@ namespace Microsoft.DotNet.Cli.Build
                 try
                 {
                     CloudBlockBlob cloudBlob = _blobContainer.GetBlockBlobReference(blob);
-                    System.Threading.Tasks.Task<string> task = cloudBlob.AcquireLeaseAsync(TimeSpan.FromMinutes(5), null);
+                    System.Threading.Tasks.Task<string> task = cloudBlob.AcquireLeaseAsync(TimeSpan.FromMinutes(1), null);
                     task.Wait();
-                    return task.Result;
+
+                    string leaseID = task.Result;
+
+                    // Create a cancelabble task that will auto-renew the lease until the lease is released
+                     _cancellationTokenSource = new CancellationTokenSource();
+                     _leaseRenewalTask = Task.Run(() => 
+                        { AutoRenewLease(this, blob, leaseID); },
+                        _cancellationTokenSource.Token);
+
+                    return leaseID;
                 }
                 catch (Exception e)
                 {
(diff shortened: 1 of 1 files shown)


build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -134,14 +146,62 @@ public IEnumerable<string> ListBlobs(string virtualDirectory)
                 }
             }
 
+            ResetLeaseRenewalTaskState();
```
brthor: Are you cancelling the lease renewal task right after creating it?

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -134,14 +146,62 @@ public IEnumerable<string> ListBlobs(string virtualDirectory)
...
+        }
+
         public void ReleaseLeaseOnBlob(string blob, string leaseId)
         {
             CloudBlockBlob cloudBlob = _blobContainer.GetBlockBlobReference(blob);
             AccessCondition ac = new AccessCondition() { LeaseId = leaseId };
             cloudBlob.ReleaseLeaseAsync(ac).Wait();
+
+            // Cancel the lease renewal task
+            ResetLeaseRenewalTaskState();
```
brthor: I think we should reset the lease renewal state prior to releasing the lease. 

What happens if the renew task runs between releasing and canceling the task?

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
gkhanna79: We are about throw an exception and thus, do not expect any lease to be acquired. Hence, we reset the Task.

brthor: Seems like this should happen at the top of the function.

brthor: I missed throw directly underneath this, expected it was returning something. Makes sense

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
gkhanna79: We can handle both scenarios - but perhaps, cancelling ahead is better?

brthor: My thoughts too, avoid the scenario where that happens as much as possible.

There'll still be the small window of releasing and renewing at the same time if the renewal is already happening at the time of cancellation, although I'm not certain the best way to handle that. Does `.Cancel()` block until the thread actually cancels? Perhaps we should also wait for the thread to join when cancelling if not.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -189,19 +189,19 @@ private static void AutoRenewLease(AzurePublisher instance, string blob, string
...
-                
+
                 Thread.Sleep(waitFor);
             }
         }
 
         public void ReleaseLeaseOnBlob(string blob, string leaseId)
         {
+            // Cancel the lease renewal task since we are about to release the lease.
+            ResetLeaseRenewalTaskState();
```
brthor: Copied from the prior commit:

There'll still be the small window of releasing and renewing at the same time if the renewal is already happening at the time of cancellation, although I'm not certain the best way to handle that. Does .Cancel() block until the thread actually cancels? Perhaps we should also wait for the thread to join when cancelling if not.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -134,11 +146,59 @@ public IEnumerable<string> ListBlobs(string virtualDirectory)
...
+
+                Thread.Sleep(waitFor);
+            }
+        }
+
         public void ReleaseLeaseOnBlob(string blob, string leaseId)
         {
+            // Cancel the lease renewal task since we are about to release the lease.
+            ResetLeaseRenewalTaskState();
+
```
dagood: It seems like the lease should be its own class, otherwise there's only one lease per `AzurePublisher`.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -134,11 +146,59 @@ public IEnumerable<string> ListBlobs(string virtualDirectory)
...
+                    waitFor = maxWait;
+                }
+                catch (Exception e)
+                {
+                    Console.WriteLine($"Retrying lease renewal on {blob}, {e.Message}");
+                    waitFor = delay;
+                }
+
+                // If the task has been requested to be cancelled, then do so.
+                token.ThrowIfCancellationRequested();
```
dagood: Why throw vs. `IsCancellationRequested`?

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -123,9 +126,18 @@ public IEnumerable<string> ListBlobs(string virtualDirectory)
...
-                    System.Threading.Tasks.Task<string> task = cloudBlob.AcquireLeaseAsync(TimeSpan.FromMinutes(5), null);
+                    System.Threading.Tasks.Task<string> task = cloudBlob.AcquireLeaseAsync(TimeSpan.FromMinutes(1), null);
                     task.Wait();
-                    return task.Result;
+
+                    string leaseID = task.Result;
+
+                    // Create a cancelabble task that will auto-renew the lease until the lease is released
+                     _cancellationTokenSource = new CancellationTokenSource();
+                     _leaseRenewalTask = Task.Run(() =>
```
dagood: Extra space at beginning of lines.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -134,11 +146,59 @@ public IEnumerable<string> ListBlobs(string virtualDirectory)
...
+        {
+            // Cancel the lease renewal task if it was created
+            if (_leaseRenewalTask != null)
+            {
+                _cancellationTokenSource.Cancel();
+                _leaseRenewalTask = null;
+            }
+        }
+
+        private static void AutoRenewLease(AzurePublisher instance, string blob, string leaseId)
```
dagood: Why is this static?

gkhanna79: It is a pattern I use to have complete isolation from the callback usertype when callback is part of the same type.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
gkhanna79: Task Cancellation requires the exception to be thrown for Cancellation and this method helps with it.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
gkhanna79: Lease is already per AzurePublisher - I am not changing the existing semantics. We can refactor/make this better in a later change.

dagood: The way I see it, pre-change `( blob, leaseId )` forms a "lease" and there's no state in `AzurePublisher`. With this change, there's lease state in the `AzurePublisher` in a way that prevents two leases at the same time.

In any case... yeah, not important, since we don't currently need more than one lease.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
dagood: What kind of isolation do you get? (I haven't seen this pattern before.)

gkhanna79: The callback cannot assume anything about the instance of the user-type it was implemented in.

gkhanna79:
@dotnet-bot test this please

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
dagood: The only difference I see is `instance.` instead of an implicit `this.`--what does this prevent the implementation from assuming? (Willing to drop this because it's not related to this fix, but I'm very curious.)

gkhanna79:
Merging this PR to check if it addresses the leaseID mismatch problem.

PR #330, opened 2016-09-23
Disable parallel restore to avoid restore timeout failure (master)

dagood:
In the new version of CLI, `--disable-parallel` includes limiting parallel http requests. Docker container builds sometimes hit restore errors when there are too many parallel requests.

Addresses https://github.com/dotnet/core-setup/issues/327 for master.

Ports https://github.com/dotnet/core-setup/pull/323 from release/1.1.0.

@gkhanna79 @wtgodbe @chcosta @mellinoe

PR #355, opened 2016-09-25
Disable parallel during build script restore

dagood:
Applies `--disable-parallel` when restoring build script packages, fixing https://github.com/dotnet/core-setup/pull/330. I missed this instance earlier when I was rebasing the fix onto latest.

@gkhanna79 @wtgodbe


dagood:
Merging this master PR so we will start validating for https://github.com/dotnet/core-setup/pull/356, the release/1.1.0 port.

PR #362, opened 2016-09-26
Add creation of /etc/paths.d to host postinstall

### LINKED ISSUE #361: PKG postinall script for the host needs to create /etc/paths.d
On macOS Sierra, users have been unable to get `dotnet` on the path. After gracious help from @the-darkvoid, the root cause was narrowed down to the fact that on a clean install of Sierra, the `/etc/paths.d` path does not exist, and thus the `/etc/paths.d/dotnet` file cannot be created. 

The installer should include this step in the post-install script via `mkdir -p`.

blackdwarf:
This will make sure that the host is on the path regardless of whether
the version of macOS creates the path by default or not.

Fixes #361, dotnet/cli#4225

/cc @brthor @mellinoe @livarcocc @leecow @richlander

### DIFF
--- a/packaging/osx/sharedhost/scripts/postinstall
+++ b/packaging/osx/sharedhost/scripts/postinstall
@@ -11,6 +11,8 @@ INSTALL_DESTINATION=$2
 chmod 755 $INSTALL_DESTINATION/dotnet
 
 # Add the installation directory to the system-wide paths
+# But first create the directory if it doesn't exist
+mkdir -p /etc/paths.d
 echo $INSTALL_DESTINATION | tee /etc/paths.d/dotnet
 
 exit 0


packaging/osx/sharedhost/scripts/postinstall
```diff
@@ -11,6 +11,8 @@ INSTALL_DESTINATION=$2
 chmod 755 $INSTALL_DESTINATION/dotnet
 
 # Add the installation directory to the system-wide paths
+# But first create the directory if it doesn't exist
+mkdir -p /etc/paths.d
```
gkhanna79: Does the installer get looked up in the path with this fix? I ask to understand why Seirra does not have this folder by default while previous OSX did?

gkhanna79: @ramarag Please review this as well.

CC @schellap

blackdwarf: @gkhanna79 not sure I follow the question? But to your other question, I don't really know. I tried to research whether Sierra changed `path_helper` and associated artifacts in any way, but no information was forthcoming.

gkhanna79: Also, assuming this is confirmed to work, please port this to release/1.1.0 branch as well.

brthor:
test this

brthor:
test OSX x64 Release Build

packaging/osx/sharedhost/scripts/postinstall
blackdwarf: Confirmed it works. On fresh install the `/etc/paths.d` does not exist. After running the PKG with this change, the folder and the `dotnet` file in it got created and the host was on the path after a successful install.

blackdwarf:
If you are all OK with this change, can I get at least one :shipit: before I merge?

gkhanna79:
LGTM!

PR #387, opened 2016-09-28
Publish the sharedframework for netcoreapp1.1

mellinoe:
Right now, we're referencing the new Microsoft.NETCore.App package, but still publishing it for `netcoreapp1.0`. We should publish it for `netcoreapp1.1` by default.

@gkhanna79 , @brthor

### DIFF
--- a/build_projects/dotnet-host-build/build.ps1
+++ b/build_projects/dotnet-host-build/build.ps1
@@ -8,7 +8,7 @@ param(
     [string]$Architecture="x64",
     [string]$TargetArch="",
     [string]$ToolsetDir="",
-    [string]$Framework="netcoreapp1.0",
+    [string]$Framework="netcoreapp1.1",
     [string[]]$Targets=@("Default"),
     [string[]]$EnvVars=@(),
     [switch]$NoPackage,
--- a/build_projects/dotnet-host-build/PrepareTargets.cs
+++ b/build_projects/dotnet-host-build/PrepareTargets.cs
@@ -52,7 +52,7 @@ namespace Microsoft.DotNet.Host.Build
                     targetRID = $"win7-{RuntimeEnvironment.RuntimeArchitecture}";
                 }
             }
-            string targetFramework = Environment.GetEnvironmentVariable("TARGETFRAMEWORK") ?? "netcoreapp1.0";
+            string targetFramework = Environment.GetEnvironmentVariable("TARGETFRAMEWORK") ?? "netcoreapp1.1";
 
             if (string.IsNullOrEmpty(configEnv))
             {


gkhanna79:
LGTM

mellinoe:
I'm still running the build for this on my machine, let me get a clean build out locally to check before we merge this.

brthor:
Assuming you've done verification of the output nupkg. 

:shipit:

mellinoe:
There are other things that need to be changed in the repo. Trying to work through them...

build_projects/dotnet-host-build/build.ps1
```diff
@@ -8,7 +8,7 @@ param(
     [string]$Architecture="x64",
     [string]$TargetArch="",
     [string]$ToolsetDir="",
-    [string]$Framework="netcoreapp1.0",
+    [string]$Framework="netcoreapp1.1",
```
gkhanna79: I think you will need to change the corresponding build.sh as well.

mellinoe: `build.sh` doesn't have a parameter like this, actually. This is only used to set the "TARGETFRAMEWORK" environment variable (which is used in PrepareTargets.cs), not for anything else. On line 122, we always publish the build projects for `netcoreapp1.0` (this variable doesn't control that).

mellinoe:
@brthor What nuget package are you referring to? I don't believe this file has any nuget package outputs. It was my understanding that it was just used as a publish target to retrieve the shared framework assets, in order to zip them up, etc.

build_projects/dotnet-host-build/build.ps1
gkhanna79: The corresponding file to change is https://github.com/dotnet/core-setup/blob/f758ae62fba9b65ed2796f4ba04c49d696683c33/build_projects/dotnet-host-build/build.sh#L122

mellinoe:
Local build is still going with my second set of changes. I think we need to change the test projects as well because they are failing with cryptic messages.

build_projects/dotnet-host-build/build.ps1
mellinoe: Right, but that is publishing the build project itself, it doesn't control anything else in the build process that I can see. That line is analogous to line 122 in this file, mentioned above. I think we should still be publishing the build exe itself for `netcoreapp1.0` (well, it really doesn't matter what we use, as long as it works).

mellinoe:
Never mind, I guess the errors and failure messages printed out during the build are expected/normal, because the build succeeded in the end and CI has so far as well.

brthor:
My assumption was that you would also be changing the target TFM(s) in the Microsoft.NETCore.App nupkg.

https://github.com/dotnet/core-setup/blob/master/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj#L29

mellinoe:
I think we want to keep that version the same. Is that right, @weshaggard ?

build_projects/dotnet-host-build/build.ps1
weshaggard: If we are setting the TARGETFRAMEWORK enviroment variable then why do we even need the default in PrepareTargets.cs? Why not error if it isn't set?

(of course currently ignoring the fact that we are using env variables at all as I really had doing that).

weshaggard:
>I think we want to keep that version the same. Is that right, @weshaggard ?

Correct we currently wanted to keep that as netcoreapp1.0 as it is only controlling the dependencies.

build_projects/dotnet-host-build/build.ps1
mellinoe: I don't know the answer to those questions, but I'd rather just quickly fix the part that is messed up than rewire that whole part of the code.

Issue #378, opened 2016-09-28
Downgrade of host packages


ericstj:
## Steps to reproduce
Restore following project.json
```json
{
  "dependencies": {
    "Microsoft.NETCore.App": "1.1.0-preview1-001069-00"
  },
  "frameworks": {
    "netcoreapp1.0": { }
  },
  "runtimes": {
    "win7-x64": { }
  }
}
```

## Expected  behavior
No warnings

## Actual behavior
Warnings:
```
warn : Detected package downgrade: Microsoft.NETCore.DotNetHostResolver from 1.2.0-beta-000984-00 to 1.1.0-preview1-001069-00
warn :  testrestore (>= 1.0.0) -> Microsoft.NETCore.App (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHostPolicy (>= 1.1.0-preview1-001069-00) -> runtime.win7-x64.Microsoft.NETCore.DotNetHostPolicy (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHostResolver (>= 1.2.0-beta-000984-00)
warn :  testrestore (>= 1.0.0) -> Microsoft.NETCore.App (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHostPolicy (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHostResolver (>= 1.1.0-preview1-001069-00)
warn : Detected package downgrade: Microsoft.NETCore.DotNetHost from 1.2.0-beta-000984-00 to 1.1.0-preview1-001069-00
warn :  testrestore (>= 1.0.0) -> Microsoft.NETCore.App (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHostPolicy (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHostResolver (>= 1.1.0-preview1-001069-00) -> runtime.win7-x64.Microsoft.NETCore.DotNetHostResolver (>= 1.1.0-preview1-001069-00) -> Microsoft.NETCore.DotNetHost (>= 1.2.0-beta-000984-00)
[...]

PR #388, opened 2016-09-28
Update badge generation to be based on TargetArch

ianhays:
It's currently based on current architecture, so we're seeing a win10-arm64 package being created with a x64 badge, making the finalize step think that x64 has been built correctly already. This resulted in the x64 build being ignored repeatedly while the arm64 build was publishing correctly.

@gkhanna79 @dagood

should resolve https://github.com/dotnet/core-setup/issues/378 for master.

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -115,6 +115,7 @@ namespace Microsoft.DotNet.Host.Build
                     {
                         "win.x86.version",
                         "win.x64.version",
+                        "win.arm64.version",
                         "ubuntu.x64.version",
                         "ubuntu.16.04.x64.version",
                         "ubuntu.16.10.x64.version",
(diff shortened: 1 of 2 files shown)


dagood:
LGTM, thanks! I'll add arm64 back to the master build now so that when you merge this, it'll build the arm64 leg during the triggered build. I'll comment here once I've merged that change to the pipeline json.

dagood:
Merged https://devdiv.visualstudio.com/DevDiv/_git/DotNet-BuildPipeline/pullrequest/38608, enabling the arm64 leg in master.

ianhays:
Thanks @dagood!

dagood:
Looks like it worked: [Latest_Packages.txt](https://github.com/dotnet/versions/blob/88c6677ef1d8b962a8fb248896c610bc317bfcab/build-info/dotnet/core-setup/master/Latest_Packages.txt) ([build](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=352929)), and the badge names in Azure look ok with x64, x86, and arm64. Thanks again. :+1:

PR #401, opened 2016-09-30
Fix release/1.0.0 auto-update: remove -Channel to fix dotnet-install CLI url

dagood:
The `-Channel preview` argument was causing dotnet-install to guess the wrong URL to install a working CLI, resulting in a 404.

This is the fix that was implemented to enable auto-update on `release/1.1.0`, and `master` will need this problem fixed too.

@gkhanna79


gkhanna79:
Change looks fine to me.

dagood:
`pack.sh` was using the old nuget endpoint, causing it to be unable to find a package [in CI](http://dotnet-ci.cloudapp.net/job/dotnet_core-setup/job/release_1.0.0/job/release_ubuntu16.04_x64_prtest/9/). I'm updating it to the v3 feed.

PR #427, opened 2016-10-06
Download RTM CLI script in Core-Setup, instead of using local copy

wtgodbe:
CC @gkhanna79 @Sridhar-MS @brthor


wtgodbe:
1.1.0 may not want the RTM CLI version, I'll try this change as we have it in master, if this first attempt fails.

gkhanna79:
I think you want what is in master and pull down the right versions for OpenSuse 42.1 and Ubuntu 16.10.

wtgodbe:
@gkhanna79 what do you mean by "and pull down the right versions for OpenSuse 42.1 and Ubuntu 16.10"? Do we need something additional here to what we have in master? CI is coming through green for this, but I can switch the channel to rel-1.0.0/use the new CLI version, if that's the right thing.

gkhanna79:
CI green is not sufficient. Since this branch has pipeline builds for the new distros, we need to ensure this change does not break bringing down the "custom" CLI @mellinoe brought up for them.

CC @dagood

wtgodbe:
@gkhanna79 just spoke with Eric M, I've changed this PR to be identical to the one I put up in master, so we're using the latest CLI version with support for the new distros.

wtgodbe:
@gkhanna79 per conversation with @mellinoe I'm changing the channel to Sdk, and the version to the same version CoreFx is pinned to (this channel/version has CLIs for OpenSUSE42.1 and Ubuntu16.10, if this CI is green I'll submit the shiproom request & put up a PR in master)

mellinoe:
This looks correct to me based on our conversation.

PR #428, opened 2016-10-06
Removing ToLower workarounds in DependencyModel Resolution.

### LINKED ISSUE #423: Need to remove more "ToLowerInvariant" workarounds in DependencyModel
There are some additional ToLowerInvariant() workarounds in the DependencyModel that we need to remove:

https://github.com/dotnet/core-setup/blob/release/1.1.0/src/Microsoft.Extensions.DependencyModel/Resolution/ResolverUtils.cs#L16-L17

https://github.com/dotnet/core-setup/blob/release/1.1.0/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs#L55

Instead of calling ToLower to build up the path, this code should use the new "path" and "hashPath" properties introduced in the DependencyContext Libraries.

This will break ASP.NET apps on case-sensitive file systems.

@joelverhagen @gkhanna79 

/cc @pranavkm

eerhardt:
These workarounds are no longer valid.  The Resolution logic should be respecting Path and HashPath properties, and falling back to straight PackageId Version values, if those are not set.

Fix #423

@pranavkm @joelverhagen 

/cc @gkhanna79 @livarcocc

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
+++ b/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
@@ -51,8 +51,12 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
                 string packagePath;
                 if (ResolverUtils.TryResolvePackagePath(_fileSystem, library, _packageCacheDirectory, out packagePath))
                 {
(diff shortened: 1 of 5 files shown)


eerhardt:
@gkhanna79 - are you good with me merging into 'master'?  Once merged I will create a PR to port to 'rel/1.1.0' and send a ship room email.  Sound good?

gkhanna79:
Please go ahead and merge in master since you have the right signoffs from the folks :)

PR #433, opened 2016-10-07
Simplify repo versioning update

gkhanna79:
Until we switch to same build system as CoreCLR and CoreFX, we need to have a simpler and central way to version the repo. This change enables it by leveraging branchinfo.txt and augmenting it for new scenarios.

@schellap @wtgodbe @ramarag PTAL

CC @weshaggard @dagood


gkhanna79:
@schellap can you PTAL?

build_projects/dotnet-host-build/PrepareTargets.cs
```diff
@@ -118,25 +118,45 @@ public static BuildTargetResult GenerateVersions(BuildTargetContext c)
 
             var branchInfo = new BranchInfo(Dirs.RepoRoot);
 
+            // Read details from branchinfo.txt for our build configuration
+            int iMajor = int.Parse(branchInfo.Entries["MAJOR_VERSION"]);
+            int iMinor = int.Parse(branchInfo.Entries["MINOR_VERSION"]);
+            int iPatch = int.Parse(branchInfo.Entries["PATCH_VERSION"]);
+            string sReleaseSuffix = branchInfo.Entries["RELEASE_SUFFIX"];
+            bool fStabilizePackageVersion = bool.Parse(branchInfo.Entries["STABILIZE_PACKAGE_VERSION"]);
```
schellap: I'd just call this `USE_MNA_STABLE_VERSION` with  comment on what `MNA` is... so it is clear what package this is used for.

schellap: NIT: there is C# and then some Hungarian languages in between :)

gkhanna79: I named it generally because we could consider using it outside of MNA if required.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -63,10 +63,10 @@ public override string ToString()
         // Full versions and package information.
         public string LatestHostBuildMajor => CommitCountString;
         public string LatestHostBuildMinor => "00";
-        public bool EnsureStableVersion => false;
-        public VerInfo LatestHostVersion => new VerInfo(1, 2, 0, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
-        public VerInfo LatestHostFxrVersion => new VerInfo(1, 2, 0, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
-        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 2, 0, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+        public bool EnsureStableVersion { get; set; }
+        public VerInfo LatestHostVersion => new VerInfo(Major, Minor, Patch, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
```
schellap: For example, if you want to release 1.2.3 and you want to bump up the host packages due to a bug fix, you don't want to bump up the repo.

gkhanna79: In which scenario do you expect the versions of various packages, produced by the repo, to differ?

schellap: Let's say we are in a release band: `1.2.3-beta`, now the current host package is `1.2.3-beta-0001`. I make a breaking change which people don't want right away, but will pick up eventually in this release. In this case, I would simply bump up the host version to `1.2.4-beta-0002` within the same release `1.2.3-beta`.

gkhanna79: @schellap We will never make breaking changes in patch version updates. Also, completely breaking changes will only be done in major version updates (in sync with SemVer semantics). Also, when we are dealing with non-Zero patch version, we are pretty much in servicing for that major.minor version combination.

schellap: @ericstj do you have any inputs on this? Repo version will be used for MNA and Host packages, effectively, with this change.

weshaggard: From a simplification point of view I like having them be the same. We of course should make sure we have the flexibility to have different versions if needed.

gkhanna79: > make sure we have the flexibility to have different versions if needed.

The flexibility exists in the change and we should define it when the need arises. In general, we would expect the versions of the packages to be always updated in sync with the M.N.App version as releases move on.

PR #434, opened 2016-10-07
Use new version of CLI so we can get working CLI for OpenSUSE

wtgodbe:
Should fix the build break in our pipelines.


wtgodbe:
CC @gkhanna79 @mellinoe @brthor 

Eric - feel free to merge this after CI is green & the new CLIs are up (or just ping me)

mellinoe:
Why did you remove `-Channel Sdk` ?

wtgodbe:
@mellinoe, According to a conversation I had with @brthor yesterday, the version string is sufficient as a key (as opposed to in the past & in 1.0.0, when we needed both a version AND a channel to comprise a key). So `-Channel Sdk` was redundant, since I was already specifying the version (correct me if I'm wrong here, Bryan).

brthor:
@mellinoe in the `rel/1.0.0` version of the install scripts, channel and version are independent concepts, and you don't need both.

in the `rel/1.0.0-preview2` of the install scripts that isn't true though.

mellinoe:
Ok, sounds good. The right thing is getting downloaded in CI, so this LGTM.

wtgodbe:
https://devdiv.visualstudio.com/DevDiv/Net%20CoreFX/_build/explorer?_a=summary&buildId=366482 here's the triggered core-setup build

PR #446, opened 2016-10-11
Revert various changes from 1.0.2

gkhanna79:
This PR reverts various changes we took in 1.0.2 but should be in 1.0.3:

1) Networking fixes
2) RID fallback graph update for Sierra

Post this change, we only have the fixes for the OSX PATH issue on Sierra.

@weshaggard @ericstj PTAL

CC @leecow @schellap @Petermarcu @dagood @wtgodbe

PR #448, opened 2016-10-11
Stabilize M.N.App 1.0.2

gkhanna79:
Stabilizing 1.0.2 package versions for the PR at https://github.com/dotnet/core-setup/pull/446

### DIFF
--- a/TestAssets/TestProjects/PortableApp/project.json
+++ b/TestAssets/TestProjects/PortableApp/project.json
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.0.2-servicing-*"
+          "version": "1.0.2"
         },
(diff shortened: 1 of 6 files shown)


gkhanna79:
@weshaggard PTAL at this as well. This is about stabilizing 1.0.2 version.

@schellap FYI

weshaggard:
Looks like there are some other places we need to update:

See https://github.com/dotnet/core-setup/commit/c3e807d7c4bce240faab72a25ffa8596d2ebef70

Also depending on if we have any crossgen fixes we need to update the build step to point at the newer version as well.

gkhanna79:
@weshaggard Monikers were already updated to 1.0.2 - see https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/shared-build-targets-utils/Utils/Monikers.cs

We are not taking any CoreCLR fixes, since 1.0.1, so I am not taking any crossgen fixes.

CC @leecow

Issue #444, opened 2016-10-11
1.0.x host is removed when installing 1.1 preview


leecow:
/host/fxr/1.0.1 is removed when installing a 1.1.0 preview Shared FX. Previous versions of the host should remain installed side-by-side. 
1. Removed /usr/local/share/dotnet
2. Installed dotnet-osx-x64.1.0.1.pkg
3. Verified /host/fxr/1.0.1
4. Installed dotnet-osx-x64.1.1.0-preview1-001100-00
5. host/fxr/ now only contains 1.1.0-preview1-001100-00 

cc/ @gkhanna79

gkhanna79:
I chatted with @schellap and he is going to take a look at this asap.

gkhanna79:
@schellap Do you have an update on what is going on here?

gkhanna79:
@schellap Please prepare a fix for release/1.0.0 branch as well for the next LTS release.

gkhanna79:
@schellap Moving this to 1.0.x milestone to track getting this in for it.

gkhanna79:
Assigning this to you @rakeshsinghranchi and @vivmishra

gkhanna79:
@vivmishra Was this fixed with https://github.com/dotnet/core-setup/pull/1838? If so, please close the issue.

vivmishra:
This is related to OSx so not fixed by #1838

gkhanna79:
Is the fix for this issue understood?

rakeshsinghranchi:
Not yet, will try to repro the issue later today.

rakeshsinghranchi:
Issue does not repro against latest build daily build of 1.0.5 and 1.1.2. 

├── LICENSE.txt
├── ThirdPartyNotices.txt
├── dotnet
├── host
│   └── fxr
│       ├── 1.0.1
│       ├── 1.1.0
│       └── 2.0.0-beta-001611-00
└── shared
    └── Microsoft.NETCore.App
        ├── 1.0.5-servicing-004885-00
        ├── 1.1.2-servicing-001491-00
        └── 2.0.0-beta-001611-00

Below PRs should have fixed the issue:
https://github.com/dotnet/core-setup/pull/453
https://github.com/dotnet/core-setup/pull/452

PR #449, opened 2016-10-12
Add version information to hostfxr component

schellap:
Add correct version info for the hostfxr component ID.

### DIFF
--- a/build_projects/dotnet-host-build/PkgTargets.cs
+++ b/build_projects/dotnet-host-build/PkgTargets.cs
@@ -31,7 +31,9 @@ namespace Microsoft.DotNet.Host.Build
             Directory.CreateDirectory(PkgsIntermediateDir);
 
             SharedHostComponentId = $"com.microsoft.dotnet.sharedhost.component.osx.x64";
-            HostFxrComponentId = $"com.microsoft.dotnet.hostfxr.component.osx.x64";
+
+            string hostFxrVersion = c.BuildContext.Get<HostVersion>("HostVersion").LockedHostFxrVersion.ToString();
+            HostFxrComponentId = $"com.microsoft.dotnet.hostfxr.{hostFxrVersion}.component.osx.x64";
 
             string sharedFrameworkNugetName = Monikers.SharedFrameworkName;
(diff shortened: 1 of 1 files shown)


build_projects/dotnet-host-build/PkgTargets.cs
```diff
@@ -31,7 +31,9 @@ public static BuildTargetResult InitPkg(BuildTargetContext c)
             Directory.CreateDirectory(PkgsIntermediateDir);
 
             SharedHostComponentId = $"com.microsoft.dotnet.sharedhost.component.osx.x64";
-            HostFxrComponentId = $"com.microsoft.dotnet.hostfxr.component.osx.x64";
+
+            string hostFxrVersion = c.BuildContext.Get<HostVersion>("HostVersion").LockedHostFxrVersion.ToString();
```
gkhanna79: Have you confirmed other packages do not have this problem and that with this fix, the issue no longer repros?

gkhanna79:
Fixes https://github.com/dotnet/core-setup/issues/444

mellinoe:
To clarify, this variable controls what this templated string is replaced with, correct?

https://github.com/dotnet/core-setup/blob/master/packaging/osx/sharedframework/shared-framework-distribution-template.xml#L20

Previously, this ID was unversioned, so any sharedframework pkg installer would overwrite that folder because they had the same ID. Now, they will have distinct ID's with different version substrings.

Is that an accurate description of the fix?

gkhanna79:
Adding @brthor as well for review.

brthor:
@mellinoe That's an accurate description as I understand it. Assuming the shared framework is installing SxS this causes hostfxr to mimic that. 

LGTM.

schellap:
Not for review yet. Closing...

build_projects/dotnet-host-build/PkgTargets.cs
schellap: Yes, it looks good.

gkhanna79:
LGTM as well.

mellinoe:
@schellap The commit seems to be made with the wrong author email.

PR #454, opened 2016-10-13
Update UWP Metapackage to reference new Microsoft.Net.Native.Compiler 1.5.0 package

zamont:
skipciplease

### DIFF
--- a/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <PropertyGroup>
-    <Version>5.2.3</Version>
+    <Version>5.2.4-beta</Version>
     <PackagePlatform>AnyCPU</PackagePlatform>
   </PropertyGroup>
 
--- a/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <PropertyGroup>
-    <Version>5.2.4-beta</Version>
+    <Version>5.3.0</Version>
     <PackagePlatform>AnyCPU</PackagePlatform>
   </PropertyGroup>
 
--- a/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
+++ b/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "Microsoft.Net.Native.Compiler": "1.5.0",
     "Microsoft.NETCore.Portable.Compatibility": "1.0.2",
     "Microsoft.NETCore.Runtime.CoreCLR": "1.0.3",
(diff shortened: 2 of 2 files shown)


zamont:
@weshaggard @ericstj 
Bringing this to your attention, in case you have any concerns.

gkhanna79:
@schellap PTAL as well.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "Microsoft.Net.Native.Compiler": "1.5.0",
```
gkhanna79: IMHO, such a change should also force update of the metapackage version.

zamont: The intention is to update the metapackage version to 5.2.3 for this change. Are you saying that will happen automatically? Or do we need to do another process to update the version?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <PropertyGroup>
-    <Version>5.2.3</Version>
+    <Version>5.2.4-beta</Version>
```
zamont: @gkhanna79 We want a prerelease nuget package so we appended beta to the 5.2.4 version. Look good?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
gkhanna79: You need to specify the version explicitly.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
gkhanna79: For the versioning for UWP metapackage, I would defer to @weshaggard and @ericstj.

In general, you are adding something fundamentally new, so updating the patch version is not sufficient. At minimum, we need to update the minor version.

Also, when do you expect to have this version of the package published and who is going to consume it and how?

zamont: @yizhang82 can correct me if I'm wrong on any of the details, but we are planning on publishing the metapackage/Microsoft.Net.Native.Compiler package at the time of Dev15 RC. The in-box template will specify version 5.2.2 of the metapackage, so developers who want to try our NuGet toolchain experience will have to manually update their metapackage.

gkhanna79: > we are planning on publishing the metapackage/Microsoft.Net.Native.Compiler package

Since UWP metapackage is built and published from the open, I am not aware of a release corresponding to which this updated version is going to be published in Dev15 RC timeframe.

@weshaggard What are your thoughts on this?

weshaggard: I agree we should bump the minor version and we also don't need to add the "-beta" as that comes from somewhere else. So this should just be "5.3.0". 

As for releasing this we will not release this until it is ready and at that point we can do it from the most appropriate release branch.

weshaggard:
```
/mnt/resource/j/workspace/dotnet_core-setup/master/debug_centos7.1_x64_prtest/pkg/Tools/Packaging.targets(812,5): error MSB4018: The "ApplyPreReleaseSuffix" task failed unexpectedly. [/mnt/resource/j/workspace/dotnet_core-setup/master/debug_centos7.1_x64_prtest/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj]
```

CI is failing because of "-beta" in your version string which is wrong. Please remove that, and don't skip CI as it needs to pass.

zamont:
Does this all check out now?

Issue #456, opened 2016-10-14
M.N.App 1.1.0 should have netcoreapp1.1 as TFM


gkhanna79:

weshaggard:
Fixed with #464

PR #462, opened 2016-10-20
Update shared framework to target netcoreapp1.1

weshaggard:
@ericstj can you please have a look Microsoft.NETCore.App changes. I essentially copied all the dependencies from the what is in release/1.0.0 branch which we will need to keep updated with whatever versions we bump in that branch. I was considering using your harvesting but I didn't figure it would map easily in this repo setup for meta-packages.

@dagood this PR replaces https://github.com/dotnet/core-setup/pull/460. 

@gkhanna79 @mellinoe could you guys please take a look.

PR #464, opened 2016-10-20
Update Microsoft.NETCore.App to target netcoreapp1.1

weshaggard:
Fix https://github.com/dotnet/core-setup/issues/456

cc @ericstj @gkhanna79 

Right now the test projects in this branch are still using netcoreapp1.0 as the target framework. I'm still debating whether or not to try and update those to netcoreapp1.1 like I'm doing in master with https://github.com/dotnet/core-setup/pull/462. What do you guys think?

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.NETCore.DotNetHost\Microsoft.NETCore.DotNetHostPolicy.pkgproj">
-      <TargetFramework>.NETCoreApp1.0</TargetFramework>
+      <TargetFramework>.NETCoreApp1.1</TargetFramework>
     </ProjectReference>
   </ItemGroup>
 
@@ -26,7 +26,7 @@
 
       <Dependency Include="@(PackageMatch -> '%(PackageId)')" Condition="'%(PackageMatch.PackageVersion)' != ''">
         <Version>%(PackageMatch.PackageVersion)</Version>
-        <TargetFramework>.NETCoreApp1.0</TargetFramework>
+        <TargetFramework>.NETCoreApp1.1</TargetFramework>
       </Dependency>
 
       <!-- List of dependencies to mark as exclude=compile -->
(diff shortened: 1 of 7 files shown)


weshaggard:
looks like I'm going to need to update the test projects (not sure why this didn't fail locally for me)
```
   One or more packages are incompatible with .NETCoreApp,Version=v1.0 (ubuntu.16.10-x64).
```

because they have the new RIDs are only supported in netcoreapp1.1 packages but they have these new runtimes. @mellinoe do you think we should split the tests (i.e. one with the new RIDS targeting netcoreapp1.1 and one without the new RIDS targeting netcoreapp1.0) or should we just bump things to 1.1. I'm leaning towards just bumping them to 1.1 but I'm curious to see what others think.

mellinoe:
I initially had tried retargeting all the tests netcoreapp1.1, but hit some other failures. I believe that some of the tests are \ hardcoded to assume that "netcoreapp1.0" is the one and only TFM for the tests, so you might end up hitting a couple of errors like that. But yeah, we will need to retarget the tests to `netcoreapp1.1`.

gkhanna79:
I concur with bumping the version.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -40,12 +40,152 @@
...
       <File Include="$(PlaceholderFile)" >
         <SkipPackageFileCheck>true</SkipPackageFileCheck>
         <TargetPath>lib/netcoreapp1.0</TargetPath>
       </File>
     </ItemGroup>
   </Target>
 
+  <!--
+    Manually include the netcoreapp1.0 dependencies.
+    If any versions change in the release/1.0.0 branch they should
```
gkhanna79: This looks extremely fragile as it will be one more manual task to account for. Can we not skip having them?

weshaggard: What do you suggest? We need to keep the dependencies consistent across versions. Ideally we would just pull them from the latest 1.0 meta-package but that requires a fair amount of work to make happen.

gkhanna79: Perhaps I am missing something - why do we need to keep them in 1.1.0 branch? Who needs these in this branch?

weshaggard: We need to produce a Microsoft.NETCore.App meta-package that will support both netcoreapp1.0 and netcoreapp1.1 and in order to do that we need to have to dependency groups. The new netcoreapp1.1 dependency set and the netcoreapp1.0 dependency set which tracks the release/1.0.0 setup. Without this people will not be able to update their Microsoft.NETCore.App meta-package and still resolve for netcoreapp1.0.

mellinoe: Can we move all of the dependencies into the pkgproj like this, instead of doing the crazy regex-matching stuff above? If we did that, we wouldn't need to be translating from one to the other in a new version, merely adding new stuff.

weshaggard: @mellinoe I would love to kill the regex hack, but we need the dependencies to still get updated by the auto-update PR's which currently in core-setup we haven't added the necessary support to update them in the pkgproj (@dagood can confirm that assumption). Even if we do that we don't gain much in this scenario as we need the netcoreapp1.0 and netcoreapp1.1 to have different versions.

weshaggard:
@gkhanna79 please go ahead and merge this when release/1.1.0 is ready to take changes.

mellinoe:
Since the Jenkins change isn't in yet (https://github.com/dotnet/core-setup/pull/466), have you been able to see if this works on Ubuntu 16.10 / openSUSE 42.1?

weshaggard:
>Since the Jenkins change isn't in yet (#466), have you been able to see if this works on Ubuntu 16.10 / openSUSE 42.1?

No but for this 1.1 change I didn't need to update the CLI, that is only part of the master PR. So my expectation is that this shouldn't be an issue.

PR #478, opened 2016-10-22
Move Microsoft.NETCore.App to 1.0.3

gkhanna79:
With the 1.0.2 LTS out, move Core-Setup and M.N.App to 1.0.3.

@weshaggard @schellap PTAL

@dagood Once we merge this PR, we will need to force a maestro update in CoreSetup 1.0.0 to pickup the latest packages from CoreCLR and CoreFX 1.0.0 branches.

FYI - @Petermarcu @leecow

PR #500, opened 2016-10-24
Publish PlatformAbstractions and DependencyModel packages to Azure

wtgodbe:
CC @gkhanna79 @dagood @livarcocc 

This change should allow us to publish the Microsoft.Dotnet.PlatformAbstractions and Microsoft.Extensions.DependencyModel packages to Azure - from there they'll be picked up by core-setup's finalizer step and published to MyGet, meaning we'll need to disable the 'nuget publisher' step in the Windows builds. To account for only wanting to publish the signed, Windows versions of these packages, we'll set a variable named "SIGNED_PACKAGES" to 'true' in the Windows signing build definitions, and 'false' in all others. 

Also, does anyone have an idea for a better target name than 'PublishSignedPackages'?

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -230,6 +230,7 @@ namespace Microsoft.DotNet.Host.Build
             nameof(PublishTargets.PublishDotnetDebToolPackage),
             nameof(PublishTargets.PublishDebFilesToDebianRepo),
             nameof(PublishTargets.PublishCoreHostPackages),
+            nameof(PublishTargets.PublishManagedPackages),
             nameof(PublishTargets.PublishSharedFrameworkVersionBadge))]
         public static BuildTargetResult PublishArtifacts(BuildTargetContext c) => c.Success();
 
@@ -274,6 +275,22 @@ namespace Microsoft.DotNet.Host.Build
                 var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}/{Path.GetFileName(file)}";
                 AzurePublisherTool.PublishFile(hostBlob, file);
                 Console.WriteLine($"Publishing package {hostBlob} to Azure.");
+            }
+
+            return c.Success();
+        }
+
+        [Target]
+        public static BuildTargetResult PublishManagedPackages(BuildTargetContext c)
+        {
+            if (EnvVars.Signed)
+            {
+                foreach (var file in Directory.GetFiles(Dirs.Packages, "*.nupkg"))
+                {
+                    var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}/{Path.GetFileName(file)}";
+                    AzurePublisherTool.PublishFile(hostBlob, file);
+                    Console.WriteLine($"Publishing package {hostBlob} to Azure.");
+                }
             }
 
             return c.Success();
--- a/build_projects/shared-build-targets-utils/Utils/EnvVars.cs
+++ b/build_projects/shared-build-targets-utils/Utils/EnvVars.cs
@@ -7,6 +7,8 @@ namespace Microsoft.DotNet.Cli.Build
     {
         public static readonly bool Verbose = GetBool("DOTNET_BUILD_VERBOSE");
 
+        public static readonly bool Signed = GetBool("SIGNED_PACKAGES");
+
         private static bool GetBool(string name, bool defaultValue = false)
         {
             var str = Environment.GetEnvironmentVariable(name);


build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -280,6 +281,23 @@ public static BuildTargetResult PublishCoreHostPackages(BuildTargetContext c)
         }
 
         [Target]
+        public static BuildTargetResult PublishSignedPackages(BuildTargetContext c)
+        {
+            string signedPackages = EnvVars.EnsureVariable("SIGNED_PACKAGES");
+            if (signedPackages.Equals("true"))
```
dagood: I didn't notice this earlier, but there's also a `EnvVars.GetBool` that would do a more extensive check, should use that I think: https://github.com/dotnet/core-setup/blob/eca152525e19a41d489cd5845afa14982653cc0a/build_projects/shared-build-targets-utils/Utils/EnvVars.cs#L10. Also allows null/empty, so we'd only need the var on the one definition we want to do this step.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -230,6 +230,7 @@ private static bool CheckIfAllBuildsHavePublished()
             nameof(PublishTargets.PublishDotnetDebToolPackage),
             nameof(PublishTargets.PublishDebFilesToDebianRepo),
             nameof(PublishTargets.PublishCoreHostPackages),
+            nameof(PublishTargets.PublishSignedPackages),
```
dagood: Hmm, would it be accurate to say `PublishManagedPackages`? A bit more exact to what's being published... although it misses out on the reason. Not sure.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -280,6 +281,23 @@ public static BuildTargetResult PublishCoreHostPackages(BuildTargetContext c)
         }
 
         [Target]
+        public static BuildTargetResult PublishSignedPackages(BuildTargetContext c)
+        {
+            string signedPackages = EnvVars.EnsureVariable("SIGNED_PACKAGES");
+            if (signedPackages.Equals("true"))
+            {
+                foreach (var file in Directory.GetFiles(Dirs.Packages, "*.nupkg"))
```
gkhanna79: Presumably, this only affects the two packages and not the existing host packages, right?

scripts/dockerrun.sh
```diff
@@ -115,6 +115,7 @@ docker run $INTERACTIVE -t --rm --sig-proxy=true \
     -e GITHUB_PASSWORD \
     -e CLI_NUGET_FEED_URL \
     -e CLI_NUGET_API_KEY \
+    -e SIGNED_PACKAGES \
```
dagood: I think this should be given the full name, `PUBLISH_SIGNED_PACKAGES` (or `PUBLISH_MANAGED_PACKAGES`) because I believe it still builds them. (But with `EnvVars.GetBool` this line shouldn't be needed.)

gkhanna79:
LGTM modulo a comment. Once verified, please port change to 1.1.0 branch as well.

build_projects/dotnet-host-build/PublishTargets.cs
wtgodbe: Yes, the existing host packages have their publish logic defined here - https://github.com/wtgodbe/core-setup/blob/c504f38fbb36f4f252c65a2f9bc7aa7e3c27bfb3/build_projects/dotnet-host-build/PublishTargets.cs#L271

build_projects/dotnet-host-build/PublishTargets.cs
wtgodbe: Good call, I'll switch that now.

build_projects/dotnet-host-build/PublishTargets.cs
wtgodbe: I think identifying what's being published is more important than identifying the reason - I'll change this

wtgodbe:
@gkhanna79 will do once CI is green. We should be seeing the same thing in 1.0.0, should I port it to there as well?

gkhanna79:
@wtgodbe Yes, I believe we should do it there as well for completeness (assuming these are built there too - can you confirm?).

wtgodbe:
@gkhanna79 yes, we're building these packages there as well

wtgodbe:
https://devdiv.visualstudio.com/DevDiv/Net%20CoreFX/_build/explorer?_a=summary&buildId=393701 If this comes out green I'll put up PRs for 1.1.0 and 1.0.0 (I'll submit to shiproom for 1.1.0, do I need to do the same for 1.0.0 @gkhanna79?)

gkhanna79:
@wtgodbe Yes, same for 1.0.0. However, do not merge the 1.0.0 PR until I merge https://github.com/dotnet/core-setup/pull/478.

wtgodbe:
Build was green, and I've confirmed the packages in question were published to Azure, alongside the pre-existing packages! I'll put through the 1.1.0 and 1.0.0 PRs now.

PR #502, opened 2016-10-24
[release/1.1.0] Enable Fedora 24 builds

mellinoe:
Fedora 24 is the last new distro we plan to support in 1.1.0. This adds the packaging projects and build work to make that happen. I needed to update the version of CLI we are downloading in order to make a small fix in the Fedora 24 bootstrap version. The contents of the 003223-6 folder are the same as the 003223-3 folder (previously used), except for that small Fedora 24 fix. We need to put everything in a new folder because we are using a "caching" service for our downloads, and it doesn't correctly handle blobs that have been changed or deleted.


gkhanna79:
LGTM

Issue #515, opened 2016-10-25
[release/1.1.0] Build code doesn't handle fedora.24 rid in all places


mellinoe:
The official build uses a couple of extra code paths to handle asset publishing. We need to fix a couple of extra places to handle the new platform name.

mellinoe:
Should be fixed by #516

PR #516, opened 2016-10-25
[release/1.1.0] Add handling for Fedora 24 in some build code.

mellinoe:
`CheckIfAllBuildsHavePublished()` fails if it encounters a name that it doesn't know about. In our case, it was failing because the new configuration's name wasn't listed in here. There was also another piece of code that I noticed which controls some moniker strings. It needed to be special-cased for Fedora 24.

@gkhanna79


mellinoe:
Fixes #515

gkhanna79:
LGTM

PR #520, opened 2016-10-26
Add registry keys to allow WER to create proper minidump

alsemenn:
Environment variables will be set to allow WER to generate proper minidumps if app fails.
See: https://github.com/dotnet/coreclr/issues/7722 for more info

### DIFF
--- a/packaging/windows/sharedframework/registrykeys.wxs
+++ b/packaging/windows/sharedframework/registrykeys.wxs
@@ -13,6 +13,32 @@
           <RegistryValue Action="write" Name="1.0.0-rc2" Type="integer" Value="1" KeyPath="yes"/>
         </RegistryKey>
       </Component>
+
+<!-- Set registry keys to allow WER to genereate correct dumps-->
+      <Component Directory="TARGETDIR">
+        <RegistryKey Root="HKLM" Key="SOFTWARE\Microsoft\Windows\Windows Error Reporting\RuntimeExceptionHelperModules">
+          <RegistryValue Action="write" Name="[DOTNETHOME]shared\$(var.FrameworkName)\$(var.FrameworkDisplayVersion)\mscordaccore.dll" Type="integer" Value="0" KeyPath="yes"/>
+        </RegistryKey>
+      </Component>
+      
+      <Component Directory="TARGETDIR">
+        <RegistryKey Root="HKLM" Key="SOFTWARE\Microsoft\Windows NT\CurrentVersion\KnownManagedDebuggingDlls">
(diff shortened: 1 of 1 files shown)


alsemenn:
The error is not due to my changes.

gkhanna79:
@dotnet-bot test Ubuntu16.04 x64 Release Build  please

gkhanna79:
@alsemenn Are you planning to get this change in master as well?

packaging/windows/sharedframework/registrykeys.wxs
```diff
@@ -13,6 +13,26 @@
           <RegistryValue Action="write" Name="1.0.0-rc2" Type="integer" Value="1" KeyPath="yes"/>
         </RegistryKey>
       </Component>
+
+<!-- Set registry keys to allow WER to genereate correct dumps-->
+      <Component Directory="TARGETDIR">
```
Petermarcu: We typically have one component per registry entry.

packaging/windows/sharedframework/registrykeys.wxs
```diff
@@ -13,6 +13,26 @@
...
+        <RegistryKey Root="HKLM" Key="SOFTWARE\Microsoft\Windows\Windows Error Reporting\RuntimeExceptionHelperModules">
+          <RegistryValue Action="write" Name="[DOTNETHOME]shared\$(var.FrameworkName)\$(var.FrameworkDisplayVersion)\mscordaccore.dll" Type="integer" Value="0" KeyPath="yes"/>
```
Petermarcu: What does $(var.FrameworkDisplayVersion) resolve to? Is it the 2 part or 3 part version?

ericstj: Typically we use [#FileID] syntax when pointing at files in the same installer.  Same for below.

ericstj: Peter let me know this is harvested, so that might be a little difficult.  Still if harvested it should have a stable ID so you could use those, but it'd be harder to read.

Petermarcu:
Lets just get these keys split into their own components and I think we'll be good to go.

packaging/windows/sharedframework/registrykeys.wxs
alsemenn: Changed.

packaging/windows/sharedframework/registrykeys.wxs
alsemenn: 1. this is "1.1.0-preview1-001104-00" part. 
2. I didn't find if I can get a full path without constructing it, but I'm not familiar with the WIX tool nor with the code base. So if there is any way to get them, I'll change to use it.

alsemenn:
I'll do the same changes to the master branch after this PR is merged.

Petermarcu:
This looks good for 1.1. For the commit to master, can you also group these related components into a component group instead of adding them to the same component group as the detection keys?

PR #526, opened 2016-10-26
Enable WinArm32 build

gkhanna79:
This change enables building Core-Setup for WinArm32.


gkhanna79:
@schellap @ianhays PTAL

CC @dagood @wtgodbe

PR #527, opened 2016-10-26
Enable the Fedora 24 build in master

mellinoe:
This is just a port of the relevant commits from release/1.1.0 into master that enable building for Fedora 24.


gkhanna79:
LGTM.

CC @schellap @ramarag

Issue #436, opened 2016-10-08
Published apps running in the root of a drive fail


billwert:
This failure seems to be related to the loop at [AppDomain.cs:3665](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/AppDomain.cs#L3665). Specifically, when a program is run from the root of the drive, APP_PATHS ends up being just the drive root (C: or whatever.) This fails Path.IsRelative and tears down the process. It seems we're too early for the managed exception to be surfaced usefully so the output is not helpful.

At first glance I thought this might be related to dotnet/cli#3697; not sure. I don't know if 'dotnet run' also fails due to this bug, as 'dotnet run' fails due to dotnet/cli#3697.

I am guessing this is a host bug as I believe the host is responsible for passing this information in.

## Steps to reproduce
0) mkdir test
1) dotnet new
2) modify project.json to publish stand alone for win10-x64
3) dotnet publish -r win10-x64 
4) subst s: <path-to-publish>
5) s:\test.exe
## Expected  behavior
program runs

## Actual behavior
D:\work\test>D:\work\test\bin\Debug\netcoreapp1.0\win10-x64\publish\test.exe
Hello World!

D:\work\test>subst t: D:\work\test\bin\Debug\netcoreapp1.0\win10-x64\publish

D:\work\test>t:\test.exe
Failed to initialize CoreCLR, HRESULT: 0x80070057

## Environment data
`dotnet --info` output:

gkhanna79:
Moving this to 1.2.0

PR #529, opened 2016-10-27
Fix get_directory to return non-empty directories

schellap:
@gkhanna79 PTAL.

The problem was get_directory was returning empty string leaving it to the callers to append the separator. It is better if it can be resilient to trailing separators, so it can append the separator when yielding a directory.

### DIFF
--- a/src/corehost/common/utils.cpp
+++ b/src/corehost/common/utils.cpp
@@ -129,14 +129,24 @@ pal::string_t get_filename(const pal::string_t& path)
 
 pal::string_t get_directory(const pal::string_t& path)
 {
-    // Find the last dir separator
-    auto path_sep = path.find_last_of(DIR_SEPARATOR);
-    if (path_sep == pal::string_t::npos)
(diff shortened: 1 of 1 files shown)


schellap:
Fixes #436

src/corehost/common/utils.cpp
```diff
@@ -129,14 +129,24 @@ pal::string_t get_filename(const pal::string_t& path)
...
+        ret.pop_back();
+    }
+
     // Find the last dir separator
-    auto path_sep = path.find_last_of(DIR_SEPARATOR);
+    auto path_sep = ret.find_last_of(DIR_SEPARATOR);
     if (path_sep == pal::string_t::npos)
     {
-        return pal::string_t(path);
+        return path;
```
schellap: PTAL

gkhanna79:
LGTM.

gkhanna79:
@schellap Should this be merged and the issue closed?

PR #530, opened 2016-10-27
Fix get_directory to return root dir with path separators

schellap:
Port of #529 

Cc @gkhanna79


schellap:
Also use fix from PR #534

gkhanna79:
Closing since this does not meet the bar.

Issue #229, opened 2016-07-27
Portable apps for binary-compatible RIDs


ericstj:
Opened from discussion on https://github.com/dotnet/corefx/pull/10348.

Consider a RID like `ubuntu.16.10-x64`.  We don't specifically cross-compile for this, instead we use a previous build of ubuntu, `ubuntu.16.04-x64`, and that's expected to work for coreclr/corefx/core-setup native components.

Folks upstack, however, can use the `ubuntu.16.10-x64` RID in their packages if they need to distinguish their own assets for this if necessary.  This will work for standalone since the user will directly choose that RID.  I don't think it works for portable apps.

Since the host hardcodes the "active RID" as the RID it was built against, `ubuntu.16.04-x64` in this case, how do we expect a portable app to ever get the running-on RID for a binary compatible distro?
https://github.com/dotnet/core-setup/blob/42de2408f4707f0ae9bba5c2e5470a3ede284f27/src/corehost/cli/deps_format.cpp#L99

/cc @mellinoe @livarcocc @ellismg @gkhanna79 @schellap

ellismg:
Just to be clear, for `ubuntu.16.10-x64` we do expect to cross compile, correct?  Looking at the existing runtime.json file, it looks like that is what we expect.  However, for something like `linuxmint.18-x64` which is parented to `ubuntu.16.04-x64`, we have this problem?

mellinoe:
Thanks for opening this.

> Just to be clear, for ubuntu.16.10-x64 we do expect to cross compile, correct?

Yep. linuxmint18 is the only one we can re-use packages for, I think.

> However, for something like linuxmint.18-x64 which is parented to ubuntu.16.04-x64, we have this problem?

Yes, I think so, and also for existing "binary-compatible rids", like linuxmint17 -> 17.3. I'm guessing that the host will just see these as ubuntu14.04 currently.

schellap:
If 2 rids are binary compatible, then wouldn't the package already provide the assets in the lowest compatible rid (ex: `win7-x64`), in which case the lower rid in the host will select these assets?

If this is not the case, for compatible rids, the host has to be built/released on the highest rid machines from the lab, so its own baked-in rid can be used to lookup into the rid fallback graph as the starting node. Otherwise, a lower rid (baked in) host will fail to use assets in a package that simultaneously provides higher compatible rids.

The runtime graph in M.N.A.deps.json will need to reflect these newer rids and the precedence in the `runtimes` section.

[...]

PR #531, opened 2016-10-27
Support Sierra as the base RID for the osx host

schellap:
@gkhanna79 PTAL.

### DIFF
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -124,6 +124,22 @@ void deps_json_t::reconcile_libraries_with_targets(
 
 pal::string_t get_own_rid()
 {
+    // For OSX, try to dynamically detect OS version and use appropriate
+    // starting RID. This is to support Sierra even though host binaries
+    // are built on El-Cap and Yosemite.
+#if defined(__APPLE__)
+    pal::os_moniker_t moniker;
+    if (pal::get_os_moniker(&moniker)) {
+        if (moniker >= pal::os_moniker_t::sierra) {
+            return pal::string_t("osx.10.12-") + get_arch();
+        }
+        if (moniker >= pal::os_moniker_t::el_capitan) {
+            return pal::string_t("osx.10.11-") + get_arch();
+        }
+    }
+#endif
+
+    // Otherwise, default to RID supplied during build time of the binary.
 #if defined(TARGET_RUNTIME_ID)
     return _STRINGIFY(TARGET_RUNTIME_ID);
 #else
--- a/src/corehost/common/pal.h
+++ b/src/corehost/common/pal.h
@@ -174,6 +174,16 @@ namespace pal
     inline bool clr_palstring(const char* cstr, pal::string_t* out) { out->assign(cstr); return true; }
 #endif
 
+#if defined(__APPLE__)
+    enum os_moniker_t
+    {
+        sierra = 16,
+        el_capitan = 15,
+    };
+
+    bool get_os_moniker(os_moniker_t* moniker);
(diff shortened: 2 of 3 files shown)


src/corehost/cli/deps_format.cpp
```diff
@@ -124,6 +124,20 @@ void deps_json_t::reconcile_libraries_with_targets(
...
+#if defined(__APPLE__)
+    pal::os_moniker_t moniker;
+    if (pal::get_os_moniker(&moniker)) {
+        if (moniker >= os_moniker_t::sierra) {
+            return pal::string_t("osx.10.12-") + get_arch();
+        }
+        if (moniker >= os_moniker_t::el_capitan) {
+            return pal::string_t("osx.10.11-") + get_arch();
+        }
+        if (moniker >= os_moniker_t::yosemite) {
```
gkhanna79: We support OSX 10.11 as the minimum OS, so we shouldnt need the check for 10.10.

src/corehost/cli/deps_format.cpp
```diff
@@ -124,6 +124,20 @@ void deps_json_t::reconcile_libraries_with_targets(
...
+            return pal::string_t("osx.10.10-") + get_arch();
+        }
+    }
+#endif
```
gkhanna79: Also, this should be "#else" and the next block of "#if" inside it.

src/corehost/common/pal.h
```diff
@@ -174,6 +174,17 @@ namespace pal
     inline bool clr_palstring(const char* cstr, pal::string_t* out) { out->assign(cstr); return true; }
 #endif
 
+#if defined(__APPLE__)
+    enum os_moniker_t
+    {
+        sierra = 16,
```
gkhanna79: I presume you confirmed this value?

src/corehost/common/pal.unix.cpp
```diff
@@ -169,6 +169,30 @@ bool pal::get_default_servicing_directory(string_t* recv)
...
+    size_t pos = s.find('.');
+    if (pos == string::npos)
+    {
+        return false;
+    }
+
+    *moniker = (os_moniker_t) stoi(s.substr(0, pos));
+    return true;
+}
+#endif
```
gkhanna79: You can remove the "#endif" and the following "#ifdef" since the _APPLE_ block is continuous.

src/corehost/cli/deps_format.cpp
gkhanna79: I see - you are trying to do fallback to use build time RID incase getosmoniker fails, is that right?

src/corehost/common/pal.unix.cpp
schellap: I introduced a new ifdef, as we are going to need these monikers for other platforms if we do dynamic detection.

src/corehost/common/pal.h
schellap: Yes, on @leecow box. Try out:

`> sysctl kern.osversion` and `osrelease`

src/corehost/cli/deps_format.cpp
schellap: Right.

src/corehost/common/pal.h
```diff
@@ -174,6 +174,17 @@ namespace pal
...
+        el_capitan = 15,
+        yosemite = 14
```
gkhanna79: Please remove the "yosemite" enum as well.

gkhanna79:
LGTM

schellap:
Partially fixes #229

PR #541, opened 2016-10-28
Upgrade nss in the Fedora 23 dockerfile

mellinoe:
This solved all of the "download timeout" problems we were seeing on Fedora 24. It's possible it will help us out on Fedora 23 as well.

@dagood, @gkhanna79


gkhanna79:
LGTM. 

@Petermarcu We will need this for an engineering infra reliability.

PR #550, opened 2016-11-01
Generate R2R images in temp folder

gkhanna79:
This will avoid locking issues when generating R2R images.

@schellap @ramarag PTAL.

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/Crossgen.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Crossgen.cs
@@ -198,6 +198,9 @@ namespace Microsoft.DotNet.Cli.Build
                 { "COMPlus_PartialNGen", "0" }
             };
 
+            // Form the dynamic path that would not collide if another instance of this is running.
+            string basePath = Path.Combine(Dirs.Intermediate, Guid.NewGuid().ToString());
+
             foreach (var file in Directory.GetFiles(pathToAssemblies))
             {
                 string fileName = Path.GetFileName(file);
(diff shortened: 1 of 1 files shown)


build_projects/shared-build-targets-utils/Utils/Crossgen.cs
```diff
@@ -198,6 +198,9 @@ public void CrossgenDirectory(string sharedFxPath, string pathToAssemblies)
                 { "COMPlus_PartialNGen", "0" }
             };
 
+            // Form the dynamic path that would not collide if another instance of this is running.
+            string basePath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
```
ellismg: We will leak this directory, which seems less than ideal.

schellap: +1. It may be better to keep this repo local? In obj?

gkhanna79: Currently, it gets generated in the intermediates folder. I have added cleanup of the folder once the crossgen is complete.

gkhanna79:
@schellap PTAL

build_projects/shared-build-targets-utils/Utils/Crossgen.cs
ellismg: +1 to @schellap's comment.  It should be simple enough as just not using Path.GetTempPath() and using obj folder, right?  You can still use the guid if there were races?

gkhanna79:
I had seen the issue in intermediate folder. But for now, I am adding the Guid folder to it, so lets see if it still results in the locking problem.

PTAL.

schellap:
Looks ideal to me!

PR #558, opened 2016-11-02
Update LibUV version in netcoreapp1.0

gkhanna79:
Since M.N.App now carries references of netcoreapp1.0, and we are updating LibUV version in LTS branch as well, we should update in the static references M.N.App 1.1.0 carries for netcoreapp1.0.

@weshaggard PTAL

@leecow @Petermarcu FYI


weshaggard:
We are going to want this in master also.

PR #564, opened 2016-11-02
Support building on alpine Linux

mellinoe:
@janvorli 
I just deleted the #ifdef that was commented out, and moved the <sys/time.h> include outside of the GLIBC define so that it gets defined for all non-Windows builds. Platforms using glibc should have the same inclusions as before.


mellinoe:
@dotnet-bot  Test Windows_NT arm Debug Build
Jenkins remoting error.

Issue #544, opened 2016-10-31
Provide installer links in the readme to the latest bits


Petermarcu:
We need to make the links to the installers that come out of this build more obvious to get. Always pointing to the latest but with an explanation on how to get older ones assuming they are still around.

gkhanna79:
@wtgodbe PTAL.

Cc @schellap

Petermarcu:
I'm going to reopen this one to track the changes I made in the comments of https://github.com/dotnet/core-setup/pull/566

PR #566, opened 2016-11-02
Add links to daily builds

gkhanna79:
@Petermarcu PTAL

CC @wtgodbe


gkhanna79:
Fixes https://github.com/dotnet/core-setup/issues/544

Petermarcu:
We lost the link on the badges to the SHA that matches the version.

Petermarcu:
Many of these options have native installers as well as archives. Should we provide archive links as well? I think so.

Petermarcu:
The Ubuntu links look funny. Can we format them as rows inside the cell?

PR #568, opened 2016-11-03
Make CheckIfAllBuildsHavePublished handle more unexpected cases

dagood:
Check to make sure only 0 or 1 badges match up with each semaphore file, and log which badges weren't found to remove the mystery from unpublished builds.

Also fixes the arm64 vs. arm problem by including the `_` delimiter when matching files to expected badges.

@gkhanna79


gkhanna79:
LGTM. you will need to rebase against my change to address merge conflicts.

PR #571, opened 2016-11-03
Changing formatting for lists of items in cells

Petermarcu:


gkhanna79:
The links are not archives but pointers to actual Debian installers that need to be installed.

Petermarcu:
Interesting, the shared framework one is. Should that one be a deb too then? I saw that was a tar.gz and assumed the others were.

gkhanna79:
The SharedFW should be a deb as well and looks like a typo on my part. Please do correct it since you are at it :)

Also, do you also want to update the list for archives for platforms for which installers are produced?

Petermarcu:
Yes, I think we should have links to both installers and archives. I'm going to get this one through to start because I don't have all the other links figured out. I'll fix the Archives wording switch tar.gz to deb in this PR.

Petermarcu:
Done. This doesn't resolve the issue I reopened but does fix some formatting that will help that issue and points to debs instead of tar.gz's for the Ubuntu Shared Framework.

Petermarcu:
@gkhanna79 , you good with this change?

gkhanna79:
LGTM.

PR #575, opened 2016-11-03
Fix copy-paste error for host/hostpolicy version

schellap:


build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -338,9 +338,9 @@ public static BuildTargetResult GenerateMSbuildPropsFile(BuildTargetContext c)
             msbuildProps.AppendLine("  <PropertyGroup>");
             msbuildProps.AppendLine($"    <Platform>{platform}</Platform>");
             msbuildProps.AppendLine($"    <DotNetHostBinDir>{Dirs.CorehostLatest}</DotNetHostBinDir>");
-            msbuildProps.AppendLine($"    <HostVersion>{hostVersion.LatestHostPolicyVersion.WithoutSuffix}</HostVersion>");
+            msbuildProps.AppendLine($"    <HostVersion>{hostVersion.LatestHostVersion.WithoutSuffix}</HostVersion>");
```
gkhanna79: Will this block 1.1.0? I would expect not since we are building all the packages and with the same versions but would be good to confirm.

gkhanna79:
LGTM

PR #596, opened 2016-11-08
Auto-update CoreFx, CoreClr dependencies in master

dagood:
This is https://github.com/dotnet/core-setup/pull/591 but without many of the upgrades. (Also see https://github.com/dotnet/core-setup/pull/593.)

This still has more upgrades than https://github.com/dotnet/core-setup/pull/460, e.g. the UWP metapackage is updated.
@weshaggard @gkhanna79 What are the desired upgrades at this point?

I'll add to this PR to configure auto-update to the desired level.


gkhanna79:
Thanks for making progress on this @dagood. Looking at this PR and https://github.com/dotnet/core-setup/pull/598/files, we should be able to build everything with the updated package versions.

The failure in 598 is the following:

`10:44:47 Unhandled Exception: System.TypeInitializationException: The type initializer for 'Microsoft.DotNet.Cli.Build.Framework.Reporter' threw an exception. ---> System.MissingMethodException: Method not found: 'System.IO.TextWriter System.Console.get_Out()'.
`

@weshaggard @tarekgh Do you know where this type moved and if we need to update projects to reference the package containing them?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
```diff
@@ -2,36 +2,36 @@
...
-    "System.Net.WebSockets.Client": "4.0.0",
-    "System.Numerics.Vectors.WindowsRuntime": "4.0.1",
-    "System.Reflection.Context": "4.0.1",
-    "System.Runtime.InteropServices.WindowsRuntime": "4.0.1",
-    "System.Runtime.Serialization.Json": "4.0.2",
-    "System.Runtime.Serialization.Primitives": "4.1.1",
-    "System.Runtime.Serialization.Xml": "4.1.1",
-    "System.Runtime.WindowsRuntime": "4.0.11",
-    "System.Runtime.WindowsRuntime.UI.Xaml": "4.0.1",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.2.0-beta-24708-01",
```
weshaggard: We should avoid updating this for now we want to closely control what gets updated. 

I also thought that the auto-update doesn't update stable versions to prerelease.

pkg/projects/Microsoft.NETCore/project.json
```diff
@@ -1,60 +1,60 @@
...
-    "System.Text.Encoding.Extensions": "4.0.11",
-    "System.Text.RegularExpressions": "4.1.0",
-    "System.Threading": "4.0.11",
-    "System.Threading.Tasks": "4.0.11",
-    "System.Threading.Tasks.Dataflow": "4.6.0",
-    "System.Threading.Tasks.Parallel": "4.0.1",
-    "System.Threading.Timer": "4.0.1",
-    "System.Xml.ReaderWriter": "4.0.11",
-    "System.Xml.XDocument": "4.0.11"
+    "Microsoft.CSharp": "4.4.0-beta-24707-01",
```
weshaggard: Similarly we should not update this.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
dagood: When VersionTools is involved, stable dependencies stay (by default), but core-setup has a custom implementation.

I'll configure it so these files aren't updated.

build_projects/update-dependencies/UpdateFilesTargets.cs
```diff
@@ -87,7 +87,14 @@ public static BuildTargetResult ReplaceProjectJson(BuildTargetContext c)
         {
             List<DependencyInfo> dependencyInfos = c.GetDependencyInfos();
 
-            IEnumerable<string> projectJsonFiles = Directory.GetFiles(Dirs.RepoRoot, "project.json", SearchOption.AllDirectories);
+            IEnumerable<string> projectJsonFiles = new string[]
```
weshaggard: While this feels like I hack I think I can live with it until we get core-setup repo changed.

weshaggard:
@gkhanna79 those errors indicate to me that something is mixing and matching versions of the framework. I will take a look at that PR to see if I can spot it.

build_projects/update-dependencies/UpdateFilesTargets.cs
dagood: Hmm, well, it's just picking some paths rather than recursively finding all, so it shouldn't seem like a hack. I wrote it in a confusing way, though, seeing it again... I'll change it to a `List<string>` and see if that helps.

gkhanna79:
LGTM.

CC @schellap @ramarag

gkhanna79:
Excellent - can you please force a new auto-update PR?

dagood:
Sure, I'll get one queued.

dagood:
Auto-PR created: https://github.com/dotnet/core-setup/pull/606

PR #599, opened 2016-11-08
Retarget Microsoft.DotNet.PlatformAbstractions to net45

mellinoe:
This project wasn't using anything that was unavailable in .NET 4.5, and I'd like to reference it in a buildtools project which targets net45.

@livarcocc


mellinoe:
I also recreated the xproj file because it was old and had the wrong name.

Issue #595, opened 2016-11-08
Add switch in source to stop publishing symbol packages to MyGet


dagood:
For release branches, we need to stop publishing symbol packages to MyGet. We need to continue publishing them for other branches.

This can be implemented with filters in https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PublishTargets.cs checking the end of the files for `.symbols.nupkg` based on the `BRANCH_NAME` or `CHANNEL` in https://github.com/dotnet/core-setup/blob/master/branchinfo.txt (as flowed through the build context).

I think it should be based off the current branch specified in that file because it's the most likely thing I can think of to be kept up to date.

Not a large task, but tracking it in an issue in case someone else needs to pick it up.

gkhanna79:
@leecow @Petermarcu Should this be done for 1.1.0? Doing so would require respinning and I would suggest we do this for 1.1.x.

CC @schellap @wtgodbe

dagood:
The alternative is to remove the package from the MyGet feed manually, and if we can't do that by normal means we can ask MyGet to help. I'll test this out this afternoon on a test feed to see if we can remove it ourselves.

We can also overwrite what's there with a blank symbol package, although that's a bit misleading.

dagood:
I didn't find any endpoint to delete the symbol package directly, but if you delete the package, the symbol package gets deleted too. So, to remove the symbol package we can just delete and reupload the library package.

PR #600, opened 2016-11-08
On release branches, don't publish symbol packages

dagood:
When the branch (according to branchinfo.txt) starts with `release/`, don't publish symbol packages.

Addresses https://github.com/dotnet/core-setup/issues/595 for master (1.2.0+) and it sounds like we'll merge it into 1.1.x (and 1.0.x?).

@gkhanna79 @wtgodbe

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -29,6 +29,8 @@ namespace Microsoft.DotNet.Host.Build
 
         private static string HostFxrNugetVersion { get; set; }
 
+        private static bool IncludeSymbolPackages { get; set; }
+
         [Target]
         public static BuildTargetResult InitPublish(BuildTargetContext c)
         {
@@ -39,6 +41,9 @@ namespace Microsoft.DotNet.Host.Build
             HostFxrNugetVersion = c.BuildContext.Get<HostVersion>("HostVersion").LockedHostFxrVersion.ToString();
             Channel = c.BuildContext.Get<string>("Channel");
             CommitHash = c.BuildContext.Get<string>("CommitHash");
+
+            // Do not publish symbol packages on a release branch.
+            IncludeSymbolPackages = !c.BuildContext.Get<string>("BranchName").StartsWith("release/");
 
             return c.Success();
         }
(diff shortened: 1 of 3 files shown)


wtgodbe:
LGTM

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -40,6 +42,9 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
             Channel = c.BuildContext.Get<string>("Channel");
             CommitHash = c.BuildContext.Get<string>("CommitHash");
 
+            // Do not publish symbol packages on a release branch.
+            IncludeSymbolPackages = !c.BuildContext.Get<string>("BranchName").StartsWith("release/");
```
dagood: It will prevent us from publishing prerelease symbol packages on a release branch, yes. I figure this might actually be a small benefit, because we'll be able to find anyone relying on MyGet for symbol packages just a bit earlier and have a chance to communicate the RTM way to get them. It also means we have more consistent published set of artifacts when a build starts making stable versions.

I could change the logic to default to branch name but also allow an override if you think publishing prerelease release-branch symbol packages will be useful. (Or it could be based on whether the branch is stable--but I haven't seen a good central source to determine that during the build yet.)

gkhanna79:
We definitely need to have a debugging story for any build that is not publicly released. Also, are you making similar changes for CoreCLR and CoreFX? If so, the same argument will be applicable there as well.

dagood:
In CoreFX, CoreCLR, and WCF I have disabled symbol package publishing in the existing release branch build definitions, and I'll be making sure we use conditionals when those are available so there's no extra thought when forking off a release branch. There is no directly analogous change due to the different build system.

> We definitely need to have a debugging story for any build that is not publicly released.

Unless I'm not understanding which builds you're talking about or what it means for a build to be publicly released, this objection applies just as much to stable-versioned symbol packages as to prerelease symbol packages, because both are only available on MyGet. (Not released.) If you think we definitely need public users to be able to access symbols on every unreleased release branch build, then we need a completely different plan--the email thread "Disabling release branch symbol package publish to MyGet" would be right place to discuss that.

Just to make sure it's clear: this PR doesn't prevent us from gathering the symbol packages internally later to give them to someone who needs them. (If it does prevent that, that's a bug.)

gkhanna79:
@dagood To confirm, the azure blob will still have the symbol packages?

dagood:
Yes, the only effect of this PR is that they wouldn't be pushed to MyGet, so there's no discoverable way to find them.

gkhanna79:
LGTM then.

PR #639, opened 2016-11-13
No need to delete temp R2R Files

gkhanna79:
This is no longer required since the files are created in intermediates folder that will get cleaned up eventually.

Also, this helps with non-deterministic locking of files when attempting delete.

@schellap PTAL


gkhanna79:
Will address cases like https://mseng.visualstudio.com/DefaultCollection/dotnetcore/_build?_a=summary&buildId=3424125

PR #658, opened 2016-11-17
Upgrade 1.1.0 to 1.1.1 servicing

schellap:
Cc @gkhanna79

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -354,7 +354,7 @@ namespace Microsoft.DotNet.Host.Build
 
                 Console.WriteLine($"Copying package {fileName} to artifacts directory {Dirs.CorehostLocalPackages}.");
             }
-            foreach (var item in hostVersion.LatestHostPackages)
+            foreach (var item in hostVersion.LatestHostPackagesToValidate)
             {
                 var fileFilter = $"runtime.{rid}.{item.Key}.{item.Value.ToString()}.nupkg";
                 if (Directory.GetFiles(Dirs.CorehostLocalPackages, fileFilter).Length == 0)
--- a/build_projects/shared-build-targets-utils/Utils/HostVersion.cs
+++ b/build_projects/shared-build-targets-utils/Utils/HostVersion.cs
@@ -63,16 +63,25 @@ namespace Microsoft.DotNet.Cli.Build
         // Full versions and package information.
         public string LatestHostBuildMajor => CommitCountString;
         public string LatestHostBuildMinor => "00";
-        public bool EnsureStableVersion => true;
-        /*
-        public VerInfo LatestHostVersion => new VerInfo(1, 1, 0, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
-        public VerInfo LatestHostFxrVersion => new VerInfo(1, 1, 0, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
-        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 1, 0, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
-        */
-        public VerInfo LatestHostVersion => new VerInfo(1, 1, 0, "", "", "", CommitCountString);
-        public VerInfo LatestHostFxrVersion => new VerInfo(1, 1, 0, "", "", "", CommitCountString);
-        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 1, 0, "", "", "", CommitCountString);
-        
+        public bool EnsureStableVersion => false;
+
+        // Comment below lines when stabilizing 1.1.1
+        public VerInfo LatestHostVersion => new VerInfo(1, 1, 1, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+        public VerInfo LatestHostFxrVersion => new VerInfo(1, 1, 1, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 1, 1, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+
+        // Uncomment below lines when stabilizing 1.1.1
+/*
+        public VerInfo LatestHostVersion => new VerInfo(1, 1, 1, "", "", "", CommitCountString);
+        public VerInfo LatestHostFxrVersion => new VerInfo(1, 1, 1, "", "", "", CommitCountString);
+        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 1, 1, "", "", "", CommitCountString);
+*/
+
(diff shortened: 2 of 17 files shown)


build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -90,7 +92,7 @@ public override string ToString()
         // Locked muxer for consumption in CLI.
         //
         public bool IsLocked = false; // Set this variable to toggle muxer locking.
```
gkhanna79: Do we not want to toggle the muxer locking?

build_projects/shared-build-targets-utils/Utils/Monikers.cs
```diff
@@ -11,9 +11,9 @@ public class Monikers
     {
         public const string SharedFrameworkName = "Microsoft.NETCore.App";
         public const string CLISdkBrandName = "Microsoft .NET Core 1.0.0 - SDK Preview 2";
```
gkhanna79: I dont think we use CLISdkBrandName - can you please remove it?

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -63,16 +63,18 @@ public override string ToString()
...
-        public VerInfo LatestHostFxrVersion => new VerInfo(1, 1, 0, "", "", "", CommitCountString);
-        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 1, 0, "", "", "", CommitCountString);
-        
+        public bool EnsureStableVersion => false;
+
+        public VerInfo LatestHostVersion => new VerInfo(1, 1, 1, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+        public VerInfo LatestHostFxrVersion => new VerInfo(1, 1, 1, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+        public VerInfo LatestHostPolicyVersion => new VerInfo(1, 1, 1, ReleaseSuffix, LatestHostBuildMajor, LatestHostBuildMinor, CommitCountString);
+
+/*
```
gkhanna79: Can you please add a comment here saying that this needs to be enabled when stabilized build is generated?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <PropertyGroup>
-    <Version>1.1.0</Version>
+    <Version>1.1.1</Version>
```
schellap: In this diff, we were producing the packages and can you clarify? We will produce packages with `1.1.1-servicing-4158`. When we service, we will turn off all other ones and produce the ones we want. So I don't think we need to pre-emptively do anything here.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
schellap: Yes, you are right!

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
gkhanna79: Without making the change similar to what @weshaggard made to packages.builds in the referenced diff, we will be generating all packages when we should not be in the servicing branch.

Thus, we should make https://github.com/dotnet/core-setup/blob/release/1.1.0/pkg/projects/packages.builds#L6 to false and add Microsoft.NETCore.App to https://github.com/dotnet/core-setup/blob/release/1.1.0/pkg/projects/packages.builds#L14 since that is the minimal set of packages that will require rebuild (as its project.json will be updated to consume the updated CoreCLR from 1.1.0 branch).

gkhanna79: You will also need to update https://github.com/dotnet/core-setup/blob/release/1.1.0/build_projects/shared-build-targets-utils/DependencyVersions.cs to reflect the new servicing versions that come out of the CoreCLR 1.1.0 branch. I have kicked off https://devdiv.visualstudio.com/DevDiv/_build/explorer?_a=summary&buildId=435359 that should publish the packages whose version you should update with.

gkhanna79:
@schellap CoreCLR has published a new package from 1.1.x branch. https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.ubuntu.14.04-x64.Microsoft.NETCore.Runtime.CoreCLR/1.1.1-servicing-24720-01 is an example.

Can you please update your PR accordingly?

gkhanna79:
@schellap Can we get this ready to go today?

gkhanna79:
@schellap Will you be able to get this fixed soon so that we can get the updated 1.1.x builds out?

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -354,7 +354,7 @@ public static BuildTargetResult PackagePkgProjects(BuildTargetContext c)
 
                 Console.WriteLine($"Copying package {fileName} to artifacts directory {Dirs.CorehostLocalPackages}.");
             }
-            foreach (var item in hostVersion.LatestHostPackages)
+            foreach (var item in hostVersion.LatestHostPackagesToValidate)
```
gkhanna79: It will be good to port this change (LatestHostPackagesToValidate) to master as well.

gkhanna79:
LGTM.

@dagood Any comments before this is merged?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
```diff
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "Microsoft.NETCore.Portable.Compatibility": "1.0.2",
-    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.0",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.1-servicing-24720-01",
```
gkhanna79: I dont think we need to update UWP for now. Correct @weshaggard ?

schellap: @weshaggard, could you confirm, otherwise we can merge once green.

gkhanna79: Given that we did update this for RTM, i think it is fine to keep your change.

weshaggard: Actually I think this is incorrect. We have to be careful with which CoreCLR package we reference for UWP. Actually it looks like a lot of the packages referenced here shouldn't have been updated. At this point we having shipped this from the 1.1 branch so it hasn't caused trouble but if we are to ship it I think there maybe some problems. 

cc @yizhang82

gkhanna79: The only place where UWP package can actually ship from, right now, is release/1.0.0 branch since it requires components from closed build. 

FWIW, UWP was not shipped from release/1.1.0.

Issue #650, opened 2016-11-17
Size Regression in Shared Framework


brianrob:
## Steps to reproduce
1. Download dotnet-obtain.ps1: https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1.
2. Run powershell .\dotnet-obtain.ps1 -SharedRuntime -InstallDir . -Channel master.
3. cd Microsoft.NetCore.App\<latest shared framework directory>.
4. dir

## Expected  behavior
Expect to have a single binary for each managed assembly in the shared framework.

## Actual behavior
For each managed assembly, there is both a .dll and a .readytorun file.

## Environment data
The specific version of the shared framework that I saw this in is 1.2.0-beta-001125-00.

PR #659, opened 2016-11-17
Fix location to generate R2R assemblies

gkhanna79:
Compute the path to generate R2R assemblies correctly.

Fixes https://github.com/dotnet/core-setup/issues/650

@schellap PTAL

CC @brianrob


schellap:
Looks better now!

Issue #723, opened 2016-12-05
Cross compiling corehost for Linux ARM32


hqueue:
This is a part of  Enabling Linux ARM32 core-setup #725.

# New Plan (Dec. 7, 2016~)

## Prerequisite
We need rootfs (issue #729) for cross-compilation. 

## Proposed approach
After rootfs for ARM32 is setup, we can build corehost for ARM using following command.
```
cd ./src/corehost
ROOTFS_DIR=<core-setup>/cross/rootfs/arm <core-setup>/src/corehost/build.sh "--arch" "arm" "--hostver" "1.1.0" "--fxrver" "1.1.0" "--policyver" "1.1.0" "--rid" "ubuntu.14.04-arm" "--commithash" "f46c58c1404fa8aaac8aed931b0862267a98c3a5" --cross
```

# Old Plan (Before Dec. 7, 2016) (Dropped)

## Steps to reproduce
You have to installed arm cross toolchain on host machine., e.g. g++-arm-linux-gnueabihf
```
cd ./src/corehost
./build.sh -xcompiler=`which arm-linux-gnueabihf-g++`
```

## Expected  behavior
Successful building corehost for Linux/ARM.

## Actual behavior
```
[  2%] Building CXX object cli/exe/CMakeFiles/dotnet.dir/__/fxr/fx_ver.cpp.o
arm-linux-gnueabihf-g++: error: unrecognized command line option ‘-fstack-protector-strong’
make[2]: *** [cli/exe/CMakeFiles/dotnet.dir/__/fxr/fx_ver.cpp.o] Error 1
make[1]: *** [cli/exe/CMakeFiles/dotnet.dir/all] Error 2
make: *** [all] Error 2
```

## Environment data
Host PC: Ubuntu 14.04 x64
Installed arm cross toolchain: 
```
$ arm-linux-gnueabihf-g++ --version
arm-linux-gnueabihf-g++ (Ubuntu/Linaro 4.8.2-16ubuntu4) 4.8.2
```
## Proposed solution
Usually ARM cross compiler version is gcc 4.7 or 4.8 on Ubuntu 14.04 when installing
cross compiler using `apt-get install`.
Let's use "-fstack-protector" instead of "-fstack-protector-strong" when compiling for Linux/ARM using GNU compilers.

I will post PR soon.

gkhanna79:
Please keep @schellap and @ramarag in PR.

gkhanna79:
CC @janvorli

[...]

PR #724, opened 2016-12-05
Use -fstack-protector when compilng for Linux/ARM

hqueue:
Relate issue: #723 

Let's use -fstack-protector instead of -fstack-protector-strong when using GNU cross
compilers for ARM, since -fstack-protector-strong  is not supported before gcc 4.9.

Signed-off-by: Hyung-Kyu Choi <hk0110.choi@samsung.com>


gkhanna79:
Is GCC 4.9 not available for us to use so as to not have this conditional?

CC @janvorli

janvorli:
@hqueue we support clang and not gcc. Why are you trying to use gcc instead?

hqueue:
> @hqueue we support clang and not gcc. Why are you trying to use gcc instead?

janvorli:
@hqueue in CoreCLR and CoreFX, we use clang for everything. So I believe we should standardize on clang here as well. @gkhanna79 do you agree with me?

hqueue:
@janvorli I also think using clang is reasonable when considering CoreCLR and CoreFX.

When using clang, there are two ways to cross-compie corehost with ARM header files and libraries.
(1) Using ARM header files and library files from rootfs (which is used to cross-compile CoreCLR for ARM)
(2) Using ARM header files and library files from installed gcc cross toolhain(e.g.  gcc-arm-linux-gnueabihf) from host machine

Becuase corehost is much smaller software compared to CoreCLR, I don't think we really need such heavy rootfs to build corehost. So I prefer 2nd way, using header fils and libaries from installed gcc cross toolchain.

What do you think of it?

janvorli:
@hqueue since CoreCLR and CoreFX use the rootfs, I think we should use the same way to stay consistent. To get the full product, all three repos need to be built, so the rootfs needs to be present on the build machine anyways. We are also planning to have a combined build that would build everything from sources of the three repos together, so consistence is important.

hqueue:
@janvorli I see. Then I think we had better to close this PR, since gcc is not a concern.
And please visit issue #725 and give us any opinion regarding enabling ARM. 

BTW do you have any plan to introduce rootfs for ARM here soon ?

gkhanna79:
I agree with @janvorli that we should have a consistent approach to build the product across the board. This includes the toolset as well. Do we not have a clang based toolset that can be used for cross-compilation instead of GCC?

I am not very familiar with what ROOTFS is used for. Is it about loading the Arm32 emulator and performing the product build within it (and thus, using the toolset within it)?

janvorli:
@gkhanna79 it is not related to the emulator. The rootfs is basically a folder with files that would be on a target machine with the other architectures if you were building it over there. 
So for example, the root level of the rootfs I have generated on my Linux box for ARM cross build looks like this:
```
ls cross/rootfs/arm/
bin  boot  dev  etc  home  lib  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```

The regular clang it capable of cross compiling. Here is an example of the command line. Please note the `-target armv7-linux-gnueabihf -mthumb -mfpu=vfpv3 --sysroot=/home/janvorli/git/coreclr/cross/rootfs/arm` part:
`/usr/bin/clang++-3.6   -DARM -DBIT64=1 -DBUILDENV_CHECKED=1 -DDEBUG -DDISABLE_CONTRACTS -DLINUX32 -DPIC=1 -DPLATFORM_UNIX=1 -DSystem_Globalization_Native_EXPORTS -DURTBLDENV_FRIENDLY=Checked -D_ARM_ -D_DBG -D_DEBUG  -Wall -Wno-null-conversion -std=c++11 -g -O0 -fPIC -I/home/janvorli/git/coreclr/bin/obj/Linux.arm.Debug/src/corefx/System.Globalization.Native -I/home/janvorli/git/coreclr/src/corefx/System.Globalization.Native -I/home/janvorli/git/coreclr/cross/rootfs/arm/usr/include/arm-linux-gnueabihf    -target armv7-linux-gnueabihf -mthumb -mfpu=vfpv3 --sysroot=/home/janvorli/git/coreclr/cross/rootfs/arm -fno-omit-frame-pointer -fms-extensions -fwrapv -fstack-protector-strong -ferror-limit=4096 -Werror -Wno-unused-private-field -Wno-unused-variable -Wno-microsoft -Wno-tautological-compare -Wno-constant-logical-operand -Wno-unknown-warning-option -Wno-invalid-offsetof -Wno-incompatible-ms-struct -fsigned-char -fPIC -o CMakeFiles/System.Globalization.Native.dir/calendarData.cpp.o -c /home/janvorli/git/coreclr/src/corefx/System.Globalization.Native/calendarData.cpp",
  "file": "/home/janvorli/git/coreclr/src/corefx/System.Globalization.Native/calendarData.cpp`

gkhanna79:
Thank you @janvorli. Do I understand it correct that ROOTFS, in the above example, is being used to invoke the clang compiler that will do a cross-compilation build for arm32? If so, is it not possible to do the same without having ROOTFS (unless ROOTFS is simply being used to have a segregated location for the cross-compilation toolset)?

janvorli:
No, you need the rootfs - it contains e.g. all the target libraries we link to compiled for the target architecture, like ARM.

gkhanna79:
Thanks Jan. 

I agree that ROOTFS approach is the right one to be enabled here as well.

Issue #725, opened 2016-12-05
Enabling Linux ARM32 for .NET Core


hqueue:
Host: Ubuntu 14.04 x64
Target: Ubuntu 14.04 or(and) 16.04 ARM, Ubuntu 14.04 ARM Softfp and Tizen

To enable core-setup for Linux ARM32, we are going to follow steps suggested in PR #712.

## ARM hardfp

**Common**
- [x] stable default clang version for cross compilation (#1411)
- [x] Use same logic to prepare rootfs across dotnet (#1432)

**CoreCLR** 
- [x] Update CoreCLR to generate nuget packages of CoreCLR (PR dotnet/coreclr#8421)
- [x] Remaining CoreCLR ARM32 issues tracked by issue dotnet/coreclr#8549.  **Still on going**

**CoreFX**
- [x] Add RID for Linux/ARM (dotnet/corefx#14161)
- [x] Add Xenial (Ubuntu 16.04) support for build-rootfs.sh (issue https://github.com/dotnet/corefx/issues/14538)
- [x] Make clang3.6 default toolset for cross arm32 build (issue https://github.com/dotnet/corefx/issues/15511)
- [x] Enable CI support for Linux Arm32 (https://github.com/dotnet/coreclr/issues/9273, https://github.com/dotnet/corefx/pull/15900)
- [x] Update CoreFX to generate nuget packages of CoreFX (issue dotnet/corefx#14298)
- [x] Enable CoreFX Pipeline build support for Linux for Arm32 and publish packages. 

**core-setup**
For core-setup, we will follow steps below.
- [x] Add support for Linux Arm32 RID (PR #712)
- [x] Preparing ROOTFS for cross-compilation Infrastructure (issue #729, PR #747)
- [x] Enable cross compiling corehost for Linux/ARM (issue #723, PR #749, #766, #812  )
- [x] Update build script to enable cross building (PR #712)
- [x] Update build script to enable downloading and packaging for Linux/ARM (PR #712)
- [x] Build nuget packages for the host and Microsoft.NETCore.App metapackage (PR #712)
- [x] Make clang3.6 default toolset for cross arm32 build (issue https://github.com/dotnet/core-setup/issues/1322)
- [x] Enable CI support for Linux Arm32 (issue https://github.com/dotnet/core-setup/issues/790)(PR https://github.com/dotnet/core-setup/pull/1512)
[...]

PR #726, opened 2016-12-05
Enable passing parameters to corehost buildscript for Linux ARM

hqueue:
Related issue: #725

To enable cross building of corehost, add environmental variable for cross compiler in build scripts.

Signed-off-by: Hyung-Kyu Choi <hk0110.choi@samsung.com>

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -164,7 +164,6 @@ namespace Microsoft.DotNet.Host.Build
             var configuration = c.BuildContext.Get<string>("Configuration");
             string rid = c.BuildContext.Get<string>("TargetRID");
             string platform = c.BuildContext.Get<string>("Platform");
-            string crossCompiler = c.BuildContext.Get<string>("CrossCompiler");
 
             // Generate build files
             var cmakeOut = Path.Combine(Dirs.CorehostLatest, "cmake");
@@ -306,11 +305,6 @@ namespace Microsoft.DotNet.Host.Build
                 buildScriptArgList.Add(rid);
                 buildScriptArgList.Add("--commithash");
                 buildScriptArgList.Add(commitHash);
-
-                if (crossCompiler != null) {
-                    buildScriptArgList.Add("--xcompiler");
-                    buildScriptArgList.Add(crossCompiler);
-                }
 
                 ExecIn(cmakeOut, buildScriptFile, buildScriptArgList);
 
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -289,37 +289,30 @@ namespace Microsoft.DotNet.Host.Build
                         throw new PlatformNotSupportedException("Target Architecture: " + platform + " is not currently supported.");
                 }
 
-                if (crossCompiler == null) {
-                    ExecIn(cmakeOut, Path.Combine(c.BuildContext.BuildDirectory, "src", "corehost", "build.sh"),
-                            "--arch",
-                            arch,
-                            "--hostver",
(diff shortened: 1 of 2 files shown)


gkhanna79:
@schellap This is broken due to the same reason as your PR.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -275,19 +276,50 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
...
+                            hostVersion.LatestHostVersion.ToString(),
+                            "--fxrver",
+                            hostVersion.LatestHostFxrVersion.ToString(),
+                            "--policyver",
+                            hostVersion.LatestHostPolicyVersion.ToString(),
+                            "--rid",
+                            rid,
+                            "--commithash",
+                            commitHash,
+                            "--xcompiler",
```
gkhanna79: Can you please refactor this using an ArgList so that we dont duplicate the implementation? Consider using https://github.com/dotnet/core-setup/blob/eca152525e19a41d489cd5845afa14982653cc0a/build_projects/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs#L10

gkhanna79:
Why do we need to pass crosscompiler as opposed to enabling the build as "./build.sh arm" and using the available cross compiler? Isnt that what we do in CoreCLR as well?

hqueue:
> Why do we need to pass crosscompiler as opposed to enabling the build as "./build.sh arm" and using the available cross compiler? Isnt that what we do in CoreCLR as well?

@gkhanna79 I made this PR to make use of existing build scipts such as   ```build_projects/dotnet-host-build/build.sh``` and ```src/corehost/buil.sh``` and we already  have to use `TARGETPLATFORM` environmental variable to sepcify target platform other than x64.

For example, with this PR, we can invoke building of ubuntu16.04-arm  as follows.
```
~/core-setup $ ./build.sh  --env-vars TARGETPLATFORM=arm,TARGETRID=ubuntu.16.04-arm,CROSSCOMPILER=`which arm-linux-gnueabihf-g++
```

I also prefer using ```./build.sh arm``` and eventually we can move to this way. However CoreCLR make use if its' own build-rootfs and dedicated cmake files for cross compilation which is not available here yet. I think we can make them for core-setup eventually.

build_projects/dotnet-host-build/CompileTargets.cs
hqueue: I applied your feedback. Thanks!

gkhanna79:
Shouldnt we use ROOTFS approach here as well?

CC @janvorli @kouvel

janvorli:
@gkhanna79 I have thought this is the same thing we have discussed in #724

hqueue:
@janvorli @gkhanna79 If you don't mind, can we merge this PR and change cross building script of corehost later?

This PR contains two changes,
(1) enable to pass target information using "TARGETPLATFORM" when building at Linux (before it always build for x64 by default and no way to change target.)
(2) enable to control missing option (--xcompile)

I will change title of this PR to "Enable to pass parameters to corehost buildscript."

Or we can abandon this PR and I can make another PR for (1).

janvorli:
@hqueue I assume that we will not need the `--xcompiler` option when we move to using rootfs, right? If I am not mistaken in this, I would like to remove that part of this change.

hqueue:
@janvorli You are right. I will  remove ```--xcompiler`` related part.

hqueue:
@gkhanna79 @janvorli  Can you squash these commits ? Or I can squash commits and force push a signle final commit again.

janvorli:
@hqueue github merge button allows me to automatically squash them at merge, so you don't need to do that.

gkhanna79:
LGTM.

Issue #729, opened 2016-12-07
Preparing ROOTFS for cross-compilation


hqueue:
As discussed in PR #724  and suggested in issue #725, we believe that we should keep consistency with other repos (e.g. CoreCLR and CoreFX) in the way we do cross compilation.

Therefore we need ROOTFS as a first step and I'm going to prepare ROOTFS following the logic from CoreCLR.

### Target
- ARM (only ARM this time, or do we need other architectures too?)
### New directories
- core-setup/cross
- core-setup/cross/arm
- core-setup/cross/arm-softfp
### New script files
- core-estup/cross/build-rootfs.sh

hqueue:
@gkhanna79 @schellap @ramarag @janvorli Looking for any comment. Thank you in advance!

gkhanna79:
For now, Arm support is sufficient @hqueue. Thank you for getting this enabled.

CC @kouvel

hqueue:
Close this issues, since PR #747 is merged.

PR #747, opened 2016-12-09
Introduce rootfs for cross-compilation

hqueue:
skip ci please

Introduce rootfs for cross compilation.  I've added scripts for ARM and ARM softfp.
Tested it with `sudo ./cross/build-rootfs.sh arm` and  `sudo ./cross/build-rootfs.sh arm-softfp` as documented in https://github.com/dotnet/coreclr/blob/master/Documentation/building/cross-building.md to keep consistency with CoreCLR and CoreFX.

Basically the code has been imported from CoreCLR and remove x86 and arm64.
rootfs is exactly same with CoreCLR, i.e. we can also build CoreCLR with this rootfs too.
Later we can refine rootfs to reduce size only with required packages if necessary.
BTW CoreFX has a different configuration.

### DIFF
--- a/cross/arm-softfp/sources.list.jessie
+++ b/cross/arm-softfp/sources.list.jessie
@@ -1,0 +1,3 @@
+# Debian (sid)   # UNSTABLE
+deb http://ftp.debian.org/debian/ sid main contrib non-free
+deb-src http://ftp.debian.org/debian/ sid main contrib non-free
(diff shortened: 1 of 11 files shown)


hqueue:
CC @schellap @ramarag @janvorli @gkhanna79 @kouvel PTAL

Related issue: #729

gkhanna79:
> Basically the code has been imported from CoreCLR and remove x86 and arm64

I was thinking about this. Since there is work going on to bring Linux for x86 (https://github.com/dotnet/coreclr/issues/8188), you may just want to keep the x86 support as well (but do not need to test it).

Aside from this, LGTM.

hqueue:
CC @gkhanna79  @schellap @ramarag  I've added x86 rootfs as suggested.

gkhanna79:
LGTM - assuming you have validated it, right?

hqueue:
@gkhanna79 Right, we have test several this PR #747 and PR #749  together and it works for ARM :)

Issue #764, opened 2016-12-12
Add '--hostbuildonly' option


jyoungyun:
## Expected  behavior

 `./build.sh --build-driver-only`

## Actual behavior
This option only builds the dotnet-host-build binary. This is useful when you want to call it directly after we modify the dotnet-host-build tool.

hqueue:
> `dotnet --hostbuildonly`
@jyoungyun  I think you meant `build_projects/dotnet-host-build/build.sh` not `dotnet ` right?

jyoungyun:
@hqueue Yes, right. I correct the content of this issue. Thank you.

PR #767, opened 2016-12-12
Add '--build-driver-only' option

jyoungyun:
This option only builds the dotnet-host-build binary.
This is useful when you want to call it directly
after we modify the dotnet-host-build tool.

I have a plan to add '--skip-hostbuild' option to disable dotnet-host-build build.
Please tell me if my suggestion does not appeal to you.

Related issue: #764

### DIFF
--- a/build_projects/dotnet-host-build/build.sh
+++ b/build_projects/dotnet-host-build/build.sh
@@ -18,6 +18,8 @@ OLDPATH="$PATH"
 REPOROOT="$DIR/../.."
 source "$REPOROOT/scripts/common/_prettyprint.sh"
 
+__BuildDriverOnly=0
+
 while [[ $# > 0 ]]; do
     lowerI="$(echo $1 | awk '{print tolower($0)}')"
     case $lowerI in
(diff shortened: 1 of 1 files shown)


build_projects/dotnet-host-build/build.sh
```diff
@@ -40,6 +42,9 @@ while [[ $# > 0 ]]; do
             # Allow CI to disable prereqs check since the CI has the pre-reqs but not ldconfig it seems
             export DOTNET_INSTALL_SKIP_PREREQS=1
             ;;
+        --hostbuildonly)
+            __HostBuildOnly=1
```
gkhanna79: I would suggest we rename this to "BuildDriver" since that is what dotnet-host-build is. Also, renaming it will alleviate any confusion with the actual host that is also built in this repo.

jyoungyun: Thank you for your suggestion. I am going to change it according to your opinion. But I don't know if this name is okay. How about '--build-driver-only' ? Do you have a better idea?

gkhanna79: @jyoungyun --build-driver-only sounds fine.

gkhanna79:
@dotnet-bot test RHEL7.2 x64 Release Build  please

gkhanna79:
@jyoungyun Is this good to merge assuming CI is green?

jyoungyun:
@gkhanna79 Yes, now All CI are fine. Thank you. :)

PR #771, opened 2016-12-12
Handfull of misc build fixes

ellismg:
While working on getting a composed end to end source build of core-setup working locally on *nix, I've developed these patches.

The major changes are using `--infer-runtimes` in more places (so we don't have to explicitly list or restore runtimes we don't care about for some projects), teaching `update-dependencies` to both work on *nix and to allow pointing it at local files for the version info instead of just remote files and having the Microsoft.NETCore.App build use the root NuGet.config instead of hard-coding it's own sources.

### DIFF
--- a/NuGet.master.config
+++ b/NuGet.master.config
@@ -1,7 +1,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <packageSources>
-    <clear />
-    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
-  </packageSources>
-</configuration>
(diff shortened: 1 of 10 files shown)


ellismg:
@gkhanna79 @AlfredoMS  Can you please take a look?  This needs a rebase on top of master which I will do after the first round of feedback (if any).

gkhanna79:
@schellap PTAL

AlfredoMS:
LGTM

ellismg:
A merge conflict is generated every time we update our dependencies because of my white-space fix-up.  When this is ready I will rebase and let CI run before merging, but I'm not going to aggressively rebase.

schellap:
LGTM, thank you!

ellismg:
Thanks @AlfredoMS and @schellap!  I rebased to fix the conflict and I'll merge once I have a clean CI.

PR #791, opened 2016-12-14
Allow upgrade when some packages not in buildinfo

dagood:
This changes dependency auto-update behavior to allow the buildinfos listed on dotnet/versions to not contain packages. Instead of writing an error and writing a garbage sentinel value into a PR, auto-update will not change the version that wasn't found.

This change can be cherry-picked into the 1.0.0 and 1.1.0 branches without conflicts. The intent is to fix auto-upgrade for 1.1.0: https://github.com/dotnet/core-setup/pull/787. I tested in 1.1.0 and it makes the expected changes without erroring because there is no new servicing Platforms package.

I'm flowing this from master to 1.1.0/1.0.0 instead of directly PRing for servicing future release branches.

@gkhanna79

### DIFF
--- a/build_projects/update-dependencies/UpdateFilesTargets.cs
+++ b/build_projects/update-dependencies/UpdateFilesTargets.cs
@@ -224,7 +224,11 @@ namespace Microsoft.DotNet.Scripts
             Regex regex = new Regex($@"{dependencyPropertyName} = ""(?<version>.*)"";");
             string newVersion = c.GetNewVersion(packageId);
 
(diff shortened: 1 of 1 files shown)


gkhanna79:
Looks fine to me. Will you be taking this to servicing branches as well (1.0.0 and 1.1.0)?

dagood:
Yep, I already made sure those cherry-picks are conflictless, but I want to make sure this looks good before making those PRs.

PR #796, opened 2016-12-15
fixed typo that interferes with control flow logic in dockerrun.sh

margaretho:


gkhanna79:
@dleeapho PTAL

dleeapho:
LGTM. FYI, build.sh specifies `DOCKER_IMAGENAME` that's why we haven't typically hit this. Thanks for catching the typo @margaretho !

PR #797, opened 2016-12-15
Limit project.json auto-upgrade scope: no UWP or miscellaneous project files (release/1.0.0)

dagood:
Ports the part of https://github.com/dotnet/core-setup/pull/596 that limits auto-update to a static list of files. Results in an auto-upgrade like https://github.com/dagood/core-setup/commit/557a0c295aa8c0e6707f419b7aa695db80a6d568 from stable.

Removes UWP auto-update noticed in https://github.com/dotnet/core-setup/pull/795. If this is merged and there's no new coreclr build, there will be no auto-upgrade changes to make.

I'll port this to release/1.1.0 too.

@gkhanna79 
/cc @weshaggard

PR #798, opened 2016-12-15
Limit project.json auto-upgrade scope: no UWP or miscellaneous project files (release/1.1.0)

dagood:
Port https://github.com/dotnet/core-setup/pull/797 to release/1.1.0. Removed some conflicting 1.1.0-specific changes that also limited which project.json files to upgrade, but they were much less restrictive.

@gkhanna79

### DIFF
--- a/build_projects/update-dependencies/UpdateFilesTargets.cs
+++ b/build_projects/update-dependencies/UpdateFilesTargets.cs
@@ -87,24 +87,11 @@ namespace Microsoft.DotNet.Scripts
         {
             List<DependencyInfo> dependencyInfos = c.GetDependencyInfos();
 
-            string[] sourceDirsToUpdate =
+            var projectJsonFiles = new List<string>
             {
-                "build_projects",
-                "pkg",
-                // Don't upgrade RTM versions in InternalAbstractions and DependencyModel.
-                // See https://github.com/dotnet/core-setup/issues/275
-                //"src",
-                "test",
-                "tools"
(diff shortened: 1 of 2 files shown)


build_projects/update-dependencies/UpdateFilesTargets.cs
```diff
@@ -87,25 +87,17 @@ public static BuildTargetResult ReplaceProjectJson(BuildTargetContext c)
         {
             List<DependencyInfo> dependencyInfos = c.GetDependencyInfos();
 
-            string[] sourceDirsToUpdate =
+            var projectJsonFiles = new List<string>
```
gkhanna79: I thought we had files like DependencyVersion.cs that were updated by Maestro and will no longer be updated. Is that correct?

dagood:
Is this auto-upgrade scope limiting still wanted? At this point there are some 1.1.0 auto-upgrade PRs merged without this change.

/cc @weshaggard

gkhanna79:
Yes, I think we want to keep this since 1.1.0 is not going to publish UWP packages and thus, shouldnt change them.

CC @weshaggard

dagood:
Ok, I'll go ahead and merge so we start limiting the auto-PRs. Here's the last 1.1.0 auto-upgrade that got merged before this change, which we can look at to hopefully see package versions that need to be changed back to stable: https://github.com/dotnet/core-setup/pull/1018.

weshaggard:
Yes I think this is wanted for now. We might later need to figure out how to auto-update multiple or we will just manually update UWP.

Issue #811, opened 2016-12-16
Incorrect PKG_RID selected during cross-build of corehost


hqueue:
## Steps to reproduce
Host: Ubuntu 14.04 x64

`./build.sh --env-vars TARGETPLATFORM=arm,TARGETRID=ubuntu.16.04-arm,CROSS=1`

## Expected  behavior
Build script correctly invoke cmake with `-DCLI_CMAKE_PKG_RID:STRING=ubuntu.16.04-arm`

## Actual behavior
cmake is invoked with `-DCLI_CMAKE_PKG_RID:STRING=ubuntu.14.04-arm`

## Reason
In the src/corehost/build.sh(https://github.com/dotnet/core-setup/blob/master/src/corehost/build.sh#L7), `-DCLI_CMAKE_PKG_RID` is defined using host machine RID even when cross building.

## Proposed solution
Use RID of rootfs for  `-DCLI_CMAKE_PKG_RID` instead of host machine RID.

hqueue:
Closing this issue, since PR #812 is merged.

PR #812, opened 2016-12-16
Select correct RID for package files when cross-building

hqueue:
Use RID of rootfs as RID for package files instead of host RID, when cross building corehost.

Previously host RID is used.

Close issue #811


hqueue:
CC @gkhanna79 @schellap @ramarag PTAL

Issue #813, opened 2016-12-16
Generate DotNetHost nupkgs for ubuntu.14.04-arm and ubuntu.16.04-arm


jyoungyun:
## Steps to reproduce
`./build.sh --verbose --env-vars CONFIGURATION=Release,TARGETPLATFORM=arm,TARGETRID=ubuntu.14.04-arm,CROSS=1,ROOTFS_DIR=/home/jyoung/git/dotnet/rootfs-coreclr/arm/`
or
`./build.sh --verbose --env-vars CONFIGURATION=Release,TARGETPLATFORM=arm,TARGETRID=ubuntu.16.04-arm,CROSS=1,ROOTFS_DIR=/home/jyoung/git/dotnet/rootfs-coreclr/arm/`

## Expected  behavior
Generate DotNetHost nupkgs for TARGETRID.

## Actual behavior
Always generate nupkgs for x64.

PR #712, opened 2016-12-02
Add new rids for Linux ARM32

hqueue:
Main issue: #725 

## Current approach

- [x] Add two RID for Linux ARM to dotnet-host-build: ubuntu.14.04-arm, ubuntu.16. 04-arm (https://github.com/dotnet/core-setup/pull/712/commits/139d22d544dada42b06bce6c08cd2c94e74f9e49, https://github.com/dotnet/core-setup/pull/712/commits/ba91959dce5a1573360fb07f02af956a5f03f9c0)
- [x] Update *.builds and *.pkgproj for corehost nupkgs (https://github.com/dotnet/core-setup/pull/712/commits/4836b7a66cee3d9a24e4dae70e3fc50a21674f6a)
- [x] Update dotnet-host-build to skip tests for Linux ARM, since it's cross building. (https://github.com/dotnet/core-setup/pull/712/commits/ec9e54fac3535fceba1a69f6c6aea9fd612d7fad)
- [x] Build nuget packages for the host and Microsoft.NETCore.App metapackage (https://github.com/dotnet/core-setup/pull/712/commits/4d6e435aed19474a90247c556e41bdfff4c19004)
- Update dotnet-host-build to support when cross crossgen is available  (TODO later)

## Old approach
I'm trying to build dotnet for rpi3 (ubuntu16.04-arm) and  I've added new rid "ubuntu.16.04-arm" to dotnet-host-build as a start point. I will also update CMake rules for corehost.

Is this a right step to enable building dotnet for ARM ?
Looking for an advice and any comment is appreciated.

PR #814, opened 2016-12-16
Generate DotNet nupkgs for ubuntu.14.04-arm and ubuntu.16.04-arm

jyoungyun:
Add runtime id for ubuntu.14.04-arm and ubuntu.16.04-arm
Currently, pkg/pack.sh does not support cross build.
By adding --rid option, pack.sh supports to generate DotNet nupkgs for arm.

Related issue : #813

### DIFF
--- a/pkg/pack.sh
+++ b/pkg/pack.sh
@@ -11,6 +11,16 @@ init_distro_name()
     fi
 }
 
+usage()
+{
+    echo "Usage: $0 --rid <Runtime Identifier>"
+    echo ""
+    echo "Options:"
+    echo "  --rid <Runtime Identifier>        Target Runtime Identifier"
+
+    exit 1
+}
+
 set -e
 
 # determine current directory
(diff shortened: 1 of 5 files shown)


jyoungyun:
@hqueue This patch generates DotNetHost nupkgs for ubuntu arm. But some parts of this patch conflict with your PRs. Let's talk about it next Monday.

gkhanna79:
@jyoungyun @hqueue A bunch of changes in this PR are the same set of changes @hqueue made in https://github.com/dotnet/core-setup/pull/712. Can you please merge the changes into a single PR that validates all packages for the host are being built as expected?

pkg/pack.sh
```diff
@@ -44,7 +71,9 @@ if [ "$(uname -s)" == "Darwin" ]; then
     __targets_param="TargetsOSX=true"
 else
     __targets_param="TargetsLinux=true"
-    init_distro_name
+    if [ -z $__distro_rid ]; then
```
schellap: When do you expect `--rid` to not be specified? It might be better to just error out here.

jyoungyun: If the pack.sh is always invoked by dotnet-build-tools, the __distro_rid will always be set. I considered the situation where the pack.sh is called directly without --rid option. However, in your opinion, I will modify this part in the next patch. Thank you.

PR #855, opened 2016-12-20
Provide flexibility in license used

ellismg:
These changes lay the groundwork in the build system to configure what license is used for artifacts that we build and publish.  In addition, it changes the default license used for our zips and tar.gz's to be the MIT license, per @richlander's request.

The general strategy is to introduced three knobs `InstallerLicenseType` `ZipFileLicenseType` and `NuPkgLicenseType` which can have one of two values `mit` or `msft`.  Depending on the value, we'll use that license when constructing the relevant artifact (msft here means the ".NET Library License").

The default for ZipFileLicenseType is now "mit" instead of "msft".

The repository now contains `msft` and `mit` license folders under `<repo-root>/resources` for different versions of some resources.  In the newly created `mit` folder, I updated the RTF EULA to be the MIT license, so if someone built with `InstallerLicenseType` set to `mit` they would get installers with the correct EULA.

I verified that the tgz's, zip's, deb's and MSI's have the correct licenses.  I will validate the PKG later today (my MacBook needs to be re-imaged).

Once I have sign-off on these changes, I'll inform shiproom and get approval to check into release/1.0.0.  I will also port these changes forward to release/1.1.0

@AlfredoMS is working on moving the build system in master to be MSBuild/BuildTools based so I will hold off on making these updates now, unless you're okay with me churning things in master @AlfredoMS.

/cc @gkhanna79 @schellap @AlfredoMS


gkhanna79:
LGTM overall. @schellap PTAL as well.

gkhanna79:
@ellismg Are you planning to get this into master and release/1.1.0 branch as well?

ellismg:
> @ellismg Are you planning to get this into master and release/1.1.0 branch as well?

Yes, once we had agreement on the approach and merged this in I was going to bring it to the other branches.

ellismg:
Solving this a different way...

PR #940, opened 2016-12-28
Fix Pipeline build break

gkhanna79:
Only set TARGETRID env var when dealing with Portable Native Binaries case.

Also, added the notion of real Target ID to BuildContext that should be used to lookup the location of artifacts folder for test runs.

### DIFF
--- a/build_projects/dotnet-host-build/PrepareTargets.cs
+++ b/build_projects/dotnet-host-build/PrepareTargets.cs
@@ -43,6 +43,8 @@ namespace Microsoft.DotNet.Host.Build
             nameof(PackDotnetDebTool))]
         public static BuildTargetResult Init(BuildTargetContext c)
         {
+            // CommonInit(c);
+
             var configEnv = Environment.GetEnvironmentVariable("CONFIGURATION");
             string platformEnv = c.BuildContext.Get<string>("Platform");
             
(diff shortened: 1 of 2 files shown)


gkhanna79:
@schellap PTAL - this is to address the pipeline build break.

gkhanna79:
CC @wtgodbe

gkhanna79:
@schellap Any inputs on this?

schellap:
This is very confusing. I wanted to chat to understand whats really going on here. The realTargetID? There seem to be 3 notions: build machine rid, the build target rid and the packaging rid. Where does this fall and an apt name would be better.

gkhanna79:
I am happy to chat but here is the issue:

Originally, I was unconditionally setting the TARGETRID environment variable so that the linux-* RID can be seen by other parts of the build system. However, on Windows, we have the following translation always being done:

https://github.com/dotnet/core-setup/pull/940/files#diff-ff058bba01b969c51b1b7f3b68cd9fe0R100

As a result, we were unconditionally setting the TARGETRID environment variable to win7-*, resulting in incorrect artifacts folder location. However, the build-definition sets a variable (in the definition) that indicates the RID to be win10-*. As a result, the signing step would look for binaries  in Win10-* location but wouldn't find them since they got dropped into win7-* location.

The fix was to only set the TARGETRID environment variable for linux-* case.

Once I fixed the above, there is no TARGETRID env var being setup. As a result, I ran into the issue of TEST_TARGETRID being set to the value of TargetRID that is setup in build context. However, on Windows, TargetRID in build context is the transformed value (see the link above) while TestTargets needs the actual RID value to lookup the artifacts folder location where binaries were dropped  and not the transformed location when TARGETRID env var is not set, as was the case prior to my change (see https://github.com/dotnet/core-setup/blob/master/build_projects/shared-build-targets-utils/Utils/Dirs.cs#L13 on how various folders are initialized).

Hence, the notion of ActualTargetRid to support lookup. I thought of passing the path but since environment variables are not good for long file paths. Hence, the actual RID.

This confusion around the transformation is what I commented in my change as well. Core-Setup build system is very fragile, IMHO.

gkhanna79:
> Where does this fall and an apt name would be better.

As explained above, the "build target RID" are two different RIDs for a Windows build - one is what is transformed to be the win7 based (explained above) and the other is the actual target rid that is used to compute subfolder names, etc.

gkhanna79:
@schellap PTAL - changed the variable name.

schellap:
:+1

gkhanna79:
@wtgodbe this is now merged. Can we queue off a build?

wtgodbe:
@gkhanna79 merging this triggered a build - https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/explorer?_a=summary&buildId=492477

Issue #849, opened 2016-12-20
Disable to generate Debian packages during the cross builds


jyoungyun:
## Steps to reproduce
```
./build.sh --env-vars DISABLE_CROSSGEN=1,CONFIGURATION=Release,TARGETPLATFORM=arm,TARGETRID=ubuntu.14.04-arm,CROSS=1,ROOTFS_DIR=/home/jyoung/git/dotnet/rootfs-coreclr/arm/
```

## Expected  behavior
Build succeed

## Actual behavior
```
dpkg-shlibdeps: error: couldn't find library libdl.so.2 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: couldn't find library libpthread.so.0 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: couldn't find library libstdc++.so.6 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: couldn't find library libm.so.6 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: couldn't find library libgcc_s.so.1 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: couldn't find library libc.so.6 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: couldn't find library ld-linux-armhf.so.3 needed by debian/dotnet-host/usr/share/dotnet/dotnet (ELF format: 'elf32-littlearm-hfabi'; RPATH: '')
dpkg-shlibdeps: error: cannot continue due to the errors listed above
Note: libraries are not searched in other binary packages that do not have any shlibs or symbols file.
To help dpkg-shlibdeps find private libraries, you might need to use -l.
dh_shlibdeps: dpkg-shlibdeps -Tdebian/dotnet-host.substvars debian/dotnet-host/usr/share/dotnet/dotnet returned exit code 2
[...]

PR #943, opened 2016-12-28
Disable GenerateDebs target for building ubuntu arm.

jyoungyun:
By using BuidlArchitectures attribute, the GenerateDebs target is disabled.
BuildArchitecture.arm enumeration is also added.

Related issue : #849

### DIFF
--- a/build_projects/Microsoft.DotNet.Cli.Build.Framework/Enumerations/BuildArchitecture.cs
+++ b/build_projects/Microsoft.DotNet.Cli.Build.Framework/Enumerations/BuildArchitecture.cs
@@ -3,6 +3,8 @@ namespace Microsoft.DotNet.Cli.Build.Framework
     public enum BuildArchitecture
     {
         x86 = 1,
-        x64 = 2
+        x64 = 2,
+        arm = 3,
+        arm64 = 4
     }
 }
--- a/build_projects/Microsoft.DotNet.Cli.Build.Framework/CurrentArchitecture.cs
+++ b/build_projects/Microsoft.DotNet.Cli.Build.Framework/CurrentArchitecture.cs
@@ -31,9 +31,35 @@ namespace Microsoft.DotNet.Cli.Build.Framework
             }
         }
 
+        public static bool Isarm
+        {
+            get
+            {
+                var archName = Environment.GetEnvironmentVariable("TARGETPLATFORM");
+                return string.Equals(archName, "arm", StringComparison.OrdinalIgnoreCase);
+            }
+        }
+
+        public static bool Isarm64
+        {
+            get
+            {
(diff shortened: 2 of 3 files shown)


build_projects/Microsoft.DotNet.Cli.Build.Framework/Enumerations/BuildArchitecture.cs
```diff
@@ -3,6 +3,7 @@ namespace Microsoft.DotNet.Cli.Build.Framework
     public enum BuildArchitecture
     {
         x86 = 1,
-        x64 = 2
+        x64 = 2,
+        arm = 3
```
gkhanna79: While you are augmenting this @jyoungyun, can you please add support for Arm64 as well?

gkhanna79:
LGTM modulo a comment.

@schellap FYI

build_projects/Microsoft.DotNet.Cli.Build.Framework/Enumerations/BuildArchitecture.cs
jyoungyun: Sure, I added arm64. But I do not add `Isarm64()` because I don't know what value you use when you build arm64 modules.

build_projects/Microsoft.DotNet.Cli.Build.Framework/Enumerations/BuildArchitecture.cs
```diff
@@ -3,6 +3,8 @@ namespace Microsoft.DotNet.Cli.Build.Framework
     public enum BuildArchitecture
     {
         x86 = 1,
-        x64 = 2
+        x64 = 2,
+        arm = 3,
+        arm64 = 4
```
gkhanna79: IsArm64 should, for now, support "arm64" as the arch string to compare for.

jyoungyun: I cound not add the function because I was not sure what you were using. Do you use `TARGETPLATFORM` environment for recognizing arm64? If you share your build command for ARM64, I can add the `Isarm64()` function.

gkhanna79: Please see https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/build.ps1#L39 - targetplatform gets sets to TargetArch which will be Arm64.

jyoungyun:
cc/ @gkhanna79 @schellap  Happy new year~

gkhanna79:
Happy New Year to you as well @jyoungyun

PR #1029, opened 2017-01-05
Prevent any refs from runtime packages

ericstj:
NuGet has a bug where it's consider RID-specific assets for compile.
https://github.com/NuGet/Home/issues/4207

Workaround that by putting a placeholder under ref.

/cc @weshaggard


ericstj:
@dotnet-bot test this please

PR #1071, opened 2017-01-08
Fix arm cross-compile target arg

stevedesmond-ca:
Build currently fails on "armv7-linux-gnueabihf: No such file or directory" running xenial (and no package/binary/file exists with that name, that I'm aware) -- I know most are testing on trusty, so we may need to branch if trusty doesn't have an "arm-linux-gnueabihf"


stevedesmond-ca:
Actually, hold off on this for now -- just noticed CoreCLR and CoreFX also have this, and I've been able to build them all along, so something else must be up.

stevedesmond-ca:
n/m, looks like my dev environment is fried

Issue #736, opened 2016-12-08
Carry uCRT binaries app-local


gkhanna79:
This is required to alleviate the dependency on an external installer (of CRT). This will be achieved in the following manner once @weshaggard has enabled a flat-file version of Microsoft.NETCore.App in  Core-Setup repo:

1) Pull in uCRT binaries in Microsoft.NETCore.App (flat-file) package
2) Update the installer to no longer chain into CRT installer.

CC @schellap @ramarag @Petermarcu

ericstj:
Can you add details about specifically which files are required and on what versions of the OS?

ericstj:
Also, it makes more sense to me that these get packaged in the CoreCLR repo.  That way they remain in sync with the actual dependencies that CoreCLR takes.  Putting this in core-setup feels to me like we're going to be in a place where CoreCLR makes an update and we need to *remember* make a corresponding update in core-setup to pick up newer bits that are in sync with what CoreCLR needs.

Adding stuff to the flat packages is trivial so that shouldn't be a deciding factor.  It's a one-time thing then gets automatically updated with maestro.  If you make it part of the CoreCLR package closure (as was done with API sets) then there isn't even a one-time hit.

gkhanna79:
@wtgodbe Assigning this to you to finish the Core-Setup side of changes once the coreCLR package is merged in by Maestro.

@Petermarcu Do you have a change for it that you wish to merge?

PR #1088, opened 2017-01-10
Update net core app to 2.0

ericstj:
Version NETCore.App to 2.0 and make it target netcoreapp2.0.

Fix the tests to build without having packages from other RIDs.

Manually add conflict resolution by using `"exclude": "all"` for all packages absorbed into NETCore.App.  This is just a workaround until we can switch to msbuild-based CLI which has the conflict resolution target.

Make NETCore.App depend on NETStandard.Library 2 so that it brings in the conflict resolution targets.

/cc @weshaggard 

This may still fail CI due to myget delay's indexing the new NETStandard.Library 2.0.0-* package.


ellismg:
> Fix the tests to build without having packages from other RIDs.

Yay!

ellismg:
@dotnet-bot test this please

(Hopefully this should pick up my changes to do archiving)

ellismg:
Appears to be working, @ericstj.  You can pull the artifacts down from details on each job (there should also be a URL to pull down zip of all artifacts for each leg).

ericstj:
Cool, thanks @ellismg

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -51,9 +51,24 @@
...
+      <TargetPath>.NETCoreApp</TargetPath>
+    </File>
+
+    <!-- Bring in Platforms for RID graph, NETStandard.Library for build-tools -->
     <Dependency Include="Microsoft.NETCore.Platforms">
       <Version>1.2.0-beta-24821-0</Version>
-      <TargetFramework>.NETCoreApp1.1</TargetFramework>
+      <TargetFramework>.NETCoreApp2.0</TargetFramework>
+    </Dependency>
+    <Dependency Include="NETStandard.Library">
```
weshaggard: Did you publish a newer version of this package with all your changes?

pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
```diff
@@ -0,0 +1,131 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
```
weshaggard: Don't we need Microsoft.NETCore.Platforms in this list?

weshaggard:
>Fix the tests to build without having packages from other RIDs.

Is the one template project.json file the only change needed? We should probably also clean up the dummy package generation now, @AlfredoMS.

gkhanna79:
@ericstj As part of bringing up the flat package, can you also look into fixing https://github.com/dotnet/core-setup/issues/736?

CC @weshaggard

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
ericstj: Yeah, currently only on dotnet-core-dev-eng.

pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
ericstj: It comes in transitively.  Any package that uses RIDs references it.

ericstj:
> Is the one template project.json file the only change needed? 

These were the changes needed: https://github.com/dotnet/core-setup/pull/1088/commits/aba3ad6e41ea8f26b915be0baf6aebdb3a9ac6a9

I suspect this may break maestro, so we may need to fix that to understand project.json.template.  I suspect corefx has the same issue.

weshaggard:
> I suspect corefx has the same issue.

Yes I think it will be broken there as well.

dagood:
@ericstj I don't think anything will break, since those test asset files aren't considered for auto-upgrade. I'm also pretty sure auto-upgrade doesn't care what the runtime keys are, so if/when runtime tokens are added to upgrade-worthy project.jsons I don't know if any changes are needed anyway. If something comes up, I'll definitely help fix it.

(FYI, Maestro is just the service that triggers "things" when a repo changes, so it has no upgrade domain knowledge--that's in buildtools, or core-setup and CLI's custom implementations.)

weshaggard:
@dagood I think the change we will need is for the change commands to also update project.json.template files like they do project.json files. We will need that for our corefx dev/eng work as well.

dagood:
@weshaggard I see, currently buildtools auto-upgrades everything in `ProjectJsonFiles` so it might be doable just to include the `.json.template` files too. The item name gets confusing then, though. Let me know if you'd like me to help test out auto-upgrade on a branch to see if they happen right.

weshaggard:
If it is that simple that will be great and I'm not worried about the property name.

PR #1089, opened 2017-01-10
Archive some artifacts as part of CI

ellismg:
This should teach the CI system archive the artifacts we care about as part of a PR.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -32,7 +32,7 @@ platformList.each { platform ->
         else {
             buildCommand = ".\\build.cmd -Configuration ${configuration} -Architecture ${architecture} -Targets Default"
         }
-        
+
     }
     else if (os == 'Windows_2016') {
         buildCommand = ".\\build.cmd -Configuration ${configuration} -Architecture ${architecture} -RunInstallerTestsInDocker -Targets Default"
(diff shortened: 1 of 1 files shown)


ellismg:
/cc @weshaggard @ericstj 

I'll merge this in tonight and do a dummy PR to make sure the jobs work as expected.  Note that CI does not do per push jobs (but we could fix that here if we wanted) so we'll have to pull the artifacts off of Jenkins for a dummy PR).

netci.groovy
```diff
@@ -61,12 +62,20 @@ platformList.each { platform ->
...
+
     if (!(os == 'Windows_NT' && architecture == 'arm')) {
         Utilities.addXUnitDotNETResults(newJob, '**/*-testResults.xml')
     }
-    
+
     Utilities.addGithubPRTriggerForBranch(newJob, branch, "${os} ${architecture} ${configuration} Build")
+
+    ArchivalSettings settings = new ArchivalSettings();
+    def archiveString = ["tar.gz", "zip", "deb", "msi", "nupkg", "exe"].collect { "artifacts/*/packages/*.${it},artifacts/*/corehost/*.${it}" }.join(",")
```
ericstj: .pkg as well for OS X?

ellismg: Good catch.  Will do.

ellismg:
@dotnet-bot test ci please

PR #1109, opened 2017-01-11
[WIP] dotnet cache host work

schellap:
Cc @ramarag, please use this runtime for your work.

The "tfm" property needs to be specified in the app.runtimeconfig.json once you publish with filter profile.

```
{
    "tfm": "netcoreapp1.1"
}
```

### DIFF
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -18,10 +18,53 @@ arguments_t::arguments_t() :
 {
 }
 
+/**
+ *
+ * Setup the shared package directories.
+ *
+ *  o %DOTNET_SHARED_PACKAGES% -- multiple delimited paths
+ *  o $HOME/.dotnet/{x86|x64}/ or %USERPROFILE%\.dotnet\{x86|x64}
+ *  o dotnet.exe relative shared packages
+ *  o Global location
+ *      Windows: C:\Program Files (x86) or
+ *      Unix: directory of dotnet on the path.
+ */
+void setup_shared_package_paths(const hostpolicy_init_t& init, const pal::string_t& own_dir, arguments_t* args)
+{
+    if (init.tfm.empty())
+    {
+        // Old (MNA < 1.1.*) "runtimeconfig.json" files do not contain TFM property.
+        return;
+    }
+
+    // Environment variable DOTNET_SHARED_PACKAGES
+    (void) get_env_shared_package_dirs(&args->env_shared_packages, get_arch(), init.tfm);
+
+    // User profile based packages
+    pal::string_t local_shared_packages;
+    if (pal::get_local_shared_package_dir(&local_shared_packages))
+    {
+        append_path(&local_shared_packages, init.tfm.c_str());
+        args->local_shared_packages = local_shared_packages;
+    }
+
+    // "dotnet.exe" relative shared packages folder
+    if (init.host_mode == host_mode_t::muxer)
+    {
+        args->dotnet_shared_packages = own_dir;
+        append_path(&args->dotnet_shared_packages, _X("packages"));
+        append_path(&args->dotnet_shared_packages, init.tfm.c_str());
+    }
+
+    // Global shared package dir
+    if (pal::get_global_shared_package_dir(&args->global_shared_packages))
+    {
+        append_path(&args->global_shared_packages, init.tfm.c_str());
+    }
+}
+
 bool parse_arguments(
-    const pal::string_t& deps_path,
-    const std::vector<pal::string_t>& probe_paths,
-    host_mode_t mode,
+    const hostpolicy_init_t& init,
     const int argc, const pal::char_t* argv[], arguments_t* arg_out)
 {
     arguments_t& args = *arg_out;
@@ -35,7 +78,7 @@ bool parse_arguments(
     auto own_name = get_filename(args.own_path);
     auto own_dir = get_directory(args.own_path);
     
-    if (mode != host_mode_t::standalone)
+    if (init.host_mode != host_mode_t::standalone)
     {
         // corerun mode. First argument is managed app
         if (argc < 2)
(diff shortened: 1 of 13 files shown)


schellap:
Cc @gkhanna79

src/corehost/cli/args.cpp
```diff
@@ -96,5 +139,7 @@ bool parse_arguments(
     pal::getenv(_X("DOTNET_HOSTING_OPTIMIZATION_CACHE"), &args.dotnet_packages_cache);
```
gkhanna79: We should remove this (and supporting code) out with this feature coming in.

src/corehost/cli/args.h
```diff
@@ -21,68 +21,65 @@ struct probe_config_t
...
-        assert(!is_roll_fwd_set() || probe_deps_json == nullptr);
         // Will not do hash match when probing a deps json.
         assert(!match_hash || probe_deps_json == nullptr);
     }
 
-    static probe_config_t svc_ni(const pal::string_t& dir, bool patch_roll_fwd, bool prerelease_roll_fwd)
+    static probe_config_t svc_ni(const pal::string_t& dir)
     {
-        return probe_config_t(dir, false, patch_roll_fwd, prerelease_roll_fwd, nullptr, true, true);
+        return probe_config_t(dir, false, nullptr, true, true, false);
```
gkhanna79: Is specifying the exact values intended to be a simplification of what happens today?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -267,9 +243,18 @@ void deps_resolver_t::setup_additional_probes(const std::vector<pal::string_t>&
     }
 }
 
-bool deps_resolver_t::probe_entry_in_configs(const deps_entry_t& entry, pal::string_t* candidate)
+/**
+ * Given a deps entry, do a probe (lookup) for the file, based on the probe config.
+ *   -- When match hash is specified, the nuget cache SHA is matched. See .sha512 files in %USERPROFILE%\.nuget.
```
gkhanna79: Didn't we need hash matching for Antares cache as implemented today, which we no longer need post this change?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -289,71 +274,52 @@ bool deps_resolver_t::probe_entry_in_configs(const deps_entry_t& entry, pal::str
...
             if (config.probe_deps_json->has_package(entry.library_name, entry.library_version) && entry.to_dir_path(probe_dir, candidate))
             {
-                trace::verbose(_X("    Probed deps json and matched [%s]"), candidate->c_str());
+                trace::verbose(_X("    Probed deps json and matched '%s'"), candidate->c_str());
                 return true;
             }
             trace::verbose(_X("    Skipping... probe in deps json failed"));
         }
-        else if (!config.is_roll_fwd_set())
+        else if (config.probe_publish_dir)
```
gkhanna79: How is roll-forward matching being done in this case? That is, why do we no longer need those two checks?

src/corehost/cli/libhost.h
```diff
@@ -194,27 +206,35 @@ struct hostpolicy_init_t
...
-        {
-            trace::error(_X("The size of the data layout used to initialize %s is %d; expected at least %d"), LIBHOSTPOLICY_NAME, input->version_lo, HOST_INTERFACE_LAYOUT_VERSION_LO);
-            return false;
-        }
+
         trace::verbose(_X("Reading from host interface version: [0x%04x:%d] to initialize policy version: [0x%04x:%d]"), input->version_hi, input->version_lo, HOST_INTERFACE_LAYOUT_VERSION_HI, HOST_INTERFACE_LAYOUT_VERSION_LO);
 
-        make_clrstr_arr(input->config_keys.len, input->config_keys.arr, &init->cfg_keys);
-        make_clrstr_arr(input->config_values.len, input->config_values.arr, &init->cfg_values);
+        if (input->version_lo >= offsetof(host_interface_t, host_mode) + sizeof(input->host_mode))
```
gkhanna79: Can you please add some comments here?

src/corehost/cli/args.cpp
schellap: Sure, I think we can remove this. Cc @pakrym

src/corehost/cli/args.h
schellap: I just removed some dead code regarding patch roll forwards for individual packages and added a parameter.

src/corehost/cli/deps_resolver.cpp
schellap: Because, we don't roll forward for individual packages anywhere. So it was all dead code.

src/corehost/cli/libhost.h
schellap: Sure, the reasoning is an old hostfxr should still load new hostpolicy. If we add TFM struct field, an old hostfxr will not provide it. The version_lo (sizeof) the old hostfxr saw at build time would be smaller and we should not attempt to read tfm in that case.

src/corehost/cli/deps_resolver.cpp
schellap: @glennc with this work, the `DOTNET_HOSTING_OPTIMIZATION_CACHE` will go away and we will use shared packages. The proposal is available here: https://github.com/dotnet/core-proposals/blob/master/DotNetCore-SharedPackageStore.md

src/corehost/cli/args.cpp
pakrym: /cc @muratg FYI

schellap: @muratg https://github.com/dotnet/core-proposals/blob/master/DotNetCore-SharedPackageStore.md This should give an idea of how to author shared packages with runtime support. We will even provide you with `dotnet cache` command to crossgen and build such a store. Cc @ramarag

ramarag:
#1330 is incorporating this change

PR #1115, opened 2017-01-11
Archive packages in pkg output

ericstj:
The pkg folder builds to pkg/bin, not artifacts.  Make sure to consider
this path as well.

/cc @ellismg @chcosta @weshaggard


ellismg:
LGTM.

ericstj:
Thanks, I think I'll need a dummy PR to get this to take effect, right?

ellismg:
Yeah, I noticed after the merge there was a problem and opened #1116 to fix.  Once that's merged I'll re-run the generator (instead of waiting for Jenkins to notice and auto-run it) and then we can do a PR.

PR #1117, opened 2017-01-11
Upgrade nss in the Fedora 23 dockerfile (port release/1.1.0 -> master)

dagood:
Port #541 to master to fix restore timeout issue. The latest repro was during `init-tools` in the pkg build.

@gkhanna79


ellismg:
LGTM

PR #1119, opened 2017-01-11
Move package dependencies to pj for maestro

ericstj:
This moves the package dependencies to the project.json.template so that
maestro can auto update them.

We can delete the hacky portion of this once we have msbuild-based CLI
that supports PackageReference.

/cc @weshaggard

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
@@ -14,6 +14,12 @@
   <ItemGroup>
     <None Include="project.json" />
   </ItemGroup>
+  <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' == ''">
+    <!-- Bring in Platforms for RID graph, NETStandard.Library for build-tools -->
+    <DependenciesToPackage Include="NETStandard.Library" />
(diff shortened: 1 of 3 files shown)


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
```diff
@@ -65,6 +71,32 @@
     <Message Importance="low" Text="%(FilesToPackage.PackagePath) -> @(FilesToPackage)" />
   </Target>
 
+  <Target Name="GetDependenciesToPackage" Condition="'@(DependenciesToPackage)' != ''" DependsOnTargets="ResolveNuGetPackages" Returns="@(_DependenciesToPackageWithVersion)">
+    <!-- hack because current nuget task doesn't return version for ReferencedNuGetPackages,
```
ellismg: Can you file an issue so we make sure we remember to pay down the debt?

ericstj: No issue is needed it will break when this happens.  We're not filing an issue to say "delete all project.jsons" right?

ellismg: Okay, makes sense.

PR #1122, opened 2017-01-12
Collection of small build improvements

ellismg:
- Get dependency updating working again
- Use NuGet.Config instead of passing an explicit list of sources
- Baseline System.ValueTuple problem when consuming a new Roslyn

### DIFF
--- a/pkg/dir.props
+++ b/pkg/dir.props
@@ -69,19 +69,16 @@
 
   <Import Project="$(ToolsDir)Build.Common.props" Condition="Exists('$(ToolsDir)Build.Common.props')" />
 
-  <ItemGroup>
-    <NuGetSource Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core-dev-eng/api/v3/index.json" />
-    <NuGetSource Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v3/index.json" />
-    <!-- needed for Microsoft.DiaSymReader.Native, potentially find a better feed -->
-    <NuGetSource Include="https:%2F%2Fapi.nuget.org/v3/index.json" />
-  </ItemGroup>
+  <PropertyGroup>
+    <NuGetConfigFile Condition="'$(NuGetConfigFile)' == ''">$(ProjectDir)/../NuGet.Config</NuGetConfigFile>
+  </PropertyGroup>
 
   <PropertyGroup>
     <DotNetTool Condition="'$(DotNetTool)' == '' and '$(OsEnvironment)'!='Unix'">$(DotnetCliPath)dotnet.exe</DotNetTool>
     <DotNetTool Condition="'$(DotNetTool)' == '' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotNetTool>
     <DotNetRestoreCommand>"$(DotNetTool)" restore</DotNetRestoreCommand>
     <DotNetRestoreCommand>$(DotNetRestoreCommand) --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))"</DotNetRestoreCommand>
-    <DotNetRestoreCommand>$(DotNetRestoreCommand) @(NuGetSource->'--source %(Identity)', ' ')</DotNetRestoreCommand>
+    <DotNetRestoreCommand>$(DotNetRestoreCommand) --configfile "$(NuGetConfigFile)"</DotNetRestoreCommand>
     <DnuRestoreCommand>$(DotNetRestoreCommand)</DnuRestoreCommand>
   </PropertyGroup>
 
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
@@ -66,7 +66,13 @@
 
     <!-- simple check to make sure we don't accidentally pull an old 
          package that should have been merged to the primary package -->
-    <Error Text="Unexpected System package %(_secondaryFilesByPackagePath.NuGetPackageId)" Condition="$([System.String]::new('%(_secondaryFilesByPackagePath.NuGetPackageId)').Contains('System.'))" />
+    <ItemGroup>
+      <PermittedSystemPackages Include="System.ValueTuple" />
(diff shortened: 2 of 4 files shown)


ellismg:
@ericstj Is bc9d130 what you were thinking of when we discussed the ValueTuple issue or do you want it done some other way?

pkg/dir.props
```diff
@@ -69,19 +69,12 @@
...
-    <NuGetSource Include="https:%2F%2Fapi.nuget.org/v3/index.json" />
-  </ItemGroup>
-
   <PropertyGroup>
     <DotNetTool Condition="'$(DotNetTool)' == '' and '$(OsEnvironment)'!='Unix'">$(DotnetCliPath)dotnet.exe</DotNetTool>
     <DotNetTool Condition="'$(DotNetTool)' == '' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotNetTool>
     <DotNetRestoreCommand>"$(DotNetTool)" restore</DotNetRestoreCommand>
     <DotNetRestoreCommand>$(DotNetRestoreCommand) --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))"</DotNetRestoreCommand>
-    <DotNetRestoreCommand>$(DotNetRestoreCommand) @(NuGetSource->'--source %(Identity)', ' ')</DotNetRestoreCommand>
+    <DotNetRestoreCommand>$(DotNetRestoreCommand) --configfile "$(ProjectDir)/../NuGet.Config"</DotNetRestoreCommand>
```
weshaggard: Do you want to make the nuget.config file its own property so we can control it by passing in another value?

ellismg: Sure, I can do this.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
```diff
@@ -66,7 +66,7 @@
 
     <!-- simple check to make sure we don't accidentally pull an old 
          package that should have been merged to the primary package -->
-    <Error Text="Unexpected System package %(_secondaryFilesByPackagePath.NuGetPackageId)" Condition="$([System.String]::new('%(_secondaryFilesByPackagePath.NuGetPackageId)').Contains('System.'))" />
+    <Error Text="Unexpected System package %(_secondaryFilesByPackagePath.NuGetPackageId)" Condition="$([System.String]::new('%(_secondaryFilesByPackagePath.NuGetPackageId)').Contains('System.')) and ('%(_secondaryFilesByPackagePath.NuGetPackageId)' != 'System.ValueTuple')" />
```
ericstj: This is ugly but I am ok with it since it is temporary and I don't expect us to ever have a *standing* exclusion.

pkg/dir.props
```diff
@@ -69,19 +69,16 @@
...
+    <NuGetConfigFile Condition="'$(NuGetConfigFile)' == ''">$(ProjectDir)/../NuGet.Config</NuGetConfigFile>
+  </PropertyGroup>
 
   <PropertyGroup>
     <DotNetTool Condition="'$(DotNetTool)' == '' and '$(OsEnvironment)'!='Unix'">$(DotnetCliPath)dotnet.exe</DotNetTool>
     <DotNetTool Condition="'$(DotNetTool)' == '' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotNetTool>
     <DotNetRestoreCommand>"$(DotNetTool)" restore</DotNetRestoreCommand>
     <DotNetRestoreCommand>$(DotNetRestoreCommand) --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))"</DotNetRestoreCommand>
-    <DotNetRestoreCommand>$(DotNetRestoreCommand) @(NuGetSource->'--source %(Identity)', ' ')</DotNetRestoreCommand>
+    <DotNetRestoreCommand>$(DotNetRestoreCommand) --configfile "$(NuGetConfigFile)"</DotNetRestoreCommand>
```
ericstj: In the past we avoided using nuget.config because of the precedence.  Folks were defining user-wide config files that happened to have the same keys as ours and disabling the feeds (usually accidentally).  This caused restore to fail (or sometimes silently select a different package version) from another feed.  

If we're going to use config, please make sure it's fool-proof.  I haven't tested but perhaps adding 
```xml
<packageSources>
  <!-- To inherit the global NuGet package sources remove the <clear/> line below -->
  <clear />
  <!-- our feeds -->
</packageSources>
<disabledPackageSources>
  <!-- ensure user settings don't disable our feeds -->
  <clear />
</disabledPackageSources>
```

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
ericstj: The not so ugly way would be to do the following:
```xml

    <!-- simple check to make sure we don't accidentally pull an old 
         package that should have been merged to the primary package -->
    <ItemGroup>
      <PermittedSystemPackages Include="System.ValueTuple" />
      <_secondarySystemPackages Include="@(_secondaryFilesByPackagePath->'%(NuGetPackageId)')" 
                                Condition="$([System.String]::new('%(Identity)').Contains('System.'))" />
      <_secondarySystemPackages Remove="@(PermittedSystemPackages)" />
    </ItemGroup>
    <Error Text="Unexpected System package(s) @(_secondarySystemPackages)" Condition="'@(_secondarySystemPackages)' != ''" />
```

pkg/dir.props
ellismg: Okay, I fixed this.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
ellismg: I was running into some problems, Eric with the above.  I think the problem may be due to the fact that we want to use the NuGetPackageId to do the filtering instead of Identity (I was finding cases where runtime packages which contributed resources named "System" were getting flagged).  I'll take a look with you in person tomorrow.  Simply changing the metadata we do the Condition on didn't causes the task to fail when it *should have* so I'll need some more MSBuild help, I fear.

ellismg:
@ericstj let me know if you want me to wait for your merge from master into dev/eng before merging this.

ericstj:
Go ahead, that's in.

PR #1137, opened 2017-01-13
Support below nupkgs build in non-Windows environment

jyoungyun:
Now, Microsoft.DotNet.PlatformAbstractions and Microsoft.Extensions.DependencyModel's nupkg does not include the necessary dlls for use. This patch should be merged for the cli build.

Related issue: #725

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -341,7 +341,9 @@ namespace Microsoft.DotNet.Host.Build
         [Target]
         public static BuildTargetResult BuildProjectsForNuGetPackages(BuildTargetContext c)
         {
-            if (CurrentPlatform.IsWindows)
+            string rid = c.BuildContext.Get<string>("TargetRID");
+
+            if (CurrentPlatform.IsWindows || rid == "ubuntu.14.04-arm" || rid == "ubuntu.16.04-arm" || rid == "tizen.4.0.0-armel" )
             {
                 var configuration = c.BuildContext.Get<string>("Configuration");
 


ericstj:
This seems OK but now we'll have multiple legs of the official build producing these.  I think we only want the Windows ones to be published since those will be the only ones signed.  @AlfredoMS @dagood I think it's just last-in wins in core-setup, or is publishing smart enough to know to only publish the Windows one?

dagood:
The build should be smart enough: publishing the managed packages to Azure depends on `SIGNED_PACKAGES` being `true`, which it is in the Windows x64 build. See https://github.com/dotnet/core-setup/pull/500

dagood:
I'm actually not sure I understand what this PR is meant to do. I'm not familiar with this part of the build, though. To me, it looks like it will enable building the packages on more platforms, but aren't the DLLs lost if the packages aren't published?

jyoungyun:
@dagood @ericstj We have a goal to release arm and armel, and have modified nupkg related code in coreclr and corefx to do this and are working on uploading nupkgs. @gkhanna79 seems to be able to give an information for uploading nupkgs. To build cli project, actually I have plans to enable cli build for arm and armel, these nupkgs need. Um, why do these nupkgs upload to azure? Is not nupkg downloaded from myget or nuget? As I know, cli project needs a shared runtime tarball which is downloaded from azure but other nupkgs are downloaded from myget. Am I wrong? And... I wanna upload a shared tarball for arm and armel to azure for building cli project. Is it possible? Thank you.

ericstj:
Thanks for the pointer @dagood. 

@jyoungyun you shouldn't need these packages in order to produce the tarball.  These aren't part of the shared framework, they just happen to live in the same repo.  CLI does have a dependency on the packages, but it can pick them up from myget.

The upload to azure we're discussing is just how our official build works.  We upload the outputs from many machines to azure, and once the entire build is done the outputs are published to myget in a single operation.

jyoungyun:
@ericstj I understood what you said. However, I doubt whether it is normal for nupkg to be created that can not be used when building on Linux. Anyway, the only thing that is uploaded is sign, I wanna make a normal nupkg when we build on another host build environment.

ericstj:
In that case it sounds like you're trying to build without using myget, is that correct?  

This is something @ellismg is working on: our build-from-source effort.  @ellismg how are you dealing with these two core-setup packages for build-from-source?

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -341,7 +341,9 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
         [Target]
         public static BuildTargetResult BuildProjectsForNuGetPackages(BuildTargetContext c)
         {
-            if (CurrentPlatform.IsWindows)
+            string rid = c.BuildContext.Get<string>("TargetRID");
+
+            if (CurrentPlatform.IsWindows || rid == "ubuntu.14.04-arm" || rid == "ubuntu.16.04-arm" || rid == "tizen.4.0.0-armel" )
```
gkhanna79: What about Debian armel?

Also, @schellap @eerhardt Do you know why is this built only for Windows?

eerhardt: It was only built on Windows before because we only ever published the NuGet packages from the Windows build, since that was the only build that signed the assemblies.

So it was a build optimization - nothing ever used/consumed these NuGet packages on non-Windows, so there was no reason to build them.

gkhanna79: Thanks @eerhardt. 

@jyoungyun Why do we need this change?

ellismg:
> @ellismg how are you dealing with these two core-setup packages for build-from-source?

I wasn't doing anything special today but I also wasn't doing anything that depended on these packages (since I'm currently stopping at core-setup and the only artifact I care about presently from core-setup is the zips).

I would have probably just removed the conditional logic completely and always built the package (and if we don't end up doing that now, I'll look into what it means to do it when I start building stuff above core-setup).

ericstj:
> removed the conditional logic completely and always built the package

Sounds reasonable to me.  @jyoungyun if your scenario is to build everything without ever going to myget then I'd just remove the condition entirely.  If instead you are OK going to myget for these packages, I'd do that instead or building them yourself since they are not arm-specific.

jyoungyun:
@ericstj I'd like to remove the condition entirely but if you are not understand why we need to build DotNetHost nupkgs for all platform, it's okay to keep the current status. But I think we need to add some comments the reason why we do not build normal nupkgs outsides Windows. Linux and other developers will not know that the generated DotNet nupkgs are not normal.

gkhanna79:
> if you are not understand why we need to build DotNetHost nupkgs for all platform

Host packages need to be built for every platform. I think the question is open for DependencyModel and PlatformAbstraction packages which are not host packages.

ericstj:
The DependencyModel and PlatformAbstraction packages are AnyCPU IL only so in the official builds we don't build them everywhere, as merely an optimization to only build what we ship.

As @ellismg and I mentioned the build from source effort requires that we are able to build all packages for a specific distro of the product from a single machine.  This gives us a reason to build on all platforms which we didn't have before.  So please just remove the condition entirely.  There is no need to add any comments since we're actually removing the special case.

PR #1148, opened 2017-01-14
Add checked-in build definitions

naamunds:
The Windows signing build definitions will be added later, because the
VSTS instance where we want to run the builds doesn't support the
current signing process, so we're working on modifying our signing
process.

In a separate PR, I plan to refactor the Linux build definitions into a
smaller number of build definitions.

/cc @chcosta @wtgodbe @dagood

### DIFF
--- a/buildpipeline/Core-Setup-PortableLinux-x64-master.json
+++ b/buildpipeline/Core-Setup-PortableLinux-x64-master.json
@@ -4,7 +4,27 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Run docker",
+      "displayName": "Docker cleanup",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "10f1f9a1-74b0-47ab-87bf-e3c9c68e8b0d",
+        "versionSpec": "0.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "type": "InlineScript",
+        "scriptPath": "",
+        "args": "",
+        "script": "echo \"Running: docker ps -a -q --filter name=$(DockerContainerName) | xargs --no-run-if-empty docker rm -f\"\ndocker ps -a -q --filter name=$(DockerContainerName) | xargs --no-run-if-empty docker rm -f",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Run build.sh in Docker container",
       "timeoutInMinutes": 0,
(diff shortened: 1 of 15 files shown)


gkhanna79:
CC @ramarag @schellap

wtgodbe:
LGTM - are there plans to refactor this later on so that we can reuse definitions?

chcosta:
I agree with Will's comment.  One of the primary advantages of PipeBuild, and a primary goal in much of the PipeBuild work, is to minimize duplicated code which has shown to be a nightmare to maintain.  Any time the definitions can be refactored in a sensible way such that they can be re-used, is going to provide you a huge benefit.

naamunds:
Yes. In a separate PR, I plan to refactor the Linux build definitions into a smaller number of build definitions. After this PR is in, I'll create a user story to track that work, for both Core-Setup and CLI.

buildpipeline/Core-Setup-PortableLinux-x64-master.json
```diff
@@ -0,0 +1,154 @@
+{
+  "build": [
+    {
```
wtgodbe: You should add a "Docker rm $(DockerContainerName)" step at the beginning of this build (as well as parameterizing the name of the container, dotnetcore-setup-build-container-rhel). We just ran into an issue where a RHEL build was killed before it could clean up its docker container, then the next RHEL build on that machine failed because it tried to create a container with the same name as the pre-existing container from the last job. See here - https://mseng.visualstudio.com/DefaultCollection/dotnetcore/_build/definitionEditor?definitionId=3545&_a=simple-process

wtgodbe: Make sure to add that step in the other RHEL job below, as well

naamunds: @wtgodbe I've added the Docker cleanup step.  Using `docker rm $(DockerContainerName)` isn't sufficient because it would fail if the container doesn't exist.  I was able to work around this by instead doing `docker ps -a -q --filter name=$(DockerContainerName) | xargs --no-run-if-empty docker rm -f`.  I had to use an inline script in the Shell++ build task for this, because the basic Run Command build task doesn't appear to support piping the `docker` command into `xargs`.

naamunds: I'll also log an issue for adopting the Docker cleanup script from buildtools, which can be done later.

naamunds: I've discovered a way to run the command in a basic Run Command build task, by using `bash -c`, so I've updated the PR to use that approach.

PR #1157, opened 2017-01-14
Post dev/eng fixes

ericstj:
Fix a few issues that have cropped up after merge to master.

1. Warning on restore of buildtools, add a global.json to prevent nuget from seeing projects under src.
2. Ensure prerelease suffix is beta.
3. Add linux-x64 to MS.NETCore.App lineup.  We still need a build out of corefx that has this as well.


ericstj:
/cc @weshaggard @chcosta @dagood @naamunds 

This won't fix the portable-linux leg yet, we need bits from corefx to do that.

PR #1180, opened 2017-01-17
Uppdate NETStandard.Library back to 2.0.0-*

ericstj:
/cc @weshaggard 

@dagood do we need to set up maestro to have this ingest the standard repo?

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/project.json.template
+++ b/pkg/projects/Microsoft.NETCore.App/project.json.template
@@ -13,7 +13,7 @@
     "Microsoft.DiaSymReader.Native": "1.4.0",
     "Libuv": "1.10.0-preview1-22036",
     "Microsoft.NETCore.Platforms": "2.0.0-beta-24917-02",
-    "NETStandard.Library": "1.6.2-beta-24913-02"
+    "NETStandard.Library": "2.0.0-beta-24914-01"
   },
   "frameworks": {
     "netcoreapp2.0": {}


ericstj:
This was regressed by https://github.com/ericstj/core-setup/commit/03ed039b36ca90e3234962b4d65cb6d480d28ee8#diff-63f91fad156c6157fe8ad7917bfe53c9 which downgraded the dependency.

weshaggard:
>@dagood do we need to set up maestro to have this ingest the standard repo?

Yes @bryanar is working on the official builds for the standard repo once that is ready and publishing then we should add the auto-update to maestro.

ericstj:
@weshaggard it's already publishing.

dagood:
The dotnet/standard builds need to publish info to https://github.com/dotnet/versions/tree/master/build-info/dotnet for auto-update to see them.

dagood:
@ericstj Hmm, are you expecting the version to be updated to the latest available from two builds, choosing between CoreFX's `1.6.2-beta-24913-02` and dotnet/standard's `2.0.0-beta-24914-01`? (Once they are both available on dotnet/versions.) Currently the assumption is that this would be a conflict, and I believe behavior is undefined. The dependency build info in dependencies.props can be configured using [`DisabledPackages`](https://github.com/dotnet/buildtools/blob/2c47cd22272e39ff42636b626f2079b59c761774/src/Microsoft.DotNet.Build.Tasks/VersionTools/BaseDependenciesTask.cs#L109) to ignore the package from CoreFX and remove the conflict, though.

ericstj:
> are you expecting the version to be updated to the latest available from two builds, 

No, the package is no longer built in CoreFx.  It is only built in standard repo.

dagood:
Cool, then this dependency should no longer regress once there's a new successful CoreFX build out, and when dotnet/standard publishes to dotnet/versions we can hook up CoreFX to auto-upgrade based on it.

ericstj:
> there's a new successful CoreFX build out

That's already happened. Are you sure there isn't an issue with maestro?  I see a mix of versions in https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt.

ericstj:
Merging this to unbreak conflict resolution.

dagood:
Ah, you're right, just 3 hours ago. I forgot that Latest_Packages has been changed to be a "put-only" file (for servicing reasons) and Last_Build_Packages is the exact results of the last build. I think we should do a one-time trim for the Latest_Packages file so that old packages aren't listed there anymore. I'll make a PR for that.

PR #1223, opened 2017-01-20
Switch dependency of NewtonSoft.Json to 6.0.4 for .NET Framework build of DependencyModel

dsplaisted:
Pre-emptively submitting this PR, we will discuss whether to move forward with it.

This switches the .NET Framework build of Microsoft.Extensions.DependencyModel to depend on version 6.0.4 of NewtonSoft.Json, instead of version 9.0.1.  This should help with assembly load conflicts we are having between the .NET SDK tasks and NuGet APIs.

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
+++ b/src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
@@ -727,4 +727,21 @@ namespace Microsoft.Extensions.DependencyModel
             public string HashPath;
         }
     }
+    
+#if NET451
+    static class JsonExtensions
+    {
+        public static bool? ReadAsBoolean(this JsonReader r)
+        {
+            string s = r.ReadAsString();
+            if (string.IsNullOrEmpty(s))
+            {
+                return null;
+            }
+
+            return bool.Parse(s);
+            
+        }
+    }
+#endif
 }
--- a/src/Microsoft.Extensions.DependencyModel/project.json
+++ b/src/Microsoft.Extensions.DependencyModel/project.json
@@ -8,17 +8,21 @@
   "dependencies": {
     "Microsoft.DotNet.PlatformAbstractions": {
       "target": "project"
-    },
-    "Newtonsoft.Json": "9.0.1"
+    }
   },
   "frameworks": {
-    "net451": {},
+    "net451": {
+      "dependencies": {
+        "Newtonsoft.Json": "6.0.4"        
+      }
+    },
     "netstandard1.3": {
       "imports": "portable-net45+wp80+win8+wpa81+dnxcore50",
       "dependencies": {
         "System.Diagnostics.Debug": "4.0.11",
         "System.Dynamic.Runtime": "4.0.11",
-        "System.Linq": "4.1.0"
+        "System.Linq": "4.1.0",
+        "Newtonsoft.Json": "9.0.1"
       }
     },
     "netstandard1.6": {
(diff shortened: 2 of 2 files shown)


eerhardt:
Why not load both Newtonsoft `6.0.4` and `9.0.1`?  NuGet get reference `6.0.4` (not sure why they need that old of a version, but whatever), and the SDK and DependencyModel can reference `9.0.1`.

I'm not sure this is a great solution because you are just pushing the issue down the line.  Won't ASP.NET projects end up with the same problem?  ASP.NET runtime assemblies reference the DependencyModel, even on `net4xx`.

/cc @davidfowl @pranavkm @pakrym

dsplaisted:
Here's my understanding of what we're encountering.  Thanks to @nguerrera, @rrelyea, and @srivatsn for helping investigate.

It's should generally be OK for different tasks in an MSBuild process to reference different versions of a shared dependency (which is strong named).  Each task will load the version of the dependency that's bundled together with it.  Since the different tasks don't share types, it should be OK to load different versions of the dependency side-by-side.

However, within each task, everything needs to depend on the same version of a given dependency.  In the case of the .NET SDK tasks, the tasks depend on Microsoft.Extensions.DependencyModel, as well as [NuGet.ProjectModel](https://www.nuget.org/packages/NuGet.ProjectModel).  Both of these in turn depend on NewtonSoft.Json.  For our tasks to work, they need to depend on the same version of NewtonSoft.Json.  This is true even if our tasks don't exchange types from NewtonSoft.Json with these dependencies or directly consume any NewtonSoft.Json APIs at all, because we can only bundle one version of the NewtonSoft.Json DLL together with the tasks.  If it doesn't have the exact same AssemblyVersion that our tasks and all their dependencies were compiled against, then there will be binding exception when .NET attempts to load a reference where the version doesn't match.  Normally in .NET the way to resolve this is to unify to a single version with binding redirects, but MSBuild doesn't support supplying binding redirects for tasks.

Currently, the .NET SDK tasks are compiled for .NET Standard 1.3, and the tasks DLL as well as its dependencies are included in the tools\netcoreapp1.0 and tools\net46 folders in the SDK.  For .NET Standard, both Microsoft.Extensions.DependencyModel, as well as NuGet.ProjectModel, depend on version 9.0.1 of NewtonSoft.Json.  So everything there should work.

However, the NuGet DLLs are multi-targeted to .NET Standard and .NET Framework.  The .NET Framework versions of the NuGet DLLs depend on version 6.0.4 of NewtonSoft.Json.  The NuGet restore tasks that ship with Visual Studio use the .NET Framework versions of the NuGet DLLs.  The .NET Framework and .NET Standard versions of these NuGet DLLs have the exact same assembly version, and the .NET Standard versions are bundled up with the .NET SDK tasks, while the .NET Framework versions are bundled up with the NuGet restore tasks.  The version that gets loaded depends on whether the .NET SDK or NuGet tasks are loaded first.  Subsequently, references to the NuGet DLLs will resolve to the already loaded version, since the assembly version is exactly the same.

We've recently added a task to the .NET SDK that runs before NuGet restore.  This is now causing the .NET Standard versions of the subset of the NuGet DLLs that we depend on to be loaded.  When the NuGet restore targets run, they load the .NET Framework versions of the rest of the NuGet DLLs.  The .NET Framework and .NET Standard versions of the NuGet DLLs depend on different versions of NewtonSoft.Json, so when they try to exchange Json types, they get a method not found exception (since the return type of the method doesn't match what the caller is looking for).

To solve this, I think that we need to package up the .NET Framework version of the NuGet DLLs with the .NET SDK tasks instead of packaging the .NET Standard versions.  This should mean that the same version of the DLLs are loaded regardless of which task loads first.  However our tasks and all their dependencies need to agree on the version of NewtonSoft.Json to use, and the NuGet DLLs depend on version 6.0.4 when targeting .NET Framework.  Hence, this PR updates Microsoft.Extensions.DependencyModel to do the same thing.

eerhardt:
Closing as we don't plan on taking this change.

PR #1241, opened 2017-01-21
Multi-level SharedFX lookup implemented.

cakine:
Framework search is done in relation to the executable directory. It would be better to be able to search for it in other directories as well. Suggested folders are the current working directory, the user location and the global .NET location. The user and global folders may vary depending on the running operational system. They are defined as follows:

User location:
•	Windows 32-bit: %SystemDrive%\Users\username\.dotnet\x86
•	Windows 64-bit: %SystemDrive%\Users\username\.dotnet\x64
•	Unix 32-bit: /home/username/.dotnet/x86
•	Unix 64-bit: /home/username/.dotnet/x64

Global .NET location:
•	Windows 32-bit: %SystemDrive%\Program Files\dotnet
•	Windows 64-bit (32-bit application): %SystemDrive%\Program Files (x86)\dotnet
•	Windows 64-bit (64-bit application): %SystemDrive%\Program Files\dotnet
•	Unix: the directory of “dotnet” defined in the system path.

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -310,6 +310,39 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
     return false;
 }
 
+void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir, const pal::string_t& fx_ver, const fx_ver_t& specified)
+{
+    trace::verbose(_X("Attempting FX roll forward starting from [%s]"), fx_ver.c_str());
+
+    std::vector<pal::string_t> list;
+    pal::readdir(fx_dir, &list);
+    fx_ver_t most_compatible = specified;
+    for (const auto& version : list)
+    {
+        trace::verbose(_X("Inspecting version... [%s]"), version.c_str());
+        fx_ver_t ver(-1, -1, -1);
+        if (!specified.is_prerelease() && fx_ver_t::parse(version, &ver, true) && // true -- only prod. prevents roll forward to prerelease.
+            ver.get_major() == specified.get_major() &&
+            ver.get_minor() == specified.get_minor())
+        {
+            // Pick the greatest production that differs only in patch.
+            most_compatible = std::max(ver, most_compatible);
+        }
+        if (specified.is_prerelease() && fx_ver_t::parse(version, &ver, false) && // false -- implies both production and prerelease.
+            ver.is_prerelease() && // prevent roll forward to production.
+            ver.get_major() == specified.get_major() &&
+            ver.get_minor() == specified.get_minor() &&
+            ver.get_patch() == specified.get_patch() &&
+            ver > specified)
+        {
+            // Pick the smallest prerelease that is greater than specified.
+            most_compatible = (most_compatible == specified) ? ver : std::min(ver, most_compatible);
+        }
+    }
+    pal::string_t most_compatible_str = most_compatible.as_str();
+    append_path(&fx_dir, most_compatible_str.c_str());
+}
+
 pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t& own_dir, const runtime_config_t& config, const pal::string_t& specified_fx_version)
 {
     // No FX resolution for standalone apps.
@@ -322,7 +355,7 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t& 
     }
     assert(mode == host_mode_t::muxer);
 
-    trace::verbose(_X("--- Resolving FX directory from muxer dir '%s', specified '%s'"), own_dir.c_str(), specified_fx_version.c_str());
+    trace::verbose(_X("--- Resolving FX directory, specified '%s'"), specified_fx_version.c_str());
     const auto fx_name = config.get_fx_name();
     const auto fx_ver = specified_fx_version.empty() ? config.get_fx_version() : specified_fx_version;
 
@@ -333,67 +366,90 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t& 
         return pal::string_t();
     }
 
-    auto fx_dir = own_dir;
-    append_path(&fx_dir, _X("shared"));
-    append_path(&fx_dir, fx_name.c_str());
+    // Multi-level SharedFX lookup will look for the most appropriate version in several locations
+    // by following the priority rank below (from 1 to 4):
+    // 1. Current working directory
+    // 2. User directory
+    // 3. .exe directory
+    // 4. Global .NET directory
+    // If it is not activated, then only .exe directory will be considered
 
-    bool do_roll_forward = false;
-    if (specified_fx_version.empty())
+    std::vector<pal::string_t> hive_dir;
+    pal::string_t env_lookup;
+
+    // Multi-level SharedFX lookup can be disabled by setting DOTNET_MULTILEVEL_LOOKUP env var to zero
+    bool multilevel_lookup = true;
+    if (pal::getenv(_X("DOTNET_MULTILEVEL_LOOKUP"), &env_lookup))
     {
-        if (!specified.is_prerelease())
+        auto env_val = pal::xtoi(env_lookup.c_str());
+        multilevel_lookup = (env_val != 0);
+    }
+
+    pal::string_t cwd;
+    pal::string_t local_dir;
+    pal::string_t global_dir;
+
+    if (multilevel_lookup)
+    {
+        if (pal::getcwd(&cwd))
         {
-            // If production and no roll forward use given version.
-            do_roll_forward = config.get_patch_roll_fwd();
+            hive_dir.push_back(cwd);
+        }
+        if (pal::get_local_dotnet_dir(&local_dir))
+        {
+            hive_dir.push_back(local_dir);
+        }
+    }
+    hive_dir.push_back(own_dir);
+    if (multilevel_lookup && pal::get_global_dotnet_dir(&global_dir))
+    {
+        hive_dir.push_back(global_dir);
+    }
+
+    for (pal::string_t dir : hive_dir)
+    {
+        auto fx_dir = dir;
+        trace::verbose(_X("Searching FX directory in [%s]"), fx_dir.c_str());
+
+        append_path(&fx_dir, _X("shared"));
+        append_path(&fx_dir, fx_name.c_str());
+
+        bool do_roll_forward = false;
+        if (specified_fx_version.empty())
+        {
+            if (!specified.is_prerelease())
+            {
+                // If production and no roll forward use given version.
+                do_roll_forward = config.get_patch_roll_fwd();
+            }
+            else
+            {
+                // Prerelease, but roll forward only if version doesn't exist.
+                pal::string_t ver_dir = fx_dir;
+                append_path(&ver_dir, fx_ver.c_str());
+                do_roll_forward = !pal::directory_exists(ver_dir);
+            }
+        }
+
+        if (!do_roll_forward)
+        {
+            trace::verbose(_X("Did not roll forward because specified version='%s', patch_roll_fwd=%d, chose [%s]"), specified_fx_version.c_str(), config.get_patch_roll_fwd(), fx_ver.c_str());
+            append_path(&fx_dir, fx_ver.c_str());
         }
         else
         {
-            // Prerelease, but roll forward only if version doesn't exist.
-            pal::string_t ver_dir = fx_dir;
-            append_path(&ver_dir, fx_ver.c_str());
-            do_roll_forward = !pal::directory_exists(ver_dir);
+            resolve_roll_forward(fx_dir, fx_ver, specified);
         }
-    }
 
-    if (!do_roll_forward)
-    {
-        trace::verbose(_X("Did not roll forward because specified version='%s', patch_roll_fwd=%d, chose [%s]"), specified_fx_version.c_str(), config.get_patch_roll_fwd(), fx_ver.c_str());
-        append_path(&fx_dir, fx_ver.c_str());
-    }
-    else
-    {
-        trace::verbose(_X("Attempting FX roll forward starting from [%s]"), fx_ver.c_str());
-
-        std::vector<pal::string_t> list;
-        pal::readdir(fx_dir, &list);
-        fx_ver_t most_compatible = specified;
-        for (const auto& version : list)
+        if (pal::directory_exists(fx_dir))
         {
-            trace::verbose(_X("Inspecting version... [%s]"), version.c_str());
-            fx_ver_t ver(-1, -1, -1);
-            if (!specified.is_prerelease() && fx_ver_t::parse(version, &ver, true) && // true -- only prod. prevents roll forward to prerelease.
-                ver.get_major() == specified.get_major() &&
-                ver.get_minor() == specified.get_minor())
-            {
-                // Pick the greatest production that differs only in patch.
-                most_compatible = std::max(ver, most_compatible);
-            }
-            if (specified.is_prerelease() && fx_ver_t::parse(version, &ver, false) && // false -- implies both production and prerelease.
-                ver.is_prerelease() && // prevent roll forward to production.
-                ver.get_major() == specified.get_major() &&
-                ver.get_minor() == specified.get_minor() &&
-                ver.get_patch() == specified.get_patch() &&
-                ver > specified)
-            {
-                // Pick the smallest prerelease that is greater than specified.
-                most_compatible = (most_compatible == specified) ? ver : std::min(ver, most_compatible);
-            }
+            trace::verbose(_X("Chose FX version [%s]"), fx_dir.c_str());
+            return fx_dir;
         }
-        pal::string_t most_compatible_str = most_compatible.as_str();
-        append_path(&fx_dir, most_compatible_str.c_str());
     }
 
-    trace::verbose(_X("Chose FX version [%s]"), fx_dir.c_str());
-    return fx_dir;
+    trace::error(_X("It was not possible to find any compatible framework version"));
+    return pal::string_t();
 }
 
 pal::string_t fx_muxer_t::resolve_cli_version(const pal::string_t& global_json)
--- a/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
+++ b/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
@@ -1,0 +1,421 @@
+﻿using System;
+using System.IO;
+using Xunit;
+using Microsoft.DotNet.InternalAbstractions;
+
+namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.MultilevelSharedFxLookup
+{
+    public class GivenThatICareAboutMultilevelSharedFxLookup
+    {
+        private RepoDirectoriesProvider RepoDirectories;
+        private TestProjectFixture PreviouslyBuiltAndRestoredPortableTestProjectFixture;
+
+        private string _currentWorkingDir;
+        private string _userDir;
+        private string _executableDir;
+        private string _cwdSharedFxBaseDir;
+        private string _userSharedFxBaseDir;
+        private string _exeSharedFxBaseDir;
+        private string _builtSharedFxDir;
+        private string _cwdSelectedMessage;
+        private string _userSelectedMessage;
+        private string _exeSelectedMessage;
+        private string _sharedFxVersion;
+
+        public GivenThatICareAboutMultilevelSharedFxLookup()
+        {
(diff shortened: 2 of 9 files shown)


cakine:
@schellap

schellap:
Cc @gkhanna79 @ramarag

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,79 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
-        append_path(&fx_dir, most_compatible_str.c_str());
-    }
-
-    trace::verbose(_X("Chose FX version [%s]"), fx_dir.c_str());
-    return fx_dir;
+	pal::string_t env_lookup;
+	std::vector<pal::string_t> hive_dir;
+
+	//Multi-level SharedFX lookup happens only if dotnet_lookup env var is set to 1
+	if (pal::getenv(_X("dotnet_lookup"), &env_lookup) && env_lookup == _X("1"))
```
schellap: On Unix env variable names are case sensitive. Use `DOTNET_OREDERED_LOOKUP`

schellap:
Pls. fix spacing and the PR feedback otherwise, this looks fine to me!

schellap:
@ramarag this has changes ported from PR https://github.com/dotnet/core-setup/pull/1109, pls. decide on the merge order between the two.

src/corehost/cli/fxr/fx_muxer.cpp
cakine: Fixed it.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,79 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
+	//Multi-level SharedFX lookup happens only if dotnet_lookup env var is set to 1
+	if (pal::getenv(_X("DOTNET_ORDERED_LOOKUP"), &env_lookup) && env_lookup == _X("1"))
+	{
+		pal::string_t cwd;
+		pal::string_t local_dir;
+		pal::string_t global_dir;
+
+		if (pal::getcwd(&cwd))
+		{
+			hive_dir.push_back(cwd);
```
gkhanna79: NIT: Extra spaces/tabs in this change.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,79 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
-        append_path(&fx_dir, most_compatible_str.c_str());
-    }
-
-    trace::verbose(_X("Chose FX version [%s]"), fx_dir.c_str());
-    return fx_dir;
+	pal::string_t env_lookup;
+	std::vector<pal::string_t> hive_dir;
+
+	//Multi-level SharedFX lookup happens only if dotnet_lookup env var is set to 1
+	if (pal::getenv(_X("DOTNET_ORDERED_LOOKUP"), &env_lookup) && env_lookup == _X("1"))
```
gkhanna79: Did you try to build the repo locally with this set to confirm that existing tests pass?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,79 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
+	for (pal::string_t dir : hive_dir)
+	{
+		auto fx_dir = dir;
+		trace::verbose(_X("Searching FX directory in [%s]"), fx_dir.c_str());
+
+		append_path(&fx_dir, _X("shared"));
+		append_path(&fx_dir, fx_name.c_str());
+
+		bool do_roll_forward = false;
+		if (specified_fx_version.empty())
```
gkhanna79: Shouldn't you use `specified` here as it already addressed the case of version to use when `specified_fx_version `is empty?

gkhanna79: Actually, why do we need this "if" check?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,79 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
+				// Prerelease, but roll forward only if version doesn't exist.
+				pal::string_t ver_dir = fx_dir;
+				append_path(&ver_dir, fx_ver.c_str());
+				do_roll_forward = !pal::directory_exists(ver_dir);
+			}
+		}
+
+		if (!do_roll_forward)
+		{
+			trace::verbose(_X("Did not roll forward because specified version='%s', patch_roll_fwd=%d, chose [%s]"), specified_fx_version.c_str(), config.get_patch_roll_fwd(), fx_ver.c_str());
```
gkhanna79: If `specified_fx_version` is not specified, what would be the message here?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,79 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
+			append_path(&fx_dir, fx_ver.c_str());
```
gkhanna79: Shouldnt this be `specified`?

src/corehost/common/pal.h
```diff
@@ -200,6 +200,8 @@ namespace pal
     bool get_own_executable_path(string_t* recv);
     bool getenv(const char_t* name, string_t* recv);
     bool get_default_servicing_directory(string_t* recv);
+	bool get_local_dotnet_dir(string_t* recv);
```
gkhanna79: Indentation issues.

src/corehost/common/pal.unix.cpp
```diff
@@ -170,6 +170,83 @@ bool pal::get_default_servicing_directory(string_t* recv)
     return true;
 }
 
+bool pal::get_local_dotnet_dir(pal::string_t* recv)
+{
+	recv->clear();
+	pal::string_t dir;
+	if (!pal::getenv("HOME", &dir))
+	{
+		struct passwd* pw = getpwuid(getuid());
```
gkhanna79: Spacing issues.

src/corehost/common/pal.unix.cpp
```diff
@@ -170,6 +170,83 @@ bool pal::get_default_servicing_directory(string_t* recv)
...
+	return ((st.st_mode & S_IEXEC) != 0);
+}
+
+static
+bool locate_dotnet_on_path(pal::string_t* dotnet_exe)
+{
+	pal::string_t path;
+	if (!pal::getenv(_X("PATH"), &path))
+	{
+		return false;
```
gkhanna79: Spacing and indentation issues

gkhanna79:
Also, can you please add tests for the new scenarios to your PR?

src/corehost/cli/fxr/fx_muxer.cpp
cakine: Actually, I did not write this piece of code. I simply copied the default behavior while looking in different locations.

I believe that this "if" forces the code to NOT roll forward when the version is defined through argument (when specified_fx_version is not empty), since we want the app to run with the exact specified version. It will ignore the version defined in the json file.

gkhanna79: I am not sure I follow. Are you suggesting you copied it from somewhere? If so, where? Either ways, since the change is part of your PR, it is considered your code and it is imperative to ensure it is correct.

Looking at the implementation of this function, are you sure you are using the right variables to determine the data?

cakine: No. I'm saying that the roll forward code was already implemented like this. I simply put it in a method and called it.

cakine: Before my changes, the default behavior was to not roll forward when the version was defined through argument. Since I did not change the roll forward code (simply moved it), this behavior did not change. I think we're supposed to do that.

src/corehost/cli/fxr/fx_muxer.cpp
gkhanna79: Can you please rename this variable to "DOTNET_MULTILEVEL_LOOKUP"?

src/corehost/cli/fxr/fx_muxer.cpp
cakine: Fixed.

src/corehost/common/pal.h
cakine: Fixed.

src/corehost/common/pal.unix.cpp
cakine: Fixed.

src/corehost/common/pal.unix.cpp
cakine: Fixed.

src/corehost/cli/fxr/fx_muxer.cpp

cakine: Yes. The build succeeded.

gkhanna79:
@ramarag PTAL

@cakine Can you please add scenario tests as well?

src/corehost/cli/fxr/fx_muxer.cpp
cakine: Actually, this code was already implemented as well (simply moved it).
I agree that it's a bit weird.

Suppose we're looking for 1.0.0 and we do not perform a roll forward.

If we specified it through argument, the message would be:
"Did not roll forward because specified version='1.0.0', patch_roll_fwd=?, chose [1.0.0]"
So we did not roll forward because we specified 1.0.0 version through argument (patch_roll_fwd does not matter).

If we did not specify it, then it will look in config if it is supposed to roll forward (let's assume it will not do it). Then the message would be:
"Did not roll forward because specified version='', patch_roll_fwd=0, chose [1.0.0]"
So we did not roll forward because patch_roll_fwd is set to 0.

I believe that's how it is supposed to be. Do you think we should change it?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,77 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
     }
 
-    auto fx_dir = own_dir;
-    append_path(&fx_dir, _X("shared"));
-    append_path(&fx_dir, fx_name.c_str());
+    pal::string_t env_lookup;
+    std::vector<pal::string_t> hive_dir;
+
+    //Multi-level SharedFX lookup happens only if dotnet_lookup env var is set to 1
+    bool multilevel_lookup = pal::getenv(_X("DOTNET_MULTILEVEL_LOOKUP"), &env_lookup) && env_lookup == _X("1");
```
gkhanna79: Can please wrap env_lookup ==.... in braces for better readability and scoping?

src/corehost/cli/fxr/fx_muxer.cpp
gkhanna79: What if specified_fx_Version was empty?

src/corehost/cli/fxr/fx_muxer.cpp

src/corehost/cli/fxr/fx_muxer.cpp
cakine: It's the second situation.
"Did not roll forward because specified version='', patch_roll_fwd=0, chose [1.0.0]"

gkhanna79:
Change looks good to me. Will you be able to get the tests incorporated as well?

cakine:
I'm just finishing the md file right now. I still need to take a look at how the HostActivation Tests are done, though.

gkhanna79:
@cakine How are the tests coming along? When do you expect we can merge this PR in with the tests?

src/corehost/common/pal.unix.cpp
```diff
@@ -170,6 +170,83 @@ bool pal::get_default_servicing_directory(string_t* recv)
...
+        {
+            *dotnet_exe = tok;
+            return true;
+        }
+        tok.clear();
+    }
+    return false;
+}
+
+bool pal::get_global_dotnet_dir(pal::string_t* recv)
```
ramarag: Try to move this next to pal::get_local_dotnet_dir,
i am also using the same functions, will be easier to correct merge conflicts

cakine:
@gkhanna79 I'm not sure because I still did not figure out a way to decide from which folder the framework was picked during activation process. When this issue is solved, I believe the tests will be easy to code.

gkhanna79:
@cakine You shouldnt be pushing updates from Maestro. Did you rebase against master correctly?

cakine:
@gkhanna79 Sorry about that. It's fixed now.

cakine:
@gkhanna79 I added a few tests for cwd and exe location. Just wanted you to take a look at what I'm doing. I still need to add tests for the user location, and test roll forwards as well.

src/corehost/common/pal.unix.cpp
cakine: Thank you. I believe it's fine now.

cakine:
@gkhanna79 I'm not sure why some checks failed, but I'll take a look at this matter as well.

test/HostActivationTests/GivenThatICareAboutSharedFxMultilevelLookup.cs
```diff
@@ -0,0 +1,194 @@
...
+            if (Directory.Exists(multilevelDir))
+            {
+                Directory.Delete(multilevelDir, true);
+            }
+
+            // Only the shared folder is necessary in cwd
+            Directory.CreateDirectory(_cwdSharedFxBaseDir);
+
+            // It's necessary to copy everything to build the fixture from exe dir
+            CopyDirectory(builtDotnet, _executableDir);
```
gkhanna79: dont you need to create _executableDir?

gkhanna79: Also, does CopyDirectory perform a recursive copy or only the copy of the contents of the specified folder?

gkhanna79: Also, can you please add some comments?

cakine: CopyDirectory will create _executableDir if it doesn't exist.

cakine: It's a recursive copy.

cakine: I'll add some comments then.

test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+            foreach (string version in availableVersions)
+            {
+                string sharedFxDir = Path.Combine(sharedFxBaseDir, version);
+                if (!Directory.Exists(sharedFxDir))
+                {
+                    throw new DirectoryNotFoundException();
+                }
+                Directory.Delete(sharedFxDir, true);
```
cakine: @gkhanna79 The exception is thrown from Directory.Delete during the first test.

test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -0,0 +1,429 @@
...
+            // SharedFxBaseDirs contain all available version folders
+            _cwdSharedFxBaseDir = Path.Combine(_currentWorkingDir, "shared", "Microsoft.NETCore.App");
+            _userSharedFxBaseDir = Path.Combine(_userDir, ".dotnet", RuntimeEnvironment.RuntimeArchitecture, "shared", "Microsoft.NETCore.App");
+            _exeSharedFxBaseDir = Path.Combine(_executableDir, "shared", "Microsoft.NETCore.App");
+
+            // Completely remove remaining files from other tests. Actually, if the tests were successfully
+            // completed, the folder is supposed to be empty. We're just playing safe
+            if (Directory.Exists(multilevelDir))
+            {
+                Directory.Delete(multilevelDir, true);
```
cakine: @gkhanna79 The other tests fail here, since this is the first time Directory.Delete is called for them (multilevelDir didn't exist during the first test).

cakine:
@mmitche @ellismg I'm facing an issue with the Fedora23 x64 build. I was not able to reproduce the bug locally (the build succeeds in my Fedora VM). Is it possible for me to access the VM where the checks are being run?

cakine:
@gkhanna79

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,67 +366,85 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
+    // 2. User directory
+    // 3. .exe directory
+    // 4. Global .NET directory
+    // If it is not activated, then only .exe directory will be considered
+
+    std::vector<pal::string_t> hive_dir;
+    pal::string_t env_lookup;
+
+    // Multi-level SharedFX lookup happens only if DOTNET_MULTILEVEL_LOOKUP env var is set to 1
+    bool multilevel_lookup = pal::getenv(_X("DOTNET_MULTILEVEL_LOOKUP"), &env_lookup) && (env_lookup == _X("1"));
```
gkhanna79: As discussed the other day, please change this to be enabled by default and use the env var to be used for opt-out to ensure we are testing the repo with the feature enabled.

gkhanna79: CC @ramarag

gkhanna79:
I am good with this change. @ramarag Do you have anything to add?

gkhanna79:
@cakine I see your newly added test failing in the log (see https://ci.dot.net/job/dotnet_core-setup/job/master/job/release_windows_nt_x64_prtest/636/consoleFull):

`11:12:18 [EXEC       <] [FAIL] [00:00:17.928] D:\j\workspace\release_windo---0c55dee9\artifacts\win81-x64\tests\artifacts\dotnetMultilevelSharedFxLookup\2\cwd> D:\j\workspace\release_windo---0c55dee9\artifacts\win81-x64\tests\artifacts\dotnetMultilevelSharedFxLookup\2\exe\dotnet.exe "--fx-version" "9999.0.0-dummy1" "D:\j\workspace\release_windo---0c55dee9\artifacts\win81-x64\tests\artifacts\5\SharedFxLookupPortableApp\bin\SharedFxLookupPortableApp.dll" exited with -2147450749`

Can you please take a look?

test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -0,0 +1,421 @@
...
+            dotnet.Exec("--fx-version", "9999.0.0-dummy1", appDll)
+                .WorkingDirectory(_currentWorkingDir)
+                .EnvironmentVariable("COREHOST_TRACE", "1")
+                .CaptureStdOut()
+                .CaptureStdErr()
+                .Execute()
+                .Should()
+                .Fail()
+                .And
+                .HaveStdErrContaining("It was not possible to find any compatible framework version");
```
cakine: @gkhanna79 Here is the failing test. It's expected to fail, so there's no problem.

gkhanna79: Thanks @cakine. Is it possible to exclude the failing output of the negative test and simply have something like "Test passed or failed" based upon the expected behaviour of the test? 

This will ensure we can keep the output clean in the log indicative of passing tests only.

cakine: I'm not sure. That's not something related to my tests. Other tests (like PortableAppActivation tests) have the same behavior. If the execution fails, then it will print the failing output (I think it happens even if there is no assertion being checked).

I believe it's purposely done like that.

cakine: In the log you can see fail outputs from other tests as well.

cakine: I think that's not an issue. The output states that the execution failed, not the test. When the test fails, an error message and a stack trace are printed.

gkhanna79:
I am planning to merge https://github.com/dotnet/core-setup/pull/1560. Once it is merged, you should rebase against it and update your negative test to invoke the Execute method as exemplified in that PR. This will keep real test passes segregated from real failures.

gkhanna79:
@cakine I have merged the PR - please rebase against it and update your test.

cakine:
@gkhanna79 Done.

PR #1281, opened 2017-01-24
Design Doc on dotnet cache

ramarag:
@gkhanna79 i am moving https://github.com/dotnet/core-proposals/blob/master/DotNetCore-SharedPackageStore.md to core-setup repo

The only difference between them is that json input format has been replaced by xml

### DIFF
--- a/Documentation/design-docs/DotNetCore-SharedPackageStore.md
+++ b/Documentation/design-docs/DotNetCore-SharedPackageStore.md
@@ -1,0 +1,216 @@
+# .NET Core - Shared Package Store
+
+## Introduction
+To enable sharing of assemblies among all machine-wide .NET Core applications, a centralized shared assembly store is needed. It enables applications to be trimmed of the shared assemblies and frees the apps from carrying them. This document will describe such an assembly store and its lookup that applications can take advantage of during development and their deployment. It also describes commands that will augment `dotnet` to compose shared package store entries and publishing mechanisms through which apps can filter assemblies to reduce disk usage.
+
+### Packages Store
+
+The package store can be either a global system-wide folder or a user local folder:
+
++ **Global**:
+    - The `dotnet` root location -- on Windows, the folder is located in `C:\Program Files (x86)\`. See layout below.
+    - In the `packages/install` folder we expect packages to be installed ONLY through platform installers like MSI, pkg, deb, apt-get etc.
+    - The package layout composed with the `dotnet cache` command (details follow) are expected to be unzipped directly into the `packages` folder. Note, that this unzip step is a manual action.
+
+```
+    - dotnet.exe
+    - shared
+        - netcoreapp2.0
+            + 2.0.0-preview2-00001
+    - packages
+        - install
+            + refs
+            + netcoreapp2.0
+            + netcoreapp2.1
+        + refs
+        + netcoreapp2.0
+        + netcoreapp2.1  
+```
+
+The layout within `netcoreapp*` folders is a NuGet cache layout.
+
++ **User**: The user local location can be used to override the global location when probing assemblies and uses a very similar layout, except that the `packages/install` folder is NOT supported in the user profile. It is only supported in the global location.
+
+```
+    - %USERPROFILE% or $HOME/
+        - .dotnet
+          - {arch=x86|x64}
+            - packages 
+                = do =
+```
+
+**OPEN**: Should we have a `dotnet` relative location as well, if operating out of a zip installation of .NET Core binaries?
+
+### Composing a runtime (non-ref) package store
+
+To compose the layout of the shared package store, we will use a dotnet command called `dotnet cache`. We expect the *hosting providers* (ex: Antares) to use the command to prime their machines and framework authors who want to provide *pre-optimized package archives* create the compressed archive layouts.
+
+The layout is composed from a list of package names and versions specified as JSON: 
+
+**Roslyn Example**
+```xml
+ <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="1.3.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic" Version="1.3.2" />
+    <PackageReference Include=""Microsoft.CodeAnalysis.VisualBasic.Features" Version="1.3.2" />
+  </ItemGroup>
+```
+
+
+and issue a command like below:
+
+```
+dotnet cache --entries packages.xml -f netcoreapp2.0 [--output C:\Foo] --rid win7-x64 --fx-version 2.0.0-preview2-00001 [--no-optimize]
+
+--framework          Specifies the TFM that the package store is applicable to
+--output       The output directory to create the package store in (default: %USERPROFILE%\.dotnet or ~/.dotnet)
+
+--no-optimize  Do not perform crossgen of the assemblies after "restore" (optimize is the default)
+
+--runtime          The runtime identifier of the target platform where these assemblies will be run
+--fx-version   The Microsoft.NETCore.App package version that will be used to run the assemblies
+
+```
+
+Hosting providers would create a `packages.xml` file corresponding to the packages that will be shared in their hosting environment and specify the file to `dotnet cache`. The file can be on the file system or from an URL. The TFM argument is used in the shared package layout described above.
+
+If `--optimize` is specified, we would precompile all the managed assets to native code in a temp folder before copying to the output folder. If crossgen is used, it would be the one acquired in the closure of the `Microsoft.NETCore.App` specified by the `--fx-version` option. Also, if no `--output` folder is specified, then the default is `~/.dotnet` or `%USERPROFILE%\.dotnet\`. The output asset files will be present in the following layout: `$HOME/.dotnet/packages/{tfm}/{package-name}/{package-version}/{asset-path}`.
+
+The output folder will be consumed by the runtime by adding to the `DOTNET_SHARED_PACKAGES_ROOTS` environment variable. See probe precedence below.
+
+# Building apps with shared packages
+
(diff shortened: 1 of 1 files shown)


Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,216 @@
...
+        + netcoreapp2.0
+        + netcoreapp2.1  
+```
+
+The layout within `netcoreapp*` folders is a NuGet cache layout.
+
++ **User**: The user local location can be used to override the global location when probing assemblies and uses a very similar layout, except that the `packages/install` folder is NOT supported in the user profile. It is only supported in the global location.
+
+```
+    - %USERPROFILE% or $HOME/
```
ramarag: @gkhanna79 
As per this doc the global location would be like
C:\Users\<user>\.dotnet\x64\packages\netcoreapp1.0\newtonsoft.json\9.0.1\lib\netstandard1.0\Newtonsoft.Json.dll

and additional probe directories will be like : 
[user specified dir]\x64\netcoreapp1.0\newtonsoft.json\9.0.1\lib\netstandard1.0\Newtonsoft.Json.dll

do we want to ensure the packages appear on both and ensure both have the same format  or just the format?
C:\Users\<user>\.dotnet\packages\x64\netcoreapp1.0\newtonsoft.json\9.0.1\lib\netstandard1.0\Newtonsoft.Json.dll
[user specified dir]\x64\netcoreapp1.0\newtonsoft.json\9.0.1\lib\netstandard1.0\Newtonsoft.Json.dll

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,216 @@
...
+## Introduction
+To enable sharing of assemblies among all machine-wide .NET Core applications, a centralized shared assembly store is needed. It enables applications to be trimmed of the shared assemblies and frees the apps from carrying them. This document will describe such an assembly store and its lookup that applications can take advantage of during development and their deployment. It also describes commands that will augment `dotnet` to compose shared package store entries and publishing mechanisms through which apps can filter assemblies to reduce disk usage.
+
+### Packages Store
+
+The package store can be either a global system-wide folder or a user local folder:
+
++ **Global**:
+    - The `dotnet` root location -- on Windows, the folder is located in `C:\Program Files (x86)\`. See layout below.
+    - In the `packages/install` folder we expect packages to be installed ONLY through platform installers like MSI, pkg, deb, apt-get etc.
```
gkhanna79: This is specifically for the folks who wish to install their custom cache.

Documentation/design-docs/DotNetCore-SharedPackageStore.md
gkhanna79: @ramarag Where does the document indicate the global location as you called out above?

The layout, as prescribed in the global location, is also applicable to user specific location modulo the `install` subfolder.

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,216 @@
...
+                = do =
+```
+
+**OPEN**: Should we have a `dotnet` relative location as well, if operating out of a zip installation of .NET Core binaries?
+
+### Composing a runtime (non-ref) package store
+
+To compose the layout of the shared package store, we will use a dotnet command called `dotnet cache`. We expect the *hosting providers* (ex: Antares) to use the command to prime their machines and framework authors who want to provide *pre-optimized package archives* create the compressed archive layouts.
+
+The layout is composed from a list of package names and versions specified as JSON:
```
dasMulli: JSON or XML? Because the following example is XML..

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,216 @@
...
+
+**Roslyn Example**
+```xml
+ <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="1.3.2" />
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic" Version="1.3.2" />
+    <PackageReference Include=""Microsoft.CodeAnalysis.VisualBasic.Features" Version="1.3.2" />
+  </ItemGroup>
+```
+
```
dasMulli: Is this really the only xml needed or does it have to be an MSBuild file? If so, does it need to reference a specific SDK?

ramarag:
skip ci please

Documentation/design-docs/DotNetCore-SharedPackageStore.md
ramarag: Yes it is a project now

Documentation/design-docs/DotNetCore-SharedPackageStore.md
ramarag: it is an xml

ramarag:
cc @blackdwarf @schellap for finalizing the spec

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,218 @@
...
+        + netcoreapp2.1  
+```
+
+The layout within `netcoreapp*` folders is a NuGet cache layout.
+
++ **User**: The user local location can be used to override the global location when probing assemblies and uses a very similar layout, except that the `packages/install` folder is NOT supported in the user profile. It is only supported in the global location.
+
+```
+    - %USERPROFILE% or $HOME/
+        - .dotnet
```
ramarag: @blackdwarf do we want to keep the '.' in '.dotnet' above. I hear that such folders on Linux become hidden.
Also should we stick to user profile ? or provide a location for machine wide common location ?

blackdwarf: @ramarag @bleroy will be able to reply to this.

bleroy: The '.' has become the convention for such tool-related folders, from ".git" to local IDE settings, so it seems appropriate here too. Such folders do indeed become hidden on Linux, but that's generally known by Linux users who know how to make them appear when needed.
User profile should be the default, but there can be an option for system-wide, that would require admin rights.
Does this help?

gkhanna79: I concur as well - keep the ".".

ramarag: ok will keep it

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,218 @@
...
+
++ `$CORE_SERVICING` on Unix or `%ProgramFiles(x86)%\coreservicing` on Windows.
++ `DOTNET_SHARED_PACKAGES_ROOTS` in the chained order
++ The user local shared package store in `%USERPROFILE%\.dotnet\packages` or `$HOME/.dotnet/`
++ The global shared package store
++ The Shared Framework directory
++ ~~`DOTNET_HOSTING_OPTIMIZATION_CACHE` deprecated in favor of `DOTNET_SHARED_PACKAGES_ROOTS`~~
++ `--additionalprobingpaths` specified in the command line
++ `runtimeOptions.additionalProbingPaths` (includes NuGet cache probe specified by the CLI for `dotnet run`)
++ Application `bin` directory
```
ramarag: Presently the Application is the last place we check for the assemblies. 
1. Should we provide a mechanism to opt out of the all the above locations ?
2. Provide a mechanism for app to win over others ?

gkhanna79:
Per the order above, app always wins. This is status quo as well, independent of this design and we need to maintain it.


Why do we need an opt out since all the above locations are optional aside from global fx and app?


Thanks,
Gaurav


Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,218 @@
...
+## Project Authoring
+We will by default treat `Microsoft.NETCore.App` as though `type: platform` is always specified, thus requiring no explicit RID specification by the user. It will be an `ERROR` to specify a RID in the csproj file using the `<RuntimeIdentifier/>` tag.
+
+## dotnet restore
+
+Because RIDs are not available until publish time, the application is treated as though it is a present day portable app and a regular restore is performed.
+
+## dotnet build
+
+`dotnet build` should treat any project as though it is a portable app and produce `runtimeconfig.json` with the `Microsoft.NETCore.App` as the framework. In addition, the `runtimeconfig.json` file should also specify the TFM field as:
```
eerhardt: >`dotnet build` should treat any project as though it is a portable app

This is not true.  `dotnet build` will build whatever it is told to build.  If someone sets the `RuntimeIdentifier`, it respects that and doesn't build a portable app.

Documentation/design-docs/DotNetCore-SharedPackageStore.md
ramarag: correction, now app wins over all the caches

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -0,0 +1,218 @@
...
+
+```dotnet publish```
+
+Publish a standalone app for the current RID:
+
+```dotnet publish --standalone```
+
+Publish a standalone app for win7-x64 filtering out the publish profile:
+
+```dotnet publish --runtime win7-x64 --profile https://asp.net/core/dev/1.2.0/profile.xml```
```
ramarag: i have replaced --profile keyword with --filter

Documentation/design-docs/DotNetCore-SharedPackageStore.md
ramarag: yes i see that this is the current behavior. Currently if a RID is specified at publish time we only get a standalone app.
There is no way to get a portable app with native assets. i have filed a bug in sdk https://github.com/dotnet/sdk/issues/559

Documentation/design-docs/DotNetCore-SharedPackageStore.md
ramarag: Sticking to the original format

PR #1287, opened 2017-01-24
Provide a switch to skip building debian packages

ellismg:
When the "Package" target is selected, we try to build the relevant
installers as well as the tarballs. On Debian, doing so requires root
because of how we factor our packages (since we have to install some
packages before building others so our dependency stuff works out).

While the package building is already somewhat conditional (you need
`debuild` installed in order to do it) we should have finer grained
switches so you can opt out of building the debs and still get
tarballs. As part of the overall source build effort and move to
buildtools we should make this whole interaction less hacky and more
configurable.

### DIFF
--- a/build_projects/dotnet-host-build/DebTargets.cs
+++ b/build_projects/dotnet-host-build/DebTargets.cs
@@ -30,7 +30,7 @@ namespace Microsoft.DotNet.Host.Build
             // Ubuntu 16.04 Jenkins Machines don't have docker or debian package build tools
             // So we need to skip this target if the tools aren't present.
             // https://github.com/dotnet/core-setup/issues/167
-            if (DebuildNotPresent())
+            if (ShouldSkipBuildDebPackages() || DebuildNotPresent())
             {
                 c.Info("Debuild not present, skipping target: {nameof(GenerateSharedHostDeb)}");
                 return c.Success();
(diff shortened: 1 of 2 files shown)


build_projects/dotnet-host-build/DebTargets.cs
```diff
@@ -267,5 +267,10 @@ private static bool DebuildNotPresent()
         {
             return Cmd("/usr/bin/env", "debuild", "-h").Execute().ExitCode != 0;
         }
+
+        private static bool ShouldSkipBuildDebPackages()
+        {
+            return Environment.GetEnvironmentVariable("DOTNET_BUILD_SKIP_DEB_PACKAGING") == "1";
```
gkhanna79: Is there a reason we are not plumbing this through command line?

ellismg: I assumed the way this would all be handled would change with the move to build tools.  I can add a switch that sets this environment variable if you'd prefer to build.sh

gkhanna79: It will be great if you can add a switch to build.sh and propagate it through to the check you added. Thanks @ellismg.

ellismg: @gkhanna79 can you take another look?  I added `--skip-deb-package-build` to build.sh as well which sets the environment variable.

AlfredoMS:
got it. it should be pretty straight forward to add it to what I'm doing

ellismg:
Thanks for the conformation, @AlfredoMS

PR #1289, opened 2017-01-24
Include crossgen in NETCore.App package

ericstj:
Previously this was included by the runtime packages, but I didn't
explicitly redistribute when I flattened them.

Adding it to the runtime-specific NETCore.App packages in tools
directory.

/cc @eerhardt @gkhanna79 @weshaggard

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
@@ -24,6 +24,10 @@
 
   <Target Name="GetFilesToPackage" DependsOnTargets="ResolveNuGetPackages" Returns="@(FilesToPackage)">
     <!-- start with all the files resolved -->
+    <PropertyGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
+      <_runtimePackagePath Condition="'%(ReferenceCopyLocalPaths.FileName)' == 'coreclr' OR '%(ReferenceCopyLocalPaths.FileName)' == 'libcoreclr'">$(PackagesDir)%(ReferenceCopyLocalPaths.NuGetPackageId)/%(ReferenceCopyLocalPaths.NuGetPackageVersion)</_runtimePackagePath>
+    </PropertyGroup>
+
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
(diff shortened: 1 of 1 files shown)


ericstj:
I'd like folks' opinion on this.  It grows the NETCore.App package by 2.5 MB (why is crossgen so huge? 😕) but it does give parity with v1.x where they were getting it in the runtime-specific core-clr packages.

weshaggard:
There are definitely a number of scenarios where cross-gen is needed so I think it is a good idea to include.

ericstj:
Ok, I'll merge this in.  We can always change it later if folks want something different.

eerhardt:
:shipit:

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
```diff
@@ -35,6 +39,9 @@
         <TargetPath Condition="'%(_FilesToPackage.IsNative)' != 'true'">runtimes/$(NuGetRuntimeIdentifier)/lib/$(PackageTargetFramework)</TargetPath>
         <TargetPath Condition="'%(_FilesToPackage.IsNative)' == 'true'">runtimes/$(NuGetRuntimeIdentifier)/native</TargetPath>
       </_FilesToPackage>
+      <_FileToPackage Include="$(_runtimePackagePath)/tools/*.*">
```
gkhanna79: Why *.* and not specifically crossgen?

ericstj: Crossgen has a different name on different OSes and didn't want to special case it.  Also if you decide to bundle more stuff there (like symbols, or other tools) we'd probably want to pick it up by default.

eerhardt:
@ericstj - I'm not seeing crossgen included in the latest NuGet packages.  Am I looking in the right spot?

https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.win7-x86.Microsoft.NETCore.App
https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.win7-x64.Microsoft.NETCore.App
https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.osx.10.10-x64.Microsoft.NETCore.App

None of those packages have the crossgen tool in them.

ericstj:
@eerhardt I'll look into that.

PR #1309, opened 2017-01-25
Add Windows build definitions with signing

naamunds:
With these changes, the build definitions will now use MicroBuild for
signing, following the example of CoreFX.  init-tools.cmd is called
explicitly following the "Compile CoreHost" step, because buildtools
are needed for the MicroBuild .props and .target files.  init-tools.cmd
would otherwise be invoked during the "PackagePkgProjects" step, but I
wanted the build to continue signing HostFxr and HostPolicy before
building PackagePkgProjects.

This commit also removes a few unused variables from the Windows ARM
definitions.

/cc @crummel @chcosta @gkhanna79 @wtgodbe

### DIFF
--- a/build_projects/dotnet-host-build/sign.proj
+++ b/build_projects/dotnet-host-build/sign.proj
@@ -1,0 +1,107 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project InitialTargets="SetSigningProperties" DefaultTargets="SignFiles" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <RepoRoot>$(MSBuildThisFileDirectory)/../../</RepoRoot>
+  </PropertyGroup>
+
+  <!-- This will be overriden if we're building with MicroBuild. -->
+  <Target Name="SignFiles">
+    <Message Text="Fake sign target.  Would sign: @(FilesToSign)" />
+  </Target>
+
+  <Import Project="$(RepoRoot)pkg/Tools/MicroBuild.Core.props" />
+  <Import Project="$(RepoRoot)pkg/Tools/MicroBuild.Core.targets" />
+
+  <Target Name="SetSigningProperties">
+    <Error Condition="'$(RID)' == ''" Text="Missing required property 'RID'." />
+    <Error Condition="'$(CertificateId)' == ''" Text="Missing required property 'CertificateId'." />
+    <PropertyGroup>
+      <!-- The OutDir and IntermediateOutputPath properties are required by MicroBuild. MicroBuild only
+           signs files that are under these paths. -->
+      <OutDir Condition="'$(OutDir)' == ''">$(RepoRoot)artifacts/$(RID)/</OutDir>
+      <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == ''">$(OutDir)</IntermediateOutputPath>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="SignCoreHost" DependsOnTargets="GetSignCoreHostFiles">
+    <CallTarget Targets="SignFiles" />
+  </Target>
+
+  <Target Name="GetSignCoreHostFiles">
+    <ItemGroup>
+      <FilesToSign Include="$(OutDir)corehost/**/hostfxr.dll">
+        <Authenticode>$(CertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(OutDir)corehost/**/hostpolicy.dll">
+        <Authenticode>$(CertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignSharedFrameworkCrossgenedAssemblies" DependsOnTargets="GetSignSharedFrameworkCrossgenedAssembliesFiles">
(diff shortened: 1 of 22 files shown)


build_projects/dotnet-host-build/sign.proj
```diff
@@ -0,0 +1,107 @@
...
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignSharedFrameworkCrossgenedAssemblies" DependsOnTargets="GetSignSharedFrameworkCrossgenedAssembliesFiles">
+    <CallTarget Targets="SignFiles" />
+  </Target>
+
+  <Target Name="GetSignSharedFrameworkCrossgenedAssembliesFiles">
+    <ItemGroup>
+      <FilesToSign Include="$(OutDir)intermediate/sharedFrameworkPublish/shared/**/Microsoft.*.dll">
```
crummel: Should this be what you define IntermediateOutputPath as instead of OutDir?  Not that it matters too much since it's a subfolder.

buildpipeline/Core-Setup-Signing-Windows-x64-master.json
```diff
@@ -0,0 +1,518 @@
...
+      "allowOverride": true
+    },
+    "STORAGE_ACCOUNT": {
+      "value": "dotnetcli"
+    },
+    "STORAGE_CONTAINER": {
+      "value": "dotnet"
+    },
+    "CONNECTION_STRING": {
+      "value": "PassedViaPipeBuild"
```
crummel: All of these secret variables should have isSecret: true set, or they won't be masked in the child legs.

buildpipeline/Core-Setup-Signing-Windows-x64-master.json
```diff
@@ -0,0 +1,518 @@
...
+      "gitLfsSupport": "false",
+      "skipSyncSource": "false",
+      "fetchDepth": "0",
+      "cleanOptions": "0"
+    },
+    "id": "https://github.com/dotnet/core-setup.git",
+    "type": "GitHub",
+    "name": "dotnet/core-setup",
+    "url": "https://github.com/dotnet/core-setup.git",
+    "defaultBranch": "naamunds/signing",
```
crummel: Don't forget to change this before checking in  :)

buildpipeline/Core-Setup-Signing-Windows-x86-master.json
```diff
@@ -0,0 +1,491 @@
+{
```
crummel: I see two substantial differences between x64/x86: x86 is missing the "sign assemblies" step on x64 line 154, and x86 does not build NuGet packages.  A) are these intentional?  B) Are they something we could consolidate and have one build definition with an architecture parameter?  C) Or does core-setup care enough about architecture to keep them separate for build history reasons even if we could consolidate?

build_projects/dotnet-host-build/sign.proj
naamunds: It shouldn't matter functionally, but I'll make that change for clarity.

buildpipeline/Core-Setup-Signing-Windows-x64-master.json
naamunds: Core-Setup depends on these variables being environment variables, so because VSTS only exports non-secret build definition variables as environment variables, they are defined as non-secrets.  This is consistent with the current build definitions.

buildpipeline/Core-Setup-Signing-Windows-x64-master.json
naamunds: Good catch.

buildpipeline/Core-Setup-Signing-Windows-x86-master.json
naamunds: Those differences are from the existing MSEng definitions. The x64 and x86 build definitions were created on the same day, and the x64 build was created with the "sign assemblies" step whereas the x86 wasn't, so I assume that that was intentional.

The "BuildProjectsForNuGetPackages" and "GenerateNuGetPackages" targets were added to only the x64 build on 8/4/16 by @livarcocc.

@wtgodbe and @gkhanna79, can you confirm if those differences are intentional?

buildpipeline/Core-Setup-Signing-Windows-x64-master.json
crummel: Could we log a bug for this?  If we really need to have them be environment variables, VSO has a environment setup build step that can do it, but I think passing them as arguments like corefx and most of the other builds would be even better.

naamunds:
I've included a change to remove "-master" from the build definition names, so that the same files can be used to build multiple branches.  This follows the pattern used by CoreFX.

buildpipeline/Core-Setup-Signing-Windows-x86-master.json
wtgodbe: Yes, as far as I know this is intentional.

wtgodbe:
LGTM

buildpipeline/Core-Setup-Signing-Windows-x64-master.json
naamunds: I'll log a user story after merging this PR: "As a Core-Setup owner, my build secrets aren't exposed in the build logs"

PR #1330, opened 2017-01-27
Lighting up Shared packages

ramarag:
@gkhanna79 @schellap @eerhardt  PTAL

This is a port of https://github.com/dotnet/core-setup/pull/1109


ramarag:
FYI @pakrym  @muratg @glennc  -  DOTNET_HOSTING_OPTIMIZATION_CACHE is going away to be replaced with [Design Doc on dotnet cache](https://github.com/dotnet/core-setup/pull/1281)

src/corehost/cli/args.h
```diff
@@ -13,76 +13,71 @@
...
-    {
-        return patch_roll_fwd || prerelease_roll_fwd;
+        trace::verbose(_X("probe_config_t: probe=[%s]  deps-json=[%p] deps-dir-probe=[%d]"),
+            probe_dir.c_str(), probe_deps_json, probe_publish_dir);
     }
 
     probe_config_t(
         const pal::string_t& probe_dir,
         bool match_hash,
-        bool patch_roll_fwd,
```
schellap: Remove match hash

src/corehost/cli/args.h
```diff
@@ -13,76 +13,71 @@
...
     }
 
     static probe_config_t cache_ni(const pal::string_t& dir)
     {
-        return probe_config_t(dir, true, false, false, nullptr, false, true);
+        return probe_config_t(dir, true, nullptr, false, true, false);
     }
     
     static probe_config_t cache(const pal::string_t& dir)
     {
```
schellap: Remove cache and cache_ni

src/corehost/cli/libhost.h
```diff
@@ -194,27 +206,35 @@ struct hostpolicy_init_t
...
-        {
-            trace::error(_X("The size of the data layout used to initialize %s is %d; expected at least %d"), LIBHOSTPOLICY_NAME, input->version_lo, HOST_INTERFACE_LAYOUT_VERSION_LO);
-            return false;
-        }
+
         trace::verbose(_X("Reading from host interface version: [0x%04x:%d] to initialize policy version: [0x%04x:%d]"), input->version_hi, input->version_lo, HOST_INTERFACE_LAYOUT_VERSION_HI, HOST_INTERFACE_LAYOUT_VERSION_LO);
 
-        make_clrstr_arr(input->config_keys.len, input->config_keys.arr, &init->cfg_keys);
-        make_clrstr_arr(input->config_values.len, input->config_values.arr, &init->cfg_values);
+        if (input->version_lo >= offsetof(host_interface_t, host_mode) + sizeof(input->host_mode))
```
schellap: Pls add a comment here as per @gkhanna79 feedback in PR #1109. That the if checks are for compat between new/old fxr/policy, and we should not garbage due to potentially shorter struct size.

src/corehost/cli/libhost.h
```diff
@@ -194,27 +206,35 @@ struct hostpolicy_init_t
...
+            init->host_mode = (host_mode_t)input->host_mode;
+        }
+        else
+        {
+            trace::error(_X("The size of the data layout used to initialize %s is %d; expected at least %d"), LIBHOSTPOLICY_NAME, input->version_lo, 
+                offsetof(host_interface_t, host_mode) + sizeof(input->host_mode));
+        }
+
+        if (input->version_lo >= offsetof(host_interface_t, tfm) + sizeof(input->tfm))
+        {
```
schellap: Here as well.

src/corehost/common/pal.unix.cpp
```diff
@@ -170,6 +170,85 @@ bool pal::get_default_servicing_directory(string_t* recv)
...
+        if (pw && pw->pw_dir)
+        {
+            dir.assign(pw->pw_dir);
+        }
+    }
+    if (dir.empty())
+    {
+        return false;
+    }
+    append_path(&dir, _X(".dotnet"));
```
schellap: Can you put this in a method called `get_local_dotnet_dir` taking an out parameter string to help @t-cakine 's work. Then append the arch shared package path.

src/corehost/common/pal.unix.cpp
```diff
@@ -170,6 +170,85 @@ bool pal::get_default_servicing_directory(string_t* recv)
...
+bool pal::get_global_shared_package_dir(pal::string_t* recv)
+{
+    recv->clear();
+    pal::string_t dotnet_exe;
+    if (!locate_dotnet_on_path(&dotnet_exe))
+    {
+        return false;
+    }
+    pal::string_t dir = get_directory(dotnet_exe);
+    append_path(&dir, _X("packages"));
```
schellap: Same here. Make a helper for global `dotnet` dir.

muratg:
@ramarag @schellap  if the env var is removed, how will dotnet find where the optimization cache is on the disk?  This is a major scenario for Azure App Service (Antares) deployments to improve startup time for App Service apps. 

cc @DamianEdwards

DamianEdwards:
@muratg this replaces that. This is effectively a new cache that's more formally supported and achieves what the existing optimization cache did plus more, e.g. reducing need to deploy the assemblies with the app at all.

muratg:
Read the [spec](https://github.com/dotnet/core-proposals/blob/master/DotNetCore-SharedPackageStore.md).  It looks like there will be a fixed global location for the ngen'd files (plus a "local" one.)  So no way to configure the location of the global one?  I wonder if we may want to specify the location still.

DamianEdwards:
@muratg by default it would be located next to the instance of dotnet.exe being invoked (aka the global location). On Antares, that could (should) be on the fast drive. If that's not enough, then yes it should be possible to set this location via an environment variable/cmd line arg.

schellap:
@muratg, paths can be chained within `$DOTNET_SHARED_PACKAGES` env. This is in the spec in host probe precedence. The architecture of the host will be appended to these paths to look for packages.

@ramarag, it might be better to correct the spec to include this and also change the probe order to reflect implementation.

src/corehost/common/pal.unix.cpp
ramarag: done

schellap: You would still need the get_*_shared_packages function to append "packages" and use the method in main code. Caio would use the helper you created for muxing the fx.

ramarag: I just moved it to [utils.cpp](https://github.com/dotnet/core-setup/pull/1330/files#diff-fe7492df9cc9bec21e24dcb50e04b9c0R256), as this is now platform independent

ramarag:
i am merging this PR for now, will followup if any new new comments become relevant

PR #1342, opened 2017-01-28
Build NuGet packages on non windows systems

ellismg:
Because some of the projects cross-compile for full framework TFMs,
the build process on non windows doesn't work (unless you have mono
and the reference assembly packages installed).

To work around this, I just build netstandard and netcoreapp TFMs when
you are not on *nix.

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -346,14 +346,17 @@ namespace Microsoft.DotNet.Host.Build
         [Target]
         public static BuildTargetResult BuildProjectsForNuGetPackages(BuildTargetContext c)
         {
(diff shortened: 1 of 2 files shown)


ellismg:
/cc @ericstj @jyoungyun

This is a generalization of the work in #1137.  I need these packages to *always* be built on any host.  Because I do not want to take a dependency on having mono and the reference assemblies installed at build time, I can't build full framework tfms, so there's a filter around what frameworks we build for and some changes in the publish logic to ensure that we only ever publish the packages from a windows build.

jyoungyun:
@ellismg I confirmed that those 2 dlls were well included even if it was Ubuntu arm build. Thank you.

ellismg:
@gkhanna79 PTAL?

PR #1347, opened 2017-01-28
Teach update-versions how to update dotnet/standard packages

ellismg:
We don't yet have versions information being written for
dotnet/standard, so we don't actually update anything unless a URL is
provided via an environment variable.


ellismg:
/cc @ericstj @dagood 

Do we expect we will set up dotnet/versions for standard soon?

dagood:
https://github.com/dotnet/standard/issues/170 is tracking that. @bryanar

bryanar:
I'll start this today. Stay tuned.

PR #1377, opened 2017-01-30
Fixing a small spelling error in an MSBuild property name.

eerhardt:
@weshaggard @ericstj 

When publishing a self-contained application, none of the framework assemblies are being published.

I'm not sure where the best place to test all this build logic is, but fixing the error for now.

/cc @erozenfeld


ericstj:
We should be able to test better in core-setup once it is on a newer MSBuild-based CLI.

PR #1383, opened 2017-01-31
Use existent version of clang in Fedora Dockerfile

naamunds:
/cc @gkhanna79 @wtgodbe @mellinoe

### DIFF
--- a/scripts/docker/fedora.24/Dockerfile
+++ b/scripts/docker/fedora.24/Dockerfile
@@ -1,6 +1,6 @@
 FROM fedora:24
 
-RUN dnf install -y bash git cmake wget which python clang-3.8.0-2.fc24.x86_64 llvm-devel-3.8.0-1.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
+RUN dnf install -y bash git cmake wget which python clang-3.8.1-1.fc24.x86_64 llvm-devel-3.8.1-2.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
                    libunwind-devel.x86_64 lttng-ust-devel.x86_64 uuid-devel libuuid-devel tar glibc-locale-source zlib-devel libcurl-devel \
                    krb5-devel openssl-devel autoconf libtool hostname
 


naamunds:
Another option for these DNF packages would be to use `clang.x86_64` and `llvm-devel.x86_64`, but I opted to continue specifying the versions to ensure that these two packages are in sync.

gkhanna79:
CC @janvorli

scripts/docker/fedora.24/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM fedora:24
 
-RUN dnf install -y bash git cmake wget which python clang-3.8.0-2.fc24.x86_64 llvm-devel-3.8.0-1.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
+RUN dnf install -y bash git cmake wget which python clang-3.8.1-1.fc24.x86_64 llvm-devel-3.8.1-2.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
```
gkhanna79: Should we specify clang, as we do for Fedora 23?

janvorli:
Please use version-less names like we do for Fedora 23. The distro has clang and llvm that always match. If we specify a version, it would be problematic since we would have to keep it up to date when Fedora does a minor update of the tools. And at the same time, updating the version would force people that have the previous minor version to upgrade.

naamunds:
@janvorli I talked to @dagood about this, and he said that there is only one stable clang package version per Fedora release.  There are two package repositories in the fedora:24 base image, "fedora" and "updates", with the former containing the stable releases.  The "fedora" repository has 3.8.0-1.fc24, and the "updates" repository has 3.8.1-1.fc24.  Because 3.8.0-1.fc24 is the only stable version for Fedora 24, I don't think the update scenarios you mentioned apply, so Davis and I prefer specifying 3.8.0-1.fc24 so that the version is explicit.

/cc @MichaelSimons @dleeapho

janvorli:
@naamunds please help me understand the benefit of specifying explicit version of clang. You've said that the reason is that you want to make sure that the llvm and clang versions match. But the package manager doesn't allow installation of non-matching versions (I've tried).

naamunds:
@janvorli If we don't specify the version, then the latest version will be used, which currently means 3.8.1-2 from the "updates" repository, but could mean a different version in the future. We want to guarantee that the Docker image can be rebuilt in the future if needed and produce an image with the same components.

janvorli:
@naamunds ok, I think I understand now. So you want to be able to produce exactly the same binaries in the future. But in that case, all the dev libraries also play a role here. So if something changes in the headers we include, the binaries we produce might differ.

dleeapho:
@janvorli you are correct that the other dev libraries would also need to be specified to arrive at a completely serviceable environment. To do that we need a comprehensive toolset story and though that isn't available yet it is normative to specify a toolset version where we reasonably can.

naamunds:
@dleeapho Agreed. I think we should keep this PR minimal and get it in ASAP to unblock the builds, and the additional work for specifying versions on the other packages can be done separately.

Issue #1322, opened 2017-01-26
Clang3.6 should be the minimal toolset for cross build


gkhanna79:
CoreCLR repo specifies clang 3.6 for cross builds (https://github.com/dotnet/coreclr/blob/master/build.sh#L747) but Core-Setup does not (see https://github.com/dotnet/core-setup/blob/master/src/corehost/build.sh). 

CC @hqueue @hseok-oh @jyoungyun

PR #1385, opened 2017-01-31
Make clang-3.6 as a default compiler for cross compilation

hqueue:
Related issue: #1322 

1. Relocate clang configuration
    Old location: build_projects/dotnet-host-build/build.sh
    New location: src/corehost/build.sh

    clang is used when building corehost, but clang configuriation is too far away from corehost.
    Therefore I've moved code to the place where clang is really necessary.

2. Make clang-3.6 as a default compiler for cross compilation
    Other than cross compilation, clang-3.5 remains as a default. (no change)


hqueue:
@gkhanna79 PTAL. I made 3.6 as a default and only compiler for cross compilation. Is it ok ?

Issue #1221, opened 2017-01-19
Host cannot find satellite assemblies listed in deps.json when using version Microsoft.NETCore.App 2.0


nguerrera:
## Steps to reproduce

* Install https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/Latest/dotnet-win-x64.latest.exe

* From Windows command prompt:
  * git clone https://github.com/nguerrera/repros
  * cd repros\SatelliteAssemblyHostIssue
  * repro.cmd

## Expected  behavior

The program runs fine.

## Actual behavior
```
Error: assembly specified in the dependencies manifest was not found -- package: 'satelliterepro', version: '1.0.0', path: 'fr/SatelliteRepro.resources.dll'
```

## Environment data

```
Microsoft .NET Core 2.0.0 - Runtime (x64) 2.0.0.1384

.NET Command Line Tools (1.0.0-rc3-004512)

Product Information:
 Version:            1.0.0-rc3-004512
 Commit SHA-1 hash:  61b7841087

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   D:\Src\repros\SatelliteAssemblyHostIssue\.dotnet_cli\sdk\1.0.0-rc3-004512
```

@gkhanna79 @ramarag

nguerrera:
cc @Eilon @pranavkm @ryanbrandenburg 

This has been blocking ASP.NET for a while. It was logged first as a CLI issue (dotnet/CLI#5172), then an SDK (dotnet/sdk#673) issue, but after preparing the repro above and talking offline, this appears to be an issue in the 2.0 host.

Eilon:
Thanks for the update. If you have a private build that you want us to try out, we can certainly do that as well to make sure the ASP.NET Core scenario is functional.

gkhanna79:
@schellap Can you please take an initial stab at this?

ryanbrandenburg:
Do we have an update on this? It's currently blocking progress on aspnet/Localization#315.

gkhanna79:
@nguerrera The repro script (in your repro repo) no longer works since the CLI it is using is not present. Trying to use latest CLI (https://dotnetcli.blob.core.windows.net/dotnet/Sdk/rel-1.0.0/dotnet-dev-win-x64.latest.zip) results in different restore errors.

Do you have a repro available offline that you can share with the CLI in question?

[...]

PR #1392, opened 2017-01-31
Fix resource assembly path computation

gkhanna79:
Fixes https://github.com/dotnet/core-setup/issues/1221

This is a regression introduced from https://github.com/dotnet/core-setup/pull/529 to fix https://github.com/dotnet/core-setup/issues/436, as a result of which extracting IETF code from resource relative path (in deps.json) would have an extra directory seperator appended.

The fix is to strip the separator out.

@ramarag PTAL.

### DIFF
--- a/src/corehost/cli/deps_entry.cpp
+++ b/src/corehost/cli/deps_entry.cpp
@@ -69,8 +69,19 @@ bool deps_entry_t::to_dir_path(const pal::string_t& base, pal::string_t* str) co
         {
             replace_char(&pal_relative_path, _X('/'), DIR_SEPARATOR);
         }
+
+        // Resources are represented as "<ietf-code>/<ResourceAssemblyName.dll>" in the deps.json.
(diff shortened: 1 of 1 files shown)


src/corehost/cli/deps_entry.cpp
```diff
@@ -69,8 +69,19 @@ bool deps_entry_t::to_dir_path(const pal::string_t& base, pal::string_t* str) co
         {
             replace_char(&pal_relative_path, _X('/'), DIR_SEPARATOR);
         }
+
+        // Resources are represented as "<ietf-code>/<ResourceAssemblyName.dll>" in the deps.json.
+        // The <ietf-code> is the "directory" in the pal_relative_path below, so extract it.
         pal::string_t ietf_dir = get_directory(pal_relative_path);
```
ramarag: NIT, itetf_dir is not used for anything, as well can be removed

ramarag:
LGTM, modulo the nit comment

PR #1393, opened 2017-01-31
Use latest stable version of clang and llvm-devel

naamunds:
As per the discussion on #1383, we should be using the stable versions of clang and llvm-devel, i.e. from the "fedora" package repository instead of the "updates" package repository.  I hadn't yet pushed the update to that PR when it was merged.

The stable version for clang and llvm-devel is 3.8.0-1.  With this change, compared to before #1383, the only change is from clang 3.8.0-2 to 3.8.0-1.

/cc @gkhanna79 @janvorli

### DIFF
--- a/scripts/docker/fedora.24/Dockerfile
+++ b/scripts/docker/fedora.24/Dockerfile
@@ -1,6 +1,6 @@
 FROM fedora:24
 
-RUN dnf install -y bash git cmake wget which python clang-3.8.1-1.fc24.x86_64 llvm-devel-3.8.1-2.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
+RUN dnf install -y bash git cmake wget which python clang-3.8.0-1.fc24.x86_64 llvm-devel-3.8.0-1.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
                    libunwind-devel.x86_64 lttng-ust-devel.x86_64 uuid-devel libuuid-devel tar glibc-locale-source zlib-devel libcurl-devel \
                    krb5-devel openssl-devel autoconf libtool hostname
 


scripts/docker/fedora.24/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM fedora:24
 
-RUN dnf install -y bash git cmake wget which python clang-3.8.1-1.fc24.x86_64 llvm-devel-3.8.1-2.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
+RUN dnf install -y bash git cmake wget which python clang-3.8.0-1.fc24.x86_64 llvm-devel-3.8.0-1.fc24.x86_64 make libicu-devel lldb-devel.x86_64 \
```
gkhanna79: This regresses the clang version used to ship Core-Setup repo binaries for 1.1.0. See https://github.com/naamunds/core-setup/blob/release/1.1.0/scripts/docker/fedora.24/Dockerfile#L3.

gkhanna79: Do we understand what is the difference between the two?

naamunds: No, I'm not sure what the difference is between clang-3.8.0-2 (which was used until #1382 was merged earlier today) and clang-3.8.0-1.  It looks like release notes are just provided for the minor versions, e.g. for 3.8 at http://releases.llvm.org/3.8.0/tools/clang/docs/ReleaseNotes.html.

I don't think leaving it as 3.8.1-1 is a viable solution, because it will likely be removed at some point, just as 3.8.0-2 was.  The two options are using the stable version from the "fedora" package repo by specifying 3.8.0-1, and using the latest version from the "updates" package repo by not specifying a version at all. As mentioned in https://github.com/dotnet/core-setup/pull/1383#issuecomment-276447836, the main concern with not specifying a version is that we want to guarantee that the Docker image can be rebuilt in the future if needed and produce an image with the same components.

I ran a build locally with these changes (using the same steps as the official Fedora 24 build), and it succeeded.  Are there any functional regressions from the clang version that we have to be aware of that wouldn't be surfaced by a build failure?

/cc @dagood @MichaelSimons @dleeapho

gkhanna79: I looked around but didnt come across any release notes for the differences. Looking at https://pkgs.org/download/clang for Fedora 24, 3.8.0-1 is the only option, so it seems like that is the only thing we can work with.

> Docker image can be rebuilt in the future if needed and produce an image with the same components.

Given the above, do you want to fix Fedora 23 dockerfile as well?

naamunds: > Given the above, do you want to fix Fedora 23 dockerfile as well?

I think that should be done as part of developing a comprehensive toolset story for these Docker images, as mentioned in https://github.com/dotnet/core-setup/pull/1383#issuecomment-276462691.  This can be done as part of https://github.com/dotnet/core-eng/issues/225 or https://github.com/dotnet/core-eng/issues/365, which I will be working on this sprint.

naamunds:
@gkhanna79 Should I go ahead and merge this PR, or are there remaining concerns?

Issue #1400, opened 2017-02-01
Ubuntu arm and tizen armel build on Docker


jyoungyun:
## Steps to reproduce
```
./cross/build-rootfs.sh arm
./build.sh --docker ubuntu.16.04 --env-vars "DISABLE_CROSSGEN=1,TARGETPLATFORM=arm,TARGETRID=ubuntu.16.04-arm,CROSS=1,ROOTFS_DIR=/opt/code/cross/rootfs/arm"
```

## Expected  behavior
Successful building core-setup

## Actual behavior
1. Disappears the required arguments
```
./build.sh --docker ubuntu.16.04 --env-vars "DISABLE_CROSSGEN=1,TARGETPLATFORM=arm,TARGETRID=ubuntu.16.04-arm,CROSS=1,ROOTFS_DIR=/opt/code/cross/rootfs/arm"
->
Argumtens~~~~:--env-vars DISABLE_CROSSGEN=1,TARGETPLATFORM=arm,TARGETRID=-arm,CROSS=1,ROOTFS_DIR=/opt/code/cross/rootfs/arm
```
2. The lack of cross build toolchain (binutils-arm-linux-gnueabihf)
```
  /usr/bin/ld: unrecognised emulation mode: armelf_linux_eabi
```

jyoungyun:
I used the exisiting Ubuntu Dockerfile for building Ubuntu ARM. If you want to create Dockerfile for Ubuntu ARM in a seperate file, please let me know.

hqueue:
@jyoungyun Although size of each image can be larger than before, I think it would be desirable to have a few docker image for maintanance. Therefore I prefer updating existing Ubuntu image as your PR.

gkhanna79:
@jyoungyun Are you doing this to enable CI support, as done for https://github.com/dotnet/core-setup/blob/master/netci.groovy#L42?

hqueue:
@gkhanna79 Not determined yet. We are trying to make use of docker for local build as well as for CI and try to evaluate any possible choices.

gkhanna79:
@hqueue In that case, once your investigations have determined the option you wish to pursue, do let me know as I would like to ensure there is a single place for Docker images that can be used across the board.

hqueue:
@gkhanna79 Sure, I will let you know when investigation is don. And glad to hear that there is a single place for Docker images. We also hope that we can make use of same Docker image across Core-Setup, CoreClr, and CoreFx, if we decided to use Docker.

[...]

PR #1401, opened 2017-02-01
Fix bug when parsing arguments in build.sh

jyoungyun:
All arguments that match the docker argument were removed by https://github.com/dotnet/core-setup/blob/master/build.sh#L36L37 lines.
```
./build.sh --docker ubuntu.16.04 --env-vars "DISABLE_CROSSGEN=1,TARGETPLATFORM=arm,TARGETRID=ubuntu.16.04-arm,CROSS=1,ROOTFS_DIR=/opt/code/cross/rootfs/arm"
->
Argumtens~~~~:--env-vars DISABLE_CROSSGEN=1,TARGETPLATFORM=arm,TARGETRID=-arm,CROSS=1,ROOTFS_DIR=/opt/code/cross/rootfs/arm
```

Related issue: #725 #1400

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,20 +24,17 @@ if [ -z "$HOME" ]; then
     mkdir -p $HOME
 fi
 
-args=( "$@" )
-
 while [[ $# > 0 ]]; do
     lowerI="$(echo $1 | awk '{print tolower($0)}')"
     case $lowerI in
         --docker)
             export BUILD_IN_DOCKER=1
             export DOCKER_IMAGENAME=$2
-            # remove docker args
-            args=( "${args[@]/$1}" )
-            args=( "${args[@]/$2}" )
             shift
             ;;
         *)
+            args+=( $1 )
+            ;;
     esac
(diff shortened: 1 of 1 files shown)


hqueue:
LGTM

build.sh
```diff
@@ -24,20 +24,17 @@ if [ -z "$HOME" ]; then
...
         --docker)
             export BUILD_IN_DOCKER=1
             export DOCKER_IMAGENAME=$2
-            # remove docker args
-            args=( "${args[@]/$1}" )
-            args=( "${args[@]/$2}" )
             shift
             ;;
         *)
+            args+=( $1 )
```
gkhanna79: Can you please elaborate on the issue you were running into?

jyoungyun: When I use `TARGETRID` env value with `--docker` option, the previous build.sh removes all matched string with `--docker` argument. This scripts want to forward the all arguments without docker to next script, so I add other arguments into `args` array.

PR #1406, opened 2017-02-01
Run Docker as current user in official builds

naamunds:
In the PortableLinux and RHEL build definitions, "docker run" is called
directly, without specifying the user as which to run.  This means that
the build artifacts which are mapped to the host have "root" as their
owner, because "root" as the default Docker user.  This prevents the
build agent's user from being able to delete those files in subsequent
builds, causing official builds to fail.

This fix runs the Docker container as the same user ID and group ID as
the host, which causes the build artifacts to have that same owner so
that they can be deleted later. This is done in a new script because
invoking the "id" command to get the user ID within the build
definition is not trivial, so this is the simplest implementation.

This is a short-term fix to unblock the builds; I am considering
removing the volume-mapping and instead copying in the files for a
longer-term solution.

/cc @gkhanna79 @wtgodbe @dagood @MichaelSimons @chcosta

### DIFF
--- a/buildpipeline/Core-Setup-PortableLinux-x64.json
+++ b/buildpipeline/Core-Setup-PortableLinux-x64.json
@@ -22,17 +22,18 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Run build.sh in Docker container",
+      "displayName": "Run build in Docker container",
       "timeoutInMinutes": 0,
       "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
         "definitionType": "task"
       },
       "inputs": {
-        "filename": "docker",
-        "arguments": "run -t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB chcosta/dotnetcore:rhel7_prereqs /bin/bash -c \"git clean -X -d -f; ./build.sh $(BuildArguments)\"",
-        "workingFolder": "",
+        "scriptPath": "scripts/dockerrun-as-current-user.sh",
+        "args": "-t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB chcosta/dotnetcore:rhel7_prereqs /bin/bash -c \"HOME=/opt/code; git clean -X -d -f; ./build.sh $(BuildArguments)\"",
+        "disableAutoCwd": "false",
+        "cwd": "",
         "failOnStandardError": "false"
       }
     }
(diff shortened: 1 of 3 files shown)


naamunds:
This change also sets HOME to `/opt/code` within the Docker container, because otherwise HOME won't be set for the user.

buildpipeline/Core-Setup-PortableLinux-x64.json
```diff
@@ -22,17 +22,18 @@
...
-        "versionSpec": "1.*",
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
         "definitionType": "task"
       },
       "inputs": {
-        "filename": "docker",
-        "arguments": "run -t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB chcosta/dotnetcore:rhel7_prereqs /bin/bash -c \"git clean -X -d -f; ./build.sh $(BuildArguments)\"",
-        "workingFolder": "",
+        "scriptPath": "scripts/dockerrun-as-current-user.sh",
```
gkhanna79: Why are we not running into this issue with other build definitions?

naamunds: The other Linux build definitions build a Dockerfile from the repo using dockerrun.sh. That scripts specifies the user when building the image at https://github.com/dotnet/core-setup/blob/master/scripts/dockerrun.sh#L92.  The PortableLinux and RHEL build definitions just pull and run an image from a Docker registry.

gkhanna79: In that case, why dont we use the RHEL docker container for these two and unify the infrastructure?

naamunds: They already both use the RHEL image.  Are you referring to consolidating the RHEL and PortableLinux build definitions into a single build definition?  I think that's a good idea and have logged https://github.com/dotnet/core-eng/issues/414 for it, but I don't think it should be done as part of this PR.

buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -22,17 +22,18 @@
...
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
         "definitionType": "task"
       },
       "inputs": {
-        "filename": "docker",
-        "arguments": "run -t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB chcosta/dotnetcore:rhel7_prereqs /bin/bash -c \"git clean -X -d -f; ./build.sh $(BuildArguments)\"",
-        "workingFolder": "",
+        "scriptPath": "scripts/dockerrun-as-current-user.sh",
+        "args": "-t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB chcosta/dotnetcore:rhel7_prereqs /bin/bash -c \"HOME=/opt/code; git clean -X -d -f; ./build.sh $(BuildArguments)\"",
```
dagood: No issue here: In case others didn't know this (I had to look it up), `HOME` doesn't need `export` here to pass along the value to the script because it's marked for export by default.

PR #1413, opened 2017-02-02
Support building on OSX 10.12

ellismg:
Since the host RID on OSX 10.12 (Sierra) was not listed in the table
of host rids to package rids, the build failed there.

While I was in the area, I added entries for other RIDs that could
come from different versions on RHEL.


gkhanna79:
Are we now having build machines for official builds to support this? Or, did you run into this on a local build @ellismg ?

gkhanna79:
Please also confirm that https://github.com/dotnet/core-setup/blob/53ccd9c957f3a47aa1a56051e286b9511e1e0248/src/corehost/cli/deps_format.cpp#L132 is returning the expected value.

ellismg:
> Are we now having build machines for official builds to support this? Or, did you run into this on a local build @ellismg ?

I hit it on a local build and did this to unblock myself.  AFAIK the build machines aren't on 10.12 (there's an internal thread about moving them forward, but no work has been done).

ellismg:
The get_own_rid function was doing the right thing.

PR #1424, opened 2017-02-02
Update Microsoft.DiaSymReader.Native version to 1.4.1

tmat:
Fixes #14627

The new package contains the same binaries as 1.4.0 and a fixed .props file.

This is the actual change in the props file: https://github.com/dotnet/roslyn-internal/commit/b9fc00eb6b5f2a9f918b1a21fbefcd6db68feec1 -- the only difference between 1.4.0 and 1.4.1

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/project.json
+++ b/pkg/projects/Microsoft.NETCore.App/project.json
@@ -39,7 +39,7 @@
     "System.Threading.Tasks.Parallel": "4.3.0",
     "System.Threading.Thread": "4.3.0",
     "System.Threading.ThreadPool": "4.3.0",
-    "Microsoft.DiaSymReader.Native": "1.4.0",
+    "Microsoft.DiaSymReader.Native": "1.4.1",
     "Libuv": "1.9.1"
   },
   "frameworks": {


tmat:
@gkhanna79 @karelz @dotnet/core-setup-contrib

gkhanna79:
Have we validated that it addresses the problem? If it does, please send email to shiproom as well.

CC @leecow

tmat:
I validated that when the new package is added to a .NET Core Console App targeting .NET Core App 1.1 I can open and read PDB, and that the output dir doesn't have the DSRN binary.

tmat:
@gkhanna79 @leecow Is it ok to merge?

leecow:
@tmat - Did the note to shiproom ever happen? I am unable to find the review request.

mikem8361:
The change looks good to me.

tmat:
This is the actual change in the props file: https://github.com/dotnet/roslyn-internal/commit/b9fc00eb6b5f2a9f918b1a21fbefcd6db68feec1 -- the only difference between 1.4.0 and 1.4.1

PR #1425, opened 2017-02-02
Fix typo in crossgen inclusion

ericstj:
I was adding to _FileToPackage rather than _FilesToPackage.

I also needed to set NuGetPackageId as we use that for deduping later,
so I refactored how we find the runtime package a bit to make that
easier.

/cc @eerhardt

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
@@ -23,9 +23,14 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
 
   <Target Name="GetFilesToPackage" DependsOnTargets="ResolveNuGetPackages" Returns="@(FilesToPackage)">
+    <ItemGroup  Condition="'$(NuGetRuntimeIdentifier)' != ''">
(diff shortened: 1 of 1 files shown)


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.depproj
```diff
@@ -39,9 +44,10 @@
         <TargetPath Condition="'%(_FilesToPackage.IsNative)' != 'true'">runtimes/$(NuGetRuntimeIdentifier)/lib/$(PackageTargetFramework)</TargetPath>
         <TargetPath Condition="'%(_FilesToPackage.IsNative)' == 'true'">runtimes/$(NuGetRuntimeIdentifier)/native</TargetPath>
       </_FilesToPackage>
-      <_FileToPackage Include="$(_runtimePackagePath)/tools/*.*">
+      <_FilesToPackage Include="$(_runtimePackagePath)/tools/*.*">
+        <NuGetPackageId>$(_runtimePackageId)</NuGetPackageId>
         <TargetPath>tools</TargetPath>
```
eerhardt: Do you also want/need the `NuGetPackageVersion`?

ericstj: I don't need it so I didn't bother adding it.  These metadata actually get ignored for packaging because they'll get overwritten in the pkgproj with the package identity and version that these bits get bundled in.

ericstj:
Checked the latest packages and they have this fix.

PR #1426, opened 2017-02-02
Include all RID-specific assets in platform manifest during official build

ericstj:
Previously I only included assets in the platform manifest for packages
being built on the current platform.

This is a problem because the platform manifest ships in a single
RID agnostic package so normally it would only have one set of files
that correspond to the RID-specific package that happened to be
built in the same build leg.

I foresaw this problem and made a switch to the package build that
would tell it to examine the assets from all RID specific packages.

This change uses that switch during the official build (identified
by environment SIGNED_PACKAGES=true).  This switch is only safe to
use in the official build where we've already built all RIDs for
dependencies since it will cause the build to download RID-specific
packages for all of its dependencies.  That would fail in a purely
vertical (RID-specific) end-to-end build, AKA: vertical build.

In the future we need to make sure that if we switch the official
build to be completely vertical we still have a build-leg that will
run after all verticals of dependencies are complete to build the
RID-less package for Microsoft.NETCore.App.

/cc @eerhardt @ellismg @dagood @chcosta @weshaggard 

Fixes https://github.com/dotnet/standard/issues/191

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -14,6 +14,11 @@
     <ShouldCreateLayout>false</ShouldCreateLayout>
     <PropsFile>$(IntermediateOutputPath)$(MSBuildProjectName).props</PropsFile>
     <PlatformManifestFile>$(IntermediateOutputPath)$(MSBuildProjectName).PlatformManifest.txt</PlatformManifestFile>
+    <!-- During an official build when we can guarantee that all RID-specific dependencies have been built,
+         restore all of those dependencies and gather the prospective content of the RID-specific Core.App
+         packages.  This is needed so that we have a complete platform manifest in the shipping version of
+         the Microsoft.NETCore.App (RID-agnostic/identity package). -->
+    <IncludeAllRuntimePackagesInPlatformManifest Condition="'$(SIGNED_PACKAGES)' == 'true'">true</IncludeAllRuntimePackagesInPlatformManifest>
   </PropertyGroup>
   
   <Choose>
@@ -107,7 +112,7 @@
     <MSBuild Projects="@(ProjectReference)"
              Condition="'%(ProjectReference.PackageTargetRuntime)' == '$(PackageRID)' OR 
                         '%(ProjectReference.BuidOnRID)' == '$(PackageRID)' OR 
-                        ('$(IncludeAllPackages)' == 'true' AND '%(ProjectReference.PackageTargetRuntime)' != '')"
+                        ('$(IncludeAllRuntimePackagesInPlatformManifest)' == 'true' AND '%(ProjectReference.PackageTargetRuntime)' != '')"
              Targets="GetPackageFiles">
       <Output TaskParameter="TargetOutputs" ItemName="SharedFrameworkRuntimeFiles" />
     </MSBuild>
--- a/pkg/tasks/GenerateFileVersionProps.cs
+++ b/pkg/tasks/GenerateFileVersionProps.cs
@@ -47,10 +47,9 @@ namespace Microsoft.DotNet.Build.Tasks
                     continue;
                 }
 
-                var filePath = file.GetMetadata("FullPath");
-                var fileName = Path.GetFileName(filePath);
+                var fileName = Path.GetFileName(file.ItemSpec);
 
-                var current = GetFileVersionData(filePath);
+                var current = GetFileVersionData(file);
 
                 FileVersionData existing;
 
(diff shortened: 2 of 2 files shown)


ericstj:
This is an interesting bug. Looks like an official build happened in the middle of building this CI run.  Since my branch was based on an older commit the version produced by the CI was -001483, but since the tests use a * in their dependencies they somehow got the one from myget and failed because the matching shared framework wasn't installed.  /cc @AlfredoMS

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -14,6 +14,11 @@
     <ShouldCreateLayout>false</ShouldCreateLayout>
     <PropsFile>$(IntermediateOutputPath)$(MSBuildProjectName).props</PropsFile>
     <PlatformManifestFile>$(IntermediateOutputPath)$(MSBuildProjectName).PlatformManifest.txt</PlatformManifestFile>
+    <!-- During an official build when we can garuntee that all RID-specific dependencies have been built,
```
eerhardt: `garuntee`

pkg/tasks/GenerateFileVersionProps.cs
```diff
@@ -112,13 +111,40 @@ public override bool Execute()
...
+                    FileVersion = FileUtilities.GetFileVersion(filePath)
+                };
+            }
+            else
             {
-                AssemblyVersion = FileUtilities.TryGetAssemblyVersion(filePath),
-                FileVersion = FileUtilities.GetFileVersion(filePath)
-            };
+                // allow for the item to specify version directly
+                Version assemblyVersion = null;
```
eerhardt: (nit) no need to set to null, since these are used in out parameters.

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -14,6 +14,11 @@
     <ShouldCreateLayout>false</ShouldCreateLayout>
     <PropsFile>$(IntermediateOutputPath)$(MSBuildProjectName).props</PropsFile>
     <PlatformManifestFile>$(IntermediateOutputPath)$(MSBuildProjectName).PlatformManifest.txt</PlatformManifestFile>
+    <!-- During an official build when we can guarantee that all RID-specific dependencies have been built,
+         restore all of those dependencies and gather the prospective content of the RID-specific Core.App
+         packages.  This is needed so that we have a complete platform manifest in the shipping version of
+         the Microsoft.NETCore.App (RID-agnostic/identity package). -->
+    <IncludeAllRuntimePackagesInPlatformManifest Condition="'$(SIGNED_PACKAGES)' == 'true'">true</IncludeAllRuntimePackagesInPlatformManifest>
```
weshaggard: Where is SIGNED_PACKAGES coming from?

weshaggard: I see from you description this comes from an env variable. Can we please make it an explicit option we pass instead of depending on env variables in our official builds?

ericstj: We're free to pass in IncludeAllRuntimePackagesInPlatformManifest if that's what you want.  I used SIGNED_PACKAGES because this is already passed in and used by the core-setup build to do a similar thing for other packages.

weshaggard: Yes I would definitely prefer to stop depending on env-variables in our builds they cause use so much grief because they are hidden away and different across setups.

eerhardt: I'm not sure I follow the ask here. Would your concern be satisfied if SIGNED_PACKAGES turns into an MSBuild property that is defaulted to 'false'. And then during official builds, we set the SIGNED_PACKAGES property to 'true'?

weshaggard: Yes I just want something that is explicitly passed.

ericstj: Well now you have two different properties and you're welcome to change the build definition...

weshaggard: @naamunds @AlfredoMS as part of the core-setup build changes can we please ensure that we are not depending on env-variables but instead we are passing in the necessary properties to the build steps. Thanks.

PR #1439, opened 2017-02-03
Support for Linux Arm32 builds

gkhanna79:
This change enables building for Ubuntu 14.04 and 16.04 for Arm32. Once this is merged, we can enable pipeline builds.

I also made few cleanup fixes to add more reliability to the build.

@ramarag @wtgodbe PTAL

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/Monikers.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Monikers.cs
@@ -38,7 +38,7 @@ namespace Microsoft.DotNet.Cli.Build
             string rid = Environment.GetEnvironmentVariable("TARGETRID") ?? RuntimeEnvironment.GetRuntimeIdentifier();
 
             // Look for expected RIDs, including Portable one, for Linux
-            if (rid.StartsWith("linux-") || rid == "ubuntu.16.04-x64" || rid == "ubuntu.16.10-x64" || rid == "fedora.23-x64" || rid == "fedora.24-x64" || rid == "opensuse.13.2-x64" || rid == "opensuse.42.1-x64" || rid == "debian.8-armel" || rid == "tizen.4.0.0-armel")
+            if (rid.StartsWith("linux-") || rid == "ubuntu.16.04-x64" || rid == "ubuntu.16.04-arm" || rid == "ubuntu.16.10-x64" || rid == "fedora.23-x64" || rid == "fedora.24-x64" || rid == "opensuse.13.2-x64" || rid == "opensuse.42.1-x64" || rid == "debian.8-armel" || rid == "tizen.4.0.0-armel")
             {
                 return $"{artifactPrefix}-{rid}.{version}";
             }
@@ -50,19 +50,20 @@ namespace Microsoft.DotNet.Cli.Build
             }
         }
 
-        public static string GetBadgeMoniker()
+        public static string GetBadgeMoniker(BuildTargetContext c)
         {
-            string rid = Environment.GetEnvironmentVariable("TARGETRID") ?? RuntimeEnvironment.GetRuntimeIdentifier();
-
+            string rid = c.BuildContext.Get<string>("TargetRID");
             if (rid.StartsWith("linux-"))
             {
-                return $"Linux_{Environment.GetEnvironmentVariable("TARGETPLATFORM") ?? CurrentArchitecture.Current.ToString()}";
+                return $"Linux_{c.BuildContext.Get<string>("Platform")}";
             }
 
             switch (rid)
             {
                 case "ubuntu.16.04-x64":
                      return "Ubuntu_16_04_x64";
+                case "ubuntu.16.04-arm":
+                     return "Ubuntu_16_04_arm";
                 case "ubuntu.16.10-x64":
                      return "Ubuntu_16_10_x64";
                 case "fedora.23-x64":
@@ -79,7 +80,7 @@ namespace Microsoft.DotNet.Cli.Build
                      return "Tizen_4_0_0_armel";
             }
 
-            return $"{CurrentPlatform.Current}_{Environment.GetEnvironmentVariable("TARGETPLATFORM") ?? CurrentArchitecture.Current.ToString()}";
+            return $"{CurrentPlatform.Current}_{c.BuildContext.Get<string>("Platform")}";
         }
 
         public static string GetDebianHostFxrPackageName(string hostfxrNugetVersion)
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -4,95 +4,7 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Delete files from $(Build.SourcesDirectory)",
-      "timeoutInMinutes": 0,
-      "task": {
-        "id": "b7e8b412-0437-4065-9371-edc5881de25b",
-        "versionSpec": "1.*",
-        "definitionType": "task"
-      },
-      "inputs": {
-        "SourceFolder": "$(Build.SourcesDirectory)",
-        "Contents": "**\n.gitignore"
-      }
-    },
-    {
-      "enabled": false,
-      "continueOnError": false,
-      "alwaysRun": false,
-      "displayName": "git clone",
-      "timeoutInMinutes": 0,
-      "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
-        "definitionType": "task"
-      },
-      "inputs": {
-        "filename": "git",
-        "arguments": "clone $(PB_RepoUrl) core-setup",
-        "workingFolder": "",
-        "failOnStandardError": "false"
-      }
-    },
-    {
-      "enabled": false,
-      "continueOnError": false,
-      "alwaysRun": false,
-      "displayName": "git checkout",
-      "timeoutInMinutes": 0,
-      "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
-        "definitionType": "task"
-      },
-      "inputs": {
-        "filename": "git",
-        "arguments": "checkout $(SourceVersion)",
-        "workingFolder": "$(Build.SourcesDirectory)/corefx",
-        "failOnStandardError": "false"
-      }
-    },
-    {
-      "enabled": false,
-      "continueOnError": false,
-      "alwaysRun": true,
-      "displayName": "Initialize tools",
-      "timeoutInMinutes": 0,
-      "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
-        "definitionType": "task"
-      },
-      "inputs": {
-        "filename": "$(Build.SourcesDirectory)/core-setup/init-tools.sh",
-        "arguments": "",
-        "workingFolder": "",
-        "failOnStandardError": "false"
-      }
-    },
-    {
-      "enabled": false,
-      "continueOnError": false,
-      "alwaysRun": false,
-      "displayName": "Docker cleanup",
-      "timeoutInMinutes": 0,
-      "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
-        "definitionType": "task"
-      },
-      "inputs": {
-        "filename": "$(Build.SourcesDirectory)/core-setup/Tools/scripts/docker/cleanup-docker.sh",
-        "arguments": "",
-        "workingFolder": "",
-        "failOnStandardError": "false"
-      }
-    },
-    {
-      "enabled": true,
-      "continueOnError": false,
-      "alwaysRun": false,
-      "displayName": "Start detached docker container",
+      "displayName": "Start detached container",
       "timeoutInMinutes": 0,
       "task": {
         "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
@@ -110,7 +22,7 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Clone repository",
+      "displayName": "Clone Repository",
       "timeoutInMinutes": 0,
       "task": {
         "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
@@ -128,7 +40,7 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Check out the specified commit",
+      "displayName": "Checkout commit",
       "timeoutInMinutes": 0,
       "task": {
         "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
@@ -146,7 +58,7 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Run build-rootfs.sh",
+      "displayName": "Build RootFS",
       "timeoutInMinutes": 0,
       "task": {
         "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
@@ -164,7 +76,7 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Run build.sh",
+      "displayName": "Run Build",
       "timeoutInMinutes": 0,
       "task": {
         "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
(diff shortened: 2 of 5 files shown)


wtgodbe:
LGTM

build_projects/shared-build-targets-utils/Utils/Monikers.cs
```diff
@@ -38,7 +38,7 @@ public static string GetProductMoniker(BuildTargetContext c, string artifactPref
             string rid = Environment.GetEnvironmentVariable("TARGETRID") ?? RuntimeEnvironment.GetRuntimeIdentifier();
 
             // Look for expected RIDs, including Portable one, for Linux
-            if (rid.StartsWith("linux-") || rid == "ubuntu.16.04-x64" || rid == "ubuntu.16.10-x64" || rid == "fedora.23-x64" || rid == "fedora.24-x64" || rid == "opensuse.13.2-x64" || rid == "opensuse.42.1-x64" || rid == "debian.8-armel" || rid == "tizen.4.0.0-armel")
+            if (rid.StartsWith("linux-") || rid == "ubuntu.16.04-x64" || rid == "ubuntu.16.04-arm" || rid == "ubuntu.16.10-x64" || rid == "fedora.23-x64" || rid == "fedora.24-x64" || rid == "opensuse.13.2-x64" || rid == "opensuse.42.1-x64" || rid == "debian.8-armel" || rid == "tizen.4.0.0-armel")
```
hqueue: Just curious :)
Are you going to enable only ubuntu.16.04-arm and not ubuntu.14.04-arm ?

build_projects/shared-build-targets-utils/Utils/Monikers.cs
```diff
@@ -50,19 +50,20 @@ public static string GetProductMoniker(BuildTargetContext c, string artifactPref
...
-                return $"Linux_{Environment.GetEnvironmentVariable("TARGETPLATFORM") ?? CurrentArchitecture.Current.ToString()}";
+                return $"Linux_{c.BuildContext.Get<string>("Platform")}";
             }
 
             switch (rid)
             {
                 case "ubuntu.16.04-x64":
                      return "Ubuntu_16_04_x64";
+                case "ubuntu.16.04-arm":
+                     return "Ubuntu_16_04_arm";
```
hqueue: Same question here :)

build_projects/shared-build-targets-utils/Utils/Monikers.cs
gkhanna79: I am enabling for both. These specialize checks (in both your questions) treat 14.04 artifact generation differently than 16.04 case in the "else" section of either.

gkhanna79:
@wtgodbe @chcosta @naamunds PTAL at the pipeline definition changes I have made.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+    },
+    "portableLinux": {
+      "value": ""
+    },
+    "CrossToolsetVersion": {
+      "value": "",
+      "allowOverride": true
+    },
+    "SkipUnmount": {
+      "value": "--SkipUnmount"
```
hqueue: Is this a parameter to build-rootfs.sh ? Then I think the option should be in lowercase, i.e. `--skipunmount`, although in it is `--SkipUnmount` help message 

This is one reason that I started refactoring build-rootfs.sh in corefx.

buildpipeline/pipeline.json
```diff
@@ -130,6 +130,29 @@
           }
         },
         {
+          "Name": "Core-Setup-CrossBuild",
+          "Parameters": {
+            "PB_Architecture": "arm"
```
wtgodbe: Just to be explicit, you should pass in values for PB_DockerTag & TargetRid here, like below (not strictly necessary, but makes readability easier).

wtgodbe:
LGTM other than comment

buildpipeline/Core-Setup-CrossBuild.json
gkhanna79: https://github.com/dotnet/core-setup/blob/master/cross/build-rootfs.sh#L28 does the lowercasing.

buildpipeline/pipeline.json
gkhanna79: Isnt that the job of default values set in the coresetup-crossbuild.json?

build_projects/shared-build-targets-utils/Utils/Monikers.cs
hqueue: I see. I understand that "else" section can handle ubuntu.14.04-arm. Is there any reason we should handle ubuntu 14.04 differently than 16.04 ?, because, at first, I thought we can handle them in same way.

buildpipeline/Core-Setup-CrossBuild.json
hqueue: Got it :) Thanks.

build_projects/shared-build-targets-utils/Utils/Monikers.cs
gkhanna79: @hqueue It is simply to maintain status quo and avoid unrequired additional changes for the moment that we can do without for cross build bringup. Over a long period, our build infrastructure is going to change for Core-Setup significantly that this piece of code will just disappear and will no longer be relevant.

buildpipeline/pipeline.json
wtgodbe: Yes, but being explicit here can make it more obvious what's being set (it might not be obvious why one job passes values for certain parameter while another doesn't) - we do this for coreFX already (https://github.com/dotnet/corefx/blob/master/buildpipeline/pipeline.json#L159), and in the CoreCLR pipelines.json file in VSO. Also relying on default values can be dangerous, in case they change - although in this case that's unlikely to happen.

gkhanna79: Look at the latest iteration - I have made all of them explicit.

build_projects/shared-build-targets-utils/Utils/Monikers.cs
hqueue: Understood and I also agree to keep changes to a minimum. Thanks :)

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+        "FilePath",
+        "SymbolStore"
+      ],
+      "daysToKeep": 10,
+      "minimumToKeep": 1,
+      "deleteBuildRecord": true,
+      "deleteTestResults": true
+    }
+  ],
+  "buildNumberFormat": "$(date:yyyyMMdd)$(rev:-rr)",
```
naamunds: It looks like this is different from the `buildNumberFormat` of the other Core-Setup build definitions, which use `$(Date:yyyMMdd)$(Rev:.r)`.

buildpipeline/Core-Setup-CrossBuild.json
hqueue: @gkhanna79 Just be curious.
Is `--SkipUnmount` (camel case) preferred over `--skipunmount` ?
Because I thought https://github.com/dotnet/core-setup/blob/master/cross/build-rootfs.sh#L28 is just fail-safe code, I've already changed help message of built-rootfs.sh in corefx. If `--SkipUnmount`, i.e. Camel case, is preferred, I will keep  `--SkipUnmount` in help meesage of build-rootfs.sh.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "reportBuildStatus": "true",
+      "fetchDepth": "0",
+      "gitLfsSupport": "false",
+      "skipSyncSource": "false",
+      "cleanOptions": "0"
+    },
+    "id": "58fa2458-e392-4373-ba2b-dd3ef0c2d7ce",
+    "type": "TfsGit",
+    "name": "DotNet-CoreFX-Trusted",
+    "url": "https://devdiv.visualstudio.com/DevDiv/_git/DotNet-CoreFX-Trusted",
```
naamunds: Why does this build definition use CoreFX as its repository instead of Core-Setup like the other build definitions?  I was expecting the whole `repository` node to be identical to the other build definitions.  I think doing so would also eliminate the need for the "git clone" and "git checkout" steps and the `PB_RepoUrl` variable.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "id": 39,
+      "name": "DotNet-Build"
+    },
+    "id": 36,
+    "name": "DotNet-Build"
+  },
+  "path": "\\",
+  "type": "build",
+  "id": 5267,
+  "name": "CoreSetup-CrossBuild",
```
naamunds: This name needs to match the name of the file, so it should be `Core-Setup-CrossBuild`.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+  "url": "https://devdiv.visualstudio.com/DefaultCollection/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_apis/build/Definitions/5267",
```
naamunds: The `url` is typically omitted for checked-in definitions.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "daysToKeep": 10,
+      "minimumToKeep": 1,
+      "deleteBuildRecord": true,
+      "deleteTestResults": true
+    }
+  ],
+  "buildNumberFormat": "$(date:yyyyMMdd)$(rev:-rr)",
+  "jobAuthorizationScope": "projectCollection",
+  "jobTimeoutInMinutes": 90,
+  "repository": {
```
naamunds: Should this build definition have `badgeEnabled` set to true like the other Core-Setup build definitions?

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+    "PB_CloudDropAccountName": {
+      "value": "dotnetbuildoutput"
+    },
+    "CloudDropAccessToken": {
+      "value": null,
+      "isSecret": true
+    },
+    "OfficialBuildId": {
+      "value": "$(Build.BuildNumber)"
+    },
```
naamunds: It doesn't look like the `OfficialBuildId` variable is used in this build definition, so I think it can be removed.  (It's already passed in by the pipebuild anyway, and I assume we want this value to match what is used for the other build legs.)

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "displayName": "Run publish-packages.sh",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "docker",
+        "arguments": "exec $(PB_DockerContainerName) $(PB_GitDirectory)/publish-packages.sh -AzureAccount=$(PB_CloudDropAccountName) -AzureToken=$(CloudDropAccessToken) -Container=$(PB_Label) -verbose -- /p:OverwriteOnPublish=true",
```
naamunds: What creates the Container specified here by `-Container=$(PB_Label)`?  Is it created during this build step? I'm wondering if it should use the build number from the pipebuild (i.e. `OfficialBuildId`) instead of the build number of this build leg.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "value": "ubuntu1404_cross_prereqs_v1",
+      "allowOverride": true
+    },
+    "PB_CloudDropAccountName": {
+      "value": "dotnetbuildoutput"
+    },
+    "CloudDropAccessToken": {
+      "value": null,
+      "isSecret": true
+    },
```
naamunds: The `CloudDropAccessToken` isn't currently passed in by the pipebuild, so you'll have to add this variable as a secret there, and then add it to the list of variables passed in via the "-g" option of the arguments to pipebuild.exe.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+    "OfficialBuildId": {
+      "value": "$(Build.BuildNumber)"
+    },
+    "PB_Label": {
+      "value": "$(Build.BuildNumber)"
+    },
+    "GitHubBranch": {
+      "value": "sni_plus_latestbuildtools"
+    },
```
naamunds: It doesn't look like the `GitHubBranch` variable is used in this build definition, so I think it can be removed.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+    "SourceVersion": {
+      "value": "HEAD",
+      "allowOverride": true
+    },
+    "PB_DockerCopyDest": {
+      "value": "$(Build.BinariesDirectory)/docker_repo"
+    },
+    "portableLinux": {
+      "value": ""
+    },
```
naamunds: What is the purpose of this `portableLinux` variable?  Will it sometimes be overridden from pipeline.json?

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "git",
+        "arguments": "checkout $(SourceVersion)",
+        "workingFolder": "$(Build.SourcesDirectory)/corefx",
```
naamunds: Should this be `core-setup` instead?  In the "Check out the specified commit" step, the `SourceVersion` corresponds to Core-Setup rather than CoreFX.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,488 @@
...
+      "alwaysRun": true,
+      "displayName": "Copy Publish Artifact: BuildLogs",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "1d341bb0-2106-458c-8422-d00bcea6512a",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "CopyRoot": "$(PB_DockerCopyDest)/corefx",
```
naamunds: Should this be `core-setup`?

gkhanna79:
@naamunds Thanks for the comments. The issues you saw were likely because the definition was built from a clone of CoreFX Crossbuild definition. I have updated it after building it over a Core-Setup Ubuntu definition that resulted in https://devdiv.visualstudio.com/DevDiv/Net%20CoreFX/_build/explorer?_a=queued&definitionId=5272 as the physical definition that I used for validation. 

The two builds for validation were https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=558978 (for Ubuntu 14.04) and https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=558980 (for Ubuntu 16.04).

Can you PTAL?

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,332 @@
...
+    "PB_DockerImageName": {
+      "value": "$(PB_DockerRepository):$(PB_DockerTag)"
+    },
+    "PB_Architecture": {
+      "value": "arm"
+    },
+    "portableLinux": {
+      "value": "",
+      "allowOverride": true
+    },
```
naamunds: I'm still wondering about the purpose of this variable.  It's used in the "Run Build" step, but it doesn't appear to ever be set to anything other than an empty string.

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -0,0 +1,332 @@
...
+      "displayName": "Build RootFS",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "sudo",
+        "arguments": "docker exec --privileged $(PB_DockerContainerName) $(PB_GitDirectory)/cross/build-rootfs.sh $(PB_Architecture) $(CrossToolsetVersion) $(SkipUnmount)",
```
naamunds: You may be able to omit `$(PB_GitDirectory)/` because it was specified as the working directory of the container.  (I'm not certain if that's how `docker exec` works, so you might have to verify it.)  This applies to some other `docker exec` commands too.

buildpipeline/Core-Setup-CrossBuild.json
gkhanna79: This will be set when we perform Portable Linux build for Arm32. They are not yet done for that architecture (but we have them enabled for x64).

gkhanna79:
Thank you all - I am merging this and will kickoff a build to see how it goes.

gkhanna79:
Queued off https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=559874&_a=summary

Petermarcu:
Are you kidding me! Are we this close to having real runtime builds for arm32? Yay!

gkhanna79:
Yes, we are close - just need to keep an eye on the engineering issues since there has been bunch of churn there (move from mseng to devdiv, opensuse being taken off, etc).

PR #1458, opened 2017-02-06
Set MACOSX_RPATH on hostfxr and hostpolicy

wjk:
On my system (using CMake 3.5.2), CMake emits a configure-time warning because the `MACOSX_RPATH` property is not set on the targets `hostfxr` and `hostpolicy`. I have gone ahead and set this property. The `MACOSX_RPATH` property is documented (by `cmake --help-property MACOSX_RPATH`) as follows:

```
MACOSX_RPATH
------------

Whether this target on OS X or iOS is located at runtime using rpaths.

When this property is set to ``TRUE``, the directory portion of
the ``install_name`` field of this shared library will be ``@rpath``
unless overridden by ``INSTALL_NAME_DIR``.  This indicates
the shared library is to be found at runtime using runtime
paths (rpaths).

This property is initialized by the value of the variable
``CMAKE_MACOSX_RPATH`` if it is set when a target is
created.

Runtime paths will also be embedded in binaries using this target and
can be controlled by the ``INSTALL_RPATH`` target property on
the target linking to this target.

Policy ``CMP0042`` was introduced to change the default value of
``MACOSX_RPATH`` to ``TRUE``.  This is because use of ``@rpath`` is a
more flexible and powerful alternative to ``@executable_path`` and
``@loader_path``.
```

Seeing as how the `dotnet` executable opens this library using `pal::load_library` (which [resolves to `dlopen`](https://github.com/dotnet/core-setup/blob/master/src/corehost/common/pal.unix.cpp#L69-L78)), this should not prove an issue with the corehost as written. However, if the libraries affected were ever to be embedded inside of a GUI application on macOS (a `*.app` bundle), this change may well affect the GUI application for the better, as the files will no longer expect to be installed in a fixed location, instead being located at load time based on the file that links against them. This enables macOS applications to be moved around on disk without breaking.

### DIFF
--- a/src/corehost/cli/dll/CMakeLists.txt
+++ b/src/corehost/cli/dll/CMakeLists.txt
@@ -59,6 +59,7 @@ if(WIN32)
     list(APPEND RESOURCES ../native.rc)
 endif()
 add_library(hostpolicy SHARED ${SOURCES} ${RESOURCES})
+set_target_properties(hostpolicy PROPERTIES MACOSX_RPATH TRUE)
 
 # Specify the import library to link against for Arm32 build since the default set is minimal
 if (WIN32 AND CLI_CMAKE_PLATFORM_ARCH_ARM)
--- a/src/corehost/cli/fxr/CMakeLists.txt
+++ b/src/corehost/cli/fxr/CMakeLists.txt
@@ -56,6 +56,7 @@ if(WIN32)
     list(APPEND RESOURCES ../native.rc)
 endif()
 add_library(hostfxr SHARED ${SOURCES} ${RESOURCES})
+set_target_properties(hostfxr PROPERTIES MACOSX_RPATH TRUE)
 
 # Specify the import library to link against for Arm32 build since the default set is minimal
 if (WIN32 AND CLI_CMAKE_PLATFORM_ARCH_ARM)


gkhanna79:
> as the files will no longer expect to be installed in a fixed location, instead being located at load time based on the file that links against them.

Do you mean based upon the file that statically links against them? hostfxr is implemented to locate hostpolicy is well known locations (either SxS or in shared FX install location), so I am not sure how this change helps.

wjk:
@gkhanna79 This change will only affect applications that link against the libraries in question by passing the `-l` flag in their linkline; it has no effect upon `dlopen()` calls. I guess I did not make that quite clear enough above.

wjk:
Note: I have just rebased this PR on top of the latest `master` commit.

gkhanna79:
@janvorli PTAL

janvorli:
@gkhanna79 I am sorry for the delay, it went out of my radar somehow. Looking now.

janvorli:

eerhardt:
@dotnet-bot test this please

janvorli:

Petermarcu:

janvorli:
@dotnet/dnceng There seems to be something broken in the CI w.r.t. CentOS7.1 x64 Debug test. It fails and clicking on details leads to a page with the following text. It seems to be consistent (in the original run and two retries): 
```
HTTP ERROR 404

Problem accessing /job/dotnet_core-setup/job/master/job/debug_centos7.1_x64_prtest/1108/. Reason:

    Not Found
```

eerhardt:
This PR is pretty old.  Since it was created, we have changed a bunch of the legs in our CI.

I'm going to close and re-open it.  Which should bring it up to date.

eerhardt:
Merging and ignoring the CentOS failure, since it is out of date.

PR #1466, opened 2017-02-06
Update "NETStandard.Library" to "2.0.0-beta-25006-01"

weshaggard:
cc @ericstj @eerhardt 

This package should have correctly signed bits from the NETStandard.Library package so the msbuild task should correctly work in VS.


eerhardt:
Should these updates be happening automatically?

weshaggard:
>Should these updates be happening automatically?

Yes we should be updating them automatically but we haven't enabled maestro yet. I was thinking to do that for both corefx and core-seutp today now that standard repo is updating the versions repo https://github.com/dotnet/versions/blob/master/build-info/dotnet/standard/master/Latest_Packages.txt.

PR #1486, opened 2017-02-08
Deprecate OpenSuse 13.2

gkhanna79:
This is to unblock CI and pipeline builds.

@wtgodbe PTAL

CC @weshaggard


wtgodbe:
LGTM - are we planning on reintroducing this later on?

weshaggard:
>LGTM - are we planning on reintroducing this later on?

I don't expect we will I think we 13.2 is reaching the end of its support life-cycle.

Issue #1504, opened 2017-02-08
Fix the RID host is built for


gkhanna79:
The highest RID host is built for is a function of the RID that comes out of https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PrepareTargets.cs#L100. I see the same in 1.1.0 as well - https://github.com/dotnet/core-setup/blob/release/1.1.0/build_projects/dotnet-host-build/PrepareTargets.cs#L50.
 
TargetRID from above is passed to CMake at https://github.com/dotnet/core-setup/blob/release/1.1.0/build_projects/dotnet-host-build/CompileTargets.cs#L191 and that is used to define TARGET_RUNTIME_RID at https://github.com/dotnet/core-setup/blob/release/1.1.0/src/corehost/cli/setup.cmake#L91 which is, then, used as the RID for the host to fallback at https://github.com/dotnet/core-setup/blob/release/1.1.0/src/corehost/cli/deps_format.cpp#L144.

In 1.0.0, I see that RID to be used by the host comes from https://github.com/dotnet/core-setup/blob/release/1.0.0/build_projects/dotnet-host-build/CompileTargets.cs#L173 that computes it from Stage0 path.

@ianhays added the win7-* check above in https://github.com/dotnet/core-setup/commit/20600ceb2e8be73bf63e72341132cf8125c8fcdb#diff-ff058bba01b969c51b1b7f3b68cd9fe0.

This is a rgeression that should be fixed in master (and would also involve ensuring that official builds are done on Win10 boxes so that we get the highest RID) and in 1.1.2 since we are close to stabilizing 1.1.1.

CC @eerhardt @ramarag @leecow @Petermarcu

eerhardt:
Repro:

Reference the attached nupkg in a `netcoreapp1.0` app
[MyRidLib.1.0.0-alpha.nupkg.zip](https://github.com/dotnet/core-setup/files/762118/MyRidLib.1.0.0-alpha.nupkg.zip)

with the following code:

```C#
        static void Main(string[] args)
        {
            new ClassLibrary.Class1().PrintMessage();

            Console.WriteLine("Hello World!");
        }
```

run on a Windows 10 machine, and you get the output:

```
F:\RIDFallback>F:\cli\.dotnet_stage0\x64\dotnet.exe run -p App\App.csproj
This is on Windows10.
Hello World!
```

now upgrade your app to `netcoreapp2.0` and run it on the same machine:

```
[...]

PR #1505, opened 2017-02-08
Set CoreHost RID to the current RID, not the target RID

ianhays:
[My commit in August](https://github.com/dotnet/core-setup/commit/20600ceb2e8be73bf63e72341132cf8125c8fcdb#diff-ff058bba01b969c51b1b7f3b68cd9fe0) changed the value we were passing to `DCLI_CMAKE_RUNTIME_ID` to always be equal to the base RID instead of the current RID as a part of getting ARM64 up and running. This commit reverts that behavior as progress towards https://github.com/dotnet/core-setup/issues/1504

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -202,7 +202,7 @@ namespace Microsoft.DotNet.Host.Build
                 List<string> cmakeArgList = new List<string>();
 
                 string cmakeBaseRid, visualStudio, archMacro, arch;
-                string ridMacro = $"-DCLI_CMAKE_RUNTIME_ID:STRING={rid}";
+                string ridMacro = $"-DCLI_CMAKE_RUNTIME_ID:STRING=win10-{platform.ToLower()}";
                 string cmakeHostVer = $"-DCLI_CMAKE_HOST_VER:STRING={hostVersion.LatestHostVersion.ToString()}";
                 string cmakeAppHostVer = $"-DCLI_CMAKE_APPHOST_VER:STRING={hostVersion.LatestAppHostVersion.ToString()}";
                 string cmakeHostPolicyVer = $"-DCLI_CMAKE_HOST_POLICY_VER:STRING={hostVersion.LatestHostPolicyVersion.ToString()}";


build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -171,7 +171,9 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
         {
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
             var configuration = c.BuildContext.Get<string>("Configuration");
-            string rid = c.BuildContext.Get<string>("TargetRID");
+            string rid = DotNetCli.Stage0.GetRuntimeId();
```
gkhanna79: I think the fix should be removing the check at https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PrepareTargets.cs#L100 instead as that will ensure we already get the correct RID when we are here.

jyoungyun: Please consider the cross build situation. When I tried to build for ubuntu arm, I passed the TargetRID value through --env-vars like below.
```
$ ./build.sh  --env-vars DISABLE_CROSSGEN=1,TARGETPLATFORM=arm,TARGETRID=ubuntu.14.04-arm,CROSS=1,ROOTFS_DIR=/home/jyoung/git/dotnet/rootfs-coreclr/arm/
```

ianhays: >I think the fix should be removing the check at 

I put in that check so that the other build products would be put under the win7 folders like they were before my changes. If I remove that check, everything will be made win10-* on a win10 machine.

If we want that then that's fine, but I was under the impression that we wanted the packages to have win7-* RIDS.

gkhanna79:
@ianhays Can you confirm what is the behaviour of 1.0.x build on win10?

build_projects/dotnet-host-build/CompileTargets.cs
ianhays: @jyoungyun is this code touched from build.sh? I thought the Unix build had it's own version of this code elsewhere?

ianhays:
>Can you confirm what is the behaviour of 1.0.x build on win10?

The TARGET_RUNTIME_ID passed to the host compilation on a win10 machine:
- versions before [my commit in August](https://github.com/dotnet/core-setup/commit/20600ceb2e8be73bf63e72341132cf8125c8fcdb#diff-ff058bba01b969c51b1b7f3b68cd9fe0) was `win10-*`
- After my change it was `win7-*`. 
- After this PR, it's `win10-* `again to match the old behavior.

build_projects/dotnet-host-build/CompileTargets.cs
gkhanna79: @jyoungyun is correct that this change will break cross-arch builds. 

> I put in that check so that the other build products would be put under the win7 folders like they were before my changes.

Which other products are you referring to Ian?

jyoungyun: @ianhays When I used below codes, cross build worked normally. I think the logic to read TARGETRID is correct only when cross build.
```
            if (string.Equals(crossEnv, "1"))                                   
            {                                                                   
                rid = c.BuildContext.Get<string>("TargetRID");                  
            }
```

ianhays: >Which other products are you referring to Ian?

The package RIDs use the TARGETRID variable. If we remove that line of code you linked then the packages we would build would be e.g. `runtime.win10-x64.Microsoft.NETCore.App.2.0.0-*` which they never have been before.

>@jyoungyun is correct that this change will break cross-arch builds.

Are cross-arch builds something new since my original changes in August were done? This PR is reverting the behavior to how it used to be before my fix accidentally touched it, so it shouldn't put us in any worse of a state than we were before my changes.

gkhanna79: Sounds good to me.

Yes, cross-arch builds came post your changes but needs to account for cross build, as @jyoungyun exemplified above.

ianhays: Is there any reason why we shouldn't always build the host for the highest windows RID regardless of the RID of the current system?

We can pass win10-* as the CMAKE_RUNTIME_ID for all windows target RIDs.

ianhays: e.g.

```
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
...
    string ridMacro = $"-DCLI_CMAKE_RUNTIME_ID:STRING=win10-{platform.ToLower()}";
...
}
```

instead of
```
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
...
    rid = c.BuildContext.Get<string>("TargetRID");
    string ridMacro = $"-DCLI_CMAKE_RUNTIME_ID:STRING={rid}";
...
}
```

gkhanna79: I would suggest something like this that will preserve cross build semantics as well:

```
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
...
    rid = $"win10-{platform.ToLower()}";
...
}
```

I would also suggest to rename the variable to "hostRID". And if you are doing this, I would suggest do this initialization in PrepareTargets.CommonInit and simply fetch it here from BuildContext.

ianhays: >I would also suggest to rename the variable to "hostRID". And if you are doing this, I would suggest do this initialization in PrepareTargets.CommonInit and simply fetch it here from BuildContext.

I agree that those changes would be nice/better, but we should really do the absolute minimum code change to get this working as intended or we may have unintended breaks. The Core-Setup repo has a **lot** of hardcoded expectations that I don't want to test. 

I pushed a minimal fix that won't break the cross-arch build as long as it passes `TargetArch`

jyoungyun:
LGTM

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -202,7 +202,7 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
                 List<string> cmakeArgList = new List<string>();
 
                 string cmakeBaseRid, visualStudio, archMacro, arch;
-                string ridMacro = $"-DCLI_CMAKE_RUNTIME_ID:STRING={rid}";
+                string ridMacro = $"-DCLI_CMAKE_RUNTIME_ID:STRING=win10-{platform.ToLower()}";
```
gkhanna79: In general, the change is fine. That said, I would suggest that this "RIDForBuildingHost" be setup in BuildContext in PrepareTargets.Init so that other parts of the build can access it if required. This ensures central initialization of such concepts instead of them being spilled all over the codebase.

And you can, then, fetch it from the BuildContext and use it.

gkhanna79:
Also, @ianhays once the correct fix is in master, please do prep a PR for 1.1.0 branch to get this fix that we will take once 1.1.1 has shipped. Filed https://github.com/dotnet/core-setup/issues/1548 and assigned to you to track it.

gkhanna79:
Why did we merge this without setting this up correctly @ianhays ?

build_projects/dotnet-host-build/CompileTargets.cs
ianhays: The buildcontext variable I left out for the reasons mentioned in the code comment and because it would only ever be used in the corehost compilation and even then only for the Windows build. We've already got a few different RID variables in `CompileTargets.cs`; adding another to only be used in the win corehost build would only add more indirection. We've already got the other baseRIDs for Windows hardcoded based on the platform (cmakeBaseRid), so it's preferable having this calculation alongside the other ones.

I'm against renaming or moving any existing variables, at least as a part of this PR. I've been burned too many times by Core-Setup

PR #1506, opened 2017-02-08
Deprecate OpenSuse 13.2 support

gkhanna79:
Remove OpenSuse 13.2 build support as it has been deprecated for 2.0 builds.

@weshaggard @eerhardt PTAL

CC @wtgodbe @dagood


gkhanna79:
Can someone PTAL as I would like to see if CoreSetup gets unblocked with this?

PR #1527, opened 2017-02-10
Fixes for Arm32 build

gkhanna79:
Specify CROSS=1 for the cross build.

Ensure that platform installers are not attempted to be published for Arm32.


gkhanna79:
@eerhardt Can you PTAL?

gkhanna79:
Fixes issues like https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=564928

eerhardt:
> Ensure that platform installers are not attempted to be published for Arm32.

Why?  Why don't we publish `apt-get` installers for arm32?

gkhanna79:
The deb generation logic appears to be using x64 tools that do not operate on arm32 binaries. This change is to get things going and then look later into what it means to deb generation for architectures like Arm and Arm64.

gkhanna79:
@eerhardt Let me know if there are any further questions I can help answer.

eerhardt:
As long as we have an issue logged and a plan to get these built, I am happy.

PR #1537, opened 2017-02-11
CrossBuild performs explicit publishing step

gkhanna79:
This change splits the crossbuild into two steps to address breaks like https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=566348 that happen when attempting to publish to myget from a Docker container:

1) Perform the actual build and publish to azure blob in container
2) Publish to myget from outstide the container as an explicit step

@dagood PTAL

### DIFF
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -49,7 +49,7 @@
       },
       "inputs": {
         "filename": "docker",
-        "arguments": "run --privileged -d -w $(PB_GitDirectory) --name $(PB_DockerContainerName) $(PB_DockerImageName) -v $(System.DefaultWorkingDirectory)$(PB_RepoName):$(PB_GitDirectory) sleep 7200",
+        "arguments": "run --privileged -d -w $(PB_GitDirectory) --name $(PB_DockerContainerName) $(PB_DockerImageName) -v $(System.DefaultWorkingDirectory)/$(PB_RepoName):$(PB_GitDirectory) sleep 7200",
         "workingFolder": "",
         "failOnStandardError": "false"
       }
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -4,6 +4,42 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
+      "displayName": "Clone Repository on host machine",
+      "timeoutInMinutes": 0,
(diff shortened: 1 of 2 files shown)


gkhanna79:
Will allow us to workaround https://github.com/NuGet/Home/issues/4286

gkhanna79:
@wtgodbe PTAL as well

gkhanna79:
We have had another build failure in Core-Setup due to this. @eerhardt @chcosta PTAL if you get a chance. 

I will merge this change later today in an attempt to address (and iterate if required) the issue.

chcosta:
I'd propose a slightly different ordering.

1. Clone Repository on host machine
2. Checkout commit on host machine
3. Start detached container
  a. Add "-v $(System.DefaultWorkingDirectory)\$(PB_RepoName):$(PB_GitDirectory)"
4. Build RootFS
5. Run Build
6. Publish Packages (from "$(System.DefaultWorkingDirectory)\$(PB_RepoName)")
7. CleanupRootFS
8. Stop Container
9. Remove Container

This would allow you to use the same file system for the build and publishing and not have to clone twice.

gkhanna79:
@chcosta Made the changes - PTAL

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -49,7 +49,7 @@
       },
       "inputs": {
         "filename": "docker",
-        "arguments": "exec $(PB_DockerContainerName) git checkout $(SourceVersion)",
+        "arguments": "run --privileged -d -w $(PB_GitDirectory) --name $(PB_DockerContainerName) $(PB_DockerImageName) -v $(System.DefaultWorkingDirectory)$(PB_RepoName):$(PB_GitDirectory) sleep 7200",
```
chcosta: I think you need a '\' between these "$(System.DefaultWorkingDirectory)$(PB_RepoName)"

chcosta: I see that it didn't show in my comment because i forgot to escape it for md.

chcosta: Otherwise, LGTM

gkhanna79: What do I need @chcosta ?

dagood:
This workaround looks reasonable and very focused on the specific issue we seem to be seeing.

(This isn't the workaround I was suggesting--I don't think it was meant to be, but just in case that isn't clear.)

PR #1552, opened 2017-02-14
Use GetFileAttributesEx to determine presence of folder or file

gkhanna79:
This Windows specific issue happens since FindFirstFile , which is used to determine file/folder presence, cannot handle arguments that terminate with trailing slash without any file pattern arguments.

A more reliable way to do the same would be to use GetFileAttributesEx that can handle this case. Also, this aligns the Windows implementation of this PAL function with what is done on non-Windows using stat function.

@eerhardt @ramarag PTAL

### DIFF
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -327,11 +327,15 @@ bool pal::file_exists(const string_t& path)
         return false;
     }
 
-    WIN32_FIND_DATAW data;
-    auto find_handle = ::FindFirstFileW(path.c_str(), &data);
-    bool found = find_handle != INVALID_HANDLE_VALUE;
-    ::FindClose(find_handle);
-    return found;
+
+    // We will attempt to fetch attributes for the file or folder in question that are
+    // returned only if they exist.
+    WIN32_FILE_ATTRIBUTE_DATA data;
+    if (GetFileAttributesExW(path.c_str(), GetFileExInfoStandard, &data) != 0) {
+        return true;
+    }
+    
+    return false;
 }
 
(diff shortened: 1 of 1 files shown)


src/corehost/common/pal.windows.cpp
```diff
@@ -327,11 +327,15 @@ bool pal::file_exists(const string_t& path)
...
-    auto find_handle = ::FindFirstFileW(path.c_str(), &data);
-    bool found = find_handle != INVALID_HANDLE_VALUE;
-    ::FindClose(find_handle);
-    return found;
+
+    // We will attempt to fetch attributes for the file or folder in question that are
+    // returned only if they exist.
+    WIN32_FILE_ATTRIBUTE_DATA data;
+    if (GetFileAttributesExW(path.c_str(), GetFileExInfoStandard, &data) != 0) {
+        return true;
```
eerhardt: In corefx, it looks like we also check the fileAttributes:

https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Win32FileSystem.cs#L340

gkhanna79: Agreed. However, the PAL api is designed to be used for validating presence of any file system artifact and not just directory (which is what it is used for right now). Hence, I didnt constraint it. This is the same semantic that existed before my change.

eerhardt: Maybe it is a poorly named method. To me, `file_exists` should return `false` when passed a directory.

gkhanna79: Agreed!

PR #1556, opened 2017-02-16
Export __base_rid to be recognized by toolchain.cmake

jyoungyun:
In an environment without gcc-arm-linux-gnueabi (such as a docker),
clang does not find a toolchain for tizen by default, and
in this case it must set the exact library location.
The CLI_CMAKE_RUNTIME_ID that is passed as an argument is not passed at check time in the clang,
so __base_rid environment is exported and used.

Related issue : #725

### DIFF
--- a/cross/armel/toolchain.cmake
+++ b/cross/armel/toolchain.cmake
@@ -16,7 +16,7 @@ add_compile_options(--sysroot=${CROSS_ROOTFS})
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -target ${TOOLCHAIN}")
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} --sysroot=${CROSS_ROOTFS}")
 
-if("${CLI_CMAKE_RUNTIME_ID}" MATCHES "tizen.*")
+if("$ENV{__CrossToolChainTargetRID}" MATCHES "tizen.*")
     set(TIZEN_TOOLCHAIN "armv7l-tizen-linux-gnueabi/4.9.2")
     include_directories(${CROSS_ROOTFS}/usr/lib/gcc/${TIZEN_TOOLCHAIN}/include/c++/)
     include_directories(${CROSS_ROOTFS}/usr/lib/gcc/${TIZEN_TOOLCHAIN}/include/c++/armv7l-tizen-linux-gnueabi)
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -167,6 +167,7 @@ fi
 
 __build_arch_lowcase=$(echo "$__build_arch" | tr '[:upper:]' '[:lower:]')
 __base_rid=$__rid_plat-$__build_arch_lowcase
+export __CrossToolChainTargetRID=$__base_rid
 
 # Set up the environment to be used for building with clang.
 if which "clang-3.5" > /dev/null 2>&1; then


hqueue:

cross/armel/toolchain.cmake
```diff
@@ -16,7 +16,7 @@ add_compile_options(--sysroot=${CROSS_ROOTFS})
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -target ${TOOLCHAIN}")
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} --sysroot=${CROSS_ROOTFS}")
 
-if("${CLI_CMAKE_RUNTIME_ID}" MATCHES "tizen.*")
```
gkhanna79: This is a CMake define - can you please help explain why it does not get propagated to the custom toolchain you point it to for Tizen build?

jyoungyun: When build with clang, toolchain.cmake will be called several times. The first two times are called with the arguments we passed, but after that, it will be called according to the rules set in clang itself. In this case, there is no defines like CLI_CMAKE_RUNTIME_ID in its own rule, so it can not set the desired toolchain value. For the same reason, coreclr / corefx is also using environment value. However, in the environment where gcc-arm-linux-gnueabi is installed, even if there is no toolchain value for Tizen, the build is done, so I uploaded the previous patch. However, the docker currently used by ci does not have the gcc library and is not normally built.

jyoungyun:
@dotnet-bot plz test OSX x64 Release Build
@dotnet-bot plz test Ubuntu x64 Release Build

cross/armel/toolchain.cmake
gkhanna79: In that case, should we simply ensure gcc-arm-linux-gnueabi is installed in the Docker image?

hqueue: @gkhanna79 I don't think using gcc-arm-linux-gnueabi is a right way, because it may mislead toolchain to use of installed gcc binaries (objects and libaries) instead of binaries in rootfs and it may cause runtime. Also options to clang, i.e. `-B` and `-L`, always have wrong value for Tizen.

In CoreCLR and CoreFX, they already use environmental variable in cmake files.
https://github.com/dotnet/coreclr/blob/master/cross/armel/toolchain.cmake#L25
https://github.com/dotnet/corefx/blob/master/cross/armel/toolchain.cmake#L18

Therefore I think using environmental variable here suggested by @jyoungyun is reasonable if there is no side-effect to other build process.

In addition, we can reduce size of Docker image for CI if we don't need gcc-arm-linux-gnueabi.

gkhanna79: Very well - in that case, I would suggest to export the variable under a different name, without change __base_rid and use that so that its intent (via its name) is explicit.

hqueue: Sure why not :)
@jyoungyun what do you think of it ?

jyoungyun: @gkhanna79 @hqueue In fact, I exported `__base_rid` because I wanted to use as much of the existing core-setup code as possible. I think it is better to use a variable declaration without `__` at export, but since it uses `__DistroRid` in CoreCLR or CoreFX, it would be nice to use this variable. What about it ?
CoreCLR:https://github.com/dotnet/coreclr/blob/master/build.sh#L77, CoreFX:https://github.com/dotnet/corefx/blob/master/src/Native/build-native.sh#L46

gkhanna79: DistroRid is a general purpose export. For this specific scenario, you probably want to name it more specifically - something like CrossToolChainTargetRID.

src/corehost/build.sh
```diff
@@ -166,7 +166,7 @@ if [ -z $__commit_hash ]; then
 fi
 
 __build_arch_lowcase=$(echo "$__build_arch" | tr '[:upper:]' '[:lower:]')
-__base_rid=$__rid_plat-$__build_arch_lowcase
+export __base_rid=$__rid_plat-$__build_arch_lowcase
```
jyoungyun: @gkhanna79 Do you want to redefine __base_rid to CrossToolChainTargetRID ? or Do you want to define CrossToolChainTargetRID again here ? I just want to export the DistroRid which is a general purpose variable and use it in toolchain.cmake file. I don't understand why CrossToolChainTargerRID should be redefined even though it is the same as __base_rid.

jyoungyun: @gkhanna79 Do I have to add the same value as `__base_rid`?

jyoungyun:
@dotnet-bot plz test OSX x64 Release Build
@dotnet-bot plz test Ubuntu x64 Release Build

src/corehost/build.sh
gkhanna79: Something like "Export __CrossToolChainTargetRID=$__base_rid" is what I would suggest.

jyoungyun:
@gkhanna79 Please take a look.

PR #1560, opened 2017-02-18
Treat negative test failures as success

gkhanna79:
Core-Setup repo has bunch of negative tests that are expected to fail. However, in the console output, they come across as failed, per the logging infrastructure, even though their failing implies passing. As a result, it is difficult to distinguish between a real failure and negative test failure.

This change will enable negative tests to pass a flag to the test infrastructure (by invoke the Execute method with true bool as argument) to indicate that they are negative tests, so that they can be marked as passed, if they fail!

@ramarag PTAL

CC @cakine

### DIFF
--- a/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
+++ b/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
@@ -74,7 +74,7 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.PortableApp
             dotnet.Exec("exec", "--runtimeconfig", runtimeConfig, appDll)
                 .CaptureStdErr()
                 .CaptureStdOut()
-                .Execute()
+                .Execute(fExpectedToFail:true)
                 .Should()
                 .Fail();
         }
@@ -195,7 +195,7 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.PortableApp
             dotnet.Exec("exec", "--depsfile", depsJson, appDll)
                 .CaptureStdErr()
                 .CaptureStdOut()
-                .Execute()
+                .Execute(fExpectedToFail:true)
                 .Should()
                 .Fail();
         }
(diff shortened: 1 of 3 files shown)


test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
```diff
@@ -74,7 +74,7 @@ public void Muxer_Exec_activation_of_Build_Output_Portable_DLL_with_DepsJson_Loc
             dotnet.Exec("exec", "--runtimeconfig", runtimeConfig, appDll)
                 .CaptureStdErr()
                 .CaptureStdOut()
-                .Execute()
+                .Execute(true)
```
ramarag: i would like these calls to use named parameters, i.e
```C#
.Execute(fExpectedToFail:true)
```
so that people reading the tests can understand the intent easily

ramarag: or still better would be to  use, the pattern followed in sdk and cli repo's fluent style

```C#
.Execute
.ShouldFail()
```

gkhanna79: Can you point me to ShouldFail in CLI? I don't see any results at https://github.com/dotnet/cli/search?utf8=%E2%9C%93&q=ShouldFail.

gkhanna79: Also, that should be equivalent to `.Should().Fail() `used in Core-Setup and if so, that does not help mark the test as passed for the negative tests.

ramarag: yes you are right, This can also be achieved by adding something like 

```C#
class Command
{
...
private bool fExpectedToFail=false;
public Command ExpectToFail()
        {
           fExpectedToFail  = true;
            return this;
        }
...
}
```
and use as 
```C#
CaptureStdOut().
ExpectToFail.
Execute()

```

PR #1568, opened 2017-02-21
Multilevel SDK Lookup

cakine:
.NETCore activation will look for the SDK version folder in the current working directory, the user location, the executable directory and the global .NET location, in the order described.

The user and global locations are defined as follows:

User location:
Windows 32-bit: %SystemDrive%\Users\username\.dotnet\x86
Windows 64-bit: %SystemDrive%\Users\username\.dotnet\x64
Unix 32-bit: /home/username/.dotnet/x86
Unix 64-bit: /home/username/.dotnet/x64

Global location:
Windows 32-bit: %SystemDrive%\Program Files\dotnet
Windows 64-bit (32-bit application): %SystemDrive%\Program Files (x86)\dotnet
Windows 64-bit (64-bit application): %SystemDrive%\Program Files\dotnet
Unix: the directory of “dotnet” defined in the system path.

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -233,6 +233,23 @@ pal::string_t get_deps_file(
 }
 
 /**
+* Multilevel Lookup is enabled by default
+*  It can be disabled by setting DOTNET_MULTILEVEL_LOOKUP env var to a value that is not 1
+*/
+bool multilevel_lookup_enabled()
+{
+    pal::string_t env_lookup;
+    bool multilevel_lookup = true;
+
+    if (pal::getenv(_X("DOTNET_MULTILEVEL_LOOKUP"), &env_lookup))
+    {
+        auto env_val = pal::xtoi(env_lookup.c_str());
+        multilevel_lookup = (env_val == 1);
+    }
+    return multilevel_lookup;
+}
+
+/**
  * Given own location, FX location, app binary and specified --depsfile and probe paths
  *     return location that is expected to contain hostpolicy
  */
@@ -375,19 +392,10 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t& 
     // If it is not activated, then only .exe directory will be considered
 
     std::vector<pal::string_t> hive_dir;
-    pal::string_t env_lookup;
-
-    // Multi-level SharedFX lookup can be disabled by setting DOTNET_MULTILEVEL_LOOKUP env var to zero
-    bool multilevel_lookup = true;
-    if (pal::getenv(_X("DOTNET_MULTILEVEL_LOOKUP"), &env_lookup))
-    {
-        auto env_val = pal::xtoi(env_lookup.c_str());
-        multilevel_lookup = (env_val != 0);
-    }
-
     pal::string_t cwd;
     pal::string_t local_dir;
     pal::string_t global_dir;
+    bool multilevel_lookup = multilevel_lookup_enabled();
 
     if (multilevel_lookup)
     {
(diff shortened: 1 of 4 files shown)


cakine:
@gkhanna79 @ramarag

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -582,36 +582,78 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, pal::stri
         trace::verbose(_X("Failed to obtain current working dir"));
     }
 
-    pal::string_t retval;
-    if (!global.empty())
+    std::vector<pal::string_t> hive_dir;
+    pal::string_t env_lookup;
+
+    // Multi-level SDK lookup can be disabled by setting DOTNET_MULTILEVEL_LOOKUP env var to zero
+    bool multilevel_lookup = true;
```
gkhanna79: This check is present in https://github.com/dotnet/core-setup/blob/c0aac4d98d3af034d8b9f8ca54ed49ec766db320/src/corehost/cli/fxr/fx_muxer.cpp#L380 as well. Can you please refactor this out to a single helper function that can be used in both the locations?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -582,36 +582,78 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, pal::stri
...
-            if (pal::directory_exists(sdk_path))
-            {
-                trace::verbose(_X("CLI directory [%s] from global.json exists"), sdk_path.c_str());
-                retval = sdk_path;
-            }
-            else
+    pal::string_t retval;
+    for (pal::string_t dir : hive_dir)
+    {
+        trace::verbose(_X("Searching SDK directory in [%s]"), dir.c_str());
```
gkhanna79: NIT: Extra blank line.

cakine: Fixed it.

src/corehost/cli/fxr/fx_muxer.cpp

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -233,6 +233,23 @@ pal::string_t get_deps_file(
...
+*/
+bool multilevel_lookup_enabled()
+{
+    pal::string_t env_lookup;
+    bool multilevel_lookup = true;
+
+    if (pal::getenv(_X("DOTNET_MULTILEVEL_LOOKUP"), &env_lookup))
+    {
+        auto env_val = pal::xtoi(env_lookup.c_str());
+        multilevel_lookup = (env_val != 0);
```
gkhanna79: Should this be ">0" to ensure negative values do not set it as well?

cakine: I can change that as well. I thought that any nonzero value would be enough to set it.

gkhanna79: I think we should make it explicit - set to 1 to enable it and not otherwise. This is the usual convention.

cakine: Updated it.

eerhardt:
> Unix: the directory of “dotnet” defined in the system path.

I consider this policy to be wrong.  Just because a command is in my path, doesn't mean that it should affect other commands.  `%ProgramFiles%` on Windows != `$PATH` on Unix.

gkhanna79:
> all of a sudden the app breaks because there isn't a dotnet on the $PATH.

I do not think the app should break if there is no dotnet on path - it should only result in us determining that there is no global location for dotnet installation.

This change is implemented at 
https://github.com/dotnet/core-setup/blob/master/src/corehost/common/pal.unix.cpp#L237 and was so since day 1 because there is no equivalent of PF folder on Unix to determine "global" locations. This is simply used as hint for it.

test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
```diff
@@ -0,0 +1,360 @@
...
+            // Add a dummy version in the exe dir
+            AddAvailableSdkVersions(_exeSdkBaseDir, "9999.0.0-dummy");
+
+            // Specified CLI version: none
+            // CWD: empty
+            // User: empty
+            // Exe: 9999.0.0
+            // Expected: 9999.0.0 from exe dir
+            dotnet.Exec("help")
+                .WorkingDirectory(_currentWorkingDir)
```
gkhanna79: Are there some negative tests we want to add @cakine?

cakine: @gkhanna79 I don't think so. The only possible negative test would be if all three folders were empty. That's troublesome to test because any version from the sdk in the global location would be picked, so the test might fail in some machines.

cakine:
I'll merge the changes tomorrow (Feb. 21st) if there are no more concerns.

gkhanna79:
Ubuntu Arm release failure is unrelated. Merging the PR.

PR #1569, opened 2017-02-22
Change dotnet-deb-tool to prefercliruntime.

eerhardt:
Also, fixing it to start publishing deb-tool to the cli-deps feed again.

We need deb-tool to prefercliruntime, so it can be ran on either 1.0 or 1.1 depending on what the CLI is running.  That way it can be invoked on distros that don't support 1.0.

@naamunds @gkhanna79 @wtgodbe @ravimeda

### DIFF
--- a/buildpipeline/Core-Setup-Ubuntu14.04-x64.json
+++ b/buildpipeline/Core-Setup-Ubuntu14.04-x64.json
@@ -101,7 +101,7 @@
       "value": "PassedViaPipeBuild"
     },
     "CLI_NUGET_FEED_URL": {
-      "value": "https://www.myget.org/F/core-setup-testing"
+      "value": "https://dotnet.myget.org/F/cli-deps"
     },
     "CLI_NUGET_API_KEY": {
       "value": "PassedViaPipeBuild"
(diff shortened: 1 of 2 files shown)


buildpipeline/Core-Setup-Ubuntu14.04-x64.json
```diff
@@ -101,7 +101,7 @@
       "value": "PassedViaPipeBuild"
     },
     "CLI_NUGET_FEED_URL": {
-      "value": "https://www.myget.org/F/core-setup-testing"
+      "value": "https://dotnet.myget.org/F/cli-deps"
```
naamunds: Should this change be made in Core-Setup-CrossBuild.json also?

eerhardt: That should be removed.  There is no reason why Core-Setup-CrossBuild.json has it.  We are only publishing the `deb-tool` from Ubuntu14.04.

eerhardt: Looking deeper into it, we are publishing it on "Ubuntu", version="14.04", which happens for both `x64` and `arm`.

We probably need a way to tell it only publish on `x64`.

I opened https://github.com/dotnet/core-setup/issues/1570 for this.

PR #1572, opened 2017-02-22
[ARM/CI]Enable ARM CI script for cross build

jyoungyun:
Related issue: #1512

PR #1512, opened 2017-02-09
[ARM/CI] Add ARM CI script for cross build

jyoungyun:
Implement ARM CI using Docker for arm and armel
Call the script in netci.groovy

Related issue : 
https://github.com/dotnet/core-setup/issues/790
https://github.com/dotnet/core-setup/issues/1400
https://github.com/dotnet/coreclr/issues/9273
https://github.com/dotnet/corefx/pull/15900

PR #1576, opened 2017-02-22
Fix build break by only publishing dotnet-deb-tool for ubuntu.14.04-x64.

### LINKED ISSUE #1570: Only publish dotnet-deb-tool for Ubuntu.14.04-x64 build
See the conversation here:

https://github.com/dotnet/core-setup/pull/1569/files/f2002f2a9eb1755ca5af0adc7e6cac305060590d#r102353639

We currently publish the deb-tool for both x64 and arm builds.  We only need to publish it once.  When we fix this, we can remove `CLI_NUGET_FEED_URL` variable from the Core-Setup-CrossBuild.json pipe build definition.

eerhardt:
Fix #1570

@naamunds @dagood @gkhanna79

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -60,6 +60,7 @@ namespace Microsoft.DotNet.Host.Build
 
         [Target]
         [BuildPlatforms(BuildPlatform.Ubuntu, "14.04")]
+        [BuildArchitectures(BuildArchitecture.x64)]
         public static BuildTargetResult PublishDotnetDebToolPackage(BuildTargetContext c)
         {
             string nugetFeedUrl = EnvVars.EnsureVariable("CLI_NUGET_FEED_URL");
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -242,12 +242,6 @@
     "GITHUB_PASSWORD": {
       "value": "PassedViaPipeBuild"
     },
-    "CLI_NUGET_FEED_URL": {
-      "value": "https://www.myget.org/F/core-setup-testing"
-    },
-    "CLI_NUGET_API_KEY": {
-      "value": "PassedViaPipeBuild"
-    },
     "PB_RepoName": {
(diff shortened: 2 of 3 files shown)


eerhardt:
test Ubuntu arm Release Build

eerhardt:
FYI - I also fixed the package version so it doesn't contain "-beta-beta-001234".  See https://dotnet.myget.org/feed/cli-deps/package/nuget/dotnet-deb-tool.

eerhardt:
test Ubuntu arm Release Build

eerhardt:
@jyoungyun @gkhanna79 - do you know why the Ubuntu arm build is failing?  Is it becuase of #1512?

I'd like to get my change in since the official builds are currently broken.

jyoungyun:
@eerhardt This issue is not related to your patch. This failure issue will be resolved after merging #1572.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -62,9 +62,13 @@ public static BuildTargetResult Publish(BuildTargetContext c)
         [BuildPlatforms(BuildPlatform.Ubuntu, "14.04")]
```
dagood: Can you add `BuildArchitecturesAttribute` rather than doing string matching? Or would the "current" platform still be x64 for the crossbuild, and only the target RID has the needed value?

eerhardt: Good call.  It looks like that works.  BuildArchitectures uses this code:

https://github.com/dotnet/core-setup/blob/1dfa45251d6706c046dff384dc81e76161f16440/build_projects/Microsoft.DotNet.Cli.Build.Framework/CurrentArchitecture.cs

Which checks the TARGETPLATFORM environment variable.

We use the same pattern on the GenerateDebs target...

eerhardt:
@dagood - I changed the condition as suggested.

build_projects/dotnet-host-build/PublishTargets.cs
dagood: Thanks, didn't notice that env var check or see the other usage, makes sense.

PR #1582, opened 2017-02-24
Use updated CrossBuild Docker Images

gkhanna79:
This switches Core-Setup to use the new Docker images that have RootFS built in.

@MichaelSimons @wtgodbe PTAL.


gkhanna79:
Confirmed in https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=586397&_a=summary that the change works as expected.

PR #1586, opened 2017-02-24
Remove the Roslyn VB support from shared framework

weshaggard:
VB support for Roslyn is a pretty large assembly which we don't
need for the mainline scenarios so we are removing from the
shared framework.

cc @ericstj @jaredpar @Petermarcu

### DIFF
--- a/TestAssets/TestProjects/PortableApp/project.json
+++ b/TestAssets/TestProjects/PortableApp/project.json
@@ -28,10 +28,6 @@
           "version": "1.3.0",
           "exclude": "all"
         },
-        "Microsoft.CodeAnalysis.VisualBasic": {
-          "version": "1.3.0",
-          "exclude": "all"
-        },
         "Microsoft.CSharp": {
           "version": "4.3.0",
           "exclude": "all"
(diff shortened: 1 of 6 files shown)


pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
```diff
@@ -5,10 +5,6 @@
       <Version>1.3.0</Version>
       <Exclude>Compile</Exclude>
     </NETCoreApp10Dependency>
-    <NETCoreApp10Dependency Include="Microsoft.CodeAnalysis.VisualBasic">
```
weshaggard: @ericstj I removed these for netcoreapp10/11 to have the conversation but I can imagine we keep these for back compat.

ericstj: What are you trying to accomplish by removing them from downlevel TFMs?

weshaggard: Nothing. As we chatted I'm going to keep those dependencies for the older target frameworks.

Issue #283, opened 2016-09-16
Transition Core-Setup build to MSBuild/BuildTools


gkhanna79:
CC @schellap

dagood:
Just to explicitly link the issues together, this is required for https://github.com/dotnet/core-setup/issues/253 (better auto-PR), https://github.com/dotnet/core-setup/issues/238 (automatically create new metapackages), and https://github.com/dotnet/core-setup/issues/259 (multiple builds per commit).

gkhanna79:
Assigning to @AlfredoMS who is working on this for build-from-source.

AlfredoMS:
My dev branch: My dev branch: https://github.com/AlfredoMS/core-setup/tree/buildtools

Work so far:
- Product build is almost complete. I still need to add crossgening

To be done:
- Packaging
- Publishing

Also I need to rebase to include the changes made by @schellap on https://github.com/dotnet/core-setup/pull/711

dagood:
@AlfredoMS Do you plan on maintaining the current CLI-style publishing logic, or changing to the way corefx/coreclr/wcf do it?

gkhanna79:
CC @wtgodbe @ramarag

AlfredoMS:
@gkhanna79 the idea is to replace all the CLI style with buildtools 

cc @ellismg

AlfredoMS:
I haven't started the work on publishing, but if there are any particular concerns please let me know soon.

AlfredoMS:
Update:

New dev branch: https://github.com/AlfredoMS/core-setup/tree/build-eng 

I expect this to build and package on all OS in the next couple of days. After that I will work on the publishing step.

AlfredoMS:
New public branch https://github.com/dotnet/core-setup/tree/buildtools
Work to be completed on this branch before merging into master.

@ellismg

AlfredoMS:
**How to use new branch**

Windows:
- run build.cmd
Mac and Linux
- run build.sh -- /p:TargetRid= [your desired target rid, e.g. osx.10.10-x64  or ubuntu.16.04-x64

you can use -verbose for msbuild verbose logging and -SkipTests=true to skip test that are broken*

To run the build step by step you can use this commands just add .sh or .cmd according to your os

sync
run build-custom -verbose -- src/corehost/build.proj
run build-custom -verbose -- build/tasks/core-setup.tasks.builds
run build-custom -verbose -- pkg/dir.proj
[...]

PR #1587, opened 2017-02-24
Moving core-setup to be msbuild based using buildtools

AlfredoMS:
**Updated 3/2**

This is a "beta" version for everyone to use and test.
Branch was rebased using master on 3/2, latest changes may not be included

Issue: https://github.com/dotnet/core-setup/issues/283

This change replaces the cli/c# based build with an msbuild/buildtools based system to be aligned with the other dotnet code repositories.

How to use:
1. clone it from this pr or from https://github.com/AlfredoMS/core-setup/tree/build-eng

Windows:
- run build.cmd
Mac and Linux
- run build.sh -- /p:TargetRid= [your desired target rid, e.g. osx.10.10-x64  or ubuntu.16.04-x64

you can use -verbose for msbuild verbose logging and -SkipTests=true to skip test that are broken*

To run the build step by step you can use this commands just add .sh or .cmd according to your os

sync
run build-custom -verbose -- src/corehost/build.proj
run build-custom -verbose -- build/tasks/core-setup.tasks.builds
run build-custom -verbose -- pkg/dir.proj
run build-custom -verbose -- src/dir.proj
run build-custom -verbose -- test/dir.proj
run build-custom -verbose -- packaging/dir.proj

for mac and linux specify the os and TargetRid
example:
run build-custom -verbose -OSGroup=OSX -- src/corehost/build.proj /p:TargetRid=osx.10.10-x64
run build-custom -verbose -OSGroup=Linux -- src/corehost/build.proj /p:TargetRid=ubuntu.16.04-x64

If you are planning on making new changes you should get familiar with the new build and bin structure.
Known issues:
- Publishing is not ready
- Tests are broken. I have a workaround but it still depends on the old build system. Use flag "-SkipTest=true"
- Build.sh needs improvements to generated targetrid or accept one as parameter
- checked in build definitions for jenkins and vsts have not been updated
- upgrade version tool has not been updated.
- generated packages will have the version suffix 0009999, this is intentional to help identify built binaries while testing the branch, will be changed for the final version.
------
@gkhanna79 @ellismg  @weshaggard @ericstj @wtgodbe @eerhardt @dagood @jyoungyun


.gitignore
```diff
@@ -1,5 +1,12 @@
 ### Repo-specific things ###
 
+# Tool Runtime Dir
+/[Tt]ools/
+
+#pkg
+pkg/version.txt
```
weshaggard: We should be generating these version files under intermediate now, so somewhere under "bin\obj"

build.proj
```diff
@@ -0,0 +1,80 @@
...
+  <Import Project="dir.props" />
+
+  <!-- required to build the projects in their specified order -->
+  <PropertyGroup>
+    <SerializeProjects>true</SerializeProjects>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <!-- To disable the restoration of packages, set RestoreDuringBuild=false or pass /p:RestoreDuringBuild=false.-->
+    <RestoreDuringBuild Condition="'$(RestoreDuringBuild)'==''">true</RestoreDuringBuild>
```
weshaggard: We will want the default for this to be false when we are done because we want to be able to build offline.

build.sh
```diff
@@ -16,42 +16,44 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
-}
-
-# Check if we need to build in docker
-if [ ! -z "$BUILD_IN_DOCKER" ]; then
-    dockerbuild "${args[@]}"
-else
-    $DIR/build_projects/dotnet-host-build/build.sh "${args[@]}"
-fi
+# Use uname to determine what the OS is.
+OSName=$(uname -s)
```
weshaggard: We should let run.exe handle these and the RID in the config.json like we do in corefx.

weshaggard: Sorry I miss-spoke the RID is computed here https://github.com/dotnet/corefx/blob/master/src/Native/build-native.sh#L24. The OSName however is computed in config.json.

dependencyversions.props
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
weshaggard: For consistency I would call this dependencies.props. It also looks like you have configured the VersionTools stuff for updating yet.

dir.props
```diff
@@ -0,0 +1,307 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
+
+  <Import Project="dependencyversions.props" />
+
+  <!-- Branch Info -->
+  <PropertyGroup>
```
weshaggard: Good. Does this mean you can delete the BranchInfo.txt file now?

dir.props
```diff
@@ -0,0 +1,307 @@
...
+    <ToolsDir Condition="'$(UseToolRuntimeForToolsDir)'=='true'">$(ToolRuntimePath)</ToolsDir>
+    <ToolsDir Condition="'$(ToolsDir)'==''">$(ProjectDir)Tools/</ToolsDir>
+    <BuildToolsTaskDir Condition="'$(BuildToolsTargets45)' == 'true' OR '$(MSBuildRuntimeType)' != 'core'">$(ToolsDir)net45/</BuildToolsTaskDir>
+    <PackagingTaskDir Condition="'$(BuildToolsTaskDir)' != ''">$(BuildToolsTaskDir)</PackagingTaskDir>
+    <DotnetCliPath Condition="'$(DotnetCliPath)'==''">$(ToolsDir)dotnetcli/</DotnetCliPath>
+  </PropertyGroup>
+
+  <!-- Setting up Versioning -->
+  <PropertyGroup>
+    <VersionComparisonDate>2016-01-01</VersionComparisonDate>
```
weshaggard: Doe we actually need a custom date or will the default date work?

dir.props
```diff
@@ -0,0 +1,307 @@
...
+    <NuGetPackageSource>@(NuGetSourceList -> '-source %(Identity)', ' ')</NuGetPackageSource>
+    <NuGetConfigCommandLine>$(NuGetPackageSource)</NuGetConfigCommandLine>
+
+    <NugetRestoreCommand>"$(NuGetToolPath)"</NugetRestoreCommand>
+    <NugetRestoreCommand>$(NugetRestoreCommand) install</NugetRestoreCommand>
+    <!-- NuGet.exe doesn't like trailing slashes in the output directory argument -->
+    <NugetRestoreCommand>$(NugetRestoreCommand) -OutputDirectory "$(PackagesDir.TrimEnd('/\'.ToCharArray()))"</NugetRestoreCommand>
+    <NugetRestoreCommand>$(NugetRestoreCommand) $(NuGetConfigCommandLine)</NugetRestoreCommand>
+    <NugetRestoreCommand>$(NugetRestoreCommand) -Verbosity detailed</NugetRestoreCommand>
+    <NugetRestoreCommand Condition="'$(OsEnvironment)'=='Unix'">mono $(NuGetRestoreCommand)</NugetRestoreCommand>
```
weshaggard: Do we need both NugetRestore and DnuRestore command properties? It would be good to only have the NugetRestore if we can eliminate the other.

dir.props
```diff
@@ -0,0 +1,307 @@
...
+    <DnuRestoreSource>@(DnuSourceList -> '--source %(Identity)', ' ')</DnuRestoreSource>
+
+    <DnuRestoreCommand>"$(DotnetToolCommand)"</DnuRestoreCommand>
+    <DnuRestoreCommand>$(DnuRestoreCommand) restore</DnuRestoreCommand>
+    <DnuRestoreCommand>$(DnuRestoreCommand) --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))" $(DnuRestoreSource)</DnuRestoreCommand>
+    <DnuRestoreCommand Condition="'$(LockDependencies)' == 'true'">$(DnuRestoreCommand) --lock</DnuRestoreCommand>
+  </PropertyGroup>
+
+  <!-- Create a collection of all project.json files for dependency updates. -->
+  <ItemGroup>
```
weshaggard: I would put this in dependencies.props.

dir.targets
```diff
@@ -0,0 +1,94 @@
...
+    <!-- We do not want to target a portable profile.
+         TODO: Make this the default in buildtools so this is not necessary. -->
+    <TargetFrameworkProfile></TargetFrameworkProfile>
+    <!-- We set this property to avoid MSBuild errors regarding not setting TargetFrameworkProfile (see above line) -->
+    <PortableNuGetMode>true</PortableNuGetMode>
+  </PropertyGroup>
+
+  <Import Project="$(ToolsDir)/Build.Common.targets" Condition="'$(UseLiveBuildTools)' != 'true'" />
+
+  <Target Name="GetLatestCommitHash">
```
weshaggard: Why did you need to override these from BuildTools?

src/corehost/build.proj
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
weshaggard: We are going to need to be able to build the host without msbuild in the picture. Looks like the majority of what you are doing here is creating a command line and generating the versioning stuff, we may need to figure out a way to do that purely from the script files.

build.sh
```diff
@@ -16,42 +16,44 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
-done
-
-# $args array may have empty elements in it.
-# The easiest way to remove them is to cast to string and back to array.
-temp="${args[@]}"
-args=($temp)
-
-dockerbuild()
-{
-    BUILD_COMMAND=/opt/code/build_projects/dotnet-host-build/build.sh $DIR/scripts/dockerrun.sh --non-interactive "$@"
```
weshaggard: Shouldn't we be able to delete the build_projects directory as part of this work?

src/dir.proj
```diff
@@ -0,0 +1,292 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="dir.props" />
+  <Import Project="..\dir.targets" />
+
+  <PropertyGroup>
+    <BuildTargets>
+      BuildProjectsForNuGetPackages;
+      PublishSharedFrameworkAndSharedHost;
```
weshaggard: Looks like most of this is about the sharedframework why not put it under the sharedframework path.

dir.props
AlfredoMS: yes!

test/HostActivationTests/project.json
```diff
@@ -3,7 +3,7 @@
   "dependencies": {
     "Microsoft.NETCore.App": {
       "type": "platform",
-      "version": "1.0.0"
+      "version": "1.1.0"
```
eerhardt: This was already existing, but we are depending on Microsoft.NETCore.App version `1.1.0`, but using TFM "netcoreapp2.0".  This seems incorrect.

eerhardt:
>   "version": "1.0.0-*",

Shouldn't the tests be updated to msbuild as well?

---
Refers to: test/HostActivationTests/project.json:2 in c009f5f. [](commit_id = c009f5f6cf967724baa5375beea86382fdf7b42d, deletion_comment = False)

test/TestUtils/RepoDirectoriesProvider.cs
```diff
@@ -28,28 +29,40 @@ public class RepoDirectoriesProvider
...
-            string corehostDummyPackages = null)
+            string corehostDummyPackages = null,
+            string dotnetSdk = null)
         {
-            _repoRoot = repoRoot ?? Path.Combine(Directory.GetCurrentDirectory(), "..", "..");
-            
-            string baseArtifactsFolder = artifacts ?? Path.Combine(_repoRoot, "artifacts");
+            _repoRoot = repoRoot ?? Directory.GetParent((Directory.GetCurrentDirectory())).Parent.FullName;
+
+            string baseArtifactsFolder = artifacts ?? Path.Combine(_repoRoot, "bin");
```
eerhardt: Should `Artifacts` be renamed here and everywhere now that this folder is no longer named "artifacts"?

test/TestUtils/RepoDirectoriesProvider.cs
```diff
@@ -28,28 +29,40 @@ public class RepoDirectoriesProvider
...
+
+            _dotnetSDK = dotnetSdk ?? Path.Combine(_repoRoot, "Tools", "dotnetcli");
 
             _targetRID = Environment.GetEnvironmentVariable("TEST_TARGETRID");
 
-            _artifacts = Path.Combine(baseArtifactsFolder, _targetRID);
+            _artifacts = Path.Combine(baseArtifactsFolder, _targetRID+".Debug");
+            if(!Directory.Exists(_artifacts))
```
eerhardt: I'm not sure I like this pattern.  If I build for "Debug" on my machine, the Debug folder is going to exist.  If I then want to build/test Release builds, I need to remember to do a "clean", or else all these places are going to pick up the wrong directory.

src/dir.props
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
+
+  <PropertyGroup>
+    <CoreHostOutputDir>$(BaseOutputRootPath)corehost\</CoreHostOutputDir>
+    <CoreHostLockedDir>$(CoreHostOutputDir)locked\</CoreHostLockedDir>
+    <SharedFrameworkSourceRoot>$(IntermediateOutputRootPath)sharedFramework\framework</SharedFrameworkSourceRoot>
```
eerhardt: Are these really global properties that need to be defined in all projects?  They don't seem like they should be global.

dir.props
```diff
@@ -0,0 +1,307 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
```
eerhardt: In the CLI, we've been following a pattern of splitting these properties up into smaller .props files.  That way "dir.props" doesn't become a dumping ground of a bunch of unrelated properties.

How do you feel about following that pattern here?

dir.props
```diff
@@ -0,0 +1,307 @@
...
+  -->
+  <PropertyGroup>
+    <!-- Temp change to make OS X build behave as a Linux build -->
+    <OsEnvironment Condition="'$(OsEnvironment)'=='' AND '$(OS)'=='OSX'">Unix</OsEnvironment>
+    <OsEnvironment Condition="'$(OsEnvironment)'==''">$(OS)</OsEnvironment>
+  </PropertyGroup>
+
+  <!-- Common repo directories -->
+  <PropertyGroup>
+    <ProjectDir>$(MSBuildThisFileDirectory)</ProjectDir>
```
eerhardt: I think `RepoRoot` is a better name for this property.  Using a property named `ProjectDir` in a .csproj is confusing because the "project" is the .csproj.

dir.props
```diff
@@ -0,0 +1,307 @@
...
+
+  <!-- list of nuget package sources passed to nuget.exe -->
+  <ItemGroup>
+    <NuGetSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-buildtools" />
+    <NuGetSourceList Include="https:%2F%2Fwww.nuget.org/api/v2" />
+  </ItemGroup>
+
+  <!-- Common nuget properties -->
+  <PropertyGroup>
+    <NuGetToolPath Condition="'$(NuGetToolPath)'==''">$(ToolsDir)NuGet.CommandLine/NuGet.exe</NuGetToolPath>
```
eerhardt: How does using `NuGet.exe` work on non-Windows?  Why not use `dotnet restore` instead - which works on all platforms?

dir.props
eerhardt: We are requiring mono to build core-setup now?  That seems wrong.

dir.props
```diff
@@ -0,0 +1,307 @@
...
+    <NugetRestoreCommand>$(NugetRestoreCommand) -OutputDirectory "$(PackagesDir.TrimEnd('/\'.ToCharArray()))"</NugetRestoreCommand>
+    <NugetRestoreCommand>$(NugetRestoreCommand) $(NuGetConfigCommandLine)</NugetRestoreCommand>
+    <NugetRestoreCommand>$(NugetRestoreCommand) -Verbosity detailed</NugetRestoreCommand>
+    <NugetRestoreCommand Condition="'$(OsEnvironment)'=='Unix'">mono $(NuGetRestoreCommand)</NugetRestoreCommand>
+  </PropertyGroup>
+
+  <!-- list of nuget package sources passed to dnu -->
+  <ItemGroup>
+    <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
+    <DnuSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
eerhardt: Can we rename all these `Dnu` variables?  That name has been out of date for over a year now.

dir.props
```diff
@@ -0,0 +1,307 @@
...
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <ConfigurationErrorMsg>$(ConfigurationErrorMsg);Unknown ConfigurationGroup [$(ConfigurationGroup)] specificed in your project.</ConfigurationErrorMsg>
+      </PropertyGroup>
+    </Otherwise>
+  </Choose>
+
+  <PropertyGroup>
+    <ExeSuffix Condition="'$(OSGroup)'=='Windows_NT'">.exe</ExeSuffix>
```
eerhardt: Move this up in the file, and then the DotnetToolCommand properties can take advantage of it:

```
    <DotnetToolCommand Condition="'$(DotnetToolCommand)'=='' and '$(OsEnvironment)'!='Unix'">$(DotnetCliPath)dotnet.exe</DotnetToolCommand>
    <DotnetToolCommand Condition="'$(DotnetToolCommand)'=='' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotnetToolCommand>
```

src/dir.proj
```diff
@@ -0,0 +1,292 @@
...
+             Directories="$(IntermediateOutputRootPath)sharedFramework;$(SharedFrameworkSourceRoot)" />
+
+    <ItemGroup>
+      <SharedFrameworkFiles Include="sharedFramework\**\*.cs" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(SharedFrameworkFiles)" DestinationFolder="$(SharedFrameworkSourceRoot)"/>
+
+    <ItemGroup>
+      <JsonLines Include='{' />
```
eerhardt: Why are we generating this file?  We already have a project.json.template file:

[https://github.com/dotnet/core-setup/blob/master/src/sharedframework/framework/project.json.template](https://github.com/dotnet/core-setup/blob/master/src/sharedframework/framework/project.json.template)

Let's just use that, and replace the 3 variables in it, instead of generating this whole file in MSBuild code.

eerhardt: Even better - why don't we move sharedframework/framework to MSBuild all up, and get rid of project.json?  Then we can just use MSBuild variables, and no need to do any replacements.

---
In reply to: [103266303](https://github.com/dotnet/core-setup/pull/1587#discussion_r103266303) [](ancestors = 103266303)

src/dir.proj
```diff
@@ -0,0 +1,292 @@
...
+      <RuntimeGraphGeneratorExe>$(runtimeGraphGeneratorOutput)/$(runtimeGraphGeneratorName)$(ExeSuffix)</RuntimeGraphGeneratorExe>
+      <DepsJson>$(SharedFrameworkNameAndVersionRoot)\$(SharedFrameworkName).deps.json</DepsJson>
+
+    </PropertyGroup>
+
+    <Exec Command="$(RuntimeGraphGeneratorExe) --project $(SharedFrameworkSourceRoot) --deps $(DepsJson) $(runtimeGraphGeneratorRuntime)"
+          EnvironmentVariables="NUGET_PACKAGES=$(PackagesDir)" />
+  </Target>
+
+  <Target Name="RestoreLockedCoreHost">
```
eerhardt: Same question here:  Why not use a .csproj for this file instead of using old/deprecated project.json?

run.cmd
```diff
@@ -0,0 +1,16 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+if not defined VisualStudioVersion (
+  if defined VS140COMNTOOLS (
+    call "%VS140COMNTOOLS%\VsDevCmd.bat"
+    goto :Run
+  )
+  echo Error: Visual Studio 2015 required.
```
eerhardt: What is the plan for moving this to VS 2017?  The dotnet/sdk repo already depends on VS 2017 being installed.

run.ps1
```diff
@@ -0,0 +1,42 @@
...
+# now execute it
+& $bootStrapperPath (Get-Location) $toolsLocalPath -DotNetInstallBranch "rel/1.0.0-preview2.1" | Out-File (Join-Path (Get-Location) "bootstrap.log")
+if ($LastExitCode -ne 0)
+{
+    Write-Output "Boot-strapping failed with exit code $LastExitCode, see bootstrap.log for more information."
+    exit $LastExitCode
+}
+
+# execute the tool using the dotnet.exe host
+$dotNetExe = Join-Path $toolsLocalPath "dotnetcli\shared\Microsoft.NETCore.App\version\dotnet.exe"
```
eerhardt: You shouldn't need to path into the shared framework to find a `dotnet.exe` to invoke.  You should just use the one that is "dotnetcli\dotnet.exe"

run.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# Use in the the functions: eval $invocation
+invocation='echo "Calling: ${FUNCNAME[0]}"'
```
eerhardt: All checked-in shell scripts should have `set -e` as their first line, so errors are handled correctly.

build.sh
```diff
@@ -16,42 +16,44 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
+        ;;
+
+    *)
+        echo "Unsupported OS $OSName detected, configuring as if for Linux"
+        __BuildOS=Linux
+        __HostOS=Linux
+        ;;
+esac
+
+$DIR/run.sh build -OSGroup=$__HostOS $@
```
eerhardt: How are we going to support building in Docker with run.sh build?  Our official builds use Docker to build on each of the platforms we target.

build.proj
```diff
@@ -0,0 +1,80 @@
...
+      BatchRestorePackages;
+      ValidateExactRestore;
+      CreateOrUpdateCurrentVersionFile;
+      BuildCustomTasks;
+      $(TraversalBuildDependsOn);
+    </TraversalBuildDependsOn>
+  </PropertyGroup>
+
+  <Target Name="BuildCustomTasks">
+      <MSBuild Projects="tasks/core-setup.tasks.builds"/>
```
eerhardt: I think a top-level directory named "tasks" is a poor name for this directory.  It feels like all our "build" specific stuff should be under a folder named "build".  So I can imagine this going under `build\tasks\`

src/corehost/build.cmd
```diff
@@ -0,0 +1,160 @@
...
+set __CMakeBinDir=""
+set __IntermediatesDir=""
+set __BuildArch=x64
+set __appContainer=""
+set __VCBuildArch=x86_amd64
+set CMAKE_BUILD_TYPE=Debug
+set "__LinkArgs= "
+set "__LinkLibraries= "
+
+:Arg_Loop
```
eerhardt: This is a lot of logic in a .cmd file.  Can we port this to powershell instead, so our build logic doesn't need to have another language?

eerhardt: Where is the corresponding `build.sh` file?

src/corehost/build.proj
```diff
@@ -0,0 +1,65 @@
...
+      <CMakeOutput Include="$(MSBuildProjectDirectory)\cli\exe\apphost\apphost" />
+      <CMakeOutput Include="$(MSBuildProjectDirectory)\cli\dll\$(HostPolicyBaseName)" />
+      <CMakeOutput Include="$(MSBuildProjectDirectory)\cli\fxr\$(DotnetHostFxrBaseName)" />      
+    </ItemGroup>
+    
+    <Copy SourceFiles="@(CMakeOutput)"
+          DestinationFolder="$(CoreHostOutputDir)"/>
+  </Target>
+
+  <Target Name="BuildeCoreHostWindows"
```
eerhardt: (nit) `BuildeCore`

src/corehost/build.proj
```diff
@@ -0,0 +1,65 @@
...
+    <!-- Run script that invokes Cmake to create VS files, and then calls msbuild to compile them -->
+    <Message Text="$(MSBuildProjectDirectory)\build.cmd $(BuildArgs)" Importance="High"/>
+    <Exec Command="$(MSBuildProjectDirectory)\build.cmd $(BuildArgs)" />
+
+    <ItemGroup>
+      <CMakeOutput Include="$(IntermediateOutputRootPath)corehost\cli\exe\dotnet\$(ConfigurationGroup)\dotnet.exe" />
+      <CMakeOutput Include="$(IntermediateOutputRootPath)corehost\cli\exe\dotnet\$(ConfigurationGroup)\dotnet.pdb" />
+      <CMakeOutput Include="$(IntermediateOutputRootPath)corehost\cli\exe\apphost\$(ConfigurationGroup)\apphost.exe" />
+      <CMakeOutput Include="$(IntermediateOutputRootPath)corehost\cli\exe\apphost\$(ConfigurationGroup)\apphost.pdb" />
+      <CMakeOutput Include="$(IntermediateOutputRootPath)corehost\cli\dll\$(ConfigurationGroup)\hostpolicy.dll" />
```
eerhardt: Why not use $(HostPolicyBaseName) and $(DotnetHostFxrBaseName) like we are in the non-windows build above?

pkg/projects/Microsoft.NETCore.DotNetHost/win/Microsoft.NETCore.DotNetAppHost.pkgproj
```diff
@@ -21,7 +21,7 @@
     <File Include="@(ArchitectureSpecificNativeFile)">
       <TargetPath>runtimes/$(PackageTargetRuntime)/native</TargetPath>
     </File>
-    <File Include="$(ProjectDir)/version.txt" />
+    <File Include="$(ProjectDir)pkg/version.txt" />
```
eerhardt: Can we make all these use `$(MSBuildProjectDirectory)` instead?  That way we don't hard-code `pkg` into all these proj files, and the allow them to be moved easier.

pkg/dir.proj
```diff
@@ -0,0 +1,39 @@
...
+        <BuildParams Condition="'$(BuildParams)' == ''">TargetsLinux=true;DistroRid=$(PackageTargetRid)</BuildParams>
+    </PropertyGroup>
+
+    <MSBuild Projects="projects/packages.builds" Properties="$(BuildParams)"/>
+
+    <ItemGroup>
+        <Nupkgs Include="$(PackageOutputPath)\**\*.nupkg" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(Nupkgs)"
```
eerhardt: Why don't the nupkgs get put in the `$(BaseOutputRootPath)corehost` in the first place?  Why have to copy them here?

dir.targets
```diff
@@ -0,0 +1,94 @@
...
+      <Output TaskParameter="ConsoleOutput" PropertyName="LatestCommit" />
+      <Output TaskParameter="ExitCode" PropertyName="LatestCommitExitCode" />
+    </Exec>
+    <!-- We shouldn't fail the build if we can't retreive the commit hash, so in this case just set it to N/A -->
+    <PropertyGroup Condition="'$(LatestCommitExitCode)'!='0'">
+      <LatestCommit>N/A</LatestCommit>
+    </PropertyGroup>
+  </Target>
+
+  <!-- Override Target rom buildtools -->
```
eerhardt: (nit) `rom`

dir.targets
```diff
@@ -0,0 +1,94 @@
...
+  <Target Name="GenerateVersionHeader"
+      Inputs="$(MSBuildProjectFile)"
+      Outputs="$(NativeVersionHeaderFile)"
+      DependsOnTargets="CreateVersionFileDuringBuild;GetLatestCommitHash"
```
eerhardt: Where does `CreateVersionFileDuringBuild` get defined?

dir.targets
```diff
@@ -0,0 +1,94 @@
...
+      Condition="'$(NativeVersionHeaderFile)'!='' and '$(GenerateVersionHeader)'=='true' and !Exists($(NativeVersionHeaderFile))">
+
+    <ItemGroup>
+      <!-- Defining versioning variables -->
+      <NativeVersionLines Include="#include &lt;Windows.h&gt;" />
```
eerhardt: Can this be moved into `src\corehost`?  It is only used by the native code in there, right?

packaging/windows/GetWix.ps1
```diff
@@ -0,0 +1,28 @@
+﻿Param(
+    [string]$WixVersion="3.10.2",
```
eerhardt: Why default this version when the caller always passes it in?  The version is duplicated here and in the props file.

packaging/dir.proj
```diff
@@ -0,0 +1,154 @@
...
+    <Exec Condition="'$(OSGroup)' != 'Windows_NT'"
+          Command='find %(OutDirs.Identity) -type f -name "*.dylib" -exec chmod 755 {} \;' />
+    <Exec Condition="'$(OSGroup)' != 'Windows_NT'"
+          Command='find %(OutDirs.Identity) -type f -name "*.so" -exec chmod 755 {} \;' />  
+    <!-- Executables (those without dots) are executable -->
+    <Exec Condition="'$(OSGroup)' != 'Windows_NT'"
+          Command='find %(OutDirs.Identity) -type f ! -name "*.*" -exec chmod 755 {} \;' />   
+    
+    
+    <!-- Copy al to combined directory -->
```
eerhardt: (nit) `al`

packaging/windows/dir.proj
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Package" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
eerhardt: Is this file used?  The osx and deb folders don't have a similar file.

src/dir.proj
```diff
@@ -0,0 +1,292 @@
...
+  </PropertyGroup>
+
+  <Target Name="Build" DependsOnTargets="$(BuildTargets)" />
+
+  <UsingTask TaskName="GetBuildArgsByFrameworks" AssemblyFile="$(BuildToolsTaskDir)core-setup.tasks.dll"/>
+
+  <Target Name="BuildProjectsForNuGetPackages">
+    
+    <PropertyGroup>
+      <pOutDir>$(IntermediateOutputRootPath)forPackaging</pOutDir>
```
eerhardt: Can we move this into a shared property since it is defined in 2 spots?  (here and when we pack the nupkgs)

src/dir.proj
```diff
@@ -0,0 +1,292 @@
...
+      <PackageProjects Include="Microsoft.Extensions.DependencyModel/project.json" />
+    </ItemGroup>
+
+    <GetBuildArgsByFrameworks
+      OSGroup="$(OSGroup)"
+      ProjectJsonPaths="@(PackageProjects)">
+      <Output ItemName="buildCmdArgs" TaskParameter="BuildArgs" />
+    </GetBuildArgsByFrameworks>
+
+    <Exec Command="$(DotnetToolCommand) build $(Args) %(buildCmdArgs.Identity)" />
```
eerhardt: I spoke with Peter P, and he mentioned that the reason we are doing this build and pack separately is for signing.  How do these assemblies get signed correctly?

src/dir.proj
```diff
@@ -0,0 +1,292 @@
...
+      <ExcludedNative Include="$(SharedFrameworkNameAndVersionRoot)\Microsoft.DiaSymReader*" />
+
+      <PlatformAssemblies Include="$(SharedFrameworkNameAndVersionRoot)\*.dll"
+                          Exclude="@(ExcludedLibraries->'$(SharedFrameworkNameAndVersionRoot)\%(Identity)');@(ExcludedNative)" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <PlatformAssembliesPaths>$(CoreLibsDir);$(SharedFrameworkNameAndVersionRoot)</PlatformAssembliesPaths>
+      <CrossgenArgs>-platform_assemblies_paths $(PlatformAssembliesPaths)</CrossgenArgs>
+      <CrossgenArgs>$(CrossgenArgs) -JITPath $(JitPath)</CrossgenArgs>
```
eerhardt: Why is this split into 2 lines?  Why not just add `-JITPath` to the line above?

src/dir.proj
```diff
@@ -0,0 +1,292 @@
...
+      <runtimeGraphGeneratorName>RuntimeGraphGenerator</runtimeGraphGeneratorName>
+      <runtimeGraphGeneratorProject>$(ProjectDir)setuptools/independent/$(runtimeGraphGeneratorName)</runtimeGraphGeneratorProject>
+      <runtimeGraphGeneratorOutput>$(IntermediateOutputRootPath)setuptools/independent/$(runtimeGraphGeneratorName)</runtimeGraphGeneratorOutput>
+      <runtimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='Windows_NT'">win</runtimeGraphGeneratorRuntime>
+      <runtimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='OSX'">osx</runtimeGraphGeneratorRuntime>
+      <runtimeGraphGeneratorRuntime Condition="'$(runtimeGraphGeneratorRuntime)'==''">linux</runtimeGraphGeneratorRuntime>      
+    </PropertyGroup>
+
+    <Error Text="Could not determine rid graph generation runtime for platform $(OSGroup)"
+          Condition="$(runtimeGraphGeneratorRuntime) == ''" />
```
eerhardt: Can this ever happen, since above you are setting this property if it is blank?

setuptools/independent/RuntimeGraphGenerator/RuntimeGraphManager.cs
```diff
@@ -24,8 +24,8 @@ public NuGet.RuntimeModel.RuntimeGraph Collect(IEnumerable<LibraryExport> export
                     if (runtimeJson != null)
                     {
                         // Convert the package-name to lower-case in the path to lookup runtime.json
-                        string lowercasedPackageName = description.Identity.Name.ToLower();
-                        string pathToPackage = export.Library.Path.Replace(description.Identity.Name, lowercasedPackageName);
+                        // string lowercasedPackageName = description.Identity.Name.ToLower();
```
eerhardt: Why comment this out?  I'm not sure I understand why we would need to change this.

Also, if this is correct, we need to update the comment above.

packaging/deb/package.targets
```diff
@@ -0,0 +1,198 @@
...
+      Overwrite="true" />
+
+    <Exec Command="$(DotnetToolCommand) restore -f $(dotnetDebToolPackageSource) --disable-parallel"
+          WorkingDirectory="$(consumingProjectDirectory)" />
+  
+  </Target>
+
+  <Target Name="GenerateDebs" 
+          DependsOnTargets="InitializeDotnetDebTool;GenerateSharedHostDeb;GenerateHostFxrDeb;GenerateSharedFrameworkDeb"
+          Condition="'$(BuildDebPkg)'=='true' and '$(TargetArchitecture)' == 'x64'" />
```
eerhardt: Who sets `$(BuildDebPkg)`?

packaging/deb/package.targets
```diff
@@ -0,0 +1,198 @@
...
+    <ItemGroup>
+        <SHFiles Include="$(InputRoot)/**/*" />
+        <SHManpages Include="$(ManPagesDir)/**/*" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(SHFiles)" DestinationFolder="$(debLayoutPackageRoot)"/>
+    <Copy SourceFiles="@(SHManpages)" DestinationFolder="$(debLayoutDocs)"/>
+
+    <!-- Replace config json variables -->
+    <WriteLinesToFile
```
eerhardt: In the CLI repo, we use this MSBuild task:  

[https://github.com/dotnet/cli/blob/rel/1.0.0/build_projects/dotnet-cli-build/ReplaceFileContents.cs#L22](https://github.com/dotnet/cli/blob/rel/1.0.0/build_projects/dotnet-cli-build/ReplaceFileContents.cs#L22)

It makes these this a lot more structured.  We can define an ItemGroup full of what should be replaced with what, and then call ReplaceFileContents.  That way we don't need these 300+ character lines of chaining `.Replace` methods together.

packaging/deb/package.targets
```diff
@@ -0,0 +1,198 @@
...
+      File="$(debLayoutDirectory)$(ConfigJsonName)"
+      Lines="$([System.IO.File]::ReadAllText('$(ConfigJsonFile)').Replace('%25SHARED_HOST_DEBIAN_VERSION%25', '$(HostVersion)').Replace('%25HOSTFXR_DEBIAN_PACKAGE_NAME%25', '$(HostFxrDebPkgName)').Replace('%25HOSTFXR_DEBIAN_VERSION%25', '$(HostVersion)').Replace('%25SHARED_FRAMEWORK_DEBIAN_PACKAGE_NAME%25', '$(SharedFxDebPkgName)').Replace('%25SHARED_FRAMEWORK_NUGET_NAME%25', '$(SharedFrameworkName)').Replace('%25SHARED_FRAMEWORK_NUGET_VERSION%25', '$(SharedFrameworkNugetVersion)').Replace('%25SHARED_FRAMEWORK_BRAND_NAME%25', '$(SharedFrameworkBrandName)'))"
+      Overwrite="true" />
+
+    <Exec Command="deb-tool -i $(debLayoutDirectory) -o $(debIntermediatesDir) -n $(DebPackageName) -v $(DebPackageVersion)"
```
eerhardt: Doesn't this need to be `dotnet deb-tool`?

setuptools/independent/RuntimeGraphGenerator/RuntimeGraphManager.cs
AlfredoMS: the packages we produces have both upper and lower case characters, this worked on windows and mac but not in linux. Eventaully all tooling and packaging will move to lowercase but for now we just avoid lower casing ourselves.

packaging/deb/package.targets
AlfredoMS: a flag passed to when invoking run.exe.
There was an issue open to make building deb packages optional

packaging/deb/package.targets
AlfredoMS: looks way better than using string,Replace. Have you thought if moving it to buildtools?

packaging/deb/package.targets
eerhardt: The default should be to build the deb packages.  And then to not build them should be the option.

packaging/deb/package.targets
eerhardt: > Have you thought if moving it to buildtools?

It can be.  The CLI doesn't use buildtools though.

setuptools/independent/RuntimeGraphGenerator/RuntimeGraphManager.cs
eerhardt: Please remove the out-of-date comment and the commented out code.

packaging/windows/dir.proj
AlfredoMS: I think this was left from a previous version on my branch. If it is no longer needed I will delete it

dir.targets
AlfredoMS: GetLatestCommitHash - is no longer a target in buildtools as it was "absorbed" by another target. I think GenerateVersionFile

GenerateVersionHeader - The version from buildtools doesn't have all the necessary defines an it was breaking the native compilation

src/corehost/build.proj
AlfredoMS: for *nix or for all OS?

I chose using msbuild to centralize parsing arguments and avoid duplicating logic in script files. otherwise we still need the same logic in build.sh and a "build-native.sh" to parse/generate a TargetRid

build.sh
AlfredoMS: build_projects contains three folder, I think we can only delete one:
-> dotnet-host-build, i think nothing in there is used anymore except by perhaps sign.proj
-> CLiFramework and shared-build-targets-utils is a dependency on how tests are run right now. This can be fixed later
-> Update dependencies can still be used as an independent tool

test/TestUtils/RepoDirectoriesProvider.cs
AlfredoMS: good point

src/dir.props
AlfredoMS: These seem to be global property in the BuildContext object of the old build system, but taking a closer look 'CoreHostOutputDir' is used both in src and test, the other ones can be moved under src.

dir.targets
weshaggard: OK we should make sure we file an issue in Buildtools to fix that there instead at a future point.

src/corehost/build.proj
weshaggard: Any new OS that we are trying to boot-strap. The native builds need to be able to build without any managed dependencies working yet like msbuild.

build.sh
weshaggard: OK. The update dependencies I expect should be replaced with the VersionTools from BuildTools now though.

test/TestUtils/RepoDirectoriesProvider.cs
weshaggard: agreed.

dir.props
weshaggard: It is always a trade-off of how many things to split up. If there are logical groups then yes it make sense to split them but we should also not just have a props file for every property.

dir.props
weshaggard: This is coming from CoreFx and is used by various things in BuildTools. It would be a larger refactoring and BuildTools update to switch it.

dir.props
weshaggard: We definitely shouldn't be dependent on mono any longer. This is probably just left over from the old corefx props.

dir.props
weshaggard: Yes please. Although I think some parts of BuildTools still use Dnu named variables but you can scope it down to just one property until Buildtools gets correctly updated. See https://github.com/dotnet/standard/blob/master/dir.props#L64 which is what I did in the standard repo.

pkg/dir.proj
weshaggard: We should stop putting packages under corehost completely now as it was only done in the past as a scoping reason but now we should just put the packages under bin\packages.

dir.props
AlfredoMS: dnurestore only calls dotnet.exe so i think it can go away

dir.props
AlfredoMS: agree

build.proj
AlfredoMS: I don't know how much we can do offline anyways since we need to restore dependencies after putting together the shared framework.

dir.targets
AlfredoMS: buildtools: https://github.com/dotnet/buildtools/blob/a3c0637b9ae8d322078aff64fae97ee2dbcfecab/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L375

packaging/deb/package.targets
AlfredoMS: ok

packaging/deb/package.targets
AlfredoMS: i'll move it into the build's custom tasks for now

src/corehost/build.proj
AlfredoMS: I chose this way because currently core-setup was depending on build_project which involved using some bootsrapper that called cli and msbuild. I understand that needs to change, if not on this PR, I'll open an issue.

src/dir.proj
AlfredoMS: build_projects\dotnet-host-build\sign.proj is called in between steps. See: https://github.com/dotnet/core-setup/blob/master/buildpipeline/Core-Setup-Signing-Windows-x64.json
Btw, i plan to move sign.proj out to the root

AlfredoMS:
I addressed some of the comments, more coming later today.
Also in the next push I will update the jenkins and vsts definitions

dir.props
eerhardt: It's still a poor name, IMO.

You could easily switch all usages of this property to something more appropriately named, and then set `$(ProjectDir)` just to keep BuildTools happy until BuildTools is fixed.

build.proj
weshaggard: We will eventually put some logic in places that will do the restoring up front and not during the build.

dir.props
weshaggard: @eerhardt if you feel strongly about it please file an issue in BuildTools and then we can push it out to the repo's after BuildTools updates.

weshaggard:
@AlfredoMS what is left to get the tests working? That will be a key part of getting this merged.

dir.props
eerhardt: https://github.com/dotnet/buildtools/issues/1367

AlfredoMS:
This PR is replaced by a new core-setup branch named 'buildtools' 

https://github.com/dotnet/core-setup/tree/buildtools

Issue #1565, opened 2017-02-20
Enable conditional rollforward when no candidate FX is found


gkhanna79:
After the patch rollforward that we do today, if the version of .NET required is not available on the machine, and DOTNET_NO_CANDIDATE_FX_ROLLFORWARD environment variable is set, we will load the next lowest version available and do patch rollforward on that version.

a.      If 1.0.1 is requested and only 1.1.0 and 1.1.1 are on the machine, then you would get 1.1.1
b.      If 1.0.1 is requested and only 1.1.1 and 1.2.0 are on the machine, then you would get 1.1.1
c.       If 1.0.1 is requested and only 2.0 and 2.0.1 are on the machine, then you would get 2.0.1

CC @glennc @DamianEdwards @leecow @bleroy

DamianEdwards:
👍

DamianEdwards:
Perhaps this should be possible via a cmd line arg and/or runtimeconfig.json setting also.

cakine:
@gkhanna79 What is the expected behavior when dealing with prereleases?

For instance:
a. If 1.0.0-dummyA is requested and only 1.1.0-dummyC and 1.1.1-dummyB are on the machine.
b. If 1.0.0 is requested and only 1.1.0-dummy  and 1.2.0 are on the machine.
c. If 1.0.0-dummy is requested and only 1.2.0-dummy and 1.1.0 are on the machine.

gkhanna79:
For pre-release versions, the current policy is to the return the next higher minimum version of the requested pre-release version is missing. There are no patch roll forwards for pre-release versions. Thus, for (a) above, it will return 1.1.0-dummyB after this feature is enabled.

For (b), since the requested version is non-prerelease, 1.2.0 will be returned after this feature is enabled (if you look at resolve_roll_forward implementation, you will see that it either deals with pre-release or non-pre-release requested version at a time - not both). 

For (c), 1.2.0-dummy will be returned after this feature is enabled.

glennc:
@gkhanna79 I don't think C is correct in the example. I think you should get 1.1.0.

Rolling forward onto a pre-release of 1.2.0 when I asked for a 1.0.0 pre-release doesn't seem right. 

Moving to a `1.0.0-echo` or to the next lowest stable would be fine. But I am not convinced about moving to another pre-release of a different version. Do you have more reasoning about why you think it should float to 1.2.0-dummy?

gkhanna79:
[...]

PR #1588, opened 2017-02-24
Roll Forward On No Candidate Fx implemented

cakine:
"Roll Forward On No Candidate Fx" is disabled by default. It can be enabled through:

1. Command line argument (--roll-forward-on-no-candidate-fx).
2. Runtimeconfig json file ('rollForwardOnNoCadidateFx' property).
3. DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX environment variable.

The conflicts will be resolved by following the priority rank described above (from 1 to 3).

If the patch roll forward we do today fails and "Roll Forward On No Candidate Fx" is set, we'll search for the next lowest production available and apply the patch roll forward again.

Fixes dotnet/core-setup#1565

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -327,40 +327,87 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
     return false;
 }
 
-void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir, const pal::string_t& fx_ver, const fx_ver_t& specified)
+void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
+    const pal::string_t& fx_ver,
+    const fx_ver_t& specified,
+    const bool& patch_roll_fwd,
+    const int& roll_fwd_on_no_candidate_fx)
 {
     trace::verbose(_X("Attempting FX roll forward starting from [%s]"), fx_ver.c_str());
 
     std::vector<pal::string_t> list;
     pal::readdir(fx_dir, &list);
     fx_ver_t most_compatible = specified;
-    for (const auto& version : list)
+    if (!specified.is_prerelease())
     {
-        trace::verbose(_X("Inspecting version... [%s]"), version.c_str());
-        fx_ver_t ver(-1, -1, -1);
-        if (!specified.is_prerelease() && fx_ver_t::parse(version, &ver, true) && // true -- only prod. prevents roll forward to prerelease.
-            ver.get_major() == specified.get_major() &&
-            ver.get_minor() == specified.get_minor())
+        if (roll_fwd_on_no_candidate_fx > 0)
         {
-            // Pick the greatest production that differs only in patch.
-            most_compatible = std::max(ver, most_compatible);
+            trace::verbose(_X("'Roll forward on no candidate fx' enabled. Looking for the least production greater than or equal to [%s]"), fx_ver.c_str());
+            fx_ver_t next_lowest(-1, -1, -1);
+            for (const auto& version : list)
+            {
+                fx_ver_t ver(-1, -1, -1);
+                if (fx_ver_t::parse(version, &ver, true) && (ver >= specified))
+                {
+                    next_lowest = (next_lowest == fx_ver_t(-1, -1, -1)) ? ver : std::min(next_lowest, ver);
+                }
+            }
+            if (next_lowest == fx_ver_t(-1, -1, -1))
+            {
+                trace::verbose(_X("No production greater than or equal to [%s] found."), fx_ver.c_str());
+            }
+            else
+            {
+                trace::verbose(_X("Found version [%s]"), next_lowest.as_str().c_str());
+                most_compatible = next_lowest;
+            }
         }
-        if (specified.is_prerelease() && fx_ver_t::parse(version, &ver, false) && // false -- implies both production and prerelease.
-            ver.is_prerelease() && // prevent roll forward to production.
-            ver.get_major() == specified.get_major() &&
-            ver.get_minor() == specified.get_minor() &&
-            ver.get_patch() == specified.get_patch() &&
-            ver > specified)
+        if (patch_roll_fwd)
         {
-            // Pick the smallest prerelease that is greater than specified.
-            most_compatible = (most_compatible == specified) ? ver : std::min(ver, most_compatible);
+            trace::verbose(_X("Applying patch roll forward from [%s]"), most_compatible.as_str().c_str());
+            for (const auto& version : list)
+            {
+                trace::verbose(_X("Inspecting version... [%s]"), version.c_str());
+                fx_ver_t ver(-1, -1, -1);
+
+                if (fx_ver_t::parse(version, &ver, true) && // true -- only prod. prevents roll forward to prerelease.
+                    ver.get_major() == most_compatible.get_major() &&
+                    ver.get_minor() == most_compatible.get_minor())
+                {
+                    // Pick the greatest production that differs only in patch.
+                    most_compatible = std::max(ver, most_compatible);
+                }
+            }
+        }
+    }
+    else
+    {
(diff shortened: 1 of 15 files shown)


cakine:
@gkhanna79

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -853,8 +905,31 @@ int fx_muxer_t::read_config_and_execute(
         append_realpath(path, &probe_realpaths);
     }
 
+    // 'No candidate fx roll forward' is disabled by default. It can be enabled through:
+    // 1. Command line argument (--no-candidate-fx-roll-forward)
+    // 2. Runtimeconfig json file ('noCandidateFxRollForward' property)
+    // 3. DOTNET_NO_CANDIDATE_FX_ROLL_FORWARD env var
+    // The conflicts will be resolved by following the priority rank described above (from 1 to 3)
+    bool no_candidate_fx_roll_fwd_enabled = config.get_no_candidate_fx_roll_fwd();
```
gkhanna79: Please add a comment indicating that env var is processed in config file processing.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -853,8 +905,31 @@ int fx_muxer_t::read_config_and_execute(
         append_realpath(path, &probe_realpaths);
     }
 
+    // 'No candidate fx roll forward' is disabled by default. It can be enabled through:
+    // 1. Command line argument (--no-candidate-fx-roll-forward)
```
gkhanna79: I was thinking about the naming - to make it slightly more intuitive from readability perspective, what do you think about "RollForwardOnNoCandidateFX"?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -853,8 +905,31 @@ int fx_muxer_t::read_config_and_execute(
...
+    // 2. Runtimeconfig json file ('noCandidateFxRollForward' property)
+    // 3. DOTNET_NO_CANDIDATE_FX_ROLL_FORWARD env var
+    // The conflicts will be resolved by following the priority rank described above (from 1 to 3)
+    bool no_candidate_fx_roll_fwd_enabled = config.get_no_candidate_fx_roll_fwd();
+    if (!no_candidate_fx_roll_fwd.empty())
+    {
+        if (no_candidate_fx_roll_fwd == _X("true"))
+        {
+            no_candidate_fx_roll_fwd_enabled = true;
```
gkhanna79: I think we should maintain this state as integer (0 or 1), instead of bool, incase we wish to support different policy levels in future for this roll forward. Bool will restrict us from going that route.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -327,40 +327,88 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
...
+                }
+            }
+            if (next_lowest == fx_ver_t(-1, -1, -1))
+            {
+                trace::verbose(_X("No production greater than or equal to [%s] found."), fx_ver.c_str());
+            }
+            else
+            {
+                trace::verbose(_X("Found version [%s]"), next_lowest.as_str().c_str());
+                most_compatible = next_lowest;
```
gkhanna79: We should update _most_compatible_ only if the requested version is missing. Take the example of 2.0.0 and 2.0.1 present on the machine, with NoCandidateRollFX enabled but PatchRollForward disabled and requested version as 2.0.0.

With your change above, _most_compatible_ becomes 2.0.1 and app executes against 2.0.1 even though patch roll forward was disabled. We should be running against 2.0.0.

gkhanna79: Actually, never mind, I read one of the conditions wrong. I think the implementation is fine above.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -327,40 +327,88 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
...
         {
-            // Pick the greatest production that differs only in patch.
-            most_compatible = std::max(ver, most_compatible);
+            trace::verbose(_X("'No candidate fx roll forward' enabled. Looking for the least production greater than or equal to [%s]"), fx_ver.c_str());
+            fx_ver_t next_lowest(-1, -1, -1);
+            for (const auto& version : list)
+            {
+                fx_ver_t ver(-1, -1, -1);
+                if (fx_ver_t::parse(version, &ver, true) &&
+                    (ver > specified || ver == specified))
```
gkhanna79: ">=" instead of two comparisons?

cakine: The ">=" operator is not defined for fx_ver_t. I can define it, though.

gkhanna79: That will be great!

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -441,12 +488,13 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t&
...
         {
-            trace::verbose(_X("Did not roll forward because specified version='%s', patch_roll_fwd=%d, chose [%s]"), specified_fx_version.c_str(), config.get_patch_roll_fwd(), fx_ver.c_str());
+            trace::verbose(_X("Did not roll forward because specified version='%s', patch_roll_fwd=%d, roll_fwd_on_no_candidate_fx=%d, chose [%s]"),
+                specified_fx_version.c_str(), config.get_patch_roll_fwd(), roll_fwd_on_no_candidate_fx_enabled, fx_ver.c_str());
             append_path(&fx_dir, fx_ver.c_str());
         }
         else
         {
-            resolve_roll_forward(fx_dir, fx_ver, specified);
+            resolve_roll_forward(fx_dir, fx_ver, specified, config.get_patch_roll_fwd(), roll_fwd_on_no_candidate_fx_enabled);
```
gkhanna79: You can factor out config.get_patch_roll_fwd and use its value instead of invoking twice (here and above).

src/corehost/cli/runtime_config.cpp
```diff
@@ -71,6 +72,21 @@ bool runtime_config_t::parse_opts(const json_value& opts)
         m_prerelease_roll_fwd = prerelease_roll_fwd->second.as_bool();
     }
 
+    auto roll_fwd_on_no_candidate_fx = opts_obj.find(_X("rollForwardOnNoCandidateFx"));
+    if (roll_fwd_on_no_candidate_fx != opts_obj.end())
+    {
+        m_roll_fwd_on_no_candidate_fx = roll_fwd_on_no_candidate_fx->second.as_bool();
```
gkhanna79: How will this property be specified in config file? Will as_bool work as expected when the value is an integer?

gkhanna79: Instead, shouldn't we not do _xtoi_ here as well, as done in other places?

src/corehost/cli/fxr/fx_muxer.cpp

src/corehost/cli/runtime_config.cpp
cakine: It will only work with boolean values. I defined it like this because that's how it was being done to patch and pre-release roll forwards properties. I tried to keep the pattern.

gkhanna79: I think there is a bit of misunderstanding - What I am suggesting is that "m_roll_fwd_on_no_candidate_fx" is maintained as an integer property so that when it is read (vi env var, command line or config), we read the value. Today, we will only support 0 or 1 but in future we could support other >1 values for different policy levels.

Bool property does not help with it. Makes sense?

cakine: I get it now. I'll fix that.

cakine: Done.

PR #1592, opened 2017-02-27
Adding linux arm to netcoreapp

Petermarcu:
@ericstj @gkhanna79


gkhanna79:
Also, please rebase the commits.

ericstj:
@gkhanna79 : because we are building RID-specific packages in core-setup now.

Keep in mind you don't "need" this, you can always reference the RID specific package, but if you want to tell the system you have a RID specific package and let user reference a RID agnostic package that data needs to live somewhere.

LGTM.

Petermarcu:
I will squash when I merge.

PR #1594, opened 2017-02-27
Update UWP metapackage for servicing

zamont:
Introduce updated S.R.Serialization.Json and S.R.Serialization.Xml packages as well as those indicated in Issue #10338


zamont:
@ericstj can you please take a look? This references dotnet/corefx#10338

ericstj:
I see this picks up some other packages built in 1.0.0 branch https://github.com/dotnet/corefx/commit/4aec5a66211db511916018d754eb87a4c2c3a26f#diff-6fe68a38964d837f8e512c4e62288dd3R1346. but not System.Linq.Expressions.  Was that intentional?

zamont:
Leaving out System.Linq.Expressions was intentional. Bringing in System.Linq.Expressions would introduce incompatibilities with the SharedLibrary for .NET Native.

weshaggard:
>Leaving out System.Linq.Expressions was intentional. Bringing in System.Linq.Expressions would introduce incompatibilities with the SharedLibrary for .NET Native.

Did you verify that? While we likely did compile it again so the binary doesn't exactly match all we did was include an rd.xml file in the assembly so it was intended to be consumed in UWP (see https://github.com/dotnet/corefx/commit/780180f171be42c1ff62e18d3aa853afb4025a3e).  If it the shared framework is in play and we need this fixed we probably need to provide the rd.xml file in another way that doesn't change the binary.

zamont:
> Leaving out System.Linq.Expressions was intentional. Bringing in System.Linq.Expressions would introduce incompatibilities with the SharedLibrary for .NET Native.

I looked into this more and got a much clearer understanding of this issue. Bringing in an updated System.Linq.Expressions NuGet package will bring in a new version of the System.Linq.Expressions.dll (v4.0.10.1). In our VS-integration for .NET Native, we explicitly check that we are using System.Linq.Expressions.dll (v.4.0.10.0). This fails as a result and we default the user out of using the SharedLibrary, which is undesirable. We should seek another way of delivering the rd.xml file to the customer.

weshaggard:
OK the changes look good to me but you should probably file another issue for the System.Linq.Expressions issue.

PR #1647, opened 2017-03-06
Add a build task to cleanup Docker.

ravimeda:
Core-Setup builds often fail due to low disk space. The most likely cause for this is the build definitions do not appear to execute `cleanup-docker.sh`, which will cleanup Docker images and containers. Hence attempting to add a build task that will perform Docker cleanup.  This task will invoke a new script `cleanup-build.sh` that will do the following:

1. Restores BuildTools package
2. Runs `pkg/Tools/scripts/docker/cleanup-docker.sh`

@gkhanna79 @chcosta @dagood @MichaelSimons  please review.


buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -31,7 +50,7 @@
       },
       "inputs": {
         "scriptPath": "scripts/dockerrun-as-current-user.sh",
-        "args": "-t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB microsoft/dotnet-buildtools-prereqs:rhel7_prereqs /bin/bash -c \"HOME=/opt/code; git clean -X -d -f; ./build.sh $(BuildArguments)\"",
+        "args": "-t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB microsoft/dotnet-buildtools-prereqs:rhel7_prereqs /bin/bash -c \"HOME=/opt/code; git clean -x -d -f; ./build.sh $(BuildArguments)\"",
```
ravimeda: @chcosta please clarify if using lowercase x is a better choice here.

chcosta: lower-case x is better.

chcosta:
It's not clear to me what the need for a separate script is.

MichaelSimons:
Agreed, If possible I would rather see init-tools and cleanup-docker as individual build steps.

---
In reply to: [284569828](https://github.com/dotnet/core-setup/pull/1647#issuecomment-284569828) [](ancestors = 284569828)

dagood:
Why do you prefer adding more steps? It seems less error-prone to me to add a cleanup "section" that we can add to as needed, rather than modifying N build definitions.

ravimeda:
I see benefits with either approach.  Having separate build tasks provides better transparency on what is being performed, and is easier to track down a failure.  On the other hand, a consolidate cleanup script reduces the maintenance of build definition file, which is a manually edited json.

chcosta:
I'm not opposed to having a script, I don't think it makes sense to add a script in the core-setup repo though, it's additional code that needs to be special case understood for just that repo, an additional point of failure, may lead to devs trying to use it with misunderstood expectations.

dagood:
I see, I agree. If we have more cleanup steps to add in after init-tools, they should be in a buildtools script. (Something that comes to mind is cleaning up docker and the agent, currently separate in buildtools.) One obstacle right now is upgrading buildtools, but that's just something to fix up or work around by adding more steps for now.

buildpipeline/Core-Setup-CentOS-x64.json
```diff
@@ -3,6 +3,43 @@
...
+      "alwaysRun": false,
+      "displayName": "Clean up Docker images and containers.",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "pkg/Tools/scripts/docker/cleanup-build.sh",
```
chcosta: cleanup-docker.sh?

ravimeda: Good catch!

buildpipeline/Core-Setup-CentOS-x64.json
```diff
@@ -3,6 +3,43 @@
...
+        "args": "",
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
```
chcosta: We should probably continue on error.  This could fail for a variety of reasons, but then later eventually succeed and I think that's ok.  we wouldn't want this cleanup stuff to cause intermittent job failures though.

ravimeda: Agreed.

ravimeda: @MichaelSimons checking if this is fine with you too.

MichaelSimons: Yes - it makes sense to me.

buildpipeline/Core-Setup-CentOS-x64.json
```diff
@@ -3,6 +3,43 @@
     {
       "enabled": true,
       "continueOnError": false,
+      "alwaysRun": true,
```
chcosta: if you are always going to run init-tools, you should also always run the cleanup stuff

ravimeda: Let's set it to false.

chcosta: I'd lean the other way.  It's cleanup stuff, and as long as we're on the machine, we should try to put the machine in a clean(er) state.

dagood: Is there a situation where the first step will only run if it is `alwaysRun`?

chcosta: I'm not certain I understand the question.  `alwaysRun=true` implies that the step will run even if another step in the build fails.  `alwaysRun=false` means that if there is a build failure in any of the steps, then this task doesn't run.  I just assumed that this cleanup step was a latter step as we do it in the other repo's, but I see that it's not.  I'd actually move this to the end so that a failure in cleanup doesn't block the build from actually running.

ravimeda: Ok. In that case let's try the following sequence:

1. `build.sh`
2. `pkg/init-tools.sh` 
3. `pkg/Tools/scripts/docker/cleanup-docker.sh` after 

No changes to 1.  But for 2 & 3, _continueOnError_ and _alwaysRun_ will be true. 
Concerns?

dagood: The reason to do cleanup first is to unblock the current agent to do the build without running out of space. We don't have much data about why these errors are happening, so I don't think we can say for sure if doing cleanup after the build will help anything. Also, I believe docker cleanup (at least) is done before the build for CoreFX and CoreCLR. (For CoreFX, as part of init-docker.)

chcosta: If cleanup is running regularly, you shouldn't be in a position where running cleanup is a prereq to building.  I also don't think we want a cleanup failure to block running the build, so I'd prefer that cleanup stuff happened at the end.  The agent cleanup stuff happens at the end in Windows builds, I don't see it enabled for Linux.

ravimeda: Let's move the two tasks post build. Both tasks should always run, and should continue on failure. I will prepare the changes.

buildpipeline/Core-Setup-CentOS-x64.json
```diff
@@ -18,6 +18,43 @@
...
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "pkg/init-tools.sh",
+        "arguments": "",
+        "workingFolder": "",
+        "failOnStandardError": "false"
```
dagood: Taking a look at the results of a build, I saw that the inputs and the versionSpec don't match the build.sh shell script call step, so this doesn't work during the build.

```
  "versionSpec": "2.*",
  "definitionType": "task"
},
"inputs": {
  "scriptPath": "build.sh",
  "args": "$(BuildArguments)",
  "disableAutoCwd": "false",
  "cwd": "",
  "failOnStandardError": "false"
```

init-tools is a no-op, though, since build.sh already initialized tools. (When we need a version of the CLI/msbuild that we can use, though, we'll need this to actually pull down new tooling.)

ravimeda: Good catch! Fix in #1687

PR #1650, opened 2017-03-07
Fix pkg and packaging targets

AlfredoMS:
-Fix pkgbuild: The issue was using the new replace file contents target instead of string.Replace this is because an empty property won't work with the custom target.

-Add missing upgrade codes when generating Msi

whit this running .\build.cmd -SkipTests=true runs succesfully to completion in a freshly clone repo.

### DIFF
--- a/pkg/projects/dir.targets
+++ b/pkg/projects/dir.targets
@@ -18,11 +18,11 @@
     <!-- Update project.json template -->
     <MakeDir Directories="$([System.IO.Path]::GetDirectoryName('$(ProjectJson)'))" />
 
-    <ReplaceFileContents
-      InputFile="project.json.template"
-      DestinationFile="$(ProjectJson)"
-      ReplacementPatterns="{RID};{TFM}"
-      ReplacementStrings="$(TargetRid);$(NuGetTargetMoniker)" />
+    <WriteLinesToFile
+      File="$(ProjectJson)"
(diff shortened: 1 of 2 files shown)


AlfredoMS:
cc @eerhardt

pkg/projects/dir.targets
```diff
@@ -18,11 +18,11 @@
     <!-- Update project.json template -->
     <MakeDir Directories="$([System.IO.Path]::GetDirectoryName('$(ProjectJson)'))" />
 
-    <ReplaceFileContents
-      InputFile="project.json.template"
-      DestinationFile="$(ProjectJson)"
-      ReplacementPatterns="{RID};{TFM}"
-      ReplacementStrings="$(TargetRid);$(NuGetTargetMoniker)" />
+    <WriteLinesToFile
```
eerhardt: Another option here:

```
    <ItemGroup>
      <_ProjectJsonReplacementValues Include="{RID}">
        <ReplacementString>$(NuGetRuntimeIdentifier)</ReplacementString>
      </_ProjectJsonReplacementValues>
      <_ProjectJsonReplacementValues Include="{TFM}">
        <ReplacementString>$(NuGetTargetMoniker)</ReplacementString>
      </_ProjectJsonReplacementValues>
    </ItemGroup>

    <ReplaceFileContents
      InputFile="project.json.template"
      DestinationFile="$(ProjectJson)"
      ReplacementItems="@(_ProjectJsonReplacementValues)" />
```

It's a little more "MSBuildy", and allows it to be extended easily.

PR #1661, opened 2017-03-08
Remove Roslyn and libuv from shared framework

eerhardt:
Roslyn and libuv are only needed for ASP.NET scenarios, and as such they shouldn't be included in the base .NET Core runtime installer. Instead, ASP.NET will create an ASP.NET runtime installer that will include ASP.NET's dependencies outside of the core runtime.

@gkhanna79 @weshaggard @ericstj 

/cc @Petermarcu @Eilon @pranavkm @jaredpar @tmat


Eilon:
cc @DamianEdwards @davidfowl FYI

eerhardt:
FYI - I plan on merging this today, unless I hear any objections by EOD.

PR #1662, opened 2017-03-08
Move Core-Setup to 1.1.2 servicing release

gkhanna79:
@weshaggard @ramarag PTAL.

CC @leecow

### DIFF
--- a/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
+++ b/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "Microsoft.NETCore.Portable.Compatibility": "1.0.2",
-    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.1-servicing-24904-01",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.2-servicing-25107-01",
     "Microsoft.NETCore.Targets": "1.1.0",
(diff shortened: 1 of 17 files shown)


gkhanna79:
@weshaggard Can you PTAL?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
```diff
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "Microsoft.NETCore.Portable.Compatibility": "1.0.2",
-    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.1-servicing-24904-01",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.1.2-servicing-25107-01",
```
weshaggard: Do we want to update UWP meta-package? I guess we have till this point but it is something that needs closer review if we plan to ship it out of this branch in the future.

src/Microsoft.Extensions.DependencyModel/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for reading `.deps` files.",
-  "version": "1.1.1",
+  "version": "1.1.2-servicing-*",
```
weshaggard: Do we actually plan to ship this or PlatformAbstractions from the servicing branches? I just checked and we did ship it with the 1.1.1 release but I would have not expected us to given we haven't changed anything here. We should try to avoid shipping these if it isn't necessary.

PR #1684, opened 2017-03-09
Fix OSX RID.

eerhardt:
The RID on OSX is incorrect.  It is missing the os name.

@gkhanna79 @ramarag

I plan on adding tests for this, but I don't have time right now and I need to get this bug fixed.  So a follow-up PR will be made to add tests.


gkhanna79:
Thanks for fixing this.

PR #1685, opened 2017-03-09
Git checkout the specific commit.

ravimeda:
In PipeBuild, build legs checkout master branch as default. This causes a problem in servicing branches. To address this behavior, build legs should checkout the specific commit.  Hence adding a task to git checkout in each PipeBuild build definition.

@dagood @chcosta @gkhanna79 please review.

### DIFF
--- a/buildpipeline/Core-Setup-CentOS-x64.json
+++ b/buildpipeline/Core-Setup-CentOS-x64.json
@@ -4,6 +4,25 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
+      "displayName": "git checkout",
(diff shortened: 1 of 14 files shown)


gkhanna79:
Does this change need to go to 1.0.0 as well?

ravimeda:
Thanks! Corresponding change for 1.0.0 is #1686

dagood:
We should do this in `master` too, for repeatability and future servicing branch forks.

gkhanna79:
agreed - @ravimeda Can you please get a PR going for that as well?

ravimeda:
Yes, #1687 is for `master`.

PR #1686, opened 2017-03-09
Git checkout the specific commit - 1.0.0

ravimeda:
Similar to #1685 But this one is for release 1.**0**.0.

In PipeBuild, build legs checkout master branch as default. This causes a problem in servicing branches. To address this behavior, build legs should checkout the specific commit. Hence adding a task to git checkout in each PipeBuild build definition.

@dagood @chcosta @gkhanna79 please review.


ravimeda:
@dagood  please could you review commit `6b880db`

PR #1687, opened 2017-03-09
Git checkout the specific commit - master

ravimeda:
These changes will ensure PRs #1685 and #1686 are avoided for future service branches.
No change is needed in Core-Setup-CrossBuild.json

@dagood @chcosta @gkhanna79 please review.

PR #1688, opened 2017-03-09
Git checkout task should be a command and not shell script - 1.1.0

ravimeda:
Git checkout task introduced through #1685 should be a VSTS Run Command and not Run Shell Script. type. Updated non-Windows definitions accordingly.

@gkhanna79 @dagood @chcosta please review.


ravimeda:
I'll make similar change to #1686 and #1687, after making sure 1.1.0 build succeeds.

PR #1689, opened 2017-03-09
Enable fallback to base host RID

gkhanna79:
Fixes https://github.com/dotnet/core-setup/issues/1620 based upon the comments at https://github.com/dotnet/core-setup/issues/1620#issuecomment-285226279.

I have also removed the capping from RID computation for Windows so that it is in sync with what we do for OSX and Linux (return what was computed without cap).

@eerhardt PTAL

### DIFF
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -122,8 +122,8 @@ void deps_json_t::reconcile_libraries_with_targets(
     }
 }
 
-// Returns the RID for the platform the host is running on.
-pal::string_t get_current_rid()
+// Returns the RID determined (computed or fallback) for the platform the host is running on.
+pal::string_t deps_json_t::get_current_rid(const rid_fallback_graph_t& rid_fallback_graph)
 {
     
     pal::string_t currentRid;
@@ -136,14 +136,26 @@ pal::string_t get_current_rid()
         }
     }
     
-    trace::info(_X("Host RID is %s"), currentRid.empty()? _X("not available"): currentRid.c_str());
+    trace::info(_X("HostRID is %s"), currentRid.empty()? _X("not available"): currentRid.c_str());
+
+    // If the current RID is not present in the RID fallback graph, then the platform
+    // is unknown to us. At this point, we will fallback to using the base RIDs and attempt
+    // asset lookup using them.
+    //
+    // We do the same even when the RID is empty.
+    if (currentRid.empty() || (rid_fallback_graph.count(currentRid) == 0))
+    {
+        currentRid = pal::get_current_os_fallback_rid() + pal::string_t(_X("-")) + get_arch();
+
+        trace::info(_X("Falling back to base HostRID: %s"), currentRid.c_str());
+    }
 
     return currentRid;
 }
 
 bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, const rid_fallback_graph_t& rid_fallback_graph)
 {
-    pal::string_t host_rid = get_current_rid();
+    pal::string_t host_rid = get_current_rid(rid_fallback_graph);
     
     for (auto& package : portable_assets->libs)
     {
--- a/src/corehost/common/pal.h
+++ b/src/corehost/common/pal.h
@@ -50,16 +50,26 @@
 
 #endif
 
-
+// When running on a platform that is not supported in RID fallback graph (because it was unknown
+// at the time the SharedFX in question was built), we need to use a reasonable fallback RID to allow
+// consuming the native assets.
+//
+// For Windows and OSX, we will maintain the last highest RID-Platform we are known to support for them as the 
+// degree of compat across their respective releases is usually high.
+//
+// We cannot maintain the same (compat) invariant for linux and thus, we will fallback to using lowest RID-Plaform.
 #if defined(_WIN32)
 #define LIB_PREFIX
 #define MAKE_LIBNAME(NAME) (_X(NAME) _X(".dll"))
+#define FALLBACK_HOST_RID _X("win10")
 #elif defined(__APPLE__)
 #define LIB_PREFIX _X("lib")
 #define MAKE_LIBNAME(NAME) (LIB_PREFIX _X(NAME) _X(".dylib"))
+#define FALLBACK_HOST_RID _X("osx.10.12")
 #else
 #define LIB_PREFIX _X("lib")
 #define MAKE_LIBNAME(NAME) (LIB_PREFIX _X(NAME) _X(".so"))
+#define FALLBACK_HOST_RID _X("linux")
 #endif
 
 #define LIBCLRJIT_NAME MAKE_LIBNAME("clrjit")
@@ -180,21 +190,17 @@ namespace pal
 
 #endif
 
-#if defined(__APPLE__)
-    enum os_moniker_t
-    {
-        sierra = 16,
-        el_capitan = 15,
-    };
-
-    bool get_os_moniker(os_moniker_t* moniker);
-#endif
-
     inline void err_flush() { std::fflush(stderr); }
     inline void out_flush() { std::fflush(stdout); }
 
     // Based upon https://github.com/dotnet/core-setup/blob/master/src/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
     pal::string_t get_current_os_rid_platform();
+    inline pal::string_t get_current_os_fallback_rid()
+    {
+        pal::string_t fallbackRid(FALLBACK_HOST_RID);
+
+        return fallbackRid;
+    }
         
     bool touch_file(const pal::string_t& path);
     bool realpath(string_t* path);
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -297,7 +297,7 @@ pal::string_t pal::get_current_os_rid_platform()
             }
 
             ridOS.append(_X("osx.10."));
-            ridOS.append(std::to_string(minorVersion));
+            ridOS.append(pal::to_string(minorVersion));
         }
     }
 
--- a/src/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
+++ b/src/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
@@ -80,9 +80,8 @@ namespace Microsoft.DotNet.PlatformAbstractions
             }
             else if (ver.Major >= 10)
             {
-                // Use Win10-* RID even for newer platforms we do not know for, so that existing Win10 assets can be used
-                // to keep apps working.
-                return "10";
+                // Return the major version for use in RID computation without applying any cap.
+                return ver.Major.ToString();
             }
             return string.Empty; // Unknown version
         }
--- a/src/corehost/cli/deps_format.h
+++ b/src/corehost/cli/deps_format.h
@@ -83,6 +83,7 @@ private:
 
     pal::string_t get_optional_path(const json_object& properties, const pal::string_t& key) const;
 
+    pal::string_t get_current_rid(const rid_fallback_graph_t& rid_fallback_graph);
     bool perform_rid_fallback(rid_specific_assets_t* portable_assets, const rid_fallback_graph_t& rid_fallback_graph);
 
     std::vector<deps_entry_t> m_deps_entries[deps_entry_t::asset_types::count];
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -272,15 +272,9 @@ pal::string_t pal::get_current_os_rid_platform()
                 }
                 else if (majorVer >= 10)
                 {
-                    switch(minorVer)
-                    {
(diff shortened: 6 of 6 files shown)


src/corehost/common/pal.unix.cpp
```diff
@@ -263,6 +263,13 @@ pal::string_t trim_quotes(pal::string_t stringToCleanup)
     return stringToCleanup;
 }
 
+pal::string_t pal::get_current_os_baserid()
```
eerhardt: Can this be defined in common code, so it doesn't need to be duplicated?

src/corehost/common/pal.h
```diff
@@ -54,12 +54,15 @@
 #if defined(_WIN32)
 #define LIB_PREFIX
 #define MAKE_LIBNAME(NAME) (_X(NAME) _X(".dll"))
+#define BASE_HOST_RID _X("win")
```
eerhardt: A thing I dislike about this approach is that a lot of times the runtime assets aren't in the base rid, instead they are in the lowest version of the base rid.

ex.

`win7-x64`
`osx.10.10-x64`

And we will be missing these assets.

Can we use the lowest OS version as the "base rid" for `win` and `osx`?  `linux` is different because we **will** have assets just in `linux-x64`.

gkhanna79: I understand. While this is straightforward for Windows (today), it is not so for even OSX. for example, we do not support 10.10 (even though our assets are in such folders). In 2.0, this will change to 10.12. In future, when Win7 retires, we may drop it as well. Linux is not different from the other two in this case.

Thus, we will need to keep this updated in the host. Base RID goal is for us to have something reasonable that when we dont know what platform we are on. Also, for us, falling to a consistent baseline makes things much easier to reason about.

eerhardt: For example, `Libuv v1.9.1` supports:

image

And all of our runtime packages all stop at `osx.10.10`:

image

So with your current code, all ASP.NET apps will break on Windows if/when there is a `win11`.

And I can imagine a lot of libraries doing the same thing we do in the runtime by stopping at `osx.10.10` instead of purely `osx`. Thus any app running on `osx.10.13` that uses these libraries will be broken.

gkhanna79: Also, on a platform that is unknown to us, we cannot guarantee even the last known "lowest RID" specific assets are guaranteed to work.

eerhardt: > Also, on a platform that is unknown to us, we cannot guarantee even the last known "lowest RID" specific assets are guaranteed to work.

We might not be able to guarantee it, because we didn't test on that platform and we never said we supported it.

But I can guarantee that your app will be broken if you require a `osx.10.10` asset when you are trying to run your app on `osx.10.13`.

gkhanna79: Windows is an outlier in this example since it maintains compat across releases. Even if we were to support Windows by keeping win7-x64 as the base RID, what would be our story for OSX and Linux as compat is not guaranteed there? And even on Windows, apps/libraries that have only, say, assets for win10-x64 but not for win7-x64 (since they added native dependency at a later point) will also not work if we set the base to win7-x64.

For unsupported platforms, we can say that DOTNET_RUNTIME_ID should be used to set the RID appropriate for each platform. IMHO, there is no clean/consistent story for them.

gkhanna79: CC @weshaggard and @Petermarcu to chime in as well.

gkhanna79: > But I can guarantee that your app will be broken if you require a osx.10.10 asset when you are trying to run your app on osx.10.13

Are you not making my point or am I missing something?

gkhanna79: The only time I see us setting the base RIDs as win7-x64/osx.10.10-x64 is if we are always going to keep them around and if we are going to assert that assets for them will work on newer version of the platforms (which we cannot guarantee for OSX anyways). At best, we keep win7-x64 forever and fallback to it.

And even then, our story is busted on Linux.

eerhardt: The main thing I want here is for when new versions of OSX come out, .NET apps aren't immediately broken by default just because the RID changed.  If Apple decides to make breaking changes, so be it, we can't help that.  But we can build our product in such a way that if there are no operating system breaking changes from osx.10.12 => osx.10.13, then .NET apps will just work.

[WWDC17](https://developer.apple.com/wwdc/) takes place June 5-9.  It would be bad if we shipped just before that and we were immediately broken by a new OS that was announced there.

Setting `DOTNET_RUNTIME_ID` is a total hack, we shouldn't use it as a crutch.  I'm glad the hack is there to get me unblocked, but it shouldn't be our default answer.

eerhardt: > Are you not making my point or am I missing something?

I'm making the point: if your app requires an asset that is under the `osx.10.10` runtime, and you are running on the imaginary `osx.10.13` OS, we will "fallback" to the `osx` RID, and we won't find that asset.  Thus your app is broken by default just because the RID changed.  But we didn't even give the `osx.10.10` asset a chance, we skipped it.  Chances are, the `osx.10.10` asset will work, unless Apple made a breaking change.

gkhanna79: I hear you. In that case, for Windows and OSX, I would rather hard-code the fallback RIDs to be the last highest RIDs known (i.e. win10-* and osx.10.12-*) so that the best attempt is made at the experience.

Linux will still be linux-x64.

What do you think?

eerhardt: I think last highest RID would meet my needs.

The other thought:  Maybe we stop using `osx.10.10` in our packages as the "base OSX" RID.  And if this asset will work on any OSX that we support, we start putting that asset under plainly `osx`.  Same goes for Windows.  Don't put assets under `win7`.  Instead put them in `win`.

I think last highest RID works - and it is what we were doing in 1.0 anyway.

gkhanna79: > I think last highest RID would meet my needs.

To be precise, this is only for Windows and OSX - Linux will fall to linux-*. @Petermarcu are you fine with this as well? 

> Maybe we stop using osx.10.10 in our packages as the "base OSX" RID. And if this asset will work on any OSX that we support, we start putting that asset under plainly osx. Same goes for Windows. Don't put assets under win7. Instead put them in win.

I think that is the plan.

gkhanna79: @eerhardt Thinking a bit more, what happens if OSX 10.13 (or win11) is incompatible with its predecessors and breaks previous assets? What would be the highest supported RID the future host would support?

eerhardt: > Thinking a bit more, what happens if OSX 10.13 (or win11) is incompatible with its predecessors and breaks previous assets?

If that was the case, we would break the chain of the RID fallback graph.  We would say that `osx.10.13` maps straight to `osx`, skipping its predecessors.  This is what happens for `ubuntu`.  `ubuntu16` => `ubuntu`, skipping `ubuntu14`.

eerhardt: > To be precise, this is only for Windows and OSX - Linux will fall to linux-*.

Correct.  That is my thought.  Since we actually have `linux` assets, falling to `linux` makes sense.

The good thing about Windows and OSX is that they are more controlled and usually have high compatibility requirements.

gkhanna79: > we would break the chain of the RID fallback graph.

And in this case, I presume we will update the base fallback RID (which we are discussing here) be updated to osx.10.13?

eerhardt: 2 thoughts:

1. No matter whether OSX 10.13 is compatible with its predecessors or not, in this plan when we officially move to support OSX 10.13, we would move the base fallback RID to be osx.10.13.
1. In the case where OSX 10.13 is incompatible with its predecessors and we still wanted to support osx.10.12 and earlier, we would need to build for OSX twice - osx.10.12 and osx.10.13 - just like we do for each major version of `ubuntu` today - ubuntu14.04 and ubuntu.16.04.

This fallback plan is not un-like what we did in `1.0` where the `TARGET_RUNTIME_ID` is supplied at build time.  The differences from `1.0` are:

1. For all platforms, we first try the actual machine's RID, if that works - great use it.  This allows us to update the `runtime.json` file, and the host automatically picks it up and works with no code changes. This change from `1.0` gives us a lot of future proofing.
2. When the machine's RID doesn't work:
  - When running on Linux, instead of using the RID of the machine we built on (since we have absolutely no idea of compatibility between this unknown Linux distro and the distro we built on), we fall back to `linux` RID, because that is the common ground for all Linux distros.
  - When on OSX and Windows, if we are running on an unknown RID, there are really only 2 possible situations:  we are on a higher version or a lower version.  If we are on a lower version, we never said we support it, and chances are we probably didn't even get this far.  But on the higher version case, I would expect things to work, barring any OS incompatibilities.  So to make it even possible for the app to work on newer versions, we start with the highest OS version we knew at build time.

Given this plan, we won't HAVE to update the host code when a new OSX version comes out.  We just need to update the `runtime.json` file, after we've done testing, and the new RID will start to work in the fallback graph.  But until that time, which takes a while, we should at least give the apps a chance to run on the new OSX.  That way we can evaluate/test on the new OSX version without making any changes at all.

gkhanna79: > We just need to update the runtime.json file, after we've done testing, and the new RID will start to work in the fallback graph.

I agree with all of the above. Specific to this statement, when runtime.json is updated, it translates to us shipping a new SharedFX that supports the new RID and, at that time, we will be updating base fallback RID to the _updated_ latest RID.

gkhanna79: @eerhardt PTAL at the updated PR - I have updated the fallback RIDs per our discussion above.

src/corehost/cli/deps_format.cpp
```diff
@@ -123,7 +123,7 @@ void deps_json_t::reconcile_libraries_with_targets(
 }
 
 // Returns the RID for the platform the host is running on.
```
eerhardt: (minor) this comment is a little out of date.

src/corehost/common/pal.h
```diff
@@ -180,21 +190,17 @@ namespace pal
...
-
-    bool get_os_moniker(os_moniker_t* moniker);
-#endif
-
     inline void err_flush() { std::fflush(stderr); }
     inline void out_flush() { std::fflush(stdout); }
 
     // Based upon https://github.com/dotnet/core-setup/blob/master/src/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
     pal::string_t get_current_os_rid_platform();
+    inline pal::string_t get_current_os_baserid()
```
eerhardt: Should we consistently say "base RID" or "fallback RID"?  Here the method is named one thing, but the const is a different thing.

src/corehost/common/pal.h
```diff
@@ -50,16 +50,26 @@
...
+// consuming the native assets.
+//
+// For Windows and OSX, we will maintain the last highest RID-Platform we are known to support for them as the 
+// degree of compat across their respective releases is usually high.
+//
+// We cannot maintain the same (compat) invariant for linux and thus, we will fallback to using lowest RID-Plaform.
 #if defined(_WIN32)
 #define LIB_PREFIX
 #define MAKE_LIBNAME(NAME) (_X(NAME) _X(".dll"))
+#define FALLBACK_HOST_RID _X("win10")
```
weshaggard: Did we not want to fallback to just "win" and "osx"?

eerhardt: See the conversation here:  https://github.com/dotnet/core-setup/pull/1689#discussion_r105297258.

If we took that approach, we would miss a lot of assets and more than likely the app wouldn't work by default, just because a new macOS version came out.

weshaggard: Sorry I missed that big conversation as it got folded away in the code changes.

While I agree we would miss some assets it was my understanding that we would change all our RID packages to be based on "win" and "osx" and we would push others to follow a similar pattern and when that happens this wouldn't be a problem any longer. To me it is the exact same issue we are trying to solve with moving to "linux".

I'm curious to get @ericstj and @Petermarcu opinion here as I'm not hung up on it but it would be nice to be consistent across platforms.

ericstj: I'd opt for the most compatible thing.  That'd be the versioned RID since it still covers the un-versioned case.

> we would push others to follow a similar pattern and when that happens this wouldn't be a problem any longer

What if folks actually had a good reason for cross-compiling by OS version and didn't want to handle that at runtime?  The reason we are able to do this is we are putting forth a lot of effort to produce distro/version-agnostic binaries.  While its fine to recommend that to customers I don't think we should mandate it unless it can truly remove a lot of complexity.

gkhanna79: Correct. The point was just not about our assets but external assets that would still be using distro-specific RIDs.

weshaggard: This same logic would apply to Linux but we are doing that differently.

gkhanna79: On Linux, there are no compat guarantees even within same distro - Ubuntu 16.04 will not run 14.04 binaries. Likewise, we cannot comment on the compat for Ubuntu 19.04, which would be new to us, about whether it works for 17.04 or 16.04.

PR #1691, opened 2017-03-10
Perform Docker cleanup after build. 1.0.0.

ravimeda:
Similar to #1647 But this one is for release 1.0.0.

@chcosta please take a look.

### DIFF
--- a/buildpipeline/Core-Setup-CentOS-x64.json
+++ b/buildpipeline/Core-Setup-CentOS-x64.json
@@ -68,8 +68,8 @@
         "definitionType": "task"
       },
       "inputs": {
-        "filename": "pkg/Tools/scripts/docker/cleanup-docker.sh",
-        "arguments": "",
+        "filename": "perl",
+        "arguments": "pkg/Tools/scripts/docker/cleanup-docker.sh",
         "workingFolder": "",
         "failOnStandardError": "false"
       }
--- a/buildpipeline/Core-Setup-CentOS-x64.json
+++ b/buildpipeline/Core-Setup-CentOS-x64.json
@@ -45,13 +45,14 @@
       "timeoutInMinutes": 0,
       "task": {
         "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
-        "versionSpec": "1.*",
+        "versionSpec": "2.*",
         "definitionType": "task"
       },
       "inputs": {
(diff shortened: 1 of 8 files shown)


buildpipeline/Core-Setup-CentOS-x64.json
```diff
@@ -36,6 +36,44 @@
...
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "pkg/Tools/scripts/docker/cleanup-docker.sh",
+        "args": "",
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
```
dagood: I tried running this def in VSTS and it failed. This step tried to run this script with `bash`, but `cleanup-docker.sh` is really a `perl` script. (The step doesn't execute `./pkg/Tools/...`, but rather `bash pkg/Tools/...`, which ignores the `#!` at the top of the file.)

Changing this to a "command line" step that runs `perl` with arg `pkg/Tools/scripts/docker/cleanup-docker.sh` works.

ravimeda: Alright. Fix in commit `fca01db`.  Same with PR #1692

buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -54,6 +54,43 @@
...
+      "alwaysRun": true,
+      "displayName": "Initialize tools",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "pkg/init-tools.sh",
```
dagood: This is a no-op because the volume-mapped Docker build initialized its tools into the agent's dir. This works for now because all we need is the script, but we'll have to fetch an appropriate CLI for the msbuild/managed cleanup steps we plan to add later. Adding a clean step before this init-tools is the way I imagine doing that.

ravimeda: Ok.

PR #1700, opened 2017-03-11
Remove corehost executable.

eerhardt:
This executable was used in the project.json based CLI.  But it is no longer used with the new MSBuild based tools.  We should remove it in 2.0.

@gkhanna79 @ramarag 

/cc @livarcocc


eerhardt:
@ellismg @chcosta - I looked through the MSBuild branch and it looks like this is already removed there. But I wanted to ensure you see this so it doesn't come back when we convert to MSBuild.

chcosta:
Thanks

PR #1701, opened 2017-03-11
Add retry on Docker image pulling step - 1.0.0

ravimeda:
Builds for Linux platforms attempt to pull the corresponding Docker image. If an attempt fails then, there is no retry. This can cause build failure.

[BuildTools ](https://github.com/dotnet/buildtools) offers `init-docker.sh` script that can retry Docker image pull step. Hence introducing this script in this branch.

Note that RHEL has a limitation that requires a RHEL image, and cannot be built using a Dockerfile.

`init-docker.sh` was manually copied, as opposed to restoring buildtools package. Copying the script was a safer approach since:
1. Version of buildtools used in this branch does not have init-docker.sh
2. This is a service branch. Upgrading to a newer version of buildtools has a greater chance for regressions.

@MichaelSimons @chcosta @gkhanna79 please review.


buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -164,6 +183,9 @@
     },
     "GITHUB_PASSWORD": {
       "value": "PassedViaPipeBuild"
+    },
+    "PB_DockerImageName": {
```
MichaelSimons: >PB_DockerImageName [](start = 5, length = 18)

Is it strange and deceiving to name this with the PB_ prefix when the variable is really local and not specified by the pipe build?

scripts/dockerrun.sh
```diff
@@ -87,8 +87,41 @@ fi
...
 # Build the docker container (will be fast if it is already built)
 echo "Building Docker Container using Dockerfile: $DOCKERFILE"
+
+# Get the name of Docker image.
+image=$(grep -i "^FROM " "$DOCKERFILE" | awk '{ print $2 }')
+
+if [ ! -z "$image" ]; then
+    echo "Pulling Docker image $image"
+    execute docker pull $image
+fi
```
MichaelSimons: I think there is value in calling out why this image is explicitly pulled versus implicitly during the build.  I could see someone not understanding this and removing the logic.

MichaelSimons:
I think it is worth noting that ini-docker/cleanup-docker is copied locally as part of this PR instead of using the centralized BuildTools version because this is a servicing branch that you do not want do disrupt by taking the latest version of build tools.

ravimeda:
Thanks @MichaelSimons Addressed the comments in `0818fb1`. PTAL.

buildpipeline/Core-Setup-RHEL7-x64.json
dagood: This convention was introduced in CoreFX during dev/eng to avoid colliding with properties in MSBuild through the env vars VSTS sets. In CoreFX, there are many of them that aren't passed in from pipeline.json.

I agree it would make our lives easier if we also had the convention you're describing.

buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -22,6 +22,29 @@
...
       "alwaysRun": false,
+      "displayName": "Initialize Docker",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        // init-docker.sh was manually copied from dotnet/buildtools, as opposed to restoring buildtools package.
```
MichaelSimons: Is this valid syntax?  I was thinking of a simple comment in the PR description.

ravimeda: Seems to be valid.  Yes, let's remove it, and only have it is PR description.

scripts/dockerrun.sh
```diff
@@ -87,8 +87,44 @@ fi
...
+    return 0
+}
+
 # Build the docker container (will be fast if it is already built)
 echo "Building Docker Container using Dockerfile: $DOCKERFILE"
+
+# Get the name of Docker image.
+image=$(grep -i "^FROM " "$DOCKERFILE" | awk '{ print $2 }')
+
+# Core-Setup uses a predefined image for RHEL.  For all other Linux platforms,
```
MichaelSimons: IMO this comment is not applicable to this specific logic.  I think the important part to capture is "Explicitly pull the base image with retry logic.  This eliminates intermittent failures during docker build caused by failing to retrieve the base image."

buildpipeline/Core-Setup-RHEL7-x64.json
dagood: I think Newtonsoft.Json handles comments ok, but they aren't part of the JSON spec. I don't think we should consider it valid in the future.

MichaelSimons: Right...and I am assuming they get lost if you were to import the definition and re-export it.  Therefore comments like this are going to be a pain in the neck to maintain/keep when using our tooling to import/export the definitions when making changes.

scripts/init-docker.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+
```
dagood: I suggest adding a GitHub link to BuildTools with commit hash and path to keep track of where this came from.

ravimeda: Something like this https://github.com/dotnet/buildtools/commit/de8abc7225c27f0b7dfd35d72deee19f7f672721 ?

dagood: I was thinking `https://github.com/dotnet/buildtools/blob/b5cc6e6ab5f71f6c0be7b730058b426e92528479/src/Microsoft.DotNet.Build.Tasks/PackageFiles/scripts/docker/init-docker.sh`

PR #1702, opened 2017-03-11
[release/1.1.0] Add retry on Docker image pull step in build

ravimeda:
Similar to #1701 But this one is for release/1.1.0

Builds for Linux platforms attempt to pull the corresponding Docker image. If an attempt fails then, there is no retry. This can cause build failure.

[BuildTools ](https://github.com/dotnet/buildtools) offers `init-docker.sh` script that can retry Docker image pull step. Hence introducing this script in this branch.

Note that RHEL has a limitation that requires a RHEL image, and cannot be built using a Dockerfile.

`init-docker.sh` was manually copied, as opposed to restoring buildtools package. Copying the script was a safer approach since:
1. Version of buildtools used in this branch does not have init-docker.sh
2. This is a service branch. Upgrading to a newer version of buildtools has a greater chance for regressions.

@MichaelSimons @chcosta @gkhanna79 please review.

### DIFF
--- a/buildpipeline/Core-Setup-RHEL7-x64.json
+++ b/buildpipeline/Core-Setup-RHEL7-x64.json
@@ -22,6 +22,25 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
+      "displayName": "Initialize Docker",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "scripts/init-docker.sh",
+        "args": "$(PB_DockerImageName)",
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
       "displayName": "Docker cleanup",
       "timeoutInMinutes": 0,
       "task": {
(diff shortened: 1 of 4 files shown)


MichaelSimons:
The comments I made to https://github.com/dotnet/core-setup/pull/1701 apply here as well.

ravimeda:
Thanks @MichaelSimons Addressed the comments through commit `50b8835`.  PTAL.

scripts/init-docker.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
+
+# Version of the script copied was -
```
MichaelSimons: Similar comment should be added to cleanup-docker.sh

buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -197,13 +219,15 @@
...
+      "skipSyncSource": "false",
+      "fetchDepth": "0",
+      "cleanOptions": "0"
     },
     "id": "https://github.com/dotnet/core-setup.git",
     "type": "GitHub",
     "name": "dotnet/core-setup",
     "url": "https://github.com/dotnet/core-setup.git",
-    "defaultBranch": "release/build_defs_1.1.0",
+    "defaultBranch": "BuildConfigurations",
```
MichaelSimons: Can you explain this change?

ravimeda: Excellent! Not sure how it crept in. Let me investigate.

ravimeda: Corresponding file in 1.0.0 look good.  So this one is due to my carelessness 👎  Will fix up.

PR #1723, opened 2017-03-13
Include symbols in core-setup zip / tarballs

chcosta:
This change preserves symbols from packages which are re-packaged into Microsoft.NETCore.App so that we can later include those symbols when producing zip / tarballs which include the shared framework.

I'm also updating buildtools in this change.  I could decouple that if it's a concern, but I updated the branch because I played around with producing a packed Microsoft.NETCore.App package before deciding it was cleaner to simply preserve the symbols outside of the package creation / publish process.

/cc @ellismg @weshaggard @ericstj


chcosta:
I'm not certain how this affects the Update Dependencies stuff in core-setup, i'm looking into that...

build_projects/shared-build-targets-utils/Utils/Crossgen.cs
```diff
@@ -14,6 +14,7 @@ public class Crossgen
         private string _jitVersion;
         private string _crossGenPath;
         private string _targetRID;
+        private static readonly string s_transportPackagePrefix = "transport";
```
ellismg: Why does this have to be a constant instead of just being inlined everywhere?  Do we expect to be changing the prefix?

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -76,13 +76,26 @@ public static string GetSharedFrameworkPublishPath(string outputRootDirectory, s
                 s_sharedFrameworkName,
                 sharedFrameworkNugetVersion);
         }
+        public static string GetNetCoreAppRuntimeLibSymbolsPath(string symbolsRoot, string sharedFrameworkRid, string sharedFrameworkTarget)
```
ellismg: NIT: Add a newline before this method.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -146,6 +159,33 @@ public void PublishSharedFramework(string outputRootDirectory, string commitHash
...
+                pdbFiles.AddRange(Directory.GetFiles(nativePdbPath));
+            }
+            if (Directory.Exists(toolsPdbPath))
+            {
+                pdbFiles.AddRange(Directory.GetFiles(toolsPdbPath));
+            }
+            foreach (string pdbFile in pdbFiles)
+            {
+                string destinationPath = Path.Combine(sharedFrameworkNameAndVersionRoot, Path.GetFileName(pdbFile));
+                if (!File.Exists(destinationPath))
```
ellismg: When would the destination file already exist?  Is this in the incremental build case?  Do we need to worry about two assemblies having the same filename and so we end up getting clashes trying to move all the PDBs into a single folder?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -126,6 +126,24 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
   
+  <Target Name="PreserveSymbols"
+          AfterTargets="CreatePackage">
+    <PropertyGroup>
+      <_PackageSymbolsProjectDir>$(PackageSymbolsBinDir)$(MSBuildProjectName)/</_PackageSymbolsProjectDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <_SymbolsFiles Include="%(File.Identity)" Condition="'%(File.Extension)' == '.pdb' OR '%(File.Extension)' == '$(SymbolFileExtension)'">
```
ellismg: For my own edification, where is `$(SymbolFileExtension)` defined?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -126,6 +126,24 @@
...
+        <DestinationFolder>$(_PackageSymbolsProjectDir)%(File.TargetPath)/</DestinationFolder>
+      </_SymbolsFiles>
+      <_PackageSymbolsTargetFrameworkDir Include="@(_SymbolsFiles->'DestinationFolder')" />
+    </ItemGroup>
+    <MakeDir Directories="%(_SymbolsFiles.DestinationFolder)" />
+    <Message Importance="High" Text="Copy %(_SymbolsFiles.Identity) -> %(_SymbolsFiles.DestinationFolder)" />
```
ellismg: Is there a reason to do this instead of just using the output for the Copy task, which already writes a message?

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
chcosta: It looks like the native assemblies publish their symbols on nix (like dbg files).  This was causing a collision and I thought it prudent to prefer whatever was already present in the publish folder..

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
chcosta: above, in this same file

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
chcosta: I thought I deleted this line, sorry, i'll clean that up.

build_projects/shared-build-targets-utils/Utils/Crossgen.cs
chcosta: I don't think we expect to, but if we have the need to, this simplifies it ever so slightly.

chcosta:
Symchk showed that we're missing symbols for these binaries, but I haven't followed up on it and don't intend to unless someone raises an alarm about it...

```
SYMCHK: libuv.dll            FAILED  - libuv.pdb mismatched or not found
SYMCHK: Microsoft.DiaSymReader.Native.amd64.dll FAILED  - Microsoft.DiaSymReader.Native.amd64.pdb mismatched or not found
SYMCHK: netstandard.dll      FAILED  - netstandard.ni.pdb mismatched or not found
SYMCHK: System.ComponentModel.Composition.dll FAILED  - System.ComponentModel.Composition.ni.pdb mismatched or not found
SYMCHK: System.Core.dll      FAILED  - System.Core.ni.pdb mismatched or not found
SYMCHK: System.Data.dll      FAILED  - System.Data.ni.pdb mismatched or not found
SYMCHK: System.dll           FAILED  - System.ni.pdb mismatched or not found
SYMCHK: System.Drawing.dll   FAILED  - System.Drawing.ni.pdb mismatched or not found
SYMCHK: System.IO.Compression.FileSystem.dll FAILED  - System.IO.Compression.FileSystem.ni.pdb mismatched or not found
SYMCHK: System.Numerics.dll  FAILED  - System.Numerics.ni.pdb mismatched or not found
SYMCHK: System.Runtime.Serialization.dll FAILED  - System.Runtime.Serialization.ni.pdb mismatched or not found
SYMCHK: System.Transactions.dll FAILED  - System.Transactions.ni.pdb mismatched or not found
SYMCHK: System.Web.dll       FAILED  - System.Web.ni.pdb mismatched or not found
SYMCHK: System.Xml.dll       FAILED  - System.Xml.ni.pdb mismatched or not found
SYMCHK: System.Xml.Linq.dll  FAILED  - System.Xml.Linq.ni.pdb mismatched or not found
```

jimcarley:
System.Transactions.dll and pdb were recently changed to System.Transactions.Local.dll and pdb.

What needs to change to get SymChk to recognize this change?

@qizhanMS

chcosta:
It looks like Microsoft.private.corefx.netcoreapp contains System.Transactions.dll, System.Transactions.Local.dll, and System.Transactions.Local.pdb.  So System.Transactions is failing symchk because the pdb isn't there.  It sounds like System.Transactions.dll shouldn't be in the corefx package at all.

qizhanMS:
For System.Transactions, System.Transactions.dll is used for "netfx" only. For .Net core, we use System.Transactions.Local.dll. We need to know which platform the SymChk is running. Also in corefx, we don't generate System.Transactions.dll. Therefore, need to find symbol from external for netfx.

chcosta:
I ran the symchk privately on my Windows machine just to ensure that repackaging the symbols didn't lead to a mismatch somewhere.

weshaggard:
>in corefx, we don't generate System.Transactions.dll. Therefore, need to find symbol from external for netfx.

Just to clarify we do generate a System.Transactions.dll in corefx that ships with .NET Core as it is just a compat shim for folks that to be able to consume binaries that were built against desktop. So we expect both System.Transactions and System.Transactions.Local to be in this package.

PR #1726, opened 2017-03-13
Remove stub pacakge generation

ellismg:
The stub package generation was from a previous model of the build
where we needed to ensure host packages for all supported platforms
existed when we restored test assets.

We no longer have this dependency, we only ever consume packages from
the current platform we are building, so we can remove all of this
code.


ellismg:
@gkhanna79 @eerhardt Thoughts?

eerhardt:
@dotnet-bot test this please

I think some new changes came in that might conflict with this change.

gkhanna79:
> We no longer have this dependency

Is this because we switched to RID split version of M.N.App? If not, then I am not sure what makes them redundant?

eerhardt:
We no longer have this dependency because of this change:  https://github.com/dotnet/core-setup/pull/1088

gkhanna79:
Sounds fine, though Test infra requires fixing as well @ellismg.

weshaggard:
LGTM once green.

fyi @chcosta

ellismg:
> Sounds fine, though Test infra requires fixing as well @ellismg.

Whoops!  I had this fixed locally but I must have dropped the commit as part of a rebase.

eerhardt:
@ellismg - are we still interested in making this change?

ellismg:
@eerhardt Whoops, yes, I still want to do this.  I think I've fixed up the test issues but I'm doing a build locally and then I'll update the PR.

eerhardt:
Thanks, @ellismg!

Issue #1597, opened 2017-02-28
Dynamically lightup of injected packages as part of application closure


gkhanna79:
Today, an application's closure is defined by its corresponding .deps.json file, which is statically computed and known at publish time. However, there are scenarios where certain packages may needed to be injected at runtime for them to become part of the application closure. 

To support this, the entity wishing to inject packages would produce a _lightup.deps.json_ and place it in a folder that mimic's the SharedFX layout. That is, the file will live in _<BaseFolder>\Shared\<SharedFXPackageName>\<SharedFrameworkVersion>\lightup.deps.json_. The host will then be pointed to _BaseFolder_  during application activation (either via an environment variable or command line argument) and will lookup _lightup.deps.json_ using the SharedFX package name/version from the application's deps.json.

The contents of this file will then be part of the deps.json merge step in the host, where the app's deps.json will still override it incase of conflicting packages (like it does today for fx.deps.json).

CC @DamianEdwards @glennc @davidfowl @ramarag @cakine @eerhardt @Petermarcu @richlander @bleroy

eerhardt:
@gkhanna79 - Can you give a more concrete user scenario?

glennc:
@eerhardt When you deploy your application to Azure App Service it will automatically log to the right place, taking into account Azure specific configuration, without you needing to remember to explicitly turn it on or configure it before a problem occurs. That's one end goal that we would enable using this.

gkhanna79:
Initial PR for this is at https://github.com/dotnet/core-setup/pull/1727

gkhanna79:
@davidfowl @DamianEdwards @glennc Do you have any thoughts on this?

DamianEdwards:
@gkhanna79 I think it's fine for this feature to not be a capability of standalone apps. I think it's in keeping with the general ethos that standalone/self-contained apps are deployed that specifically because they don't want outside influences on them.

gkhanna79:
@DamianEdwards, in that case, I would like to ensure that feature does not lightup for standalone scenarios. Are you fine with this enforcement?

Also, can you folks please chime in on https://github.com/dotnet/core-setup/pull/1727#discussion_r106680983 as well?

DamianEdwards:
I'm fine with this not working for stand-alone apps.

PR #1630, opened 2017-03-04
Custom deps feature implemented [IN PROGRESS].

cakine:
It's now possible to inject packages at runtime through custom .deps.json files. The files must be specified through DOTNET_CUSTOM_DEPS environment variable or through --custom-deps command line argument. It's important to notice that the command line argument overwrites the environment variable value.

For instance, if we wanted to load AppB during AppA execution:
dotnet --custom-deps ...\AppB.deps.json AppA.dll

It's possible to specify multiple files:
dotnet --custom-deps ...\AppB.deps.json;...\AppC.deps.json;...\AppD.deps.json AppA.dll

A directory path can be specified:
dotnet --custom-deps base_dir AppA.dll
Every .deps.json file contained in base_dir\shared\fx_name\fx_version will be picked.

It's also possible to mix directory and file paths:
dotnet --custom-deps base_dir1;...\AppB.deps.json;base_dir2 AppA.dll

The changes were manually tested, but it's still necessary to add unit tests to ensure that the new feature correctly follows the expected behavior.

Fixes dotnet/core-setup#1597

Issue #635, opened 2016-11-12
Cannot resolve runtime and native assets from dependency context of a class library


piotrpMSFT:
Moving from https://github.com/dotnet/cli/issues/4005 on behalf of @natemcmaster.

-----

I am attempting to solve the problem of creating an AssemblyLoadContext from a class library project (https://github.com/aspnet/EntityFramework/issues/5320) with only 3 inputs: dependency context, nuget package directory, and application base path. (NB: no project context)
## Problem

The build trims "runtimeTargets" from the deps.json file. In order to execute code from an ALC, we need all runtime and native assets.
## Steps to reproduce

Create and compile netstandard project

``` c#
public class Class1 {}
```

``` json
{
    "dependencies": {
        "System.Data.SqlClient": "4.1.0"
    },
    "frameworks": {
        "netstandard1.3": {}
    }
}
```
## Expected

``` json
  "runtime.native.System.Data.SqlClient.sni/4.0.0": {
        "dependencies": {
          "runtime.win7-x64.runtime.native.System.Data.SqlClient.sni": "4.0.1",
          "runtime.win7-x86.runtime.native.System.Data.SqlClient.sni": "4.0.1"
        }
      },
      "runtime.win7-x64.runtime.native.System.Data.SqlClient.sni/4.0.1": {
        "runtimeTargets": {
          "runtimes/win7-x64/native/sni.dll": {
            "rid": "win7-x64",
            "assetType": "native"
          }
        }
      },
      "runtime.win7-x86.runtime.native.System.Data.SqlClient.sni/4.0.1": {
        "runtimeTargets": {
          "runtimes/win7-x86/native/sni.dll": {
            "rid": "win7-x86",
            "assetType": "native"
          }
        }
      },
```
## Actual

``` json
      "runtime.native.System.Data.SqlClient.sni/4.0.0": {
        "dependencies": {
          "runtime.win7-x64.runtime.native.System.Data.SqlClient.sni": "4.0.1",
          "runtime.win7-x86.runtime.native.System.Data.SqlClient.sni": "4.0.1"
        }
      },

      "runtime.win7-x64.runtime.native.System.Data.SqlClient.sni/4.0.1": {},
      "runtime.win7-x86.runtime.native.System.Data.SqlClient.sni/4.0.1": {},
```

cc @davidfowl

gkhanna79:
[...]

PR #1727, opened 2017-03-13
Lightup Deps.json feature

gkhanna79:
Ported from PR https://github.com/dotnet/core-setup/pull/1630 and fixes https://github.com/dotnet/core-setup/issues/1597


gkhanna79:
@eerhardt @ramarag PTAL

pakrym:
How does this work with `Microsoft.Extensions.DependencyModel` .deps file loading? It only supports discovering two deps files: application and frameworks (see https://github.com/dotnet/core-setup/blob/master/src/Microsoft.Extensions.DependencyModel/DependencyContextPaths.cs)

gkhanna79:
It does not @pakrym. This is intended to be a scoped change only to support Azure lightup scenarios.

pakrym:
There may be unexpected consequences, for example if Azure Lightup brings controller and MVC uses runtime assembly scan to discover controllers it would not get discovered. Just wanted to make sure this is known.

gkhanna79:
I think any assembly injected using this mechanism will not be discoverable via DependencyContext, from my naive understanding of it.

Either ways, I believe you own DependencyContext, so you may want to followup on this.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -918,6 +920,13 @@ int fx_muxer_t::read_config_and_execute(
         roll_fwd_on_no_candidate_fx_val = pal::xtoi(roll_fwd_on_no_candidate_fx.c_str());
     }
 
+    pal::string_t additional_deps_serialized = additional_deps;
+    if (additional_deps_serialized.empty())
+    {
+        // additional_deps_serialized stays empty if DOTNET_ADDITIONAL_DEPS env var is not defined
+        pal::getenv(_X("DOTNET_ADDITIONAL_DEPS"), &additional_deps_serialized);
```
eerhardt: We only support **one** additional deps file?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -403,6 +417,69 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
...
+
+    if (m_portable)
+    {
+        for (pal::string_t json_file : m_additional_deps_files)
+        {
+            m_additional_deps.push_back(std::unique_ptr<deps_json_t>(
+                new deps_json_t(true, json_file, m_fx_deps->get_rid_fallback_graph())));
+        }
+    }
+    else
```
eerhardt: We support additional deps on standalone apps?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -403,6 +417,69 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
...
+
+    if (additional_deps_serialized.empty())
+    {
+        return;
+    }
+
+    pal::string_t additional_deps_path;
+    pal::stringstream_t ss(additional_deps_serialized);
+
+    // Process the delimiter seperated custom deps files.
```
eerhardt: (nit) spelling `seperated`

TestAssets/TestProjects/LightupClient/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+                throw new FileNotFoundException("Exception: Failed to load the lightup assembly!", ex);
+            }
+
+            if (asmGreet == null)
+            {
+                throw new FileNotFoundException("Failed to load the lightup assembly!");
+            }
+            else
+            {
+                Console.WriteLine("Lightup assembly successfully loaded!");
```
eerhardt: Maybe we should call into the `Greet.Hello` method and write its result to the console.  That proves that the assembly is loaded and code can be executed in it.

TestAssets/TestProjects/LightupLib/project.json
```diff
@@ -0,0 +1,386 @@
...
+        "Microsoft.NETCore.App": {
+          "type": "platform",
+          "version": "2.0.0-beta-*"
+        },
+        "Newtonsoft.Json": "9.0.1-beta1",
+        "Microsoft.CodeAnalysis.Analyzers": {
+          "version": "1.1.0",
+          "exclude": "all"
+        },
+        "Microsoft.CSharp": {
```
eerhardt: ?? What is all this?

TestAssets/TestProjects/LightupLib/project.json
```diff
@@ -0,0 +1,386 @@
+{
+
+  "buildOptions": {
+    "emitEntryPoint": false
+  },
+  "dependencies": {},
+  "frameworks": {
+    "netcoreapp2.0": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
```
eerhardt: I think this "lib" should be a netstandard library.

eerhardt: Actually, thinking more about this.  I've spoken with @ericstj in the past about this type of scenario.  See [https://github.com/dotnet/core-setup/issues/635#issuecomment-284516681](https://github.com/dotnet/core-setup/issues/635#issuecomment-284516681) for a write-up on why using a `netstandard` library's .deps.json file at runtime is not correct.

Are the consumers of this feature aware that they should be producing these "Light up" deps.json files using netcore TFMs?

---
In reply to: [106215579](https://github.com/dotnet/core-setup/pull/1727#discussion_r106215579) [](ancestors = 106215579)

src/corehost/cli/fxr/fx_muxer.cpp
gkhanna79: EnvVar supports same semantic for specification as the command line. 

Also the general pattern is to have command line always override the envVar since it is an explicit action on part of the app.

src/corehost/cli/deps_resolver.cpp
gkhanna79: Multiple are supported for both.

TestAssets/TestProjects/LightupLib/project.json
gkhanna79: Yes - the consumers are targeted folks (ASP.NET) and they are aware of it.

TestAssets/TestProjects/LightupLib/project.json
```diff
@@ -0,0 +1,386 @@
+{
```
gkhanna79: > I think the "light up" library should bring in a new dependency that the "LightupClient" doesn't have.

This is what LightupClient is doing - it is bringing in LightLib that it does not have a reference to.

src/corehost/cli/deps_resolver.cpp
gkhanna79: Fixed.

TestAssets/TestProjects/LightupLib/project.json
gkhanna79: Actually, I did test manually with lib as netstandard2.0 binary. I wanted to get the initial review going, hence so. I am making changes to get the library to build in the same manner.

TestAssets/TestProjects/LightupLib/project.json
eerhardt: I'm talking about LightupLibrary bringing in ITS own dependency.  Loading 1 assembly usually works fine (MSBuild does this).  But bringing it that assemblies dependency closure is usually a problem.

anurse:
Could we have a flag in `runtimeconfig.json` to allow me, as an app developer, to turn this feature off completely, even if the environment variables are set?

I feel like some app developers may want to be a little paranoid and disable this loading completely, to avoid having to worry about it at all. It may be a little overkill (and it's certainly not perfect), but it feels a little cleaner to have a complete opt-out that remains in the control of the app developer.

In ASP.NET, our "plugin loading" system which will consume this additional deps file also has this "opt-out" functionality, but there could be cases where even loading the new deps file could be problematic to app developers, so having a global opt-out just seems like a Good Thing here ;).

Ideally, this `runtimeconfig.json` flag would be propagated from an MSBuild Property too ;)

(/cc @DamianEdwards @glennc)

gkhanna79:
Having a flag to override a feature that is opt-in by design sounds extremely odd to me.

DamianEdwards:
It's not opt-in for the app, it's opt-in for the server/hoster.

anurse:
> It's not opt-in for the app, it's opt-in for the server/hoster.

Exactly. If I as an App Developer don't want hosters messing with my loaded app, I think it should be possible to disable this feature.

gkhanna79:
But then why would they need to unconditionally apply to all apps? Why can't the app tell the host not do any "injection of plugins" so that the host only specifies the plugin deps for the apps that have not opted out.

I do not see why the host has to participate in enabling override for something that is optional by design. IMHO, the server/hoster should do the right thing in determining to whom they can apply the flag.

CC @richlander

anurse:
It definitely shouldn't apply to all apps, I'm talking about a way for an _individual_ app to set a setting in their own `runtimeconfig.json` file that tells the host not to load the additional deps files.

gkhanna79:
> It definitely shouldn't apply to all apps, I'm talking about a way for an individual app to set a setting in their own runtimeconfig.json file that tells the host not to load the additional deps files.

My understanding is that an application does not ever know that such an injection may happen - such an injection is _magic._ This was one of the reasons why the idea to include plugin dependencies in the app.deps.json, at the time of the app publish, was turned down since the app would then need to know about this.

What you are proposing above translates to app knowing about this even though we are claiming it is _magical injection_. I would assert that this is contradictory - either you make it explicit for the app to bring in (or not) the dependencies or handle things for them. The opt-in nature of the feature enables the server to exclude the injection from happening for apps that may choose to request.

eerhardt:
> IMHO, the server/hoster should do the right thing in determining to whom they can apply the flag.

Agreed.  If the web host wants to allow individual apps from opting out of this behavior, the host should have the flag to allow individual apps from opting out.  Then the host doesn't set the Environment Variable, and it doesn't pass in `--additional-deps`.  That's how the opt-out should work.  If someone set the variable, then this behavior kicks in.  Don't set the variable if you don't want this behavior.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -369,6 +369,20 @@ bool deps_resolver_t::resolve_tpa_list(
         }
     }
 
+    // If additional deps files were specified that need to be treated as part of the
+    // application, then add them to the mix as well.
+    for (const auto& additional_deps : m_additional_deps)
+    {
+        auto additional_deps_entries = additional_deps->get_entries(deps_entry_t::asset_types::runtime);
```
eerhardt: const additional_deps_entries?

Same for `entry` below.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -403,6 +417,69 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
...
+            if (pal::file_exists(additional_deps_path))
+            {
+                m_additional_deps_files.push_back(additional_deps_path);
+            }
+        }
+        // If it's not a single deps file, we'll assume it's a directory
+        else
+        {
+            // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
+            append_path(&additional_deps_path, _X("shared"));
```
eerhardt: Do we have a test for this? passing in a directory of .deps.json files.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -403,6 +417,69 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
...
+            append_path(&additional_deps_path, fx_name.c_str());
```
eerhardt: What happens when this is a stand alone app?  will fx_name and fx_ver be filled out?  Where do those values comes from in that scenario?

gkhanna79: For standalone apps, as you said, there is no fx_ver/fx_name that is available since that information comes from runtimeconfig.json that does not have it for standalone publish. 

Also, since standalone apps already published in context of a given FX/Ver/RID, it is reasonable to expect the user to know that information and point explicitly to corresponding json files.

gkhanna79: I do think adding a check for standalone to enforce that would be helpful - I will make that change.

eerhardt: My concern is that if Azure App Service only ever sets the base directory, then standalone apps will never get the "light up" injected into it.

TestAssets/TestProjects/LightupLib/project.json
gkhanna79: I have added a test that brings in dependency of LightupLib and also looks up lightup deps.json from shared location.

src/corehost/cli/deps_resolver.cpp
gkhanna79: Fixed

src/corehost/cli/deps_resolver.cpp
gkhanna79: Is there a way to get the M.N.App information from app.deps.json for a standalone app?

eerhardt: Today - no.  The fact that it **doesn't** have this information is what tells the host it is a standalone app.

gkhanna79: Well, deps.json for a standalone app does have M.N.App as a dependency, which contains the name and version. Isn't that the version it is published for? Given that M.N.App is special in many ways across the platform, can this information not be treated specially and extracted from there to allow lookup?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -403,6 +417,75 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
...
+            }
+        }
+    }
+
+    if (m_portable)
+    {
+        for (pal::string_t json_file : m_additional_deps_files)
+        {
+            m_additional_deps.push_back(std::unique_ptr<deps_json_t>(
+                new deps_json_t(true, json_file, m_fx_deps->get_rid_fallback_graph())));
```
eerhardt: I just thought of another issue with standalone apps.  We don't have the `get_rid_fallback_graph()` information for standalone apps.  Which means that we won't be picking native assets out of the cache correctly.  If we are running on `win10` and there is only an asset for `win7`, it won't get picked up.

So most "Lightup" native assets won't be resolved correctly when running a standalone app.

gkhanna79: Seems like you are right - and that is because the assets would be in a custom location like cache (or nuget cache).

eerhardt: Or do we get the rid_fallback_graph from the app.deps.json instead of the fx_deps?

gkhanna79: RID falback only comes from fx.deps.json.

eerhardt: Oh right.  Now I remember, in the app.deps.json, it has the asset path:

```
        "native": {
          "runtimes/win7-x64/native/libuv.dll": {}
        }
```

Which is how the host knows how to load it.  The "fallback" logic happens in NuGet, not at runtime.

src/corehost/cli/deps_resolver.cpp
eerhardt: We could do something like that, but it wouldn't work in all scenarios.  Since the app is standalone, the user could have:

1. Not referenced M.N.App in 1.x.  (I know this isn't a mainline scenario, but it is possible.)
2. Use a pre-release version of M.N.App that the host doesn't have installed, thus doesn't have "light up".deps.json files for it.  For example, someone could take a 2.0.0-preview1 standalone app and publish to Azure today.
3. Use their own build-from-source version of M.N.App, which again the host would have no idea about.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -403,6 +417,75 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
...
+        // 
+        // Standalone apps do not carry that information but are published
+        // incontext of a FX/Ver/RID. Thus, explicit path to deps.json
+        // must be specified for them.
+        else if (m_portable)
+        {
+            // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
+            append_path(&additional_deps_path, _X("shared"));
+            append_path(&additional_deps_path, fx_name.c_str());
+            append_path(&additional_deps_path, fx_ver.c_str());
```
eerhardt: How do patch versions work here?  Has the `fx_ver` all ready been "roll-forwarded" to the latest installed shared framework?  Or if the app explicitly referenced `1.1.0`, is `fx_ver` still have `1.1.0`, even though the app is going to run on `1.1.1`, since that is the highest patch installed?

My concern is:  will Azure need to list out EVERY possible patch version, because some apps were targeting `1.1.0`, and some were targeting `1.1.1`.  So in order to get all apps to "light up", they will need `1.1.0` and `1.1.1` directories.

gkhanna79: We will be looking up the version that app requested for (which is what `fx_ver` corresponds to). There are no roll-forwards in this picture since lightup does not have to co-relate to premise of rollforward (i.e. apply patches).

You are correct, though, that different apps can explicitly callout various patch versions and thus, Azure/Asp.net would need to have folders for all of them. Even if we were to use the roll forwarded version, Azure/ASP.NET would still need to list out all versions. 

@DamianEdwards @glennc @davidfowl What all versions are you planning to list out?

gkhanna79: @DamianEdwards @glennc @davidfowl Any inputs on this?

gkhanna79: @DamianEdwards @davidfowl @glennc Pinging again to comment on this. I would like to get this PR merged in and this is the only outstanding issue to be understood.

gkhanna79:
@eerhardt PR is updated to only support the feature for portable apps.

src/corehost/cli/deps_resolver.cpp
glennc: Unless there is an easy way to make it so that we only need the `1.1.1` version then in this case I am OK with Antares just having files for all the versions. It's less than ideal but I don't think it is a deal breaker. @DamianEdwards is ok with it as well.

gkhanna79: Thanks for the confirmation @glennc. At this point, all open issues are addressed. If I don't hear of anything else, I will merge this change by EOD today.

test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
```diff
@@ -0,0 +1,188 @@
...
+                .And
+                .HaveStdOutContaining("Exception: Failed to load the lightup assembly!");
+        }
+
+        /*
+         * This method is needed to workaround dotnet build not placing the host from the package
+         * graph in the build output.
+         * https://github.com/dotnet/cli/issues/2343
+         */
+        private static void ReplaceTestProjectOutputHostInTestProjectFixture(TestProjectFixture testProjectFixture)
```
eerhardt: This method doesn't appear to be used.  Plus it is duplicated with [https://github.com/dotnet/core-setup/blob/492bb0cc31e1a85affa6e8758466af5cdd798d09/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs#L151](https://github.com/dotnet/core-setup/blob/492bb0cc31e1a85affa6e8758466af5cdd798d09/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs#L151)

TestAssets/TestProjects/LightupLib/project.json
eerhardt: Do we still need all these extra dependencies?  We were able to remove them for the LightupClient project.

gkhanna79: Yes, without them, I have been running into failures in bringing the dependency of the lib.

eerhardt: Oh, right. Because there is no "conflict resolution" between the 2.0 `NETCore.App` and 1.x `System.*` packages.

TestAssets/TestProjects/LightupClient/project.json
```diff
@@ -0,0 +1,17 @@
+{
```
davidfowl: This repo still uses xproj?

eerhardt: Yes.  And it still uses the CLI build framework 😉 

There is an effort to move this repo to MSBuild.  See https://github.com/dotnet/core-setup/tree/buildtools

PR #1728, opened 2017-03-14
Adding apphost.exe

ramarag:
@gkhanna79 @eerhardt @ericstj PTAL

Once i bring the necessary changes from cli, i will remove dotnet.exe from the package

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -63,6 +63,10 @@
     <Dependency Include="@(NETCoreApp11Dependency)">
       <TargetFramework>.NETCoreApp1.1</TargetFramework>
     </Dependency>
+    <!-- references the host packages -->
+    <Dependency Include="Microsoft.NETCore.DotNetAppHost">
+       <Version>$(AppHostFullVersion)</Version>
+    </Dependency>
 
     <!-- Add a placeholder to make sure NuGet understands we still support older versions,
          even though they don't have ref-assets -->
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -470,7 +470,8 @@ bool deps_resolver_t::resolve_probe_dirs(
         {
             // For standalone apps, apphost.exe will be renamed. Do not use the full package name
             // because of rid-fallback could happen (ex: CentOS falling back to RHEL)
-            if (ends_with(entry.library_name, _X(".Microsoft.NETCore.App"), false) && (entry.asset_name == _X("dotnet") || entry.asset_name == _X("apphost")))
+            if (((entry.asset_name == _X("dotnet")) && ends_with(entry.library_name, _X(".Microsoft.NETCore.App"), false)) ||
+                ((entry.asset_name == _X("apphost")) && ends_with(entry.library_name, _X(".Microsoft.NETCore.DotNetAppHost"), false)))
             {
                 trace::warning(_X("Warning: assembly specified in the dependencies manifest was not found -- package: '%s', version: '%s', path: '%s'"), 
                     entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -91,6 +91,11 @@
                                     Condition="Exists('$(DotNetHostBinDir)/dotnet$(SymbolFileExtension)')">
       <IsSymbolFile>true</IsSymbolFile>
     </ArchitectureSpecificNativeFile>
+    <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/apphost$(ApplicationFileExtension)" />
```
gkhanna79: I do not follow the purpose of this change @ramarag. M.N.App does not need to have AppHost within it since the latter is only supposed to be used for standalone scenarios. 

Given that, instead of it being pulled in here, have we considered restoring the ApHost package explicitly, to get AppHost executable, as part of the standalone app publish? Doing this will also enable us to remove the following dependency while maintaining the versioning semantic I was describing to @eerhardt in your other PR:

https://github.com/dotnet/core-setup/blob/master/pkg/projects/Microsoft.NETCore.DotNetHost/Microsoft.NETCore.DotNetHostResolver.pkgproj#L19

ramarag: i might have misunderstood your comments. I thought you said that we may not version AppHost along with M.N.App. So now how do you decide upon the version of AppHost, do we hard code it with sdk ?

Currently we are using the App host from M.N.App, which has been working well so long, my take away from the other thread was to stick to this

gkhanna79: Few things:

- Apphost should never be part of SharedFX. If you look at Senthil's original PR, it was being explicitly deleted from being part of SharedFX. If it is part of SharedFX now, it is likely because of RID split package work that @ericstj was working upon.

- AppHost has nothing to do with SharedFX versioning and is only required to be part of Standalone publish. In this sense, it is sort of an artifact used only by SDK (or SDK experiences).

- Dotnet.exe versioning is the one that needs to be controlled with respect to the SharedFX we generate.

ramarag:
I have removed the host pieces including hostpolicy.dll from M.N.App package and instead am referring to DotNetHostPolicy package.

@eerhardt @ericstj @gkhanna79 can you take a look

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -85,28 +90,6 @@
     </ProjectReference>
   </ItemGroup>
   
-  <ItemGroup Condition="'$(PackageTargetRuntime)' != ''">
-    <ArchitectureSpecificNativeFile Include="$(DotNetHostBinDir)/dotnet$(ApplicationFileExtension)" />
```
eerhardt: I don't see the $(ApplicationFileExtension) property being used any more.  Can it be removed?

eerhardt: The same goes for

LibraryFilePrefix
LibraryFileExtension
SymbolFileExtension

---
In reply to: [106176307](https://github.com/dotnet/core-setup/pull/1728#discussion_r106176307) [](ancestors = 106176307)

src/corehost/cli/deps_resolver.cpp
```diff
@@ -470,7 +470,8 @@ bool deps_resolver_t::resolve_probe_dirs(
         {
             // For standalone apps, apphost.exe will be renamed. Do not use the full package name
             // because of rid-fallback could happen (ex: CentOS falling back to RHEL)
-            if (ends_with(entry.library_name, _X(".Microsoft.NETCore.App"), false) && (entry.asset_name == _X("dotnet") || entry.asset_name == _X("apphost")))
+            if (((entry.asset_name == _X("dotnet")) && (ends_with(entry.library_name, _X(".Microsoft.NETCore.App"), false) || ends_with(entry.library_name, _X(".Microsoft.NETCore.DotNetHost"), false))) ||
```
eerhardt: When does the `dotnet` asset come in the `Microsoft.NETCore.App` library anymore?  It should only come in the `Microsoft.NETCore.DotNetHost` library, right?

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -64,6 +64,11 @@
       <TargetFramework>.NETCoreApp1.1</TargetFramework>
     </Dependency>
 
+    <!-- references the host packages -->
+    <Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
```
gkhanna79: Is there a reason this reference is a Dependency as opposed to P2P reference like https://github.com/dotnet/core-setup/blob/release/1.1.0/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj#L11, which was removed in https://github.com/dotnet/core-setup/commit/81e058c627c963ac996177b36a2ab8ecd66a8095#diff-e2c52444a9da4a5eae65d36e5d3e6aba?

gkhanna79: Also, when you build the repo locally:

1) Can you confirm if the dependency being injected corresponds to the package (of HostPolicy) you built?

2) Can you confirm that the RID split package contains the hostpolicy corresponding to the package that got referenced in (1) above and does not carry hostfxr and dotnet/apphost.exe?

3) Can you confirm that sharedFrameworkPublish folder does not carry hostfxr and dotnet/apphost.exe within shared\M.N.App\<version> folder?

src/corehost/cli/deps_resolver.cpp
gkhanna79: Same question as @eerhardt above.

ramarag: was too conservative, removing

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
ramarag: done

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
ramarag: will remove the files from sharedFrameworkPublish as part of followup PR, when sdk is stabilized

gkhanna79: > will remove the files from sharedFrameworkPublish as part of followup PR, when sdk is stabilized

Is SDK consuming the artifacts within that folder?

PR #1747, opened 2017-03-15
Retry Docker image pull, and cleanup - master

ravimeda:
Recent build failures happened either due to -

1. Attempt to pull Docker image fails, and there is no retry.
2. Low disk space likely because Docker images were not cleaned up by earlier builds.

This fix attempts to add a retry on Docker image pull step. In case of RHEL, which requires a predefined image, `init-docker.sh` script is used to reliably pull down the image on behalf of the build. `init-tools.sh` task restores BuildTools package that brings in `init-docker.sh`.  Hence the task is moved up in the `Core-Setup-RHEL7-x64.json`

@MichaelSimons @dagood @chcosta @gkhanna79 please review.

### DIFF
--- a/buildpipeline/Core-Setup-RHEL7-x64.json
+++ b/buildpipeline/Core-Setup-RHEL7-x64.json
@@ -20,6 +20,44 @@
     },
     {
       "enabled": true,
+      "continueOnError": true,
+      "alwaysRun": true,
+      "displayName": "Initialize tools",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "pkg/init-tools.sh",
+        "args": "",
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Initialize Docker",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "pkg/Tools/scripts/docker/init-docker.sh",
+        "args": "$(DockerImageName)",
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
       "continueOnError": false,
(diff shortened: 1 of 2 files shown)


ravimeda:
For reference, service branch fixes are in PRs #1701 #1702

buildpipeline/Core-Setup-RHEL7-x64.json
```diff
@@ -20,55 +20,74 @@
...
       }
     },
     {
       "enabled": true,
-      "continueOnError": true,
-      "alwaysRun": true,
-      "displayName": "Initialize tools",
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Docker cleanup",
```
MichaelSimons: There is no need for this step with the introduction of init-docker.  It cleans up all containers.

ravimeda: Agree. Let me state here, this post build cleanup happens in other Linux build definitions. However, in those build definitions there is no `init-docker.sh` usage.

ravimeda: Ok.

ravimeda:
Thanks!

scripts/dockerrun.sh
```diff
@@ -87,8 +87,43 @@ fi
 #  VSO
 [ ! -z "$BUILD_BUILDID" ] && DOTNET_BUILD_CONTAINER_NAME="${BUILD_BUILDID}-${BUILD_BUILDNUMBER}"
 
+# Executes a command and retries if it fails.
+# NOTE: This function is the exact copy from init-docker.sh.
+# Reason for not invoking init.docker.sh directly is since that script 
+# also performs cleanup, which we do not want in this case.
```
dagood: To me that seems like a reason to split that part of `init-docker.sh` into, say, `pull-image.sh`, instead. @MichaelSimons

ravimeda: Or simpler `init-docker.sh' can accept a boolean to indicate whether cleanup should be skipped or not.

dagood: I think composition is simpler in this case, but it does seem trivial for now.

MichaelSimons: @dagood, good point on the split.  That being said, wouldn't this also require a change so that the build tools are available at this point in the build for the dev-workflow?  This comes back to the problem that it is not drop dead easy to share common components.  Perhaps the approach of auto-coping the shared scripts into each repo would be the best here.

dagood: @MichaelSimons That's true, they aren't normally restored until midway into the build, inside the container. I think this comment should read something like this, then:

> Reason for not invoking init-docker.sh directly is that it is not readily available at this point in the build.

(Maybe it'd even link to a core-setup issue tracking that.)

I agree that checked-in-script auto-update makes a lot of sense here. I think we could probably get that going before we get BuildTools segmented in a way that works here.

Issue #1757, opened 2017-03-16
dockerrun.sh explicit pre-build image pull uses wrong path: doesn't append /Dockerfile


dagood:
When building using a dockerfile, for example:

    build.sh --skip-prereqs --configuration Release --docker ubuntu.14.04 --targets Default

The build emits this in the logs:

    grep: scripts/docker/fedora.23: Is a directory

[dockerrun.sh#L118](https://github.com/dotnet/core-setup/blob/19829e68d80a6ca249d230a09f7da8a90c0fb60d/scripts/dockerrun.sh#L118) should be trying to grep `scripts/docker/fedora.23/Dockerfile`. The `DOCKERFILE` variable is set to the directory, rather than the Dockerfile itself.

@ravimeda @MichaelSimons

MichaelSimons:
@ravimeda - Is this issue fixed?  Can it be closed?

ravimeda:
Yes, PRs are as follows -

- Master  https://github.com/dotnet/core-setup/pull/1759
- 1.1.0  https://github.com/dotnet/core-setup/pull/1769
- 1.0.0  https://github.com/dotnet/core-setup/pull/1767

PR #1759, opened 2017-03-16
Fix the path to Dockerfile.

ravimeda:
Path to Dockerfile was the containing directory and not the the file. Update the path to include the filename.

https://github.com/dotnet/core-setup/issues/1757

I'll send similar PRs for the two servicing branches.

### DIFF
--- a/scripts/dockerrun.sh
+++ b/scripts/dockerrun.sh
@@ -115,7 +115,7 @@ execute() {
 echo "Building Docker Container using Dockerfile: $DOCKERFILE"
 
 # Get the name of Docker image.
-image=$(grep -i "^FROM " "$DOCKERFILE" | awk '{ print $2 }')
+image=$(grep -i "^FROM " "$DOCKERFILE/Dockerfile" | awk '{ print $2 }')
 
 # Explicitly pull the base image with retry logic. 
 # This eliminates intermittent failures during docker build caused by failing to retrieve the base image.
--- a/scripts/dockerrun.sh
+++ b/scripts/dockerrun.sh
@@ -37,7 +37,7 @@ while [[ $# > 0 ]]; do
             echo "Usage: $0 [-d|--dockerfile <Dockerfile>] [-i|--image <ImageName>] <Command>"
             echo ""
             echo "Options:"
-            echo "  <Dockerfile>    The path to the Dockerfile to use to create the build container"
+            echo "  <Dockerfile>    The path to the folder that contains Dockerfile to use to create the build container"
             echo "  <ImageName>     The name of an existing Dockerfile folder under scripts/docker to use as the Dockerfile"
             echo "  <Command>  The command to run once inside the container (/opt/code is mapped to the repo root; defaults to nothing, which runs the default shell)"
             exit 0
--- a/scripts/dockerrun.sh
+++ b/scripts/dockerrun.sh
@@ -37,7 +37,7 @@ while [[ $# > 0 ]]; do
             echo "Usage: $0 [-d|--dockerfile <Dockerfile>] [-i|--image <ImageName>] <Command>"
             echo ""
             echo "Options:"
-            echo "  <Dockerfile>    The path to the folder that contains Dockerfile to use to create the build container"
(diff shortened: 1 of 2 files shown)


MichaelSimons:
@ravimeda, Can you explain what steps you have taken to test this locally?

ravimeda:
Performed the following in a terminal window:

1. DOCKERFILE="\repos\core-setup\scripts\docker\centos.7".  Note there is no slash at end.
2. Run `image=$(grep -i "^FROM " "$DOCKERFILE" | awk '{ print $2 }')`
3. Above step fails saying path is a directory.
4. Run `image=$(grep -i "^FROM " "$DOCKERFILE/Dockerfile" | awk '{ print $2 }')`
5. Command succeeds. `image` is set to `centos:7.1.1503`

Please let me know if other tests are needed.

ravimeda:
For context, in testing the change in PR #1747, I set `DOCKERFILE` to the file and not the directory.  Given the name of the variable, I assumed it pointed to the file. This makes me wonder if we should rename the variable accordingly.

scripts/dockerrun.sh
```diff
@@ -115,7 +115,7 @@ execute() {
 echo "Building Docker Container using Dockerfile: $DOCKERFILE"
 
 # Get the name of Docker image.
-image=$(grep -i "^FROM " "$DOCKERFILE" | awk '{ print $2 }')
+image=$(grep -i "^FROM " "$DOCKERFILE/Dockerfile" | awk '{ print $2 }')
```
dagood: I think we probably want to support both scenarios: `DOCKERFILE` being a path to the dir, as well as a direct path to the Dockerfile. As far as I can tell, the rest of the scripts work fine with either, and I don't think we should enforce the `Dockerfile` name convention.

I think a solution would be to check for `$DOCKERFILE` being a dir, and if it is, try `$DOCKERFILE/Dockerfile`.

MichaelSimons:
@ravimeda, I think it would be best to test all of your changes as they exist in dockerrun.sh.  To do this I think you should be invoking the script itself.  In this particular case you may decide to comment out the docker run section for testing purposes.  You should also consider how to exercise all code paths such as the retry logic.  You can easily do this by disabling your network or passing in a non-existent docker image.  The bottom line is that we need to make certain our changes are addressing the target problem and adding stability to the build system.

ravimeda:
Thanks @MichaelSimons Tested 'dockerrun.sh' in following cases:

1. No arguments
2. Specified -i, -d or both.  Valid and invalid values.
3. Retry logic for image pulling by turning off internet

Updated the description of Dockerfile argument to indicate that it is path to the folder containing the file.

scripts/dockerrun.sh
ravimeda: `DOCKERFILE` should point to the directory containing 'Dockerfile'. Else the following command will fail.  
`docker build --build-arg USER_ID=$(id -u) -t $DOTNET_BUILD_CONTAINER_TAG $DOCKERFILE`
If we want to pass a file then `-t` should be replaced with `-f`

scripts/dockerrun.sh
```diff
@@ -37,7 +37,7 @@ while [[ $# > 0 ]]; do
             echo "Usage: $0 [-d|--dockerfile <Dockerfile>] [-i|--image <ImageName>] <Command>"
             echo ""
             echo "Options:"
-            echo "  <Dockerfile>    The path to the Dockerfile to use to create the build container"
+            echo "  <Dockerfile>    The path to the folder that contains Dockerfile to use to create the build container"
```
dagood: \<nit> "contains **a** Dockerfile"

scripts/dockerrun.sh
MichaelSimons: @ravimeda - FYI if we want to pass a file then we would need to add the `-f` option and specify the `Name of the Dockerfile (Default is 'PATH/Dockerfile')`  The `-t` option is the `Name and optionally a tag in the 'name:tag' format (default [])`.  That has to remain as it is.

ravimeda:
Thanks!

ravimeda:
https://github.com/dotnet/core-eng/issues/561

PR #1761, opened 2017-03-16
adding tfm to host package reference

ramarag:
following up on https://github.com/dotnet/core-setup/pull/1728

@ericstj @gkhanna79 @eerhardt PTAL?

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -66,6 +66,7 @@
     <!-- references the host packages -->
     <Dependency Include="Microsoft.NETCore.DotNetAppHost">
        <Version>$(AppHostFullVersion)</Version>
+      <TargetFramework>.NETCoreApp2.0</TargetFramework>
     </Dependency>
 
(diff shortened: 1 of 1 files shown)


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -66,6 +66,7 @@
     <!-- references the host packages -->
     <Dependency Include="Microsoft.NETCore.DotNetAppHost">
        <Version>$(AppHostFullVersion)</Version>
+      <TargetFramework>.NETCoreApp2.0</TargetFramework>
```
eerhardt: (Nit) is the white space lined up here?

gkhanna79: Do we want to have similar change for the other host packages as well?

ramarag: this is a temp fix will clean it up when i do the proper one

ramarag: @gkhanna79 basically every nuspec file only selects one group of dependecies. The way we group dependencies for M.N.App is tfm

PR #1798, opened 2017-03-21
Fix intermediate path for package projects

chcosta:
win10 assets were not getting packaged because the intermediate path was not getting set correctly

/cc @ericstj @weshaggard

### DIFF
--- a/pkg/projects/dir.props
+++ b/pkg/projects/dir.props
@@ -3,7 +3,6 @@
   <Import Project="..\dir.props" />
   <PropertyGroup>
     <PackagePlatform>AnyCPU</PackagePlatform>
-    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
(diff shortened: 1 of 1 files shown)


pkg/projects/dir.props
```diff
@@ -3,6 +3,10 @@
   <Import Project="..\dir.props" />
   <PropertyGroup>
     <PackagePlatform>AnyCPU</PackagePlatform>
+    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
```
weshaggard: Is configuration not set somewhere else?

weshaggard:
It isn't clear to me that this fixes the issue or not. Where were these set before your other change?

chcosta:
Yes, those values were present before my change, I'm adding them back - https://github.com/dotnet/core-setup/blob/89a066a05c4e2d98c885c9a376125478dc8492b6/pkg/projects/dir.props

Private validation has shown that this does fix the issue.  

I'll look into whether Configuration is necessary here or not.

PR #1802, opened 2017-03-21
Add assemblies which are not getting signed

chcosta:
Fixes https://github.com/dotnet/core-setup/issues/1801

I don't have a good way to validate this, but looking at the code, it seems reasonable that this was the cause of these assemblies not getting signed.  

/cc @livarcocc @eerhardt @dagood @gkhanna79

### DIFF
--- a/build_projects/dotnet-host-build/sign.proj
+++ b/build_projects/dotnet-host-build/sign.proj
@@ -47,6 +47,18 @@
       <FilesToSign Include="$(IntermediateOutputPath)sharedFrameworkPublish/shared/**/Microsoft.*.dll">
         <Authenticode>$(CertificateId)</Authenticode>
       </FilesToSign>
+      <FilesToSign Include="$(IntermediateOutputPath)sharedFrameworkPublish/shared/**/mscorlib.dll">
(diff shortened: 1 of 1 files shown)


dagood:
@crummel Do you have context on this?

crummel:
@dagood I don't know how these files would have been signed previously as they're not included in the old build definitions but this change looks fine to me.

build_projects/dotnet-host-build/sign.proj
```diff
@@ -47,6 +47,18 @@
       <FilesToSign Include="$(IntermediateOutputPath)sharedFrameworkPublish/shared/**/Microsoft.*.dll">
         <Authenticode>$(CertificateId)</Authenticode>
       </FilesToSign>
+      <FilesToSign Include="$(IntermediateOutputPath)sharedFrameworkPublish/shared/**/mscorlib.dll">
```
weshaggard: @chcosta this is one place to update once you switch to buildtools. We should do the signing similar to corefx but we need to do it after we run crossgen on the binaries but before packaging.

Issue #1816, opened 2017-03-22
Use init-docker in CrossBuild


dagood:
CrossBuild doesn't use `init-docker` before trying to use an image. As a result, the ordinary Docker logic is to pull without retries, which often fails. We need to work out how to properly include this script in the CrossBuild.

It isn't straightforward to include because in `master` we want to use init-docker from BuildTools, but we don't have BuildTools at this point in the build.

PR #1817, opened 2017-03-22
Run CrossBuild as agent user to avoid creating undeletable files

dagood:
Create `docker-as-current-user.sh` for both `run` and `exec`. Use this script in the CrossBuild so that files created in the mapped volume are owned by the agent user. This allows VSTS to clean up.

Since the user isn't root anymore, I had to set `HOME` to something other than `/`. `export`ing the agent account's home into the container results in a `HOME` that doesn't exist, so I added that as another condition to use the custom `HOME` code in build.sh. (Trying to simply clear `HOME` to activate the custom code results in `/` being set, and I didn't want to use the workaround of sending a random garbage value.)

I also moved the mapped container git dir from `/root/...` to `/opt/code/...` for permissions.

@MichaelSimons @gkhanna79 PTAL

I noticed when I was testing that this def doesn't have any pull-retry logic, so it failed to pull the image quite often in my test runs. Created https://github.com/dotnet/core-setup/issues/1816 for this, but it might be tracked somewhere else, too. /cc @ravimeda

### DIFF
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -61,14 +61,15 @@
       "displayName": "Start detached container",
       "timeoutInMinutes": 0,
       "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
         "definitionType": "task"
       },
       "inputs": {
-        "filename": "docker",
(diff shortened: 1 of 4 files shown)


buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -61,32 +61,34 @@
       "displayName": "Start detached container",
```
MichaelSimons: This is not directly related to your change, but I'm not seeing why we have 4 docker steps - start container, build in container, stop container, remove container.  Seems like this could be accomplished in one step.

dagood: Agreed. IIRC this is because this def was copied from CoreFX and tweaked for Core-Setup.

dagood:
@dotnet-bot test Fedora23 x64 Debug Build

PR #1818, opened 2017-03-22
Fix the reading of host_interface_t to be backwards compatible.

AustinWise:
PR #1727 added some new fields to host_interface_t but did not correctly check the size of the struct before reading the fields.


AustinWise:
@gkhanna79 Not a problem, thanks for making this open source! It's fun to follow along and also I get some good ideas to improve code quality at work.

Issue #1722, opened 2017-03-13
Microsoft.NETCore.App 2.0 should no longer depend on Microsoft.NETCore.DotNetHost


eerhardt:
End users will no longer consume the host that is contained in `Microsoft.NETCore.DotNetHost`.  It no longer makes sense to have every .NET Core console app depend on this host, and have tooling that just takes it out of the build/publish content.

Another approach we could take:

No longer build the `Microsoft.NETCore.DotNetAppHost` package.  Instead, just put the "AppHost" in the existing `Microsoft.NETCore.DotNetHost` package, replacing the existing `dotnet.exe` asset with the `apphost.exe`.

Going a step further, if we renamed `apphost.exe` to `dotnet.exe` in the package, then tooling wouldn't even need that much smarts in it.  The tooling can simply say:

1. Rename the resolved `dotnet.exe`file to `AppName.exe`.  (We do this today)
2. `[if TFM >= netcoreapp2.0]` EmbedAppNameInHost.

gkhanna79:
I agree that doing this will definitely simplify things in SDK. However, it also requires thinking a bit more about how Core-Setup platform installers would need to be produced, how we specify the version of DotnetHost package to be used for platform installer generation, to name a few.

> Instead, just put the "AppHost" in the existing Microsoft.NETCore.DotNetHost package

We cannot do this as explained at https://github.com/dotnet/sdk/pull/978#discussion_r105721813. I think better approach would be to have HostFXR have a dependency on AppHost package but with that, we need to resolve the issue described above in this comment.

Issue #1674, opened 2017-03-08
HostPolicy servicing maybe broken after RID split package support


gkhanna79:
See https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/fxr/fx_muxer.cpp#L57

CC @ericstj

gkhanna79:
This is a forcing function to understand how the following will work in 2.0.0:

1) Hammer servicing for M.N.App constituents
2) Ensure that RID split version of M.N.App only gets the locked versions of the host (or only gets the components that are serviced).

CC @weshaggard @eerhardt @ramarag

ericstj:
I think @ramarag was undoing the change that merged the host into NETCore.App.  I don't care much what you do here, just make your scenarios work and understand that there is a desire to flatten the package graph for .NET Core App.  I think you can make either work and there are likely additional changes to make the merge work.

weshaggard:
We need to be sure that we still have a working hammer servicing story for other things beyond just the hostpolicy. We need to make sure it works for everything in the flat NETCore.App package.

ericstj:
That just works.  They are still in packages and the path is based on the package path.  Hammer servicing has always been 1:1 with the package ID, version, and path in the package.

gkhanna79:
@ramarag I am assigning this to you since you are looking into fixing this.

gkhanna79:
To close the loop, have we confirmed that fx.deps.json contains the data expected for this validation?



Thanks,
Gaurav


-------- Original message --------

ramarag:
I have verified the entry in fx.deps.json and it now points to hostpolicy package.

Just for the sake of completeness, the version of hostpolicy in the  fx.deps.json is whatever that was specified in the M.N.App package and the assembly that is put in the shared framework is from the locked version in the core-setup repo, They could Diverge!!!

gkhanna79:
I am not aware of this behavior being valid even before Eric's changes. Where do you see that happening? It is a bug.



Thanks,
Gaurav


-------- Original message --------

Issue #274, opened 2016-09-15
Cleanup dotnet.exe presence from SharedFW


gkhanna79:
CC @blackdwarf

Issue #217, opened 2016-07-16
Fix dependency of DotNetHostResolver to resolve to DotNetAppHost


schellap:
## Steps to reproduce

dotnet new
dotnet restore
## Expected  behavior

The project.lock.json contains only apphost.exe and not dotnet.exe
## Actual behavior

The project.lock.json contains both the executables
## Environment data

`dotnet --info` output:

Blocked on https://github.com/dotnet/cli/issues/3850

schellap:
Remove https://github.com/dotnet/core-setup/blob/cc564bd08a1f76be06b5d4a9bfc6dc0eb593d613/pkg/projects/Microsoft.NETCore.DotNetHost/Microsoft.NETCore.DotNetHostResolver.pkgproj#L16

gkhanna79:
@ramarag I take it you will use this issue to track the packaging resolution work you are doing, right?

CC @eerhardt @ericstj

ramarag:
will be fixed by https://github.com/dotnet/core-setup/pull/1819

PR #1819, opened 2017-03-23
cleanup apphost dependencies

ramarag:
Removes dotnet.exe from M.N.App and disables dotnet.exe from being used as a standalone app host


ramarag:
cc @gkhanna79 @eerhardt PTAL

ramarag:
fixes 
https://github.com/dotnet/core-setup/issues/1722
https://github.com/dotnet/core-setup/issues/1674
https://github.com/dotnet/core-setup/issues/274
https://github.com/dotnet/core-setup/issues/215
https://github.com/dotnet/core-setup/issues/217
https://github.com/dotnet/core-setup/issues/174

setuptools/independent/DepsProcessor/project.json
```diff
@@ -16,6 +16,7 @@
   "frameworks": {
     "netcoreapp1.0": {
       "imports": [
+        "netstandard1.5",
```
eerhardt: This is unncessary.  `netcoreapp1.0` already references `netstandard1.5`.  You don't need to import it explicitly.

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -154,20 +183,17 @@ private static void ReplaceTestProjectOutputHostInTestProjectFixture(TestProject
...
-            if (!File.Exists(testProjectHost) || !File.Exists(testProjectHostPolicy))
+            if (!File.Exists(testProjectHostPolicy))
             {
                 throw new Exception("host or hostpolicy does not exist in test project output. Is this a standalone app?");
             }
 
-            var dotnetHost = Path.Combine(dotnet.GreatestVersionSharedFxPath, $"dotnet{testProjectFixture.ExeExtension}");
             var dotnetHostPolicy = Path.Combine(dotnet.GreatestVersionSharedFxPath, $"{testProjectFixture.SharedLibraryPrefix}hostpolicy{testProjectFixture.SharedLibraryExtension}");
-            var dotnetHostFxr = Path.Combine(dotnet.GreatestVersionSharedFxPath, $"{testProjectFixture.SharedLibraryPrefix}hostfxr{testProjectFixture.SharedLibraryExtension}");
+            var dotnetHostFxr = Path.Combine(dotnet.GreatestVersionHostFxrPath, $"{testProjectFixture.SharedLibraryPrefix}hostfxr{testProjectFixture.SharedLibraryExtension}");
```
eerhardt: This whole method should be removed.  The referenced bug is closed because it was project.json specific. 

```
        /*
         * This method is needed to workaround dotnet build not placing the host from the package
         * graph in the build output.
         * [https://github.com/dotnet/cli/issues/2343](https://github.com/dotnet/cli/issues/2343)
         */
```

This DOES happen now, so this whole workaround should go away.

ramarag: actually this bug is still functional in core-setup repo, i could not find a way to get the project.json tooling with the fix, We should remove it when we upgrade our project system completely

build_projects/dotnet-host-build/sign.proj
```diff
@@ -35,6 +35,9 @@
       <FilesToSign Include="$(OutDir)corehost/**/hostpolicy.dll">
         <Authenticode>$(CertificateId)</Authenticode>
       </FilesToSign>
+      <FilesToSign Include="$(OutDir)corehost/**/dotnet.exe">
```
ramarag: @dagood @crummel is this all that takes for signing dotnet.exe ?

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -234,15 +241,8 @@ private void GenerateRuntimeGraph(DotNetCli dotnetCli, string destinationDeps)
             }
         }
 
-        private void CopyHostArtifactsToSharedFramework(string sharedFrameworkNameAndVersionRoot, string hostFxrVersion)
+        private void CopyHostArtifactsToSharedFramework(string sharedFrameworkNameAndVersionRoot)
         {
-            File.Copy(
-                Path.Combine(_corehostLockedDirectory, HostArtifactNames.DotnetHostBaseName),
```
eerhardt: We put the host in the shared framework in v1 because that was how you specified that you wanted to run the application on a specific shared framework. `C:\Program Files\dotnet\shared\Microsoft.NETCore.App\1.0.1\dotnet.exe` to ensure the app was using the 1.0.1 shared fx.

Do we no longer need this?  Can the version be specified through the command line args?

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
eerhardt: Please update the comment.

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -107,6 +107,35 @@ public void Running_Publish_Output_Standalone_EXE_with_Unbound_AppHost_Fails()
...
+            var appExe = fixture.TestProject.AppExe;
+
+            string hostExeName = $"dotnet{Constants.ExeSuffix}";
+            string builtHost = Path.Combine(RepoDirectories.HostArtifacts, hostExeName);
+            File.Copy(builtHost, appExe, true);
+
+            int exitCode = Command.Create(appExe)
+                .CaptureStdErr()
+                .CaptureStdOut()
+                .Execute(fExpectedToFail:true)
```
eerhardt: Should we also test that an appropriate message is written to stderr?

src/corehost/corehost.cpp
```diff
@@ -126,12 +126,12 @@ pal::string_t resolve_fxr_path(const pal::string_t& own_dir)
...
         }
     }
-    // TODO: Issue #215 Do not allow dotnet to load hostfxr side-by-side.
+    
     pal::string_t fxr_path;
     if (library_exists_in_dir(own_dir, LIBFXR_NAME, &fxr_path))
     {
         trace::info(_X("Resolved fxr [%s]..."), fxr_path.c_str());
-        return fxr_path;
+        trace::error(_X("A fatal error was encountered. This executable can not be used to activate a standalone app"));
```
eerhardt: `can not` is one word: `cannot`.

setuptools/independent/DepsProcessor/RuntimeReference.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    public class RuntimeReference
+    {
+        public static List<RuntimeLibrary> RemoveReferences(IReadOnlyList<RuntimeLibrary> runtimeLibraries, IReadOnlyList<string> refname)
+        {
+            List<RuntimeLibrary> result = new List<RuntimeLibrary>();
+
+            foreach (var runtimeLib in runtimeLibraries)
+            {
+                if (string.IsNullOrEmpty(refname.FirstOrDefault(elem => runtimeLib.Name.ToLower().Contains(elem))))
```
eerhardt: Don't use `ToLower`, use OrdinalIgnoreCase string comparison.  Repeated elsewhere.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
ramarag: @gkhanna79 is this a valid scenario ? if so then we will have to look for a different way to block standalone activation from dotnet.exe

@eerhardt then what was the point in adding the following in runtimeconfig.json ? 
``` 
"framework": {
      "version": "2.0.0-beta-001791-00"
    }
```

setuptools/independent/DepsProcessor/RuntimeReference.cs
eerhardt: Also - don't use `Contains`.  The names need to equal.

setuptools/independent/DepsProcessor/RuntimeReference.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.DotNet.ProjectModel;
+using Microsoft.DotNet.ProjectModel.Graph;
+using Microsoft.Extensions.DependencyModel;
+using NuGet.Frameworks;
+
+namespace DepsProcessor
+{
+    public class RuntimeReference
+    {
+        public static List<RuntimeLibrary> RemoveReferences(IReadOnlyList<RuntimeLibrary> runtimeLibraries, IReadOnlyList<string> refname)
```
eerhardt: `refname` is a bad name for this parameter.  It should be plural, since it is a list.  And it should have full names, no abbreviations.

Since this is references we are talking about, how about just `names`?

setuptools/independent/DepsProcessor/RuntimeReference.cs
ramarag: no we want runtime specific assets to be removed

ramarag: are you suggesting using ?
```
runtimeLib.Name.IndexOf(elem, StringComparison.OrdinalIgnoreCase) >= 0 
```

build_projects/dotnet-host-build/sign.proj
crummel: That looks good.  If you want to confirm you can run the same commands as the official build:
`build_projects\dotnet-host-build\build.ps1 -Configuration Release -Targets Prepare,CompileCoreHost`, then
`msbuild build_projects\dotnet-host-build\sign.proj /t:SignCoreHost /p:Configuration=Release /p:RID=win81-x86`
and look for the "Fake sign target.  Would sign:" message on your local build to make sure it includes dotnet.exe.  I don't know if this applies, but since we were just looking at this in CoreFX, this is Authenticode only, you need a \<StrongName\> line if you want that as well.

setuptools/independent/DepsProcessor/Program.cs
```diff
@@ -75,11 +78,19 @@ public static int Main(string[] args)
                 var graph = manager.Collect(exporter.GetDependencies(LibraryType.Package));
                 var expandedGraph = manager.Expand(graph, runtimes);
 
+                var trimmedRuntimeLibraries = context.RuntimeLibraries;
+
+                if (runtimepackagesToBeRemoved != null && runtimepackagesToBeRemoved.Count > 0)
+                {
+                    runtimepackagesToBeRemoved = runtimepackagesToBeRemoved.Select(x => x.ToLower()).ToList(); ;
```
eerhardt: Please don't use ToLower.  Use ignore case when doing the comparisons.  Also, no reason to call ToList.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -459,12 +459,14 @@ public static BuildTargetResult RestoreLockedCoreHost(BuildTargetContext c)
...
             var lockedHostFxrVersion = hostVersion.LockedHostFxrVersion.ToString();
+            var lockedHostVersion = hostVersion.LockedHostVersion.ToString();
             string currentRid = HostPackageSupportedRids[c.BuildContext.Get<string>("TargetRID")];
             string framework = c.BuildContext.Get<string>("TargetFramework");
 
             string projectJson = $@"{{
   ""dependencies"": {{
-      ""Microsoft.NETCore.DotNetHostResolver"" : ""{lockedHostFxrVersion}""
+      ""Microsoft.NETCore.DotNetHostResolver"" : ""{lockedHostFxrVersion}"",
+      ""Microsoft.NETCore.DotNetHost"" : ""{lockedHostVersion}"",
```
gkhanna79: I presume we are explicitly adding reference to Dotnethost here since it is no longer in the transitive closure of HostResolve, right? If so, it will be good to add a comment indicating the same.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -139,19 +139,18 @@ public void PublishSharedFramework(string outputRootDirectory, string commitHash
             PublishMutationUtilties.CleanPublishOutput(
                 sharedFrameworkNameAndVersionRoot,
                 "framework",
+                new string[] { "framework", "apphost", "hostfxr", "dotnet" },
```
gkhanna79: I do not think that "framework" needs to be in this array as that is the framework.deps.json whose deletion is managed by the two bools you pass in after this array.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
gkhanna79: dotnet.exe from a specific SharedFX version was used to run a _portable_ app against that version of the framework and not a standalone app. I am not aware of any scenario requiring this and, for 2.0, it is fine to remove it. However, we should keep an eye out to understand if anyone is impacted with this breaking change.

gkhanna79: > Can the version be specified through the command line args?

Invoking dotnet.exe that is sibling to the _shared_ folder (which contains M.N.App\<version>) with _-fx-version_ argument and specifying the version will enable the same behavior.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -215,16 +214,24 @@ private void GenerateRuntimeGraph(DotNetCli dotnetCli, string destinationDeps)
             }
             if (!string.IsNullOrEmpty(runtimeGraphGeneratorRuntime))
             {
-                var runtimeGraphGeneratorName = "RuntimeGraphGenerator";
-                var runtimeGraphGeneratorProject = Path.Combine(Dirs.RepoRoot, "setuptools", "independent", runtimeGraphGeneratorName);
-                var runtimeGraphGeneratorOutput = Path.Combine(Dirs.Output, "setuptools", "independent", runtimeGraphGeneratorName);
+                var depsProcessorName = "DepsProcessor";
```
gkhanna79: Is there a specific reason we renamed this? This is a depsprocessor but of a special kind - it generated RID fallback graph.

ramarag: look at the implementation we are doing more than fallback graph generation

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
ramarag: look at the implementation, framework.dll is not deleted

src/corehost/corehost.cpp
```diff
@@ -126,12 +126,12 @@ pal::string_t resolve_fxr_path(const pal::string_t& own_dir)
             return ret_path;
         }
     }
-    // TODO: Issue #215 Do not allow dotnet to load hostfxr side-by-side.
+    
     pal::string_t fxr_path;
     if (library_exists_in_dir(own_dir, LIBFXR_NAME, &fxr_path))
     {
         trace::info(_X("Resolved fxr [%s]..."), fxr_path.c_str());
```
gkhanna79: You should remove the trace::info call since we are not trying to resolve it.

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
gkhanna79: Yes, that is a good idea so that we are not affected by false negative scenarios.

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -107,6 +107,35 @@ public void Running_Publish_Output_Standalone_EXE_with_Unbound_AppHost_Fails()
...
+                .ExitCode;
+
+            if (CurrentPlatform.IsWindows)
```
gkhanna79: Message maybe a better option as @eerhardt suggested above.

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
gkhanna79: > does not place correct hostpolicy is still functional in core-setup repo

Are you suggesting that publishing a standalone app, with sources as local folders that contain the built packages, does not restore the built hostpolicy? If so, we should fix it as part of this PR without adding debt of getting another change in and making it clean in a single go.

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
gkhanna79: I have looked at the implementation @ramarag. Its is an invalid entry to be had in the array - why is it present as there is no such file?

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
gkhanna79: Like I said, it is a special kind of deps processor.

test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
ramarag: The only fix is to update the toolset, this workaround does the right thing

NuGet.Config
```diff
@@ -6,6 +6,7 @@
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" /> 
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json" />
```
weshaggard: what requires corefxlab? We should avoid having this feed in our list.

setuptools/independent/DepsProcessor/project.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "NuGet.RuntimeModel": "3.5.0-rc-1285",
     "NuGet.Versioning": "3.5.0-rc-1285",
-    "System.CommandLine": "0.1.0-e160119-1",
+    "System.CommandLine": "0.1.0-e170322-6",
```
weshaggard: If this is the reason we needed the corefxlab feed then we should either downgrade or move this package to the dotnet-core feed instead.

gkhanna79: @weshaggard I believe this is used to build a tool that is used only during the repo build. Is your concern that using CoreFXLab will break BuildFromSource?

weshaggard: My concern is that we might end up polluting our official core-setup builds from packages on that feed without realizing it. That feed is for experiential things and we need to be certain we don't take a shipping dependency on any of those.

PR #1827, opened 2017-03-23
Remove Linux distro-specific dependencies

eerhardt:
Removing the Linux distros from the runtime.json dependency list. This will cause anyone using self-contained applications to use the portable linux assets by default.

@weshaggard @gkhanna79 @ericstj 

/cc @Petermarcu

### DIFF
--- a/pkg/projects/dir.props
+++ b/pkg/projects/dir.props
@@ -108,6 +108,20 @@
     </BuildRID>
   </ItemGroup>
 
+  <!-- RIDs to remove from the runtime dependencies, so the lineup packages don't have distro specific dependencies.
+       Instead, the linux-x64 runtime will be used on all linux distros by default. 
+       NOTE: There isn't a linux-arm build yet, so we can't remove any arm RIDs. -->
+  <ItemGroup>
+    <LinuxDistroRIDNonDependency Include="debian.8-x64" />
+    <LinuxDistroRIDNonDependency Include="fedora.23-x64" />
+    <LinuxDistroRIDNonDependency Include="fedora.24-x64" />
+    <LinuxDistroRIDNonDependency Include="opensuse.42.1-x64" />
+    <LinuxDistroRIDNonDependency Include="rhel.7-x64" />
+    <LinuxDistroRIDNonDependency Include="ubuntu.14.04-x64" />
+    <LinuxDistroRIDNonDependency Include="ubuntu.16.04-x64" />
+    <LinuxDistroRIDNonDependency Include="ubuntu.16.10-x64" />
+  </ItemGroup>
+
   <ItemGroup Condition="'$(HasRuntimePackages)' != 'false'">
     <_project Include="@(BuildRID)">
       <Platform Condition="'%(Platform)' == ''">x64</Platform>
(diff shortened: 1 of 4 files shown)


pkg/projects/dir.props
```diff
@@ -108,6 +108,20 @@
     </BuildRID>
   </ItemGroup>
 
+  <!-- RIDs to remove from the runtime dependencies, so the lineup packages don't have distro specific dependencies.
+       Instead, the linux-x64 runtime will be used on all linux distros by default. 
+       NOTE: There isn't a linux-arm build yet, so we can't remove any arm RIDs. -->
```
gkhanna79: Can you please file an issue for this? I have brought up linux-arm CoreCLR yesterday (https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.linux-arm.Microsoft.NETCore.Runtime.CoreCLR) and we should track this work.

pkg/projects/dir.props
```diff
@@ -108,6 +108,20 @@
...
+  <ItemGroup>
```
weshaggard: Should we make this inclusive instead of exclusive? That way we don't have anything creep in that we don't want.

weshaggard:
Do we have a follow-up issue to track the clean up of this along with the removal of the specific RIDs once we verify the portable linux stuff works everywhere we want?

pkg/projects/dir.targets
```diff
@@ -89,4 +89,20 @@
               ItemName="Dependency" />
     </MSBuild>
   </Target>
+
+  <!-- Remove the linux distro specific dependencies so the portable linux RID is used by default. -->
+  <Target Name="RemoveLinuxDistroDependencies"
+          Condition="'@(RuntimeDependency)' != ''"
+          AfterTargets="DetermineRuntimeDependencies"
+          Outputs="%(LinuxDistroRIDNonDependency.Identity)">
```
weshaggard: What do we need to do batching here? I would expect we could do this all in the item group below.

pkg/projects/dir.targets
```diff
@@ -89,4 +89,20 @@
...
+    <PropertyGroup>
+      <CurrentLinuxDistroRID>%(LinuxDistroRIDNonDependency.Identity)</CurrentLinuxDistroRID>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <_RuntimeDependencyToRemove Include="@(RuntimeDependency)"
+                                  Condition="'%(RuntimeDependency.TargetRuntime)' == '$(CurrentLinuxDistroRID)'" />
```
weshaggard: Something along the lines of this should allow you to remove the batching from target.
`Condition="'%(RuntimeDependency.TargetRuntime)' == '%(CurrentLinuxDistroRID.Identity)'"`

pkg/projects/dir.props
eerhardt: Changed to inclusive.

pkg/projects/dir.targets
eerhardt: With the change to inclusive, I was able to get rid of the batching.

pkg/projects/dir.props
eerhardt: https://github.com/dotnet/core-setup/issues/1832

eerhardt:
Closing in favor of https://github.com/dotnet/core-setup/pull/1837

PR #1831, opened 2017-03-23
Update the Microsoft.NETCore.UniversalWindowsPlatform package for building 5.3.x versions

zamont:
- Update Release/1.0.0 to create the UWP MP 5.3.2 that has the same dependencies as UWP MP 5.3.1 (including Microsoft.NETCore 5.0.2) plus the new Microsoft.Net.Native.Compiler 1.6.1
- Update descriptions.json to specify Visual Studio 2017 requirement when using the UWP metapackage from VS.


pkg/projects/descriptions.json
```diff
@@ -36,7 +36,7 @@
     },
     {
         "Name": "Microsoft.NETCore.UniversalWindowsPlatform",
-        "Description": "Provides a set of packages that can be used when building Universal Windows applications on .NETCore.",
+        "Description": "Provides a set of packages that can be used when building Universal Windows applications on .NETCore.\nWhen using Visual Studio this package requires Visual Studio 2017 or later.",
```
zamont: For context: Our Microsoft.Net.Native.Compilers package is only supported on Visual Studio 2017 or later. Since the UWP metapackage is now responsible for bringing this package in, we want to make sure developers using Visual Studio 2015 are aware that upgrading to 5.3.x will not work for them.

pkg/projects/Microsoft.NETCore/Microsoft.NETCore.pkgproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <PropertyGroup>
-    <Version>5.0.3</Version>
+    <Version>5.0.2</Version>
```
weshaggard: What version of this are you shipping from the 1.0.0 branch? We should probably ensure we don't conflict with each other.

zamont: We are shipping 5.0.2 in the 1.0.0 branch as well. Our previous UWP MP 5.3.1 contained a reference to Microsoft.NETCore 5.0.2 which is why I made this change. It seems like we don't have much choice except to keep this at 5.0.2 because 5.0.3 is still pre-release.

weshaggard: OK

zamont: Should I temporarily remove my change to pkg/stable.packages.props to make sure we don't produce a Microsoft.NETCore? Or is that unnecessary?

zamont:
@gkhanna79 PTAL

pkg/projects/Microsoft.NETCore/Microsoft.NETCore.pkgproj
weshaggard: I think it is fine to keep it stable here but I do worry that we might end up building and shipping it from both these branches. So please make sure you test this and ensure that the package doesn't end up in the output folder otherwise it will get picked up and published.

zamont: Building this locally results in the following nupkgs in the core-setup/pkg/bin/packages directory:

- Microsoft.NETCore.App.1.1.2-servicing-*
- Microsoft.NETCore.UniversalWindowsPlatform.5.3.2-servicing-*

Is that satisfactory? Also, when should I mark UWP 5.3.2 as stable?

weshaggard: That seems good. You shouldn't mark it stable until we have one official build with the prerelease label that you can test with and sign-off on before making it stable.

zamont: Sounds good. Let me know if you have any more concerns on this PR. I'll also wait for Gaurav to comment on the PR.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json
```diff
@@ -1,36 +1,41 @@
...
-    "System.Net.WebSockets.Client": "4.3.0",
-    "System.Numerics.Vectors.WindowsRuntime": "4.3.0",
-    "System.Reflection.Context": "4.3.0",
-    "System.Runtime.InteropServices.WindowsRuntime": "4.3.0",
-    "System.Runtime.Serialization.Json": "4.3.0",
-    "System.Runtime.Serialization.Primitives": "4.3.0",
-    "System.Runtime.Serialization.Xml": "4.3.0",
-    "System.Runtime.WindowsRuntime": "4.3.0",
-    "System.Runtime.WindowsRuntime.UI.Xaml": "4.3.0",
+    "Microsoft.NETCore.Runtime.CoreCLR": "1.0.3",
```
gkhanna79: Should this be updated to reference CoreCLR package that shipped out of 1.0.x branch?

zamont: For the next release, I think we should. For UWP MP 5.2.3/5.3.1, we did not update the CoreCLR package so they remained at version 1.0.3.

zamont: Unless you think now would be an appropriate time to update this and have a newer CoreCLR package dependency than UWP MP 5.2.3/5.3.1.

PR #1836, opened 2017-03-23
Consolidate linux build defs for easier maintenance

### LINKED ISSUE #1555: Refactor checked-in build definitions to share common code
The checked-in build definitions (at https://github.com/dotnet/core-setup/tree/master/buildpipeline) share a lot of logic and could be consolidated into a smaller number of build definitions to make maintenance easier. This would follow the pattern used by CoreFX.

/cc @wtgodbe @gkhanna79 @chcosta @mmitche @MichaelSimons

MichaelSimons:
I need to make some changes to the build definitions and didn't want to make the changes in so many places so I thought it would be easier to just fix this issue first.

Fixes #1555


buildpipeline/Core-Setup-Linux.json
```diff
@@ -175,12 +181,12 @@
         "SymbolStore"
       ],
       "daysToKeep": 2,
-      "minimumToKeep": 10,
+      "minimumToKeep": 1,
```
dagood: Is this intentional, or just a difference between them? I would prefer 10, although historically I know the CLI-split defs tend to have short retention policies.

buildpipeline/pipeline.json
```diff
@@ -16,109 +16,151 @@
...
           }
         },
         {
-          "Name": "Core-Setup-OSX-x64",
+          "Name": "Core-Setup-Linux",
+          "Parameters": {
+            "PB_DockerOS": "ubuntu.14.04",
+            "REPO_ID": "562fbfe0b2d7d0e0a43780c4",
+            "REPO_USER": "dotnet",
+            "REPO_PASS": "PassedViaPipeBuild",
```
dagood: Interesting, I kinda like that this is here, so the cases where it's needed are actually distinguished. But to be clear, this line has no effect, right?

buildpipeline/Core-Setup-Linux.json
MichaelSimons: All of the other definitions except for Fedora23 were specifying 1 so I felt this was what I should use.  I have no opinion - you tell me.

buildpipeline/pipeline.json
MichaelSimons: I waffle on this, maybe it is more confusing than valuable.  It is coming from Global.

[ (Global)'CLI_NUGET_API_KEY' = '********'(Global)'CONNECTION_STRING' = '********'(Global)'GITHUB_PASSWORD' = '********'(Global)'NUGET_API_KEY' = '********'(Global)'OfficialBuildId' = '20170323-06'(Global)'PB_SignType' = 'real'(Global)'REPO_PASS' = '********'(Global)'SourceBranch' = 'consolidate-linux-build-def'(Global)'SourceVersion' = '68fd9453af79193321a50af58b6c0e7074b9d936''BuildConfiguration' = 'Release''PUBLISH_TO_AZURE_BLOB' = 'false''PB_DockerOS' = 'ubuntu.14.04''REPO_ID' = '562fbfe0b2d7d0e0a43780c4''REPO_USER' = 'dotnet''REPO_SERVER' = 'azure-apt-cat.cloudapp.net''CLI_NUGET_FEED_URL' = 'https://dotnet.myget.org/F/cli-deps' ]

Maybe I should remove it.

MichaelSimons: I removed them.

dagood: Sounds good. If we could filter which pipeline variables go to which builds, then something like this definitely would be valuable, but not yet. (That would have even more value in a case like CoreFX where the NuGet API key is only needed by a single build leg.)

buildpipeline/Core-Setup-Linux.json
dagood: Let's use 10, so we have a reasonable amount of log/artifact history.

MichaelSimons: Done.

buildpipeline/pipeline.json
```diff
@@ -16,109 +16,146 @@
...
             "Type": "build/product/",
             "Platform": "x64"
           }
         },
         {
-          "Name": "Core-Setup-OSX-x64",
+          "Name": "Core-Setup-Linux",
+          "Parameters": {
+            "PB_DockerOS": "ubuntu.14.04",
+            "REPO_ID": "562fbfe0b2d7d0e0a43780c4",
```
gkhanna79: why do some entries have additional parameters than the others?

MichaelSimons: These distros produce and publish Debian packages.  These parameters are used in this process and are only applicable to these legs.  Because some of them are unique, I though this would be the best way to specify them.

PR #1837, opened 2017-03-23
Remove Linux distro-specific dependencies

eerhardt:
Removing the Linux distros from the runtime.json dependency list. This will cause anyone using self-contained applications to use the portable linux assets by default.

### DIFF
--- a/pkg/projects/dir.props
+++ b/pkg/projects/dir.props
@@ -57,19 +57,11 @@
 
   <ItemGroup>
     <OfficialBuildRID Include="alpine.3.4.3-x64" />
-    <OfficialBuildRID Include="debian.8-x64" />
     <OfficialBuildRid Include="debian.8-armel">
       <Platform>armel</Platform>
     </OfficialBuildRid>
-    <OfficialBuildRID Include="fedora.23-x64" />
-    <OfficialBuildRID Include="fedora.24-x64" />
     <OfficialBuildRID Include="linux-x64" />
-    <OfficialBuildRID Include="opensuse.42.1-x64" />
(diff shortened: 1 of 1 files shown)


eerhardt:
@weshaggard @gkhanna79 @ericstj @chcosta PTAL

eerhardt:
> I'm not sure how it fixed your other testing and shared framework issues.

It fixed it because when we build the `Microsoft.NETCore.App` identity package on the native distro (say ubuntu), the runtime.json still contains the `ubuntu` mapping. @ericstj pointed out that this is how it works today on a non-official RID.

I sent out an email to discuss what we want to do officially here for developers who want to use the native assets for the specific distro.

pkg/projects/dir.props
```diff
@@ -57,19 +57,11 @@
 
   <ItemGroup>
     <OfficialBuildRID Include="alpine.3.4.3-x64" />
-    <OfficialBuildRID Include="debian.8-x64" />
     <OfficialBuildRid Include="debian.8-armel">
       <Platform>armel</Platform>
     </OfficialBuildRid>
-    <OfficialBuildRID Include="fedora.23-x64" />
```
gkhanna79: Will this break official Core-Setup builds since they are still invoked for these platforms?

eerhardt: No.  The RID that the machine is building for gets added to this list implicitly.  See https://github.com/dotnet/core-setup/blob/master/pkg/projects/dir.props#L87-L91

The latest fedora 23 build passed:  https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=634406

PR #1838, opened 2017-03-24
Issue #1574: Enable new MSI upgrade code to support proper ref counting for SxS install of HostFxr (1.0)

vivmishra:
- Enable HostFxrMsi to install side by side between releases and build-to-build. This fix is for 1.0.
- This change will now enable the HostFxrMsi to be generated with a new upgrade code with every build.

- Reference Issue: #1574

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -364,9 +364,6 @@ namespace Microsoft.DotNet.Host.Build
             var dotnetCli = DotNetCli.Stage0;
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
             var sharedFrameworkNugetVersion = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
-
-            // For Windows (MSI) we want to pick the MSI version of LockedHostFxrVersion till the time we do not ship an updated hostfxr (to 1.0.1) (#1574)
-            //var hostFxrVersion = hostVersion.LockedHostFxrVersion.ToString();
             var hostFxrVersion = (CurrentPlatform.Current == BuildPlatform.Windows) ? hostVersion.LockedHostFxrMSIVersion.ToString() : hostVersion.LockedHostFxrVersion.ToString();
             var commitHash = c.BuildContext.Get<string>("CommitHash");
 
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -364,8 +364,7 @@ namespace Microsoft.DotNet.Host.Build
             var dotnetCli = DotNetCli.Stage0;
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
             var sharedFrameworkNugetVersion = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
-            //We provide a different locked hostFxr version for Windows (MSIs) instead of LockedHostFxrVersion for issue #1574. This we can update when we have the next binary change for the hostfxr.dll (from 1.0.1)
-            var hostFxrVersion = (CurrentPlatform.Current == BuildPlatform.Windows) ? hostVersion.LockedHostFxrMSIVersion.ToString() : hostVersion.LockedHostFxrVersion.ToString();
+            var hostFxrVersion = hostVersion.GetLockedPlatformInstallerVersion().ToString();
             var commitHash = c.BuildContext.Get<string>("CommitHash");
 
             var sharedFrameworkPublisher = new SharedFrameworkPublisher(
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -364,7 +364,7 @@ namespace Microsoft.DotNet.Host.Build
             var dotnetCli = DotNetCli.Stage0;
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
             var sharedFrameworkNugetVersion = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
-            var hostFxrVersion = hostVersion.GetLockedPlatformInstallerVersion().ToString();
+            var hostFxrVersion = hostVersion.GetLockedHostFXRPlatformInstallerVersion().ToString();
             var commitHash = c.BuildContext.Get<string>("CommitHash");
 
             var sharedFrameworkPublisher = new SharedFrameworkPublisher(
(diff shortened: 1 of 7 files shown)


gkhanna79:
@vivmishra Were you able to confirm if, with this fix, we need to explicitly pass a new version or will refcount take care of the issue?

vivmishra:
The initial change will fix the basic scenario like below but will not fix more complex scenarios where the bad version of the MSI is installed later. To fix that we will have to update HostFXR version (folder).  

Following will work with initial change:
-------------------------------------------
1.	Bad 1.0.4  -> Fixed 1.0.X -> Uninstall any                --- No Issues. Ref counts correctly so install or uninstall in any order is fine. Same with 1.1.
2.	Bad 1.0.4 -> Fixed 1.1.X -> Uninstall any                --- No Issues. 1.0/1.1 HostFXRs now install correctly in different folders now with ref counting. Earlier 1.1 used to trample 1.0, and in reverse install order 1.0 hostfxr was never installed.
3.	Bad 1.1.X -> Fixed 1.0.X->Uninstall any                --- No Issues. 1.0/1.1 HostFXRs now install correctly in different folders now with ref counting. 

Following will fail with the original fix:
-------------------------------------------
4.	Fixed 1.0.X -> Bad 1.0.4 -> Bad 1.1.X                       --- HostFXRs for 1.0  is busted as 1.1 tramples the same. Install order of bad 1.0.4 is a problem here.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -106,7 +106,7 @@ public override string ToString()
         // These versions are used when generating platform installers.
         //
         public bool IsLocked = true; // Set this variable to toggle muxer locking.
-        public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 1, "", "", "", CommitCountString) : LatestHostFxrVersion;
+        public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 2, "", "", "", CommitCountString) : LatestHostFxrVersion;
```
gkhanna79: This issue is really a platform installer bug and should not require a nuget package update (doing so should happen only if we fixed a bug in hostfxr and want to fix a new one).

Can we force a version just for the platform installer?

vivmishra: Reverted the change to LockedHostFxrVersion to 1.0.1. Now the HostFXR MSI name is back to "dotnet-hostfxr-win-x86.1.0.1.msi" and the folder in which the hostfxr installs is updated to "1.0.5". 

Since, 1.0.5 is a new folder the scenario #4 above is fixed.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -72,7 +72,7 @@ public override string ToString()
         // These versions should only be incremented in a servicing release if the package in question
         // is being updated.
         public VerInfo LatestHostVersion => new VerInfo(1, 0, 1, "", "", "", CommitCountString);
-        public VerInfo LatestHostFxrVersion => new VerInfo(1, 0, 1, "", "", "", CommitCountString);
+        public VerInfo LatestHostFxrVersion => new VerInfo(1, 0, 5, "", "", "", CommitCountString);
```
gkhanna79: The point I was making is that we do not change these values unless we are updating the nuget package and that happens if hostfxr.dll is changing.

In this case, we are not fixing any issue in hostfxr.dll, so we dont have to change the version here. Instead we should change the value passed to https://github.com/dotnet/core-setup/pull/1838/files#diff-6a92b73d69f92c6c99098e4630dcb696R142.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -107,6 +107,8 @@ public override string ToString()
         //
         public bool IsLocked = true; // Set this variable to toggle muxer locking.
         public VerInfo LockedHostFxrVersion => IsLocked ? new VerInfo(1, 0, 1, "", "", "", CommitCountString) : LatestHostFxrVersion;
+        //We add a different LockedHostFxrVersion for the MSI as the package needs to deviate for issue #1574. This we can update or remove when we have a binary change for the hostfxr.dll (from 1.0.1)
```
gkhanna79: I would suggest adding a bool here (e.g. fExplicitHostFXRMSIVersion) and set it to true. And then, based upon whether the bool is true or not, either set it to explicit version or the version of LockedHostFXRVersion (this will allow us to simply use this bool to make a switch in future without needing to fix the locations you are changing right now).

vivmishra: New changes have been submitted based on the feedback. 

To summarize: 
 - All the 4 scenarios are now covered.
 - We have added a new variable LockedHostFxrMSIVersion (1.0.5) for the duration when the MSI package has to deviate from others. When we take a fix to the hostfxr.dll itself in future then we can revert to both MSI and other packages to contain the same version (1.0.6 or later) of the package for hostfxr.
- The MSI build related variables have been renamed to correctly reflect the internal MSI version and the MSI file version. 
- For publishing we will continue to publish the packages to Azure blob to the same base version (1.0.1) folder.

build_projects/dotnet-host-build/CompileTargets.cs
```diff
@@ -364,7 +364,8 @@ public static BuildTargetResult PublishSharedFrameworkAndSharedHost(BuildTargetC
             var dotnetCli = DotNetCli.Stage0;
             var hostVersion = c.BuildContext.Get<HostVersion>("HostVersion");
             var sharedFrameworkNugetVersion = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
-            var hostFxrVersion = hostVersion.LockedHostFxrVersion.ToString();
+            //We provide a different locked hostFxr version for Windows (MSIs) instead of LockedHostFxrVersion for issue #1574. This we can update when we have the next binary change for the hostfxr.dll (from 1.0.1)
```
gkhanna79: I would suggest adding comment to the tune of "Get version to be used for MSI generation" to explain the overall purpose of this change.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -107,6 +107,9 @@ public override string ToString()
...
+        public bool fExplicitHostFXRMSIVersion = true; // Set this variable to override LatestHostFxrVersion for MSI packages only. This can be set to false when we fall back to using just LockedHostFxrVersion for all.
```
gkhanna79: The last sentence should be something like "This should be set to false when we no longer need to override the MSI version to be different from the HostFXR nuget package version".

Issue #1575, opened 2017-02-22
Uninstalling 1.1.0-preview1 runtime breaks 1.0.1 runtime - Fix in 1.1.x


rakeshsinghranchi:
## Steps to reproduce

1. Install [1.0.1 runtime](https://go.microsoft.com/fwlink/?LinkID=827515)
(I see that the `C:\Program Files\dotnet\host\fxr\1.0.1` directory gets created)
2. Install [1.1.0-preview1 runtime](https://go.microsoft.com/fwlink/?LinkID=831444)
3. Uninstall 1.1.0-preview1 runtime
(There is no longer a `C:\Program Files\dotnet\host` directory)
## Expected  behavior
```
$ dotnet

Microsoft .NET Core Shared Framework Host

  Version  : 1.0.1
  Build    : cee57bf6c981237d80aa1631cfe83cb9ba329f12

Usage: dotnet [common-options] [[options] path-to-application]

Common Options:
  --help                           Display .NET Core Shared Framework Host help.
  --version                        Display .NET Core Shared Framework Host version.

Options:
  --fx-version <version>           Version of the installed Shared Framework to use to run the application.
  --additionalprobingpath <path>   Path containing probing policy and assemblies to probe for.

Path to Application:
  The path to a .NET Core managed application, dll or exe file to execute.

If you are debugging the Shared Framework Host, set 'COREHOST_TRACE' to '1' in your environment.

To get started on developing applications for .NET Core, install .NET SDK from:
  http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
```

## Actual behavior
```
$ dotnet
A fatal error occured, the required library hostfxr.dll could not be found at C:\Program Files\dotnet
```

## Environment data
`dotnet --info` output:
```
A fatal error occured, the required library hostfxr.dll could not be found at C:\Program Files\dotnet
```

rakeshsinghranchi:
Refer similar fix in 2.0.0 : https://github.com/dotnet/core-setup/pull/1571#pullrequestreview-23143709

vivmishra:
Fixed and merged.

PR #1839, opened 2017-03-24
Issue #1575: Enable new MSI upgrade code to support proper ref counting for SxS install of HostFxr (1.1)

vivmishra:
- Enable HostFxrMsi to install side by side between releases and build-to-build. This fix is for 1.1.
- This change will now enable the HostFxrMsi to be generated with a new upgrade code with every build.

- Reference Issue: #1575

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -421,7 +421,7 @@ namespace Microsoft.DotNet.Host.Build
             string sharedFrameworkNugetVersion = c.BuildContext.Get<string>("SharedFrameworkNugetVersion");
             string sharedFrameworkRid = c.BuildContext.Get<string>("TargetRID");
             string sharedFrameworkTarget = c.BuildContext.Get<string>("TargetFramework");
-            var hostFxrVersion = hostVersion.GetLockedPlatformInstallerVersion().ToString();
+            var hostFxrVersion = hostVersion.GetLockedHostFXRPlatformInstallerVersion().ToString();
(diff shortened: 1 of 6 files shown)


vivmishra:
**Following scenario will work this change:**
1. Bad 1.0.4 -> Fixed 1.0.X -> Uninstall any --- No Issues. Ref counts correctly so install or uninstall in any order is fine. Same with 1.1.
2. Bad 1.0.4 -> Fixed 1.1.X -> Uninstall any --- No Issues. 1.0/1.1 HostFXRs now install correctly in different folders now with ref counting. Earlier 1.1 used to trample 1.0, and in reverse install order 1.0 hostfxr was never installed.
3. Bad 1.1.X -> Fixed 1.0.X->Uninstall any --- No Issues. 1.0/1.1 HostFXRs now install correctly in different folders now with ref counting.
4. Fixed 1.0.X -> Bad 1.0.4 -> Bad 1.1.X --- HostFXRs for 1.0 is busted as 1.1 tramples the same. Install order of bad 1.0.4 is a problem here.

**To summarize the fix:**
1. All the 4 scenarios are covered.
2. We have added a new variable LockedHostFxrMSIVersion (1.1.2) for the duration when the MSI package has to deviate from others. When we take a fix to the hostfxr.dll itself in future then we can revert to both MSI and other packages to contain the same version (1.1.3 or later) of the package for hostfxr.
3. The MSI build related variables have been renamed to correctly reflect the internal MSI version and the MSI file version.
4. For publishing we will continue to publish the packages to Azure blob to the same base version (1.1.0) folder.

gkhanna79:

PR #1841, opened 2017-03-24
Remove unneccessary 1.0 compat keys from 2.0

Petermarcu:
This key was there to deal with a backwards versioning problem in 1.0. Not needed in 2.0 anymore.


Petermarcu:
@rakeshsinghranchi @ericstj

PR #1845, opened 2017-03-24
Add uniform docker init and cleanup as well as VSTS Agent cleanup

### LINKED ISSUE #1816: Use init-docker in CrossBuild

MichaelSimons:
Fixes #1816 
Related dotnet/core-eng#589

Modified the cross build, rhel, portable linux, and linux build definitions to do the following:

1. Setup a sandbox location to run init-tools on the host machine at the beginning of each build.  The sandbox location is used so that it does not interfere with the build within a container step that volume maps in the repo.  The sandbox location also allows the post build cleanup logic to reference the tools.
2. Run Init-Docker at the beginning of all builds.
3. Created a VSTS Agent cleanup step at the end of the build.
4. Refactored the cross build definition to use a simpler Docker pattern for building.
5. Renamed all of the containers to indicate which build definition they are for.  This helps in tracking down any resources left behind on the build machines.

### DIFF
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -58,7 +58,27 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Start detached container",
+      "displayName": "Create host machine tools sandbox",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "5bfb729a-a7c8-4a78-a7c3-8d717bb7c13c",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "SourceFolder": "pkg",
+        "Contents": "**",
+        "TargetFolder": "$(DockerHost_Sandbox)",
+        "CleanTargetFolder": "false",
+        "OverWrite": "false",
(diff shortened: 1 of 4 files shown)


buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -286,6 +288,15 @@
     "SourceVersion": {
       "value": "",
       "allowOverride": true
+    },
+    "DockerHost_Sandbox": {
+      "value": "$(Build.StagingDirectory)/HostSandbox"
```
dagood: I'm conflicted on this, vs. `Build.BinariesDirectory` (and enabling cleanup of that dir). The doc has a pretty specific purpose for `Build.StagingDirectory`, but `Binaries` is pretty specific, too. I'm fine leaving it Staging, but I wish there was a clear fit.

MichaelSimons: Agreed this is not optimal.  Leaving as is.

PR #1854, opened 2017-03-27
Mark the UWP MP 5.3.2 package as stable

zamont:
Mark the UWP MP 5.3.2 package as stable


zamont:
@weshaggard @gkhanna79 I've tested the 5.3.2-servicing metapackage that was produced on Friday and am ready to mark this as stable

PR #1869, opened 2017-03-28
[release/1.1.0] Only publish when destination deliberately specified

dagood:
This lets us configure which publish steps are performed. Before, there were only two modes: publish nothing or publish everything. "Publish everything" fails if some publish destinations are not configured. After this PR, unconfigured destinations are skipped. This will let us set up a new internal build that only publishes to blob storage. Official builds will continue unchanged.

---

This will require some PipeBuild definition changes to pass in the parameters that enable publishing.

Converted `PublishCoreHostPackagesToFeed` to targets so it can share the `EnvironmentAttribute` logic.

Workaround for https://github.com/dotnet/corefx/issues/17614: `EnvironmentAttribute` now evaluates "ENV_VAR_EMPTY_WORKAROUND" the same as null/empty.

I had this branch in `dotnet` so I could run test builds on it.

### DIFF
--- a/buildpipeline/Core-Setup-Ubuntu16.04-x64.json
+++ b/buildpipeline/Core-Setup-Ubuntu16.04-x64.json
@@ -142,7 +142,7 @@
       "value": "ENV_VAR_EMPTY_WORKAROUND"
     },
     "REPO_ID": {
-      "value": "ENV_VAR_EMPTY_WORKAROUND"
+      "value": "575f40f3797ef7280505232f"
     },
     "REPO_USER": {
       "value": "ENV_VAR_EMPTY_WORKAROUND"
--- a/buildpipeline/Core-Setup-Ubuntu16.04-x64.json
+++ b/buildpipeline/Core-Setup-Ubuntu16.04-x64.json
@@ -126,32 +126,32 @@
       "value": "--skip-prereqs --configuration $(BuildConfiguration)  --docker ubuntu.16.04 --targets Default"
     },
     "CONNECTION_STRING": {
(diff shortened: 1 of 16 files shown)


dagood:
I'm going to make the corresponding changes to the existing 1.1.0 pipebuild def tomorrow morning and merge this. The pipebuild def changes look backward compatible so we can revert this PR if it causes problems. If that works, I'll move on to https://github.com/dotnet/core-setup/pull/1870 [port to 1.0.0].

buildpipeline/Core-Setup-Ubuntu16.04-x64.json
```diff
@@ -126,32 +126,32 @@
...
-      "value": "PassedViaPipeBuild"
+      "value": "ENV_VAR_EMPTY_WORKAROUND"
     },
     "GITHUB_PASSWORD": {
-      "value": "PassedViaPipeBuild"
+      "value": "ENV_VAR_EMPTY_WORKAROUND"
     },
     "REPO_ID": {
-      "value": "575f40f3797ef7280505232f"
+      "value": "ENV_VAR_EMPTY_WORKAROUND"
```
dagood: I noticed `REPO_ID` is different for each distro, so I have to add it back to the legs.

dagood:
Comparing the build triggered by merging this PR and the previous one, it seems to have worked fine. It published to the NuGet feeds and linux package repos it's supposed to. It also updated the versions repo as expected: https://github.com/dotnet/versions/commit/74a985a966a22a737d09cd5abc03dc7d830e58f5.

PR #1870, opened 2017-03-28
[release/1.0.0] Only publish when destination deliberately specified

dagood:
This will require some PipeBuild definition changes to pass in the parameters that enable publishing.

See https://github.com/dotnet/core-setup/pull/1869


dagood:
> You may want to validate a build behaves as expected in 1.1 rather than trying to fix things in 2 different places in parallel.

Yeah, you're right. This is a cherry-pick with some conflict resolution, no need to have it reviewed now.

gkhanna79:
@dagood Can you please give me a summary of this change?

dagood:
@gkhanna79 Added one to the description of https://github.com/dotnet/core-setup/pull/1869.

dagood:
I've updated this port from the 1.1.0 PR and made the (backward-compatible) 1.0.0 PipeBuild definition changes.

PR #1887, opened 2017-03-30
Exclude symbols from runtime packages

### LINKED ISSUE #1884: runtime.RID.Microsoft.NETCore.App packages contain all the symbols
## Steps to reproduce
Look at the latest package on https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.opensuse.42.1-x64.Microsoft.NETCore.App

## Expected  behavior
It should be a reasonable size.

## Actual behavior
It is 90MB, since it contains all the symbol files.

/cc @ericstj @gkhanna79 @chcosta

chcosta:
Fixes #1884 

Prevent symbols from being include in lib packages.

/cc @eerhardt @ericstj

### DIFF
--- a/pkg/projects/dir.props
+++ b/pkg/projects/dir.props
@@ -49,6 +49,11 @@
     </Otherwise>
   </Choose>
 
+  <ItemGroup>
+    <AdditionalLibPackageExcludes Condition="'$(SymbolFileExtension)' != ''" Include="%2A%2A\%2A$(SymbolFileExtension)" />
+    <AdditionalSymbolPackageExcludes Condition="'$(LibraryFileExtension)' != ''" Include="%2A%2A\%2A.a;%2A%2A\%2A$(LibraryFileExtension)" />
+  </ItemGroup>
+
   <PropertyGroup Condition="'$(PackageRID)' == ''">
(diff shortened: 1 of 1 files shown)


chcosta:
I've validated this on Windows. Feel free to merge it or not since it won't regress functionality. I won't  be be able to do Linux validation today though

pkg/projects/dir.props
```diff
@@ -49,6 +49,11 @@
     </Otherwise>
   </Choose>
 
+  <ItemGroup>
+    <AdditionalLibPackageExcludes Condition="'$(SymbolFileExtension)' != ''" Include="%2A%2A\%2A$(SymbolFileExtension)" />
```
eerhardt: Why `%2A`?  Why not use `*`?

ericstj: Because that would trigger globbing.  This is supposed to represent the literal pattern string.  Not files that live on disk.

eerhardt: That is totally not obvious to someone just looking at this code.

In all honesty, it would probably be best to not use `Include/ItemSpec` in MSBuild in these situations.  There are just too many bugs that pop up because MSBuild has certain semantics on `Include`.

Issue #1886, opened 2017-03-29
DependencyContextLoader should load all dependency files it is given to it


eerhardt:
## Steps to reproduce
After https://github.com/dotnet/sdk/pull/1053 is merged, it is possible for an app to run on the shared framework, but not be "portable".  In one of these apps, if you inspect the `DependencyContext.Default`

## Expected  behavior
It should have all the framework assemblies, since it runs on the shared framework.

## Actual behavior
It doesn't contain the framework assemblies.  This is because of [this check for IsPortable](https://github.com/dotnet/core-setup/blob/master/src/Microsoft.Extensions.DependencyModel/DependencyContextLoader.cs#L74) which is not true for these types of apps.

## Notes
We should just load all the dependency files passed through AppContext by the host.  Then when other deps files get merged into the app (see #1597 and https://github.com/dotnet/core-setup/issues/624#issuecomment-288245699), the DependencyModel automatically knows about these new deps files.

Issue #1017, opened 2017-01-04
DependencyContext.Default throws when Assembly.GetEntryAssembly returns null


jbogard:
I have some code that depends on DependencyContext.Default (an extension to IServiceCollection that scans DependencyContext.Default for assemblies. MVC Core gets around this by assuming the IHostingEnvironment.ApplicationName is the entry point assembly name.

MVC Core tooling assumes that DependencyContext.Default can be null (but it can't).

## Steps to reproduce
Create an AppDomain and do not set its entry assembly, for example in a unit test:

```c#
[Fact]
public void ShouldNotThrowForNullEntryAssembly()
{
    DependencyContext defaultContext = null;
    Action act = () => defaultContext = DependencyContext.Default;

    act.ShouldNotThrow();
    defaultContext().NotBeNull();
}
```

## Expected behavior
That it not throw, perhaps with a null object representation that can be Merged with other assemblies

## Actual behavior
`DependencyContextLoader.Load` throws a NullReferenceException

## Environment data
Windows 10 x64

moozzyk:
@gkhanna79 - I don't think I am the right assignee for this bug

gkhanna79:
@schellap @pakrym Do either of you own this in Core-Setup repo?

schellap:
@eerhardt any idea who this should go to?

eerhardt:
> any idea who this should go to?

Probably someone from the CLI team (either me or @livarcocc). Or possibly @pakrym, if he wants to take it.

eerhardt:
I'm fixing this to return "null" when `Assembly.GetEntryAssembly()` returns null.

Unfortunately, a "null object" can't really be returned because of the way the `Merge` logic works.  It uses the TargetInfo and CompilationOptions of the context it was invoked on.  See https://github.com/dotnet/core-setup/blob/master/src/Microsoft.Extensions.DependencyModel/DependencyContext.cs#L72-L78.

Callers will need to check for `null` returning from `DependencyContext.Default`, and if they want to "merge" with a different context, do a null check, and just use the other context instead.

Issue #622, opened 2016-11-10
DependencyModel uses wrong package cache environment variable name


piotrpMSFT:
Moved from https://github.com/dotnet/cli/issues/3034 for @pakrym. 

I verified the issue still exists, though the code has moved:
https://github.com/dotnet/core-setup/blob/master/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs#L75

----

In PR https://github.com/dotnet/cli/pull/2156 `DOTNET_PACKAGES_CACHE` variable name was changes to `DOTNET_HOSTING_OPTIMIZATION_CACHE`
https://github.com/dotnet/cli/pull/2156/files#diff-76ff25fcb7bb5726c0738b88686cf16dR111

But DependencyModel usage (https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs#L71) was not updated, so MVC Razor compiler won't use package cache for view compilation when only `DOTNET_HOSTING_OPTIMIZATION_CACHE` is set.

@schellap @muratg

gkhanna79:
@piotrpMSFT @muratg @pakrym Who owns DependencyModel that lives in Core-Setup so that this issue can be routed to them?

muratg:
Assigning to @gkhanna79 and @eerhardt based on our discussion last week.

eerhardt:
To follow up here, this environment variable completely went away in 2.0, thus the whole resolver is now deleted.

The new support for resolving assemblies from the `dotnet cache` locations will be completed by https://github.com/dotnet/core-setup/issues/624#issuecomment-288245699

Issue #624, opened 2016-11-10
Support nuget's "fallback package target" feature in DependencyContext


piotrpMSFT:
Moving from https://github.com/dotnet/cli/issues/3569 on behalf of @muratg.

NuGet now supports a secondary package target to improve `dotnet restore` performance. The idea is to populate the secondary folder for Azure Web Sites to improve kudu restore performance.

This WI is to complete the work and add the support for this secondary package folder in DependencyContext as well. This would fix the `dotnet run` scenario for MVC runtime view compilation as well as `dotnet test` scenarios. 

This can be done for CLI preview3. We have punted the work from CLI preview2 in the core war today.

cc @piotrpMSFT @Petermarcu  @Eilon @davidfowl

gkhanna79:
Who is the owner for DependencyContext to whom this bug should be routed?

CC @schellap

muratg:
@piotrpMSFT @Petermarcu  is there someone on your teams that can take this on? @pakrym can do a code review, but he'll be busy with ASP.NET tasks.

gkhanna79:
@muratg Historically, @piotrpMSFT 's team and folks in ASP.NET have been involved in design and implementation of this component, IIRC. Will @pakrym not be able to bring this up for 1.2.0 Preview next year?

muratg:
Assigning to @gkhanna79 and @eerhardt based on our discussion last week.

@gkhanna79 This is one of the bugs that we need in 2.0.0.

eerhardt:
Assigning to @ramarag to fix along with https://github.com/dotnet/sdk/issues/1002, which is basically the same issue.

@ramarag - the high-level plan here is that if the native dotnet host knows how to load assemblies from a given place, then DependencyModel should be able to load assemblies from that place as well.  So in my mind, the work necessary is:

1. The dotnet host will pass all `additional-probing-paths` it knows about into managed code, probably through the `AppContext.GetData` API where it already passes information to managed code.
2. DependencyModel has the ability to look at these additional probing paths, and resolve compilation assemblies from them, just like it does for `NUGET_PACKAGES` and `%USERPROFILE%\.nuget\packages` today.
3. The locations of any `dotnet cache` locations should be passed to managed code as well, so DependencyModel can probe those locations.
[...]

PR #1907, opened 2017-03-31
A few simple DependencyModel bugs

### LINKED ISSUE #617: Remove DependencyModel RuntimeLibrary hacks for xunit
https://github.com/dotnet/cli/pull/1932 will introduce some small hacks to DependencyModel in order to allow coreclr.xunit to run. Once that's in, fix xunit and remove the hacks

Moved from https://github.com/dotnet/cli/issues/1998 on behalf of @anurse.

eerhardt:
I batched up 4 little bugs to the DependencyModel, since I thought it would be easier to review them all at once, instead of bugging everyone 4 separate times.

It may be easiest to review commit-by-commit.

Fixes #617, #622, #1017, and #1886


global.json
```diff
@@ -1,3 +1,4 @@
 {
-    "projects": [ "src", "test", "setuptools", "build_projects" ]
+  "projects": [ "src", "test", "setuptools", "build_projects" ],
```
davidfowl: 2 spaces!? 😄

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
```diff
@@ -42,37 +41,19 @@ public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> ass
 
             if (!string.IsNullOrEmpty(_packageCacheDirectory))
             {
-                var hashSplitterPos = library.Hash.IndexOf('-');
```
pakrym: Why is hash checking code gone?

global.json
eerhardt: Blame VS 2015

src/Microsoft.Extensions.DependencyModel/DependencyContextPaths.cs
```diff
@@ -7,32 +12,49 @@ namespace Microsoft.Extensions.DependencyModel
     internal class DependencyContextPaths
     {
         private static readonly string DepsFilesProperty = "APP_CONTEXT_DEPS_FILES";
```
pakrym: Why don't we use only `APP_CONTEXT_DEPS_FILES`? I was thinking maybe it worth it to return just `string[]` from this class, it would make sure we are loading `.desp` files in the same order that muxer does.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
eerhardt: Looking more at that change, this environment variable went away completely.  :(

src/Microsoft.Extensions.DependencyModel/DependencyContextPaths.cs
eerhardt: The reason is because there is a place where we use the SharedRuntime property to find the directory of the framework. And I didn't like that we assume a specific index was the SharedFx deps file. There is an AppContext parameter for the SharedFx deps file, so I felt it was correct to use it instead. Now the only order we assume is that the first deps file is the entry assembly's.

But I can change it so the SharedRuntime property is populated from the FX_Deps parameter. And in the loader, we don't use the SharedRuntime property.  Instead use the full string[] that was passed by APP_CONTEXT_DEPS_FILES in that order.

The SharedFx deps file would be in both, assuming the host passed it in both.

Thoughts?

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
eerhardt: The full fix for the issue of the host and DependencyModel being out of sync should be addressed by https://github.com/dotnet/core-setup/issues/624#issuecomment-288245699

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
```diff
@@ -42,37 +41,19 @@ public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> ass
...
+                    return true;
                 }
             }
             return false;
         }
 
         internal static string GetDefaultPackageCacheDirectory(IEnvironment environment)
         {
-            return environment.GetEnvironmentVariable("DOTNET_PACKAGES_CACHE");
+            return environment.GetEnvironmentVariable("DOTNET_HOSTING_OPTIMIZATION_CACHE");
```
gkhanna79: This should not be required for 2.0 since the cache feature is replacing it.

src/Microsoft.Extensions.DependencyModel/DependencyContextPaths.cs
pakrym: 👍 I like it.

test/Microsoft.Extensions.DependencyModel.Tests/PackageCacheResolverTest.cs
```diff
@@ -171,7 +181,7 @@ public void UsesPathProperty()
         private IEnvironment GetDefaultEnvironment()
         {
             return EnvironmentMockBuilder.Create()
-                .AddVariable("DOTNET_PACKAGES_CACHE", CachePath)
+                .AddVariable("DOTNET_HOSTING_OPTIMIZATION_CACHE", CachePath)
```
gkhanna79: This should not be required.

test/Microsoft.Extensions.DependencyModel.Tests/PackageCacheResolverTest.cs
```diff
@@ -40,19 +40,28 @@ public void SkipsNonPackage()
...
-                .Contain(library.Hash)
-                .And.Contain(library.Name);
+            resolver.TryResolveAssemblyPaths(library, assemblies)
+                .Should().BeTrue();
+            assemblies.Should().NotBeEmpty();
         }
 
         [Fact]
-        public void ChecksHashFile()
+        public void IgnoresHashFile()
```
gkhanna79: Is this hash check used for Antares scenario (and thus, we may not need it)?

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCacheCompilationAssemblyResolver.cs
eerhardt: I see that now.  

My plan is that I delete this whole file in this change. Then when @ramarag makes his change (referenced above) he will make this whole scenario work correctly with the new cache feature.

eerhardt:
I've addressed all feedback.  PTAL and let me know what you think.

PR #1909, opened 2017-03-31
Extract ID and VersionID independent of their order

### LINKED ISSUE #1903: RID calculation in CoreHost is broken
On a debian linux I see the following when enabling `COREHOST_TRACE=1`:

```
HostRID is 8debian.-x64
Falling back to base HostRID: linux-x64
```

Meanwhile dotnet --info shows this:

```
Runtime Environment:
 OS Name:     debian
 OS Version:  8
 OS Platform: Linux
 RID:         debian.8-x64
```

This causes RID fallback to fail when the assets don't exactly match the target RID. I ran into this because we moved libuv out of the shared framework (which is the reason we never hit this before). 

## Repro steps

1. Get a machine with docker
2. `docker run -it --rm microsoft/dotnet-nightly:2.0.0-beta-sdk`
3. Run `dotnet --info` confirm the RID is `debian.8-x64`
4. Now make a new console application using `mkdir app && cd app && dotnet new console`
5. Run `dotnet restore && dotnet build`
6. Set COREHOST_TRACE via `export COREHOST_TRACE=1`
7. Run the applicaion and pipe strerr to a file `dotnet bin/Debug/netcoreapp2.0/app.dll 2> out.txt`
8. Execute `cat out.txt | grep HostRID`

## Expected

```
cat out.txt | grep HostRID
HostRID is debian.8-x64
```

## Actual

```
cat out.txt | grep HostRID
HostRID is 8debian.-x64
Falling back to base HostRID: linux-x64
```

/cc @gkhanna79

gkhanna79:
@eerhardt PTAL

Fixes #1903 

CC @davidfowl

### DIFF
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -383,13 +383,7 @@ pal::string_t pal::get_current_os_rid_platform()
 
                 if (fFoundID && fFoundVersion)
                 {
-                    // Form the RID using the values we got
-                    ridOS.append(valID);
-                    ridOS.append(_X("."));
-                    ridOS.append(valVersionID);
-
-                    // Remove any double-quotes
-                    ridOS = trim_quotes(ridOS);
+                    // We have everything we need to form the RID - break out of the loop.
                     break;
                 }
 
@@ -399,6 +393,23 @@ pal::string_t pal::get_current_os_rid_platform()
 
             // Close the file now that we are done with it.
             fsVersionFile.close();
+
+            if (fFoundID)
+            {
+                ridOS.append(valID);
+            }
+            
+            if (fFoundVersion)
+            {
+                ridOS.append(_X("."));
+                ridOS.append(valVersionID);
+            }
+
+            if (fFoundID || fFoundVersion)
+            {
+                // Remove any double-quotes
+                ridOS = trim_quotes(ridOS);
+            }
         }
     }
 
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -347,7 +347,9 @@ pal::string_t pal::get_current_os_rid_platform()
         {
             pal::string_t line;
             pal::string_t strID(_X("ID="));
+            pal::string_t valID;
             pal::string_t strVersionID(_X("VERSION_ID="));
+            pal::string_t valVersionID;
 
             bool fFoundID = false, fFoundVersion = false;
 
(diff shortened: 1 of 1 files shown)


src/corehost/common/pal.unix.cpp
```diff
@@ -357,23 +359,35 @@ pal::string_t pal::get_current_os_rid_platform()
...
-                    fFoundVersion = true;
+                    size_t pos = line.find(strVersionID);
+                    if ((pos != std::string::npos) && (pos == 0))
+                    {
+                        valVersionID.append(line.substr(11));
+                        fFoundVersion = true;
+                    }
                 }
 
                 if (fFoundID && fFoundVersion)
```
eerhardt: This doesn't match the managed code exactly.

https://github.com/dotnet/core-setup/blob/master/src/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs#L34

Each part will get appended whether the other part is found or not.

gkhanna79: Sure - but I am not sure we want to maintain consistency in missing data. Are you suggesting we do the same (no place anything if either one is not found) in managed code as well?

eerhardt: I think having this logic as consistent as possible is a thing we should strive for. It's bad if managed code thinks the RID is different than what the host thinks.

gkhanna79: I think that will be true no matter what. Computed RID is not always same as RID the host will use for it's fallback graph (e.g. unknown distro - computed RID will be different from what host will use - linux-*).

Can DependencyModel determine, given a RID, whether it is in fallback graph or not? If not, then we cannot have that consistency (which does not exist in positive case in 1.x as well), let alone having consistency for bad case.

eerhardt: Managed code can see the fallback graph from the FX_DEPS file that is passed through AppContext.

Regardless, the code that computes the RID for a machine should be consistent. If it is ever different it is going to cause a debugging head-ache.  `dotnet --info` will print one RID, which may appear to work.  But the host uses a different RID, and no one will have any idea why.

gkhanna79: Even if I make the native code compute the RID in the manner managed code does, for cases when we are on an unknown distro or on a platform whose version we do not know, host will use a RID that is different from the computed one. In that case, I do not follow your concern around debugging.

gkhanna79: Done the mimic - check it out.

Issue #1824, opened 2017-03-23
HTTP 400 "The specified block list is invalid." when publishing to blob storage


dagood:
In the official build, we sometimes get a 400 error when trying to publish artifacts onto Azure blob storage. https://devdiv.visualstudio.com/DevDiv/_build?buildId=631398:

```
<RequestResult>
  <HTTPStatusCode>400</HTTPStatusCode>
  <HttpStatusMessage>The specified block list is invalid.</HttpStatusMessage>
  [...]
  <Error>
    <Code>InvalidBlockList</Code>
    <Message>The specified block list is invalid.
```

```
...
at System.Threading.Tasks.Task.Wait()
at Microsoft.DotNet.Cli.Build.AzurePublisher.PublishFile(String blob, String file)
at Microsoft.DotNet.Host.Build.PublishTargets.PublishCoreHostPackages(BuildTargetContext c)
at Microsoft.DotNet.Cli.Build.Framework.BuildContext.ExecTarget(BuildTarget target)
```

---

Approaches:

 1. Remove our parallel overwrites: https://github.com/dotnet/core-setup/issues/1994.
 2. Catch and ignore this specific exception.
 3. Use more semaphores to avoid concurrent writes.

markwilkie:
Do you know if there’s an issue tracking retries (or other hardening) for this Davis?

dagood:
There isn't, but it sounds reasonable to fix this that way because it seems pretty random. This issue could track it, since it's specific to the core-setup build system.

dagood:
@MattGal pointed out that Core-Setup is using a very old version of `WindowsAzure.Storage`, `6.2.2-preview` (Dec 2015). Latest: `8.1.1` (Feb 2017). We should try updating it.

dagood:
The first build with https://github.com/dotnet/core-setup/pull/1935 merged worked. I'm going to proactively close this, in case this issue never repros.

dagood:
https://github.com/dotnet/core-setup/issues/1994 is the underlying problem, but I think we can safely ignore this exception: it indicates that some other build successfully copied the same file, so the publish that encountered the error was unnecessary in the first place.

dagood:
To make sure "2. Catch and ignore this specific exception." is a valid solution I checked two of the repros we've had to make sure the files weren't corrupted by the conflict. They weren't. I also wasn't able to find any web search results for `corrupt The specified block list is invalid` or anything similar.

eerhardt:
[...]

PR #1935, opened 2017-04-03
Upgrade WindowsAzure.Storage to 8.1.1

dagood:
Potentially a fix for https://github.com/dotnet/core-setup/issues/1824.

After the upgrade, I had to change the `UploadFromFileAsync` call to remove a parameter that no longer exists and add in a `System.Threading.Thread` reference to get the build projects to compile.

### DIFF
--- a/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
+++ b/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
@@ -55,7 +55,7 @@ namespace Microsoft.DotNet.Cli.Build
         public void PublishFile(string blob, string file)
         {
             CloudBlockBlob blockBlob = _blobContainer.GetBlockBlobReference(blob);
-            blockBlob.UploadFromFileAsync(file, FileMode.Open).Wait();
+            blockBlob.UploadFromFileAsync(file).Wait();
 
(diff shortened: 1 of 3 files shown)


build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -55,7 +55,7 @@ public void PublishArchive(string archiveFile, string channel, string version)
         public void PublishFile(string blob, string file)
         {
             CloudBlockBlob blockBlob = _blobContainer.GetBlockBlobReference(blob);
-            blockBlob.UploadFromFileAsync(file, FileMode.Open).Wait();
+            blockBlob.UploadFromFileAsync(file).Wait();
```
gkhanna79: Are the two calls equivalent? Also, do you think it is worth doing a private official build with this?

dagood: Yes, the `FileMode` param was removed and hard-coded to `Open`. I couldn't find anything official mentioning it, so I checked the IL and confirmed that other than that, it's identical.

I'm not aware of any known-good way to do a private build on the master branch right now, so I don't think it's worth it. It seems a lot more reasonable to make the change and revert if needed.

PR #1967, opened 2017-04-05
Generating UAP Metapackage for the new framework

joperezr:
PTAL:  @weshaggard @ericstj @chcosta 

FYI: @joshfree 

This will create the UAP Metapackage for the new uap10.1 framework. @ericstj what do we want to do with the version of the UAP Metapackage? for now I'm just using 5.3.0 which is what we use to have, but I'm sure that we will want to change that to be a prerelease instead similar to what we do for netcoreapp.


pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.builds
```diff
@@ -1,11 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <HasRuntimePackages>false</HasRuntimePackages>
+  	<OverridePackageRIDs>true</OverridePackageRIDs>
```
weshaggard: nit: spacing.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json.template
```diff
@@ -0,0 +1,16 @@
+{
+  "dependencies": {
+    "Microsoft.Net.Native.Compiler": "1.6.0",
+    "Microsoft.Private.CoreFx.UAP": "4.4.0-preview1-25205-01",
```
weshaggard: What about the other packages like WCF?

weshaggard: We should probably diff the file list before and after this change of the meta-package to ensure there are no missing files (it is of course expected there will be new files).

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
```diff
@@ -1,39 +1,71 @@
...
+    <OmitDependencies>true</OmitDependencies>
+    <SkipValidatePackage>true</SkipValidatePackage>
+    <ShouldCreateLayout>false</ShouldCreateLayout>
+    <PropsFile>$(IntermediateOutputPath)$(MSBuildProjectName).props</PropsFile>
+    <PlatformManifestFile>$(IntermediateOutputPath)$(MSBuildProjectName).PlatformManifest.txt</PlatformManifestFile>
+    <PreventImplementationReference Condition="'$(PackageTargetRuntime)' != ''">true</PreventImplementationReference>
+    <!-- During an official build when we can guarantee that all RID-specific dependencies have been built,
+         restore all of those dependencies and gather the prospective content of the RID-specific Core.App
+         packages.  This is needed so that we have a complete platform manifest in the shipping version of
+         the Microsoft.NETCore.App (RID-agnostic/identity package). -->
```
weshaggard: Microsoft.NETCore.App ?

weshaggard:
cc @zamont 

@joperezr as part of this change can you also delete Microsoft.NETCore.pkgproj as it is no longer needed after this change.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
```diff
@@ -0,0 +1,125 @@
...
+
+    <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
+      <!-- RID-specific: include all runtime files -->
+      <_FilesToPackage Include="@(ReferenceCopyLocalPaths)">
+        <!-- ResolveNugetPackageAssets doesn't preserve the asset type (native), 
+             calculate it by looking for native in the path -->
+        <IsNative Condition="$([System.String]::new('%(Identity)').ToLowerInvariant().Replace('\', '/').Contains('/native/'))">true</IsNative>
+      </_FilesToPackage>
+      <_FilesToPackage>
+        <TargetPath Condition="'%(_FilesToPackage.IsNative)' != 'true'">runtimes/$(NuGetRuntimeIdentifier)/lib/$(PackageTargetFramework)</TargetPath>
```
weshaggard: Do we actually have any native binaries? Is clrcompression perhaps the only one?

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json.template
joperezr: that's a good point, I'm not sure how the netcoreapp is pulling WCF in, since I basically tried to mimic what we do over there. cc @ericstj

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
joperezr: good catch

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.builds
joperezr: will fix

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
joperezr: as of now we have clrcompression.dll and Microsoft.DiaSymReader.Native.amd64.dll

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json.template
ericstj: NETCore.App doesn't include WCF, its only part of UWP.  You should be including WCF here as a redist in the flat meta-package.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
ericstj: I don't think DiaSymReader should be part of UWP, but double check what was there before to be sure.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
```diff
@@ -0,0 +1,125 @@
...
+        <TargetPath Condition="'%(_FilesToPackage.IsNative)' == 'true'">runtimes/$(NuGetRuntimeIdentifier)/native</TargetPath>
+      </_FilesToPackage>
+      <_FilesToPackage Include="$(_runtimePackagePath)/tools/*.*">
```
ericstj: This was for crossgen we can probably exclude that from UWP package.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
```diff
@@ -0,0 +1,125 @@
...
+  <Target Name="GetFilesToPackage" DependsOnTargets="ResolveNuGetPackages" Returns="@(FilesToPackage)">
+    <ItemGroup  Condition="'$(NuGetRuntimeIdentifier)' != ''">
+      <_runtimeCLR Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.FileName)' == 'coreclr' OR '%(ReferenceCopyLocalPaths.FileName)' == 'libcoreclr'" />
+    </ItemGroup>
+    <!-- start with all the files resolved -->
+    <PropertyGroup Condition="'@(_runtimeCLR)' != ''">
+      <_runtimePackageId>%(_runtimeCLR.NuGetPackageId)</_runtimePackageId>
+      <_runtimePackageVersion>%(_runtimeCLR.NuGetPackageVersion)</_runtimePackageVersion>
+      <_runtimePackagePath>$(PackagesDir)/$(_runtimePackageId)/$(_runtimePackageVersion)</_runtimePackagePath>
+    </PropertyGroup>
```
ericstj: If we do remove crossgen as I suggest below, you can get rid of this stuff.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
ericstj: If this does hang around you might want to be a bit more intentional about not doing this for aot.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
```diff
@@ -0,0 +1,125 @@
...
+      <_filesByPackagePath Include="@(_primaryFilesByPackagePath);@(_secondaryFilesByPackagePath)" />
+
+      <!-- transform back to source -->
+      <FilesToPackage Include="@(_filesByPackagePath->'%(SourcePath)')" />
+    </ItemGroup>
+
+    <!-- simple check to make sure we don't accidentally pull an old 
+         package that should have been merged to the primary package -->
+    <ItemGroup>
+      <PermittedSystemPackages Include="System.ValueTuple" />
```
ericstj: I think we can remove this exception (and likely can do so for netcoreapp too).  I believe ValueTuple is being included in flat package now.

pkg/projects/dir.props
```diff
@@ -96,6 +96,38 @@
     </BuildRID>
   </ItemGroup>
 
+  <!-- ================== Section for Universal Windows Platform package ================== -->
```
ericstj: This doesn't need to be here.  Just put it in some dir.props in the UWP folder.  Also you could just set both of these before importing.  The PackageRID property group already has a condition to do nothing if set,  you could put the same sort of thing on the OfficialBuildRID itemgroup.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/project.json.template
joperezr: Fixed.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
joperezr: it wasn't there before, so I'll go ahead and remove.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
joperezr: Copy/Paste mistake, we don't need it so I'll go ahead and remove.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
joperezr: Done.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
joperezr: Removed, I did however need to add all of the WCF packages in here. Also, note that I had to do a filter in order to remove System.IO.Compression from the closure which was getting brought in by WCF

pkg/projects/dir.props
joperezr: If I move it, then it would require a major refactoring because there are `ITEMGROUPS` bellow this that depend on what I'm setting, and there are `ITEMGROUPS` above this that are required in order to define my itemgroup

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
```diff
@@ -27,28 +27,20 @@
...
-    </PropertyGroup>
-
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
       <!-- RID-specific: include all runtime files -->
       <_FilesToPackage Include="@(ReferenceCopyLocalPaths)">
         <!-- ResolveNugetPackageAssets doesn't preserve the asset type (native), 
              calculate it by looking for native in the path -->
         <IsNative Condition="$([System.String]::new('%(Identity)').ToLowerInvariant().Replace('\', '/').Contains('/native/'))">true</IsNative>
       </_FilesToPackage>
+      <!-- Exclude System.IO.Compression files as they get pulled into the closure by WCF -->
```
ericstj: We shouldn't need to do this.  If something from Compression is winning it means we're either missing something from our rid-specific corefx package or there is a bug here: https://github.com/dotnet/core-setup/pull/1967/commits/1a99f0846db37aac669c4f023314d9b9e7ca80b1#diff-e5fdfbf0dc7f5037aa5538f270da8d8eR65

ericstj: Actually I think I understand what's going on here.  We previously shipped a clrcompression.dll in the lib folder.  https://github.com/dotnet/corefx/commit/d0ae14f6136fbc4b3af428b71523758c983c24f0

This means that the IsNative condition is actually the problem https://github.com/dotnet/core-setup/pull/1967/commits/1a99f0846db37aac669c4f023314d9b9e7ca80b1#diff-e5fdfbf0dc7f5037aa5538f270da8d8eR35

What I'd do is add another line above.
```
<IsNative Condition="'%(FileName)%(Extension)' == 'clrcompression.dll'">true</IsNative>

```

joperezr: Done

pkg/projects/dir.props
joperezr: This is fixed now, as we agreed offline, I separated the package rids into individual props files that can be overwritten.

pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.depproj
gkhanna79: Correct, there is no crossgen usage for UWP.

joperezr:
@weshaggard @ericstj Is this good to merge?

joperezr:
@ericstj where should we log an issue about that?

ericstj:
I'd just email @joshfree and @MattWhilden about it.  Not sure where this targets file lives after the Microsoft.Net.Native.Compiler package refactoring.

Issue #1559, opened 2017-02-17
Port Nuget lowercasing change to 1.0.X


gkhanna79:
@rrelyea @emgarten This is for 1.0.5

CC @eerhardt @piotrpMSFT @livarcocc @srivatsn @leecow

joelverhagen:
This is my commit to 1.1.0 that implemented this: https://github.com/dotnet/core-setup/commit/e5d476944462bce31a61852cf8fa1692f7683e7f

eerhardt:
`1.0.5` is open for check-in, correct?  Should we make this change now?

gkhanna79:
Yes, we should get it in now.

emgarten:
Great! I'll take a look at merging Joel's change.

emgarten:
@gkhanna79 @eerhardt which branch should this go into for 1.0.5? master?

gkhanna79:
release/1.0.0 and once your PR is prepared, please send email to shiproom.

PR #1976, opened 2017-04-06
Update HostPolicy package

gkhanna79:
Porting of the nuget lowering change (https://github.com/dotnet/core-setup/issues/1559) modified Hostpolicy.dll but the PR didnt incorporate the change to increment its version and produce a new package. 

This change fixes the issue. I have confirmed that the produced M.N.App has a dependency on HostPolicy package 1.0.5.

@eerhardt PTAL

CC @emgarten


gkhanna79:
OSX failure looks like infra issue.

Issue #1950, opened 2017-04-04
Installers are not published to Latest blob path for 1.0.x and 1.1.x servicing 


rakeshsinghranchi:
## Steps to reproduce
Go to Readme page : https://github.com/dotnet/core-setup 
Install 1.1.x or 1.0.x installer for Windows 
## Expected  behavior
Installer should be from latest daily build

## Actual behavior
Downloaded installer is of 1.1.0 RTM and 1.0.0 RTM 

## Environment data

dagood:
The blobs in `dotnet/release/1.1.0/Installers/Latest/` have last modified dates that correspond with the latest 1.1.x build. This makes me think the build is using the version `1.1.0` rather than `1.1.x-servicing-x-x` as the copy source, making it copy the 1.1.0 build to Latest during every build.

It looks like `SharedFrameworkNugetVersion` having the wrong value in https://github.com/dotnet/core-setup/blob/a29f0361ef3f30a50dccbc961d6e44533f587234/build_projects/dotnet-host-build/PublishTargets.cs#L114 it would cause this issue, but I can't find any indication in the rest of the build logs that it's wrong.

I want to add build logging to Azure uploads to see what it's actually doing, first.

dagood:
From the logs, I see that the first `Installers` copy actually works, e.g.

> Copying blob 'release/1.1.0/Installers/**1.1.2-servicing-001480-00**/dotnet-win-x64.1.1.2-servicing-001480-00.exe' to 'release/1.1.0/Installers/Latest/dotnet-win-x64.latest.exe'

Then it gets stomped over by the [second `Installers` copy](https://github.com/dotnet/core-setup/blob/7d020e76689fa0e312a27294c7c6bc035e5511a7/build_projects/dotnet-host-build/PublishTargets.cs#L117):

> Copying blob 'release/1.1.0/Installers/**1.1.0**/dotnet-win-x64.1.1.0.exe' to 'release/1.1.0/Installers/Latest/dotnet-win-x64.latest.exe'

That source `1.1.0` is `SharedHostNugetVersion`
 * [which is set by](https://github.com/dotnet/core-setup/blob/7d020e76689fa0e312a27294c7c6bc035e5511a7/build_projects/dotnet-host-build/PublishTargets.cs#L40) `c.BuildContext.Get<HostVersion>("HostVersion").LockedHostVersion`
[...]

PR #1981, opened 2017-04-06
[release/1.1.0] Swap host and framework installer copies to Latest

dagood:
This tweaks the overwrite order to keep the just-built installers, if there is overlap.

(cherry picked from commit 5647411ba1687b50b9e2b522d1551c68daaba8f1)

See https://github.com/dotnet/core-setup/issues/1950


dagood:
OSX CI is waiting on a download, and it seems likely it will fail like the 1.0.0 version of this PR did. Merging now.

PR #1982, opened 2017-04-06
Pretend existing VCRedist is successful install.

gkhanna79:
Port of https://github.com/dotnet/cli/pull/6265.

@Petermarcu PTAL

### DIFF
--- a/packaging/windows/sharedframework/bundle.wxs
+++ b/packaging/windows/sharedframework/bundle.wxs
@@ -42,6 +42,7 @@
                        ProductName="$(var.Crt_ProductName)" 
                        Size="$(var.Crt_Size)" 
                        Version="$(var.Crt_Version)" />
+        <ExitCode Behavior="success" Value="1638" />
       </ExePackage>
       <MsiPackage SourceFile="$(var.SharedFXMsiSourcePath)">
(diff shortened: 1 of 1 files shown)


Petermarcu:
LGTM

Petermarcu:
Go ahead and merge since it only impacts Windows and we know Mac CI is having trouble today.

packaging/windows/sharedframework/bundle.wxs
```diff
@@ -42,6 +42,7 @@
                        ProductName="$(var.Crt_ProductName)" 
                        Size="$(var.Crt_Size)" 
                        Version="$(var.Crt_Version)" />
+        <ExitCode Behavior="success" Value="1638" />
```
rakeshsinghranchi: Burn should not try to install higher version at all. Detect Condition ( https://github.com/gkhanna79/core-setup/blob/2f91187544afc8d39a14f1e181fde394accf1bf2/packaging/windows/sharedframework/bundle.wxs#L32) is expected to return success and skip installation of the item ( in this case VC Redist ). Detection Condition need to be looked deeper in this case.

gkhanna79: Evidence indicates that corrupted installation of a version that not in that link is what results in the failure.

PR #1986, opened 2017-04-07
Cachemessaging

ramarag:
Gives better message when runtime store is missing

@eerhardt @gkhanna79 @DamianEdwards PTAL

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
+++ b/src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
@@ -143,6 +143,7 @@ namespace Microsoft.Extensions.DependencyModel
                     Debug.Assert(compilationLibrary.Type == runtimeLibrary.Type);
                     Debug.Assert(compilationLibrary.Path == runtimeLibrary.Path);
                     Debug.Assert(compilationLibrary.HashPath == runtimeLibrary.HashPath);
+                    Debug.Assert(compilationLibrary.RuntimeStoreManifestName == null);
                 }
 
                 var library = (Library)compilationLibrary ?? (Library)runtimeLibrary;
@@ -335,6 +336,11 @@ namespace Microsoft.Extensions.DependencyModel
             if (library.HashPath != null)
             {
                 libraryJson.Add(new JProperty(DependencyContextStrings.HashPathPropertyName, library.HashPath));
+            }
+
+            if (library.RuntimeStoreManifestName != null)
+            {
+                libraryJson.Add(new JProperty(DependencyContextStrings.RuntimeStoreManifestPropertyName, library.RuntimeStoreManifestName));
             }
 
             return libraryJson;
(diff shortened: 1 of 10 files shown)


gkhanna79:
Can we please add tests along with the change?

ramarag:
added tests

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs
```diff
@@ -338,6 +341,7 @@ public void MergesRuntimeAndCompileLibrariesForPortable()
             library.Should().HavePropertyValue("serviceable", true);
             library.Should().HavePropertyValue("path", "PackagePath");
             library.Should().HavePropertyValue("hashPath", "PackageHashPath");
+            library.Should().HavePropertyValue("runtimeStoreManifestName", "placeHolderManifest.xml");
```
eerhardt: Can we have at least 1 "read" test as well?

src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
```diff
@@ -337,6 +338,12 @@ private JObject WriteLibrary(Library library)
                 libraryJson.Add(new JProperty(DependencyContextStrings.HashPathPropertyName, library.HashPath));
             }
 
+            if (library.RuntimeStoreManifestName != null)
+            {
+                libraryJson.Add(new JProperty(DependencyContextStrings.RuntimeStoreManifestPropertyName, library.RuntimeStoreManifestName));
+            }
+
```
eerhardt: (nit) unnecessary blank line.

src/Microsoft.Extensions.DependencyModel/Library.cs
```diff
@@ -18,7 +18,6 @@ public class Library
             : this(type, name, version, hash, dependencies, serviceable, path: null, hashPath: null)
         {
         }
-
         public Library(string type,
```
eerhardt: (nit) class members should have a single blank line between them.

src/corehost/cli/deps_format.cpp
```diff
@@ -69,6 +69,7 @@ void deps_json_t::reconcile_libraries_with_targets(
 
         pal::string_t library_path = get_optional_path(properties, _X("path"));
         pal::string_t library_hash_path = get_optional_path(properties, _X("hashPath"));
+        pal::string_t filter_project_list = get_optional_path(properties, _X("runtimeStoreManifestName"));
```
eerhardt: I think it would make sense to keep a consistent name here as what it is called in the deps.json file.  It gets too confusing when we rename the same concept for no reason.

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs
ramarag: added a test for read as well

PR #1997, opened 2017-04-07
Add UAP meta-package project.json to the update template

joperezr:
cc: @dagood @chcosta 

Adding the UAP project.json template to the list of files that will get updated when running updatedependencies.ps1


joperezr:
Merging now since OSX build is still hanging. Will address any feedback later.

Issue #1977, opened 2017-04-06
Sign arm32/arm64 Windows binaries in build pipeline


wtgodbe:
Today we only sign x64 & x86 - we should be signing arm32 & arm64 as well. This could be done as part of refactoring the pipeline .json blobs to use shared definitions.

CC @gkhanna79 @dagood @leecow @crummel

gkhanna79:
We should get this in place before feature complete if we can.

wtgodbe:
@MichaelSimons do you know of any reason we're not already doing this?

MichaelSimons:
@wtgodbe - this is outside my realm.  @dagood, @crummel - can you answer Will's question?

wtgodbe:
Also, does anyone know why the environment variable SIGNED_PACKAGES gets set in the x64 build (https://github.com/dotnet/core-setup/blob/master/buildpipeline/Core-Setup-Signing-Windows-x64.json#L461), but not in the x86 build?

gkhanna79:
Is that a regression from the original change that brought in checked in build definitions for Core-Setup (e.g. https://github.com/dotnet/core-setup/commit/22e89978651525436965791f11613ebd0cb4edc6#diff-3867ee2c3921d2e2988404fcb10bb64e)?

I do not see SIGNED_PCKAGES in either 1.0.0 and 1.1.0 build definitions for WinX86.

@ravimeda What are your thoughts?

@wtgodbe Can you please take a look at the 1.1.2 binaries on myget to confirm if they are actually signed or not? Likewise for 1.0.5.

wtgodbe:
@weshaggard do you know what SIGNED_PACKAGES is doing? looks like it's at least used here https://github.com/dotnet/core-setup/blob/69eeba979b44cee29b84cfd64193d838eb572c09/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj#L17, along with a couple of other places

dagood:
If things haven't changed, setting `SIGNED_PACKAGES` in only one build leg is intentional--it's an indicator of the single build that should publish its managed NuGet packages, to avoid builds overwriting each other. It's not (afaik) an indicator of whether or not the build attempts to sign its binaries.

[...]

PR #1999, opened 2017-04-07
SIGNED_PACKAGES should not be set for PortableBuilds

gkhanna79:
skip ci please

Based upon the discussion in https://github.com/dotnet/core-setup/issues/1977, specifically at https://github.com/dotnet/core-setup/issues/1977#issuecomment-292621831, I am disabling SIGNED_PACKAGES to be set in Portable WinX64 build definition.

@dagood @wtgodbe PTAL

CC @crummel

### DIFF
--- a/buildpipeline/pipeline.json
+++ b/buildpipeline/pipeline.json
@@ -158,7 +158,8 @@
         {
           "Name": "Core-Setup-Signing-Windows-x64",
           "Parameters": {
-            "PB_PortableBuild": "-portable"
+            "PB_PortableBuild": "-portable",
+            "SIGNED_PACKAGES": "false"
           },
           "ReportingParameters": {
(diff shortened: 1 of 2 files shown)


buildpipeline/pipeline.json
```diff
@@ -158,7 +158,8 @@
         {
           "Name": "Core-Setup-Signing-Windows-x64",
           "Parameters": {
-            "PB_PortableBuild": "-portable"
+            "PB_PortableBuild": "-portable",
+            "SIGNED_PACKAGES": "false"
```
dagood: I recommend setting it to false by default, and only setting it to true for the nonportable x64 build. The x86 entries below should also both be false.

gkhanna79: I have updated it to be false by default and true only for X64 build.

PR #2001, opened 2017-04-08
Issue #1645: Enable Host to carry shared component (registry and files) (1.0)

vivmishra:
- Enable Host Msi to install as a shared component between releases and build-to-build. 
- This change will now enable the Host Msi to be generated with a new upgrade code with every build and thus do proper ref counting.
- On uninstall of higher version product the shared host (dotnet.exe) and its registry entry will remain on the box, till the time the last product is uninstalled. 

Reference Issue: #1645


packaging/windows/host/host.wxs
```diff
@@ -35,9 +34,13 @@
         </Directory>
     </Fragment>
     <Fragment>
-        <ComponentGroup Id="InstallFiles">
+        <ComponentGroup Id="InstallSharedHostandDetectionKeys">
```
gkhanna79: Will this work in the following scenario on a clean machine?

1) Install 1.0.4
2) Install with fix
3) Uninstall 1.0.4

Does the new host remain? Also, what happens if, in (3), we uninstall host with the fix - does the old host remain?

vivmishra: Both those scenarios work, i.e. the host remains on the box unless the last version of the product is uninstalled. 

There is a more complex scenario like below that will not be addressed by this:
1.Install with fix
2.Install 1.0.4
3.Install 1.0.2
4.Uninstall 1.0.4  -- Registry is busted, but host (dotnet.exe) remains.
5.Uninstall 1.0.5  -- host is removed busting 1.0.2. 

In step 5, Uninstalling 1.0.2 first instead of 1.0.5 does not remove the host.

gkhanna79: What does "Registry is Busted" means?

vivmishra: For step 4, the following regkey is removed. (taking x86 as an example) 
     HKEY_LOCAL_MACHINE\SOFTWARE\dotnet\Setup\InstalledVersions\x86\sharedhost

The hostfxr and shared framework regkeys are not impacted.

gkhanna79: Thanks - I am not sure if the ordering issue can be fixed without creating a new hierarchy.

vivmishra:
Merged

Issue #1646, opened 2017-03-06
Enable Host to carry shared component (registry and files) - 1.1.X fix


vivmishra:
This issue for fixing the 1.1.x issue as per https://github.com/dotnet/core-setup/pull/1615

## Broken scenario :
Install Runtime Bundle 2 ( higher version ) 
Install Runtime Bundle 1 ( lower version )
Uninstall Runtime Bundle 2 

### Expected result :
-- Host (dotnet.exe) from Bundle2 should exist 
Actual result:
-- Host ( dotnet.exe) is removed

## Issue
As of today , Host MSIs are major upgradable ( sharing same keypath and upgrade code ) . This is causing uninstallation of the previously installed lower version of the product. Also, bundle carrying host msi ( Runtime or SDK) skips installation of the lower version host and thus doesn't add the refcount to the component. 

This fix updates the upgrade code for every build and ensures that regkeys are part of shared component too so that only higher version payloads ( dotnet.exe,regkey entry for version ) are preserved during SxS installation.

Similar, fix needs to be made in 1.0 and 1.1 release branch. 

Scenario verified and passed  : 

Build two version of 2.0 runtime bundles locally and tested the following scenario
Bundle1 : lower version
Bundle 2 : higher version

## Scenario:
Scenario 1:
1.	Install Bundle 1 
2.	Install Bundle 2
3.	Verify bundle installed under program files 
4.	Verify bundle.exe under program files\dotnet\ has version from bundle2
5.	Verify the registry key has version from Bundle 2 
6.	Open a command prompt and run dotnet.exe from anywhere ( dotnet.exe path should be present in the system path variable )
7.	Uninstall Bundle 2 
8.	Verify Bundle 2 versioned dotnet.exe and *.txt files remain in the machine and registry key has bundle 2 version 
9.	Verify dotnet.exe remains in the path environment variable
10.	Uninstall Bundle 1
11.	Verify all artifacts of Bundle 1 and Bundle 2 are removed from machine
12.	Verify dotnet.exe is removed from path environment

Scenario 2 
1.	Install Bundle 2
2.	Install Bundle 1
3.	Verify bundle is installed under program files ( focus on program files\donet.exe and license.txt , thirdpartynotices.txt ) . Verify that dotnet.exe and registry key has Bundle 2 version
4.	Open a command prompt and run dotnet.exe from anywhere ( dotnet.exe path should be present in the system path variable )
5.	Uninstall Bundle 2
[...]

PR #2002, opened 2017-04-08
Issue #1646: Enable Host to carry shared component (registry and files) (1.1)

vivmishra:
- Enable Host Msi to install as a shared component between releases and build-to-build. 
- This change will now enable the Host Msi to be generated with a new upgrade code with every build and thus do proper ref counting.
- On uninstall of higher version product the shared host (dotnet.exe) and its registry entry will remain on the box, till the time the last product is uninstalled. 

Reference Issue: #1646


vivmishra:
Merged

PR #2015, opened 2017-04-10
Skipping UWP package build on non-Windows

### LINKED ISSUE #2014: runtime.win10*.Microsoft.NETCore.UniversalWindowsPlatform packages are getting built on linux/osx builds
## Steps to reproduce
build dotnet/core-setup on a linux or OSX machine.

## Expected  behavior
We should not be producing `runtime.win10` packages, since this isn't building on win10.

## Actual behavior
```
eerhardt@eerhardt-Virtual-Machine:~/git/core-setup/artifacts/ubuntu.14.04-x64/corehost$ ls -al
...
-rwxrw-r-- 1 eerhardt eerhardt   883676 Apr 10 11:21 runtime.ubuntu.14.04-x64.Microsoft.NETCore.DotNetHostPolicy.2.0.0-preview1-001948-                   00.symbols.nupkg
-rwxrw-r-- 1 eerhardt eerhardt   703114 Apr 10 11:21 runtime.ubuntu.14.04-x64.Microsoft.NETCore.DotNetHostResolver.2.0.0-preview1-00194                   8-00.nupkg
-rwxrw-r-- 1 eerhardt eerhardt   703114 Apr 10 11:21 runtime.ubuntu.14.04-x64.Microsoft.NETCore.DotNetHostResolver.2.0.0-preview1-00194                   8-00.symbols.nupkg
-rwxrw-r-- 1 eerhardt eerhardt  7794176 Apr 10 11:23 runtime.win10-x64-aot.Microsoft.NETCore.UniversalWindowsPlatform.5.3.0-preview1-00                   1948-00.nupkg
-rwxrw-r-- 1 eerhardt eerhardt 16199962 Apr 10 11:23 runtime.win10-x64-aot.Microsoft.NETCore.UniversalWindowsPlatform.5.3.0-preview1-00                   1948-00.symbols.nupkg
-rwxrw-r-- 1 eerhardt eerhardt 19468525 Apr 10 11:24 runtime.win10-x64.Microsoft.NETCore.UniversalWindowsPlatform.5.3.0-preview1-001948                   -00.nupkg
-rwxrw-r-- 1 eerhardt eerhardt 60734115 Apr 10 11:24 runtime.win10-x64.Microsoft.NETCore.UniversalWindowsPlatform.5.3.0-preview1-001948                   -00.symbols.nupkg
```

joperezr:
fixes #2014 

cc: @ericstj @dagood @eerhardt 

Skipping UAP package build when building for non-Windows.

### DIFF
--- a/pkg/packages.builds
+++ b/pkg/packages.builds
@@ -3,7 +3,8 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <ItemGroup>
-    <Project Include="$(MSBuildThisFileDirectory)\projects\*\*.builds" />
+  	<ProjectExclusions Include="$(MSBuildThisFileDirectory)\projects\*\Microsoft.NETCore.UniversalWindowsPlatform.builds" Condition="'$(OSEnvironment)' != 'Windows_NT'" />
+    <Project Include="$(MSBuildThisFileDirectory)\projects\*\*.builds" Exclude="@(ProjectExclusions)" />
   </ItemGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.traversal.targets))\dir.traversal.targets" />


pkg/packages.builds
```diff
@@ -3,7 +3,8 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
   <ItemGroup>
-    <Project Include="$(MSBuildThisFileDirectory)\projects\*\*.builds" />
+  	<ProjectExclusions Include="$(MSBuildThisFileDirectory)\projects\*\Microsoft.NETCore.UniversalWindowsPlatform.builds" Condition="'$(OSEnvironment)' != 'Windows_NT'" />
```
eerhardt: This seems like the wrong coupling to me.

Shouldn't the `Microsoft.NETCore.UniversalWindowsPlatform` project do the skipping itself?  It would make more sense to have this "don't build me on non-windows" logic be in that project instead of being up in the higher directories.

joperezr: If we do this at the individual project level it would still require adding conditions on shared code, since if we disable it on the `Microsoft.NETCore.UniversalWindowsPlatform.builds` file it would only disable the identity package build, but not the RID specific ones. They get built based on the [shared dir.props code](https://github.com/dotnet/core-setup/blob/master/pkg/projects/dir.props#L71-L76) where a condition would make less sense to me, which is why I thought that the best approach was to exclude the builds file entirely one level up.

joperezr: BTW, we use the same approach in [corefx](https://github.com/dotnet/corefx/blob/master/src/src.builds#L5-L7) for similar situations.

ericstj: I think this is fine to unblock the build.  I do think we should take a closer look at OfficialBuildRID mechanism and its use by both projects and builds.  Perhaps it should just create items and we should leave it to the proj/builds to decide how to use those.  That stuff has grown organically and we haven't looked at how to make it generically usefull holistically yet.

Issue #1719, opened 2017-03-13
Core-Setup build should not rely on Dockerfile within the repository


ravimeda:
## Steps to reproduce

1. Perform an official build.

## Expected  behavior
Rather than consuming (or maintaining) a Dockerfile within a repository, it is more efficient to specify the image name required through PB_DockerImageName variable in a build task, and let PipeBuild pull the appropriate image on behalf of the build.  Investigate the possibility of reusing the shared images in `microsoft/dotnet-buildtools-prereqs`.

Given the scope of this change, do not consider this for existing servicing branches i.e., 1.0.0 and 1.1.0.

## Actual behavior
Builds relies on Dockerfiles located within the repository at `/core-setup/tree/master/scripts/docker`

wtgodbe:
@MichaelSimons can you confirm that this has already been done?

MichaelSimons:
@wtgodbe - Nothing has been done on this.  A dup was just logged yesterday - #2031.

rakeshsinghranchi:
Please consider these docker files as well : https://github.com/dotnet/core-setup/tree/master/scripts/docker

gkhanna79:
We should do this for 2.0 @wtgodbe.

CC @eerhardt

wtgodbe:
Resolved by https://github.com/dotnet/core-setup/commit/6887ab556bc8302390782711dcdd75b75e769cf5

PR #2027, opened 2017-04-11
Build Native Debian installers consuming Portable Linux binaries

rakeshsinghranchi:
This work enables building debian installers ( Ubuntu 14.04/16.04/16.10  and Debian 8 ) consuming Linux Portable binaries.
Change involves : 
1. Updating existing build definition for Portable Linux and adding build steps to generate and publish installers for aforementioned distors and versions
2. Stop building native installers building native binaries for each aforementioned distro

Note: There is separate step to build docker image and then run the build for each distro version. There is a goal to consume the docker image from docker registry than building during the vso build step.


build_projects/dotnet-host-build/PackageTargets.cs
```diff
@@ -266,5 +269,25 @@ private static void FixPermissions(string directory)
...
+
+            // set to default path if PORTABLE_BINARY_LOCATION environment variable doesn't exist
+            if(preBuiltPortableBinaryPath == null)
+            {
+                sharedFrameworkPublishPath = Dirs.SharedFrameworkPublish;
+            }
+            else 
+            {
+                Console.WriteLine($"Installers will package binaries from path set by PORTABLE_BINARY_LOCATION environment variable :{preBuiltPortableBinaryPath}");
+                sharedFrameworkPublishPath = preBuiltPortableBinaryPath;
```
rakeshsinghranchi: Is it okay to use Console.WriteLine() here or expected to call BuildContext.warn() ? I really want to add the log here.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -288,6 +298,13 @@ private static bool CheckIfAllBuildsHavePublished(BuildTargetContext c)
         public static BuildTargetResult PublishArtifacts(BuildTargetContext c) => c.Success();
 
         [Target(
+            nameof(PublishTargets.PublishInstallerFilesToAzure),
+            nameof(PublishTargets.PublishDotnetDebToolPackage),
+            nameof(PublishTargets.PublishDebFilesToDebianRepo),
+            nameof(PublishTargets.PublishSharedFrameworkVersionBadge))]
```
rakeshsinghranchi: One can use the Environment variable check to determine a target to be called or not.  I have followed this approach of grouping the applicable targets.

scripts/dockerbuild.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
```
rakeshsinghranchi: Added a dockerbuild.sh file , future goal is to eliminate the step of building docker images during the build process,rather consume from docker registry. @gkhanna79 has suggested opening an issue related to this.

scripts/dockerbuild.sh
```diff
@@ -0,0 +1,101 @@
...
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ "$SOURCE" != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+cd "$DIR/.."
+
+INTERACTIVE="-i"
```
eerhardt: Where is this variable used?  I don't see it being used.

rakeshsinghranchi: Not used, will remove.

build_projects/dotnet-host-build/PackageTargets.cs
```diff
@@ -266,5 +269,25 @@ private static void FixPermissions(string directory)
                 FS.FixModeFlags(directory);
             }
         }
+
+        private static string GetSharedFrameworkPublishPath()
```
chcosta: I guess it wasn't clear to me that you were making changes to the build system itself, when you said packaging, I interpreted that as the stuff under the pkg folder.  Most of this will be going away in the next week or two.  What's the priority on getting this done?  Would it make sense to wait on this, or to make parallel changes in my fork based on buildtools?

rakeshsinghranchi: CCing @gkhanna79 @Petermarcu  to provide more details regarding the priority . As of now, I am targeting 4/14 (EOW) to get this change in.

buildpipeline/Core-Setup-PortableLinux-x64.json
```diff
@@ -97,8 +97,180 @@
...
+      "alwaysRun": false,
+      "displayName": "Copy built Portable binaries to staging directory",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "5bfb729a-a7c8-4a78-a7c3-8d717bb7c13c",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "SourceFolder": "$(Build.SourcesDirectory)/artifacts/linux-x64/intermediate/sharedFrameworkPublish",
```
eerhardt: Why are we doing this copy? Why not just point the PORTABLE_BINARY_LOCATION to here?

buildpipeline/Core-Setup-PortableLinux-x64.json
```diff
@@ -97,8 +97,180 @@
...
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
       "continueOnError": true,
-      "alwaysRun": true,
+      "alwaysRun": false,
```
eerhardt: ?? Is this good?  I think we've seen resources run out on the machines if we don't do this.

scripts/dockerbuild.sh
```diff
@@ -0,0 +1,101 @@
...
+#  Jenkins
+[ ! -z "$BUILD_TAG" ] && DOTNET_BUILD_CONTAINER_NAME="${BUILD_TAG}-${DOCKER_TAG}"
+#  VSO
+[ ! -z "$BUILD_BUILDID" ] && DOTNET_BUILD_CONTAINER_NAME="${BUILD_BUILDID}-${BUILD_BUILDNUMBER}-${DOCKER_TAG}"
+
+# Executes a command and retries if it fails.
+# NOTE: This function is the exact copy from init-docker.sh.
+# Reason for not invoking init.docker.sh directly is since that script 
+# also performs cleanup, which we do not want in this case.
+execute() {
```
eerhardt: Is there any way to reuse this logic so it isn't copied with `dockerrun.sh`?

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: Because 'git clean -xdf' is called for each step of building docker image. Git clean is required and removes the artifacts folder. Copying to staging directory to preserves the folder.

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: Thanks for pointing this out,will find out more details about alwaysRun usage and update the value accordingly.

build_projects/dotnet-host-build/PackageTargets.cs
gkhanna79: Yes, we should get this in for 4/14 and any new infra would rebase against this to pick it up.

buildpipeline/Core-Setup-PortableLinux-x64.json
```diff
@@ -97,8 +97,180 @@
...
+      "displayName": "Build Ubuntu.14.04 docker image",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "scripts/dockerbuild.sh",
+        "args": "--non-interactive -d $(Build.SourcesDirectory)/scripts/docker/ubuntu.14.04 -t ubuntu.14.04-coresetup",
```
gkhanna79: Can we use the Docker images @MichaelSimons has created instead of using the scripts to create one?

build_projects/dotnet-host-build/PackageTargets.cs
```diff
@@ -266,5 +269,25 @@ private static void FixPermissions(string directory)
...
+        {
+            string sharedFrameworkPublishPath = string.Empty;
+            
+            string preBuiltPortableBinaryPath=Environment.GetEnvironmentVariable("PORTABLE_BINARY_LOCATION")?.Trim();
```
gkhanna79: This should be read in Prepare.Init and set within the build context and then you should use BuildContext to read the value here.

gkhanna79: Also, I would name the variable to "PORTABLE_BUILD_STAGING_LOCATION".

scripts/dockerbuild.sh
MichaelSimons: I would think dockerrun.sh could call into this script to build and therefore no logic would need to be duplicated.

buildpipeline/Core-Setup-PortableLinux-x64.json
MichaelSimons: Can you explain why it is necessary to git clean -xdf in each build step?  This seems wrong.

---
In reply to: [110995126](https://github.com/dotnet/core-setup/pull/2027#discussion_r110995126) [](ancestors = 110995126)

buildpipeline/Core-Setup-PortableLinux-x64.json
```diff
@@ -97,8 +97,180 @@
...
+      "displayName": "Build Debian Installer in Ubuntu.14.04 docker container",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "scripts/dockerrun-as-current-user.sh",
+        "args": "--rm $(DockerBuildEnvironmentVariables) -e CLI_NUGET_API_KEY -e CLI_NUGET_FEED_URL -e CLI_NUGET_SYMBOLS_FEED_URL -v $(Build.SourcesDirectory):/opt/code/ -v $(Build.StagingDirectory)/sharedFrameworkPublish/:/opt/sharedFrameworkPublish/ -w /opt/code ubuntu.14.04-coresetup /bin/bash -c \"HOME=/opt/code; git clean -X -d -f; ./build.sh --skip-prereqs --configuration Release --targets $(InstallerOnlyBuildArguments)\"",
```
MichaelSimons: I realize this is an existing pattern but is setting HOME necessary?  Does specifying the working directory achieve the same?

MichaelSimons: Why are the CLI_NUGET... variable needed only for ubuntu.14.04?

buildpipeline/Core-Setup-PortableLinux-x64.json
MichaelSimons: This should not be changed.  Same for the other cleanup tasks.

buildpipeline/Core-Setup-PortableLinux-x64.json
```diff
@@ -222,6 +394,33 @@
...
+    "REPO_USER": {
+      "value": "dotnet"
+    },
+    "REPO_PASS": {
+      "value": "PassedViaPipeBuild"
+    },
+    "REPO_SERVER": {
+      "value": "azure-apt-cat.cloudapp.net"
+    },
+    "DockerBuildEnvironmentVariables": {
```
MichaelSimons: The name is a little confusing since it is not used for the `Run build in Docker container` step

buildpipeline/Core-Setup-PortableLinux-x64.json
eerhardt: Typically NuGet starts throwing errors if $HOME isn't set.  That's why the pattern exists.

```
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
13:17:17 Parameter name: path1
13:17:17    at System.IO.Path.Combine(String path1, String path2, String path3)
13:17:17    at NuGet.Common.NuGetEnvironment.GetFolderPath(SpecialFolder folder)
13:17:17    at NuGet.Common.NuGetEnvironment.GetFolderPath(NuGetFolderPath folder)
13:17:17    at NuGet.Configuration.SettingsUtility.GetHttpCacheFolder()
13:17:17    at NuGet.Configuration.NuGetPathContext.Create(ISettings settings)
```

buildpipeline/Core-Setup-PortableLinux-x64.json
chcosta: +1 since this is something I've done in my refactoring but I had intended to run by @gkhanna79

rakeshsinghranchi: My understanding is that we want to get rid of building docker image on the fly as part of the build step. See issue : https://github.com/dotnet/core-setup/issues/1719  . It will be easier to remove this step and consume the prebuilt images in the build step. Once issue 1719 is resolved, we can simply remove this step.

build_projects/dotnet-host-build/PackageTargets.cs
rakeshsinghranchi: Agreed,will make this change.

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: This is required because PublishDebToolPackage Target requires it. see here : https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PublishTargets.cs#L64

scripts/dockerbuild.sh
rakeshsinghranchi: My understanding is that we want to get rid of building docker image on the fly as part of the build step. See issue : #1719 . It will be easier to remove this step and consume the prebuilt images from dockerhub in the build step. Once issue 1719 is resolved, we can simply remove this script.

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: Thanks Michael. will take care of this.

scripts/dockerbuild.sh
MichaelSimons: While I agree 1719 is the best long term fix, the reality is it will probably be some time before that happens.  Duplicating this script logic is problematic in the interim.  Refactoring it now looks pretty easy.

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: cc @gkhanna79  
if I don't call git clean step and artifacts are left behind from the previous build , I have noticed that coreclr fails to initialize in the prepare step. @gkhanna79  has more insights on why this happens.  The same approach of calling  git clean is done for portable Linux workflow https://github.com/dotnet/core-setup/blob/master/buildpipeline/Core-Setup-PortableLinux-x64.json#L92 here.

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: @chcosta, @MichaelSimons  : do we have docker images that can be pulled from docker hub? My understanding is that #1719  is tracking the required work.

scripts/dockerbuild.sh
rakeshsinghranchi: @MichaelSimons , are you suggesting updating dockerrun.sh to replace the docker build step at https://github.com/dotnet/core-setup/blob/master/scripts/dockerrun.sh#L127 with call to dockerbuild.sh . dockerbuild.sh generates its own custom tag and image name. Please help clarify the suggestion.

MichaelSimons: yes that is what I am suggesting.  Can't dockerrun.sh just `dockerbuild.sh -t $DOTNET_BUILD_CONTAINER_TAG -d $DOCKERFILE`.  e.g. pass in the custom tag is wants.

scripts/dockerbuild.sh
```diff
@@ -0,0 +1,101 @@
...
+# This eliminates intermittent failures during docker build caused by failing to retrieve the base image.
+if [ ! -z "$image" ]; then
+    echo "Pulling Docker image $image"
+    execute docker pull $image
+fi
+
+docker build --build-arg USER_ID=$(id -u) -t $DOCKER_TAG $DOCKERFILE
+
+set +e
+docker rm -f $DOTNET_BUILD_CONTAINER_NAME
```
MichaelSimons: This like seems irrelevant here.

buildpipeline/Core-Setup-PortableLinux-x64.json
MichaelSimons: Yes we have docker images that can be used.  It is not clear to me what the requirements are for what the images contain.  The Dockerfiles are [here](https://devdiv.visualstudio.com/DevDiv/_git/DotNetCore?path=%2Fdockerfiles&version=GBmaster&_a=contents).  The images are [here](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/tags/).

scripts/dockerbuild.sh
rakeshsinghranchi: do you mean Line# 97 and 98 ?

MichaelSimons: Line# 98 - $DOTNET_BUILD_CONTAINER_NAME does not appear in the script.

buildpipeline/Core-Setup-PortableLinux-x64.json
rakeshsinghranchi: I am looking for images equivalent to the docker file. for e.g. Ubuntu 14.04 as in https://github.com/dotnet/core-setup/blob/master/scripts/docker/ubuntu.14.04/Dockerfile  . Container will be used for building debian package.

rakeshsinghranchi: Docker image for building debian packages are not available yet. Updated #1719 added reference to Dockerfile to required packages for building debian packages.

rakeshsinghranchi:
@MichaelSimons @gkhanna79 @eerhardt  - All the feedbacks are addressed. Please review the changes.

build_projects/dotnet-host-build/PackageTargets.cs
```diff
@@ -266,5 +269,25 @@ private static void FixPermissions(string directory)
...
             }
         }
+
+        private static string GetSharedFrameworkPublishPath(BuildTargetContext c)
+        {
+            string sharedFrameworkPublishPath = string.Empty;
+            
+            string preBuiltPortableStagingPath=c.BuildContext.Get<string>("PortableBuildStagingLocation");
+
+            // set to default path if PORTABLE_BUILD_STAGING_LOCATION environment variable doesn't exist
```
gkhanna79: I think you can update the comment to be more reflective of the intent - "If we are not generating distro specific installers for portable build, then we wont have access to staging location and thus, will use default   binary location where SharedFX was published".

scripts/dockerbuild.sh
rakeshsinghranchi: Taken care of this.

buildpipeline/Core-Setup-PortableLinux-x64.json
MichaelSimons: Any thoughts on this?

scripts/dockerbuild.sh
MichaelSimons: Shouldn't the `set +e` have been kept?

rakeshsinghranchi:
@dotnet-bot Test this please

PR #2028, opened 2017-04-11
[WIP] Stabilize for 1.0.5

gkhanna79:
Stabilize Core-Setup to 1.0.5

PR #2029, opened 2017-04-11
[WIP] Stabilize for 1.1.2

gkhanna79:
Stabilize packages for 1.1.2

### DIFF
--- a/TestAssets/TestProjects/PortableApp/project.json
+++ b/TestAssets/TestProjects/PortableApp/project.json
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.1.2-servicing-*"
+          "version": "1.1.2"
         },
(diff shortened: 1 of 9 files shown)


gkhanna79:
@eerhardt PTAL

Failures in CI are expected.

eerhardt:
Do you need to update the branchinfo.txt like you are in https://github.com/dotnet/core-setup/pull/2028/files?

gkhanna79:
There are slight differences in the two branches and I have noticed that it was not changed earlier as well, and pre-release suffix was omitted as expected.

Overall, I think we should simplify a lot of complexity in this infrastructure :(

eerhardt:
> Overall, I think we should simplify a lot of complexity in this infrastructure :(

Agreed.  We have already taken a huge leap forward here in the dotnet/cli repo when we moved to MSBuild.  Hopefully we can do the same when this repo is converted.

PR #2039, opened 2017-04-12
Probe paths

ramarag:
The host passes  the provided probing paths to AppContext and these can be queried by the DependencyModel when it is resolving the assemblies

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
+++ b/src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
     public class PackageCompilationAssemblyResolver: ICompilationAssemblyResolver
     {
         private readonly IFileSystem _fileSystem;
-        private readonly string _nugetPackageDirectory;
+        private readonly string[] _nugetPackageDirectories;
 
         public PackageCompilationAssemblyResolver()
             : this(EnvironmentWrapper.Default, FileSystemWrapper.Default)
@@ -19,64 +19,85 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
         }
 
         public PackageCompilationAssemblyResolver(string nugetPackageDirectory)
-            : this(FileSystemWrapper.Default, nugetPackageDirectory)
+            : this(FileSystemWrapper.Default, new string[] { nugetPackageDirectory })
         {
         }
 
         internal PackageCompilationAssemblyResolver(IEnvironment environment,
             IFileSystem fileSystem)
-            : this(fileSystem, GetDefaultPackageDirectory(environment))
+            : this(fileSystem, GetDefaultProbeDirectories(environment))
         {
         }
 
-        internal PackageCompilationAssemblyResolver(IFileSystem fileSystem, string nugetPackageDirectory)
+        internal PackageCompilationAssemblyResolver(IFileSystem fileSystem, string[] nugetPackageDirectories)
         {
             _fileSystem = fileSystem;
-            _nugetPackageDirectory = nugetPackageDirectory;
+            _nugetPackageDirectories = nugetPackageDirectories;
         }
 
-        private static string GetDefaultPackageDirectory(IEnvironment environment) => 
-            GetDefaultPackageDirectory(RuntimeEnvironment.OperatingSystemPlatform, environment);
+        private static string[] GetDefaultProbeDirectories(IEnvironment environment) =>
+            GetDefaultProbeDirectories(RuntimeEnvironment.OperatingSystemPlatform, environment);
 
-        internal static string GetDefaultPackageDirectory(Platform osPlatform, IEnvironment environment)
+        internal static string[] GetDefaultProbeDirectories(Platform osPlatform, IEnvironment environment)
         {
-            var packageDirectory = environment.GetEnvironmentVariable("NUGET_PACKAGES");
+#if !NETSTANDARD1_3            
+#if NETSTANDARD1_6
+            var probeDirectories = AppContext.GetData("PROBING_DIRECTORIES");
+#else
+            var probeDirectories = AppDomain.CurrentDomain.GetData("PROBING_DIRECTORIES");
+#endif
 
-            if (!string.IsNullOrEmpty(packageDirectory))
-            {
-                return packageDirectory;
-            }
+           var listOfDirectories = probeDirectories as string;
 
-            string basePath;
-            if (osPlatform == Platform.Windows)
-            {
-                basePath = environment.GetEnvironmentVariable("USERPROFILE");
-            }
-            else
-            {
-                basePath = environment.GetEnvironmentVariable("HOME");
-            }
-            if (string.IsNullOrEmpty(basePath))
-            {
-                return null;
-            }
-            return Path.Combine(basePath, ".nuget", "packages");
+           if (!string.IsNullOrEmpty(listOfDirectories))
+           {
+               return listOfDirectories.Split(new char [] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries );
+           }
+#endif
+
+           var packageDirectory = environment.GetEnvironmentVariable("NUGET_PACKAGES");
+
+           if (!string.IsNullOrEmpty(packageDirectory))
+           {
+               return new string[] { packageDirectory };
+           }
+
+           string basePath;
+           if (osPlatform == Platform.Windows)
+           {
+               basePath = environment.GetEnvironmentVariable("USERPROFILE");
+           }
+           else
+           {
+               basePath = environment.GetEnvironmentVariable("HOME");
+           }
+
+           if (string.IsNullOrEmpty(basePath))
+           {
+               return new string[] { string.Empty };
+           }
+
+           return new string[] { Path.Combine(basePath, ".nuget", "packages") };
+
         }
 
         public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> assemblies)
         {
-            if (string.IsNullOrEmpty(_nugetPackageDirectory) ||
+            if (_nugetPackageDirectories == null || _nugetPackageDirectories.Length == 0 ||
                 !string.Equals(library.Type, "package", StringComparison.OrdinalIgnoreCase))
             {
                 return false;
             }
 
-            string packagePath;
-
-            if (ResolverUtils.TryResolvePackagePath(_fileSystem, library, _nugetPackageDirectory, out packagePath))
+            foreach (var directory in _nugetPackageDirectories)
             {
-                assemblies.AddRange(ResolverUtils.ResolveFromPackagePath(_fileSystem, library, packagePath));
-                return true;
+                string packagePath;
+
+                if (ResolverUtils.TryResolvePackagePath(_fileSystem, library, directory, out packagePath))
+                {
+                    assemblies.AddRange(ResolverUtils.ResolveFromPackagePath(_fileSystem, library, packagePath));
+                    return true;
+                }
             }
             return false;
         }
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -164,6 +164,18 @@ void deps_resolver_t::setup_shared_store_probes(
     }
 }
 
+pal::string_t deps_resolver_t::get_probe_directories()
+{
+    pal::string_t directories;
+    for (const auto& pc : m_probes)
+    {
+        directories.append(pc.probe_dir);
+        directories.push_back(PATH_SEPARATOR);
+    }
+
+    return directories;
+}
+
 void deps_resolver_t::setup_probe_config(
     const hostpolicy_init_t& init,
     const arguments_t& args)
--- a/test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
+++ b/test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
@@ -24,8 +24,8 @@ namespace Microsoft.Extensions.DependencyModel.Tests
                 .AddVariable("NUGET_PACKAGES", PackagesPath)
                 .Build();
 
-            var result = PackageCompilationAssemblyResolver.GetDefaultPackageDirectory(Platform.Unknown, environment);
-            result.Should().Be(PackagesPath);
+            var result = PackageCompilationAssemblyResolver.GetDefaultProbeDirectories(Platform.Unknown, environment);
+            result.Should().Contain(PackagesPath);
         }
 
 
@@ -36,8 +36,8 @@ namespace Microsoft.Extensions.DependencyModel.Tests
                 .AddVariable("USERPROFILE", "User Profile")
                 .Build();
 
-            var result = PackageCompilationAssemblyResolver.GetDefaultPackageDirectory(Platform.Windows, environment);
-            result.Should().Be(Path.Combine("User Profile", ".nuget", "packages"));
+            var result = PackageCompilationAssemblyResolver.GetDefaultProbeDirectories(Platform.Windows, environment);
+            result.Should().Contain(Path.Combine("User Profile", ".nuget", "packages"));
         }
 
         [Fact]
@@ -47,8 +47,8 @@ namespace Microsoft.Extensions.DependencyModel.Tests
                 .AddVariable("HOME", "User Home")
                 .Build();
 
-            var result = PackageCompilationAssemblyResolver.GetDefaultPackageDirectory(Platform.Linux, environment);
-            result.Should().Be(Path.Combine("User Home", ".nuget", "packages"));
+            var result = PackageCompilationAssemblyResolver.GetDefaultProbeDirectories(Platform.Linux, environment);
+            result.Should().Contain(Path.Combine("User Home", ".nuget", "packages"));
         }
 
         [Fact]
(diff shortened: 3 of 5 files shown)


ramarag:
@gkhanna79 @eerhardt PTAL

eerhardt:
Fixes #624 and https://github.com/dotnet/sdk/issues/1002

test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
```diff
@@ -24,8 +24,8 @@ public void ShouldUseEnvironmentVariableToGetDefaultLocation()
                 .AddVariable("NUGET_PACKAGES", PackagesPath)
                 .Build();
 
-            var result = PackageCompilationAssemblyResolver.GetDefaultPackageDirectory(Platform.Unknown, environment);
-            result.Should().Be(PackagesPath);
+            var result = PackageCompilationAssemblyResolver.GetDefaultProbeDirectories(Platform.Unknown, environment);
```
eerhardt: I see now what you were asking about last night w.r.t. testing DependencyModel with the latest Microsoft.NETCore.App.

Would it be possible to get some sort of end-to-end test here?  Basically a 2.0 app that references DependencyModel, call it with `--additionalprobingpath`, and ensure that the PackageCompilationAssemblyResolver can find assemblies in that probing path?

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
```diff
@@ -11,74 +11,95 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
...
 
-            if (!string.IsNullOrEmpty(packageDirectory))
-            {
-                return packageDirectory;
-            }
+           var listOfDirectories = probeDirectories as string;
+
+           if (!string.IsNullOrEmpty(listOfDirectories))
+           {
+               return listOfDirectories.Split(new char [] {Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries );
```
eerhardt: I'm wondering if we should ensure that the NUGET_PACKAGES path from below is always appended to the end of this list (using a hashset, so it isn't duplicated).  That way, even if the NuGet cache wasn't passed as a probing path, it is still used like it was before.

Thoughts?

test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
```diff
@@ -60,7 +60,7 @@ public void ResolvesAllAssemblies()
                 .Build();
             var library = F.Create(assemblies: F.TwoAssemblies);
 
-            var resolver = new PackageCompilationAssemblyResolver(fileSystem, PackagesPath);
+            var resolver = new PackageCompilationAssemblyResolver(fileSystem, new string[] {PackagesPath});
```
pakrym: Add tests that use multiple probe directories.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
```diff
@@ -11,74 +11,95 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
...
+           }
+#endif
+
+           var packageDirectory = environment.GetEnvironmentVariable("NUGET_PACKAGES");
+
+           if (!string.IsNullOrEmpty(packageDirectory))
+           {
+               return new string[] {packageDirectory};
```
pakrym: Space after curly brace everywhere.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
```diff
@@ -11,74 +11,95 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
...
+        private static string[] GetDefaultProbeDirectories(IEnvironment environment) =>
+            GetDefaultProbeDirectories(RuntimeEnvironment.OperatingSystemPlatform, environment);
 
-        internal static string GetDefaultPackageDirectory(Platform osPlatform, IEnvironment environment)
+        internal static string[] GetDefaultProbeDirectories(Platform osPlatform, IEnvironment environment)
         {
-            var packageDirectory = environment.GetEnvironmentVariable("NUGET_PACKAGES");
+#if !NETSTANDARD1_3            
+#if NETSTANDARD1_6
+            var probeDirectories = AppContext.GetData("PROBING_DIRECTORIES");
```
pakrym: Put `AppContext.GetData/AppDomain.CurrentDomain.GetData` as a method on `IEnvironment` and add tests.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
pakrym: If `NUGET_PACKAGES` is defined `dotnet run` would pas it to muxer, right?

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
```diff
@@ -11,74 +11,95 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
...
-            {
-                return null;
-            }
-            return Path.Combine(basePath, ".nuget", "packages");
         }
 
         public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> assemblies)
         {
-            if (string.IsNullOrEmpty(_nugetPackageDirectory) ||
+            if (_nugetPackageDirectories == null || _nugetPackageDirectories.Length < 1 ||
```
pakrym: ` == 0`

test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
pakrym: Functional tests would be very nice here, this is not very common scenario.

eerhardt: > this is not very common scenario.

It is becoming more and more common.  For example, in the latest 2.0 SDK, the "first run"/"offline" experience uses a fallback folder instead of populating the NuGet cache.  So by default, everyone will be using fallback folders + NuGet cache.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
eerhardt: Yes, I think in the normal cases this would "just work" as coded.

I was thinking about the fact that this class would always probe the NuGet cache no matter what.  Now there are cases where it won't. But I think that is OK because the host won't probe the NuGet cache either in those cases.

test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
pakrym: I would still prefer having functional test for it.

eerhardt: I'm not disagreeing.  I was actually making that same point.  This is a very common scenario and we should have tests that ensure it doesn't break.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
```diff
@@ -11,74 +11,95 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
...
+#endif
+
+           var packageDirectory = environment.GetEnvironmentVariable("NUGET_PACKAGES");
+
+           if (!string.IsNullOrEmpty(packageDirectory))
+           {
+               return new string[] {packageDirectory};
+           }
+
+           string basePath;
```
gkhanna79: Why are we constructing the nuget package path? This is an implementation detail of nuget and could change in future and at that point, this API will break.

IMHO, if the env-var is not available, we should use null for nuget package location.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -164,6 +164,18 @@ void deps_resolver_t::setup_shared_store_probes(
     }
 }
 
+pal::string_t deps_resolver_t::get_probe_directories()
+{
+    pal::string_t directories;
+    for (const auto& pc : m_probes)
```
gkhanna79: This change does not look correct. The goal of this change is to list the additional probing paths. m_probes contains all sort of probing locations (store, FX - which is flat, servicing location, etc). Additional probing path is one of the various kind of probe locations - see https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/deps_resolver.cpp#L202

I think you should be filtering out additional probing paths and only adding them here. Also, the property APP have is that they have a nuget layout, which is not true for all other probing locations.

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
eerhardt: A thing to always remember is that DependencyModel is used on desktop as well as .NET Core.  Also, it is used on netcoreapp1.0 and netcoreapp1.1 where the host won't be passing in the probing paths.

So we should not be changing this logic at all.  This is existing logic, which @ramarag is preserving in the case where `PROBING_DIRECTORIES` are not getting passed to the DependencyModel.

gkhanna79: So, it is well understood that this is a dependence on an implementation detail of nuget (which may not change frequently but nonetheless)?

src/corehost/cli/deps_resolver.cpp
eerhardt: I agree that the FX and probing locations shouldn't be needed here.

But is there ever a scenario where the "runtime store" locations will be needed by managed code?  I know that there aren't compile assemblies in there today, but the design doc talks about having them some day. And there may be a case where managed code wants to find runtime assemblies as well.  So I'd error on the side of making "runtime store" locations available to managed code.  See # 3 here:  https://github.com/dotnet/core-setup/issues/624#issuecomment-288245699

src/Microsoft.Extensions.DependencyModel/Resolution/PackageCompilationAssemblyResolver.cs
eerhardt: > So, it is well understood that this is a dependence on an implementation detail of nuget 

Yes.  DependencyModel can't have a hard dependency on NuGet, so it can't call it for this information.

But in general, the problem you are pointing out is why we are doing this change.  NuGet added more locations to where it knows how to find assets (the "fallback" folder). This change is allowing the DependencyModel knowledge of these new locations.

src/corehost/cli/deps_resolver.cpp
gkhanna79: I think RuntimeStore location exposure is fine - especially since it follows the same layout as APP (which is the nuget file layout) - and because Store location is a more first class APP.

mlorbetske:
Do we think we'll be able to get this in in time for this to be depended on for the assemblies that go in the ASP.NET package cache tomorrow? /cc @Eilon

src/corehost/cli/deps_resolver.cpp
ramarag: I agree some locations don't make sense, but i would rather not put more logic in the host. This will also function in addition to what @eerhardt mentioned, a way to expose what paths host is actually trying to look up assemblies.

gkhanna79: > but i would rather not put more logic in the host

I disagree - we need to put the right logic in the host, not just what gets everything. Otherwise, we will run into support scenarios where we will be forced to fix issues when people start to take dependence on this and when certain paths wont make sense. For example, even if you pass everything to the API user, they have no way of knowing which can be used in a nuget layout (for runtime compilation) vs not.

I would like us to do the right thing here.

ramarag: what support cost are we talking about, both the servicing and FX location are actually populated by us ? and any other location is just like a nuget cache location

gkhanna79: When users assume the layout of the folders and they dont match expectation. this API is about exposing the paths that were specified with --additionalprobingpath or are implicit version of it (e.g. store location). This is not about servicing or FX locations. We should ensure the API places the user in the pit of success - not additional complexity.

What are your concerns in fixing the host to account for this, as this was expected to be in place to ensure the change is correct?

gkhanna79:
@ramarag have you looked at the non-Windows failures above?

ramarag:
@gkhanna79 non windows builds apart from portable linux are not producing packages for Microsoft.Extensions.DependencyModel. Is it explicitly disabled for non-windows ?

gkhanna79:
I vaguely recall so.

ramarag:
2 options here:

1. Find how the packaging is disabled and enable it, Is there a reason it is disabled?
2. Check in without the end to end functional test

ramarag:

gkhanna79:
I believe it is disabled since it is platform neutral, so only built once. @eerhardt may know as well.

For now, file an issue to enable the tests early next week, but get the change in to get it to folks around.


Thanks,
Gaurav

ramarag:
done

mikeharder:
This issue is blocking ASP.NET so we are eager to test the fix.  Which build of the SDK and/or Runtime should have this fix?

Petermarcu:
Looks like the last good build of the runtime was the commit right before this and the subsequent ones failed with what looks like infra issues. I just kicked off another build. Hopefully it will be green.

Petermarcu:
The build I kicked off completed with these changes in it. You can grab the latest from here: https://github.com/dotnet/core-setup

mikeharder:
The ASP.NET build fetches the latest runtime version from https://dotnetcli.blob.core.windows.net/dotnet/master/dnvm/latest.sharedfx.win.x64.version, which still contains build 2013.  We can workaround this in the ASP.NET build for now.

mikeharder:
The latest version of Microsoft.NETCore.App at https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App is also 2013, so I don't think there's any workaround.

rakeshsinghranchi:
There seems to be issue with build Finalize step where packages are pushed to NuGet Feed and update the latest version.  However,all the latest build artifacts are available in Azure blob storage. You can find the latest build from 
http://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/2.0.0-preview1-002020-00/Microsoft.NETCore.App.2.0.0-preview1-002020-00.nupkg . Let me know if this unblocks you for now.

mikeharder:
If the issues with the finalize step can be resolved tomorrow we would rather wait.  The ASP.NET build relies on the MyGet feed and it would be some work to change this.

danroth27:
I just installed the latest CLI, but I'm still seeing this issue:

```
~\Desktop\templates2\WebApp2> dotnet --info
.NET Command Line Tools (2.0.0-preview1-005815)

Product Information:
 Version:            2.0.0-preview1-005815
 Commit SHA-1 hash:  a150055dfb

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0-preview1-005815\

  Version  : 2.0.0-preview1-002021-00
  Build    : e9456614cfe65b6dc7c5326a597337159142c84c
```

```
~\Desktop\templates2\WebApp2> dotnet new mvc
...
The template "ASP.NET Core Web App" was created successfully.
...
Restore succeeded.
~\Desktop\templates2\WebApp2> dotnet run

Unhandled Exception: System.InvalidOperationException: Can not find compilation library location for package 'Microsoft.NETCore.App'
```

ramarag:
@danroth27  please check the version of Microsoft.Extensions.DependencyModel, it should be at least 2.0.0-preview1-002021 for this feature to kick in.

You can find the version in the generated app.deps.json

danroth27:
Ah, I'm actually getting the 1.1.0 version of that package. Looks like this is being cased by a bad dependency in MVC:

```json
      "Microsoft.AspNetCore.Mvc.Core/2.0.0-preview1-final": {
        "dependencies": {
          "Microsoft.AspNetCore.Authorization": "2.0.0-preview1-final",
          "Microsoft.AspNetCore.Hosting.Abstractions": "2.0.0-preview1-final",
          "Microsoft.AspNetCore.Http": "2.0.0-preview1-final",
          "Microsoft.AspNetCore.Mvc.Abstractions": "2.0.0-preview1-final",
          "Microsoft.AspNetCore.ResponseCaching.Abstractions": "2.0.0-preview1-final",
          "Microsoft.AspNetCore.Routing": "2.0.0-preview1-final",
          "Microsoft.Extensions.DependencyModel": "1.1.0",
          "Microsoft.Extensions.FileProviders.Abstractions": "2.0.0-preview1-final",
          "Microsoft.Extensions.Logging.Abstractions": "2.0.0-preview1-final",
          "Microsoft.Extensions.PlatformAbstractions": "2.0.0-preview1-final",
          "NETStandard.Library": "2.0.0-preview1-25211-01",
          "System.Buffers": "4.3.0",
          "System.Diagnostics.DiagnosticSource": "4.4.0-preview1-25205-01"
        },
```

eerhardt:
Yes, MVC will have to update its dependency to Microsoft.Extensions.DependencyModel to be the latest 2.0.0 build.  It looks like it did according to https://dotnet.myget.org/feed/aspnetcore-ci-dev/package/nuget/Microsoft.AspNetCore.Mvc.Core/.  The latest builds have a dependency to the new DependencyModel version.

An optional workaround, in the meantime, is to add `<PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.0-preview1-*" />` to your .csproj.

PR #2044, opened 2017-04-12
Use cannot instead of can not in error text

danroth27:
While "can not" is technically OK, "cannot" is the more common form and looks more professional.


eerhardt:
Thanks!

PR #2046, opened 2017-04-12
Remove the use of --infer-runtimes during restore

ellismg:
--infer-runtimes has been deprecated for a long time, but we continued to
use it because there was no pressing reason to move away from it. I've
been trying to build core-setup on newer versions of RHEL, where the RID
has changed from rhel.7.2-x64 to rhel.7.3-x64. Since our long term plan
(which will happen for 2.0) is to unify all RHEL 7 rids to rhel.7-<arch>,
I had hoped I could use DOTNET_RUNTIME_ID to specifcy a RID of rhel.7-x64
and have everything work.

However, the use of infer-runtimes is not playing nicely with that, so now
is the time to stop using this feature. We only have a handful of projects
now where we care about restoring for a specific rid (the build system,
the update scripts and a tool run during build time to generate some deps
files for the shared framework) so it was easy enough to get everyone on a
project.json.template plan.

Long term, the build project and update-dependencies scripts will be
removed as we move over to MSBuild + BuildTools. I expect that we'll also
just use a custom MSBuild task in the future to replace the DepsProcessor
tool and then we can delete all of this stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -297,3 +297,8 @@ test/PackagedCommands/Consumers/*/project.json
 
 # VS generated files
 launchSettings.json
+
+# Generated project.json files (the project.json.template file is the authoritative copy)
+setuptools/independent/DepsProcessor/project.json
+build_projects/dotnet-host-build/project.json
+build_projects/update-dependencies/project.json
--- a/build_projects/dotnet-host-build/PrepareTargets.cs
+++ b/build_projects/dotnet-host-build/PrepareTargets.cs
@@ -385,7 +385,16 @@ namespace Microsoft.DotNet.Host.Build
                 .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "setuptools", "dotnet-deb-tool"))
                 .Execute()
                 .EnsureSuccessful();
-            dotnet.Restore("--verbosity", "verbose", "--disable-parallel", "--infer-runtimes")
+
+            var independentToolsRoot = Path.Combine(c.BuildContext.BuildDirectory, "setuptools", "independent");
+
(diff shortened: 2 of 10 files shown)


ellismg:
@chcosta I hate touching this stuff right now, but I sort of need it to unblock some stuff I want to do. I think that you may not actually have to absorb any of this stuff once we are on buildtools, but it's possible that the change to how we restore the stuff in the `setuptools` folder needs to be brought over.  Let me know if I can help somehow.

ellismg:
/cc @omajid, with this change the following works on RHEL 7.3:

`DOTNET_RUNTIME_ID=rhel.7-x64 ./build.sh`

Previously build.sh would complain about being unable to find rhel.7.3 packages. Setting the environment variable + this patch will fix that issue.

ellismg:
The OSX build failure is due to differences in sed or grep behavior between Linux and OSX.  I'll dig in.

.gitignore
```diff
@@ -297,3 +297,8 @@ test/PackagedCommands/Consumers/*/project.json
 
 # VS generated files
 launchSettings.json
+
+# Generated project.json files (the project.json.template file is the authoritative copy)
+setuptools/independent/DepsProcessor/project.json
+build_projects/dotnet-host-build/project.json
+build_projects/update-dependencies/project.json
```
dagood: Would it be best to remove the project.jsons we have checked in now to avoid confusion?

dagood: Nevermind--didn't catch that the project.json modifications were moves.

ellismg:
@dotnet-bot test Fedora24 x64 Debug Build please

eerhardt:
The other option here would have been to convert all 3 of these tools into "framework dependent" tools (i.e. `type: platform`), thus not requiring a RID at all.

PR #2059, opened 2017-04-13
Revert "Merge pull request #2046 from ellismg/remove-infer-runtimes"

ellismg:
This reverts commit 3e354ac6f6a4efe241928ac03fc5ee740d7c7f61, reversing
changes made to 4e4d0a2046227332678d40c8e80f1b89a74e7956.

It appears that after this change we started to hit some flakyness in
our official builds and the thought is that this commit may have been
the issue.

Reverting for now to try to unblock the official builds and we'll
reapply this once we understand the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -297,8 +297,3 @@ test/PackagedCommands/Consumers/*/project.json
 
 # VS generated files
 launchSettings.json
-
-# Generated project.json files (the project.json.template file is the authoritative copy)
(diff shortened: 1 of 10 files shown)


Petermarcu:
Go ahead and commit.

ellismg:
BTW, @Petermarcu I was able to reproduce this locally, and I'm digging in.

PR #2062, opened 2017-04-14
Fix null ref during build and bring back my --infer-runtime changes

ellismg:
The first commit fixes the issue that was causing the `--infer-runtimes` change to break intermittently, the second commit reverts the revert of the `--infer-runtimes` change.

If there's a way you can launch an official build on this without merging, it may be a good idea, but my testing locally makes me believe that I have addressed the issue.


Petermarcu:
Lets go with it and I'll kick off another build. Thanks for digging into it!

PR #2066, opened 2017-04-14
Remove alpine from official build rid list.

eerhardt:
alpine isn't supported by corefx, so we can't build core-setup supporting it.

@mellinoe @ericstj @weshaggard @gkhanna79


eerhardt:
This is necessary to unblock official builds since they are failing now that there are no corefx packages for alpine.

gkhanna79:
We didn't build alpine in coresetup, right? If so, lgtm.


Thanks,
Gaurav

PR #2072, opened 2017-04-14
Revert "Fix pack.cmd ERRORLEVEL early exits"

dagood:
Unblocks the build: arm64 builds depend on pack.cmd errors being ignored

This reverts commit d6d8a172fccab36486fa23e17bbe920d2868c21d. (https://github.com/dotnet/core-setup/pull/2067)

We need to figure out what the right fix is to the arm64 builds before putting this back. Right now the ignored errors are causing the arm64 builds to "skip" creating UWP packages.

### DIFF
--- a/pkg/pack.cmd
+++ b/pkg/pack.cmd
@@ -15,24 +15,22 @@ if /i "%1" == "-portable"             (set __PortableBuildArgs=true&shift&goto A
 
 :: Initialize the MSBuild Tools
 call "%__ProjectDir%\init-tools.cmd"
-if NOT [!ERRORLEVEL!]==[0] goto :Error
 
 :: Restore dependencies mainly to obtain runtime.json
 pushd "%__ProjectDir%\deps"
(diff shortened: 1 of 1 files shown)


weshaggard:
We should just skip the UWP package build on arm64.

weshaggard:
Just add a check for arm64 here https://github.com/dotnet/core-setup/blob/master/pkg/packages.builds#L6.

gkhanna79:
@dagood Are you planning to update your PR and get the change merged? The current active build is going to fail as well.

weshaggard:
I pushed https://github.com/dotnet/core-setup/commit/a6505a9a8d4866dcb651e19bd8e0ade7af1f8ad5 to try and fix the issue. I'm going home for the weekend but hopefully this allows core-setup builds to succeed. If not we can revisit on Monday.

dagood:
Thanks, @weshaggard, looks like it's been working.

Issue #2081, opened 2017-04-17
Reduce occurences of build publishing race condition


chcosta:
We are occasionally hitting this failure in our official builds

> From: Davis Goodin 
Sent: Monday, April 17, 2017 9:49 AM
To: dotnet-bot <dn-bot@microsoft.com>; .NET Core Builds <dncbuilds@microsoft.com>
Subject: RE: DevDiv Build 20170417-04 failed
Looking into this. At first glance, it’s a publish semaphore locking issue, which is normally caused by two builds happening too close together:
Build failed: System.Exception: Unable to acquire lease on master/Binaries/sharedFxPublishSemaphore
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=682013

I’d think this could be avoided  by moving this code, https://github.com/dotnet/core-setup/blob/master/build_projects/dotnet-host-build/PublishTargets.cs#L89, around a bit to something more like

``` C#
                string leaseId;

                // Prevent race conditions by dropping a version hint of what version this is. If we see this file
                // and it is the same as our version then we know that a race happened where two+ builds finished 
                // at the same time and someone already took care of publishing and we have no work to do.
                if (AzurePublisherTool.IsLatestSpecifiedVersion(targetVersionFile))
                {
                    return c.Success();
                }
                else
                {
                    leaseId = AzurePublisherTool.AcquireLeaseOnBlob(semaphoreBlob);

                    Regex versionFileRegex = new Regex(@"(?<CommitHash>[\w\d]{40})");
```

dagood:
Doesn't this introduce the race condition again?

I'm also not sure if this solves the issue because the code here is for multiple build legs of the same pipebuild finishing at the same time. (Comment terminology is outdated.) The build issue is when multiple builds of different pipebuilds (different commits) finish at the same time, and are both are in that `else` block.

chcosta:
I titled the issue "reduce", not "eliminate".  I think this would help reduce the occurrences where we try to acquire a lease but are already on the specified version.  You're right, this would only apply to the current Pipebuild instance.  For fixing multiple pipebuild instances (multiple commits merged close together), we'd probably have to do something with how maestro schedules jobs.  

I think it will take quite a bit of testing to determine if this would make the situation better or worse.  I'm completely ok with closing this if the current logic is a non-issue for build legs within the same pipebuild instance.

chcosta:
[...]

PR #2085, opened 2017-04-17
Block build finalization to prevent race conditions

gkhanna79:
Block build finalization step if another build is in the process of doing so.

@chcosta @dagood PTAL

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using Microsoft.DotNet.Cli.Build.Framework;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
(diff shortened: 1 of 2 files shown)


build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -83,21 +85,81 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
             {
                 string targetContainer = $"{Channel}/Binaries/Latest/";
                 string targetVersionFile = $"{targetContainer}{CommitHash}";
+                string flagFinalizationInProgress = $"{targetContainer}/Finalizing";
                 string semaphoreBlob = $"{Channel}/Binaries/sharedFxPublishSemaphore";
+                
                 AzurePublisherTool.CreateBlobIfNotExists(semaphoreBlob);
 
-                string leaseId = AzurePublisherTool.AcquireLeaseOnBlob(semaphoreBlob);
```
dagood: It looks like we already have the looping logic you added below in `AcquireLeaseOnBlob` by passing the `maxWaitDefault` parameter. Can't we just pass `TimeSpan.FromSeconds(1800)` into this function call to extend the timeout?

gkhanna79: That loop magic is to keeping trying to acquire the lease on the blob. Here, I am trying to not even get into that loop if we have finalization going on for a different build - effectively at two different levels and if this works, we may want to remove that loop completely from lease acquisition.

gkhanna79:
Another build failed with same issues - @dagood I would like to get the changes merged to see if they help. Are you fine with getting these merged?

dagood:
Something messed up and I didn't get an email about your thread reply, sorry for the delay.

I still don't see what this does differently from the existing race condition prevention code except have a longer timeout. I'll see if I have any code-level comments in case you want to get this merged anyway.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -83,21 +85,81 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
...
+                while(true)
                 {
-                    AzurePublisherTool.ReleaseLeaseOnBlob(semaphoreBlob, leaseId);
-                    return c.Success();
+                    Stopwatch stopWatch = new Stopwatch();
+                    stopWatch.Start();
+
+                    // Wait for 30 mins to get access to finalization flag - if it does not come through, then we will
+                    // fail the build.
+                    bool fIsAnotherFinalizationInProgress = false;
```
dagood: `f` prefix doesn't seem like standard C#.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -219,14 +219,21 @@ public void ReleaseLeaseOnBlob(string blob, string leaseId)
             cloudBlob.ReleaseLeaseAsync(ac).Wait();
         }
 
-        public bool IsLatestSpecifiedVersion(string version)
+        public bool DoesBlobReferenceExist(string version)
```
dagood: `version` -> `blob`?

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -83,21 +85,81 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
...
+                    if (fIsAnotherFinalizationInProgress)
+                    {
+                        // We timed out waiting for finalization - throw an exception
+                        throw new InvalidOperationException("Timed out while waiting for finalization flag to be cleared from another build!");
+                    }
+
+                    // Prevent race conditions by dropping a version hint of what version this is. If we see this file
+                    // and it is the same as our version then we know that a race happened where two+ builds finished 
+                    // at the same time and someone already took care of publishing and we have no work to do.
+                    if (AzurePublisherTool.DoesBlobReferenceExist(targetVersionFile))
```
dagood: Having this outside where we have the lock makes me uncomfortable. Although in our current situation (long publishes) it wouldn't cause a race condition, if publish suddenly took less time than the `AcquireLeaseOnBlob` default timeout I think it could.

gkhanna79: We are not using the blob, for which lease is acquired, in this check. This is also what @chcosta proposed to move around.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
gkhanna79: There is nothing version specific in the implementation of the function - it simply checks for  a string reference in the blob.

build_projects/dotnet-host-build/PublishTargets.cs
dagood: I don't see why it makes a difference which blob is locked. I agree this is similar to what @chcosta suggested, but in that case it was more clear-cut and I explained the race condition I see here: https://github.com/dotnet/core-setup/issues/2081#issuecomment-294537361. I think this PR mitigates the race condition by having the "polling lock" before getting a lease, but the core problem seems the same.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
dagood: I should have been clearer--I was suggesting renaming the `version` parameter to `blob` to match the new name of the method.

chcosta:
This work will need to be done again and/or ported when core-setup buildtools work gets merged.

PR #2090, opened 2017-04-17
Removing Company Info from apphost.exe

ramarag:
This removes the mention of microsoft in the binary details

@gkhanna79 @eerhardt PTAL

### DIFF
--- a/build_projects/dotnet-host-build/CompileTargets.cs
+++ b/build_projects/dotnet-host-build/CompileTargets.cs
@@ -78,11 +78,30 @@ namespace Microsoft.DotNet.Host.Build
             var buildNumberMajor = hostVer.VerRsrcBuildMajor;
             var buildNumberMinor = hostVer.VerRsrcBuildMinor;
             var buildDetails = $"{semVer}, {commitHash} built by: {System.Environment.MachineName}, UTC: {DateTime.UtcNow.ToString()}";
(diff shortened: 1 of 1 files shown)


ramarag:
After this change AppHost's properties will look like : 

image

gkhanna79:
Can we also cleanup the Description, File Version and Product version @ramarag?

ramarag:
@gkhanna79 i would keep them, why do you think we should strip them off ?

gkhanna79:
They are reflective of our build and have no semblence to the app in which apphost will be used for. Thus, it will be good to clear them out.

gkhanna79:
We should get this in for Preview1 build. @ramarag can we get this in by early next week?

ramarag:
@gkhanna79 i believe the information we have left in the binary will help us trouble shoot when and if we ship another version of apphost.exe, I understand we never want to change, but we should allow for them to be easily distinguishable

gkhanna79:
I believe we can determine the version based upon the toolset used to build the application or the .NET Core App that is targeted. From an end-user perspective, we really do not want to have any such information, to avoid the case of multiple standalone hosts having the same information that could confuse locked down scenarios from locking all of them when they intended to lock down just one.

ramarag:
closing this as versioning tooling might change with @chcosta' changes

PR #2091, opened 2017-04-17
Try longer lease acquisition

gkhanna79:
@dagood PTAL - Lets try with just longer timeout. If this does not help, I will look into merging the other PR.


gkhanna79:
Merging longer timeout to see if it helps.

dagood:
LGTM. For more than a couple merges near each other we'll need a more scalable approach eventually.

PR #2093, opened 2017-04-18
Adding Developer stores when unzipped CLI's are used

ramarag:
Adding a path like 

<dotnet_relative_path>\store\<arch>\<tfm> so that when developers uzip the CLI and point to <dotnet_relative_path>\store as the store path, the store feature will actually kick in 

@gkhanna79 @eerhardt PTAL

fixes the originally filed bug https://github.com/dotnet/sdk/issues/928

### DIFF
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -22,11 +22,11 @@ arguments_t::arguments_t() :
  * Setup the shared store directories.
  *
  *  o %DOTNET_SHARED_STORE% -- multiple delimited paths
- *  o $HOME/.dotnet/{x86|x64}/ or %USERPROFILE%\.dotnet\{x86|x64}
- *  o dotnet.exe relative shared store
+ *  o $HOME/.dotnet/{x86|x64}/tfm or %USERPROFILE%\.dotnet\{x86|x64}\<tfm>
+ *  o dotnet.exe relative shared store\<arch>\<tfm>
  *  o Global location
  *      Windows: C:\Program Files (x86) or
- *      Unix: directory of dotnet on the path.
+ *      Unix: directory of dotnet on the path.\<arch>\<tfm>
  */
 void setup_shared_store_paths(const hostpolicy_init_t& init, const pal::string_t& own_dir, arguments_t* args)
 {
@@ -52,12 +52,14 @@ void setup_shared_store_paths(const hostpolicy_init_t& init, const pal::string_t
     {
         args->dotnet_shared_store = own_dir;
         append_path(&args->dotnet_shared_store, RUNTIME_STORE_DIRECTORY_NAME);
+        append_path(&args->dotnet_shared_store, get_arch());
         append_path(&args->dotnet_shared_store, init.tfm.c_str());
     }
 
     // Global shared store dir
     if (get_global_shared_store_dir(&args->global_shared_store))
     {
+        append_path(&args->global_shared_store, get_arch());
         append_path(&args->global_shared_store, init.tfm.c_str());
     }
 }


src/corehost/cli/args.cpp
```diff
@@ -54,6 +54,17 @@ void setup_shared_store_paths(const hostpolicy_init_t& init, const pal::string_t
         append_path(&args->dotnet_shared_store, RUNTIME_STORE_DIRECTORY_NAME);
         append_path(&args->dotnet_shared_store, init.tfm.c_str());
     }
+    
+    // "dotnet.exe" relative shared store folder, adding the arch so that unzipped developer scenario works
+    if (init.host_mode == host_mode_t::muxer)
```
gkhanna79: Can we merge this block with the immediately preceding muxer check block that also operates upon store primitives?

gkhanna79: Is this to support the layout of runtime store that is packaged with CLI in a zip/tar?

gkhanna79: Why does CLI attach an arch to the location? If the arch is not specified, then wont this merge into the earlier "shared store" location relative to dotnet.exe?

ramarag: This is to support developers who download a zipped cli to try out this feature.
I have merge the condition, the reason for adding the path, is at cli runtime store creation time we always add architecture to the output path

src/corehost/cli/args.cpp
```diff
@@ -53,6 +53,16 @@ void setup_shared_store_paths(const hostpolicy_init_t& init, const pal::string_t
         args->dotnet_shared_store = own_dir;
         append_path(&args->dotnet_shared_store, RUNTIME_STORE_DIRECTORY_NAME);
         append_path(&args->dotnet_shared_store, init.tfm.c_str());
+
+        //Also add an architecture specific path to allow for unzipped developer cli experience
```
eerhardt: If we take this approach, the comment at the top of this method should be updated.

src/corehost/cli/args.cpp
```diff
@@ -53,6 +53,16 @@ void setup_shared_store_paths(const hostpolicy_init_t& init, const pal::string_t
...
+        //where developer points runtime store operation output to "dotnet.exe" relative shared store folder
+
+        pal::string_t local_developer_store = own_dir ;
+        append_path(&local_developer_store, RUNTIME_STORE_DIRECTORY_NAME);
+        append_path(&local_developer_store, get_arch());
+        append_path(&local_developer_store, init.tfm.c_str());
+        local_developer_store.push_back(PATH_SEPARATOR);
+        (args->env_shared_store).push_back(local_developer_store);
```
eerhardt: One thing I dislike about this approach is that it really interweaves where we are looking in a fashion that I don't think makes sense:

o %DOTNET_SHARED_STORE% -- multiple delimited paths
o dotnet.exe relative shared store - with architecture appended     `<-- you are adding this`
o $HOME/.dotnet/{x86|x64}/ or %USERPROFILE%\.dotnet\{x86|x64}
o dotnet.exe relative shared store - no architecture
o Global location

gkhanna79: The ordering being described above by @eerhardt also does not lineup with the multi-level lookup we have for SharedFX and SDK (see https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/multilevel-sharedfx-lookup.md). 

Why do we need the entry @eerhardt calls out above to be added. Why cannot it be the non-architecture entry (penultimate in the list) and everything just works? Arch is only required for userprofile case.

eerhardt: My thoughts are:

1. We shouldn't have the non-architecture locations anymore.  We should just always append the architecture.  Like we talked about in https://github.com/dotnet/sdk/issues/928#issuecomment-289089579.
2. We need to make the change for the global location as well, since it uses a non-architecture location.  Again, don't have the non-architecture locations at all.  It simplifies everything.

gkhanna79: Agreed about (1) above. It was not reflected in the change and hence, I was not sure how it was being planned - hence, my questions. To account for the global location to have the arch, I presume platform installers for the Store artifacts would need to carry the change, right? Also, looking up using Arch is specific to store location only.

ramarag: @eerhardt good catch for global location, we could have picked up the wrong architecture there.
I have added architecture for all probing locations now

gkhanna79:
@ramarag What are the next steps for this? Shouldn't this be in Preview1 build?

gkhanna79:
@eerhardt @ramarag Should this be ported to Preview1 since ASP.NET is going to use this feature?

eerhardt:
> Should this be ported to Preview1 since ASP.NET is going to use this feature?

I think we should bring it in.  But my reasoning is that this would technically be a breaking change.

gkhanna79:
It is breaking relative to what?

eerhardt:
> It is breaking relative to what?

We are moving the "store" directory that we are finding.

Before this change, we would find it here:

`C:\Program Files\dotnet\store\netcoreapp2.0`

After this change, we find it here:

`C:\Program Files\dotnet\store\x64\netcoreapp2.0`

gkhanna79:
So it is breaking relative to what we had earlier under development, right? If so, is this just about communicating the change? Is it observable to external folks (e.g. ASP.NET)? 

I think it is important to get this change in for Preview1 so that Preview Antares is on the right plan from get-go.

eerhardt:
It is important to get into Preview1 now, so we don't have to make a breaking change AFTER Preview1.  Both for ASP.NET and any other consumer of this feature.

> So it is breaking relative to what we had earlier under development, right?

Yes, we have never released this feature.  But if we shipped Preview1 without this change, it would be a breaking change to make it later.

gkhanna79:
And is this change an implementation detail or user-visible?

Either ways, @ramarag please get a release/2.0.0 PR going and send email to shiproom about it.

eerhardt:
> And is this change an implementation detail or user-visible?

Since it involves where someone would install these files on the file system, this is user-visible.

PR #2094, opened 2017-04-18
Clean up OSX installer.

eerhardt:
Remove unused resources from sharedhost and hostfxr pkgs.
Remove unused conclusion.html from sharedframework product.
Start using the welcome.html page for the sharedframework product.

@livarcocc @gkhanna79 @ramarag


gkhanna79:
@dotnet-bot test this please

eerhardt:
> I presume you have tested the change.

Yes, I just ran through the installer and everything worked.

The thing I noticed is that on the Welcome page we have:

https://github.com/dotnet/core-setup/blob/master/packaging/osx/sharedframework/resources/en.lproj/welcome.html#L10

```
Microsoft .NET Core Host
```

I think we want to change that to say `Microsoft .NET Core Runtime` to match the title at the top of the installer.

If you agree, I'll make that change as well.

gkhanna79:
I agree.


Thanks,
Gaurav

eerhardt:
```
08:46:39 C:\build\work\eca3d12b\wix3\src\ext\UIExtension\wixlib\LicenseAgreementDlg.wxs(32) : error LGHT0103 : The system cannot find the file 'D:\j\workspace\release_windo---0c55dee9\packaging\osx\sharedhost\resources\en.lproj\eula.rtf'.
```

Oh, you've got to be kidding me.

Petermarcu:
Seems we have some cross build pollution:(

eerhardt:
> Seems we have some cross build pollution:(

The Windows installer code is reading/using the eula's from the `OSX` directory.

I guess I'll add the eula's back.

Petermarcu:
Or put them in a place that makes more sense for Windows?

eerhardt:
> Or put them in a place that makes more sense for Windows?

Some of the resources are shared between Windows and OSX (background and sharedframework eula).  Since this has been status quo for over a year, I didn't feel it was worth changing now.  Also, this is how the CLI does it, so I'd have to move all this stuff into a "common" location in both repos to keep them consistent.

Just doesn't feel worth the effort.

Issue #1721, opened 2017-03-13
build-rootfs.sh does not always set the exit code to a non-zero value when hitting a failure


MichaelSimons:
The [build-rootfs.sh](https://github.com/dotnet/core-setup) script does not set the exit code correctly upon most failures.  This is problematic when invoking directly because the output is very lengthy and you can easily miss errors when scanning.  More importantly, because of this it is not practical to invoke this script programmatically and determine if it succeeds.

Sample error I would expect to set the exit code to 1
```
$ ./cross/build-rootfs.sh arm xenial --skipunmount
...
Err:101 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf libgettextpo-dev armhf 0.19.7-2ubuntu3
  Hash Sum mismatch
Get:102 http://ports.ubuntu.com/ubuntu-ports xenial-updates/main armhf libicu-dev armhf 55.1-7ubuntu0.1 [8343 kB]
Get:103 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf libllvm3.6v5 armhf 1:3.6.2-3ubuntu2 [7542 kB]
Get:104 http://ports.ubuntu.com/ubuntu-ports xenial-updates/main armhf libpython2.7 armhf 2.7.12-1ubuntu0~16.04.1 [936 kB]
Get:105 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf liblldb-3.6 armhf 1:3.6.2-3ubuntu2 [6494 kB]
Get:106 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf libffi-dev armhf 3.2.1-4 [160 kB]
Get:107 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf llvm-3.6-runtime armhf 1:3.6.2-3ubuntu2 [40.1 kB]
Get:108 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf llvm-3.6 armhf 1:3.6.2-3ubuntu2 [1197 kB]
Get:109 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf libtinfo-dev armhf 6.0+20160213-1ubuntu1 [66.3 kB]
Get:110 http://ports.ubuntu.com/ubuntu-ports xenial/main armhf llvm-3.6-dev armhf 1:3.6.2-3ubuntu2 [11.3 MB]
[...]

PR #2116, opened 2017-04-19
Exit with failure when build-rootfs.sh fails

wtgodbe:
Attempts to address https://github.com/dotnet/core-setup/issues/1721

CC @MichaelSimons @gkhanna79

### DIFF
--- a/cross/build-rootfs.sh
+++ b/cross/build-rootfs.sh
@@ -128,20 +128,46 @@ fi
 if [ -d "$__RootfsDir" ]; then
     if [ $__SkipUnmount == 0 ]; then
         umount $__RootfsDir/*
+        if [ $? -ne 0 ]; then
+            echo "Failed to unmount RootfsDir."
+            exit 1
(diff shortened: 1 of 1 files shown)


cross/build-rootfs.sh
```diff
@@ -128,20 +128,52 @@ fi
...
+        fi
     fi
     rm -rf $__RootfsDir
 fi
 
 if [[ -n $__LinuxCodeName ]]; then
+    __InstallFailed=0
     qemu-debootstrap --arch $__UbuntuArch $__LinuxCodeName $__RootfsDir $__UbuntuRepo
+    if [ $? -ne 0 ]; then
+        __InstallFailed=1
```
gkhanna79: Can we, instead, jump to a label immediately instead of proceeding to the next command?

wtgodbe: @gkhanna79 just updated the change, is that better?

PR #2125, opened 2017-04-20
Update update-dependencies to use VersionTools.

eerhardt:
This allows us to use all the goodness of the build-tools features.  For example, when a new build comes out, and an existing update-dependencies PR exists, it will update the existing PR.

/cc @gkhanna79


eerhardt:
I tested it locally on my machine.  You can just run the update-dependencies script locally.

The updates all worked correctly.  The only thing that wasn't tested was creating the PR, but that is handled by VersionTools now.

PR #2135, opened 2017-04-21
Specify portable RID for Windows build for binaries to be signed.

gkhanna79:
@eerhardt PTAL.

This should specify the correct RID to be used for signing purposes (specified at https://github.com/dotnet/core-setup/blob/master/buildpipeline/Core-Setup-Signing-Windows-x64.json#L490) for portable builds.


gkhanna79:
CC @dagood @wtgodbe

gkhanna79:
Merging since this does not affect CI but only official builds.

Issue #1994, opened 2017-04-07
Apparent published output overlap between build legs


dagood:
This is the most likely cause of https://github.com/dotnet/core-setup/issues/1824:

> "Failed due to “The specified block list is invalid.”"

When multiple files are published to the same blob at the same time, they fail because the change references the state of the blob when the copy begins.

First, I want to port https://github.com/dotnet/core-setup/pull/1951 to master so that we can tell exactly what each leg is publishing.

If it turns out there is no overlap, I'll close this.

/cc @gkhanna79 @MattGal

gkhanna79:
Sounds good to me.

dagood:
There is a lot of overlap. For source/dest info: https://gist.github.com/dagood/a5d06e2ffaebeb3a9e92c06087ea82c2

```
2 for 'master/Binaries/2.0.0-preview1-001943-00/dotnet-hostfxr-win-x64.2.0.0-preview1-001943-00.zip'
2 for 'master/Binaries/2.0.0-preview1-001943-00/dotnet-sharedframework-symbols-win-x64.2.0.0-preview1-001943-00.zip'
2 for 'master/Binaries/2.0.0-preview1-001943-00/dotnet-sharedframework-win-x64.2.0.0-preview1-001943-00.zip'
2 for 'master/Binaries/2.0.0-preview1-001943-00/dotnet-win-x64.2.0.0-preview1-001943-00.zip'
2 for 'master/Binaries/2.0.0-preview1-001943-00/Microsoft.DotNet.PlatformAbstractions.2.0.0-preview1-001943.nupkg'
2 for 'master/Binaries/2.0.0-preview1-001943-00/Microsoft.DotNet.PlatformAbstractions.2.0.0-preview1-001943.symbols.nupkg'
2 for 'master/Binaries/2.0.0-preview1-001943-00/Microsoft.Extensions.DependencyModel.2.0.0-preview1-001943.nupkg'
2 for 'master/Binaries/2.0.0-preview1-001943-00/Microsoft.Extensions.DependencyModel.2.0.0-preview1-001943.symbols.nupkg'
11 for 'master/Binaries/2.0.0-preview1-001943-00/Microsoft.NETCore.App.2.0.0-preview1-001943-00.nupkg'
11 for 'master/Binaries/2.0.0-preview1-001943-00/Microsoft.NETCore.App.2.0.0-preview1-001943-00.symbols.nupkg'
[...]

PR #2153, opened 2017-04-24
Only publish identity packages from a single build leg (windows x64).

### LINKED ISSUE #2136: Need to set SIGNED_PACKAGES on portable windows builds

eerhardt:
Also, fail publishing if an existing blob exists - no more overwriting azure storage blobs.  We need to ensure each leg publishes distinct packages so we don't get into race conditions in publishing.

Fix #2136, #1994 and #1824
@ericstj @dagood @chcosta @gkhanna79 @wtgodbe

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -309,12 +309,25 @@ namespace Microsoft.DotNet.Host.Build
         {
             foreach (var file in Directory.GetFiles(Dirs.CorehostLocalPackages, "*.nupkg"))
             {
-                var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}/{Path.GetFileName(file)}";
-                AzurePublisherTool.PublishFile(hostBlob, file);
-                Console.WriteLine($"Publishing package {hostBlob} to Azure.");
+                if (!IsRidAgnosticPackage(file) || EnvVars.PublishRidAgnosticPackages)
+                {
+                    var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}/{Path.GetFileName(file)}";
+                    AzurePublisherTool.PublishFile(hostBlob, file);
+                    Console.WriteLine($"Publishing package {hostBlob} to Azure.");
+                }
+                else
+                {
+                    Console.WriteLine($"Skipping publishing package {file} to Azure because it is RID-agnostic and EnvVars.PublishRidAgnosticPackages is not set to true.");
+                }
             }
 
             return c.Success();
+        }
+
+        private static bool IsRidAgnosticPackage(string packagePath)
+        {
+            var fileName = Path.GetFileName(packagePath);
+            return !fileName.StartsWith("runtime.", StringComparison.OrdinalIgnoreCase);
         }
 
         [Target]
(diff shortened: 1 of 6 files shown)


buildpipeline/Core-Setup-Signing-Windows-x64.json
```diff
@@ -477,7 +477,10 @@
     "GITHUB_PASSWORD": {
       "value": "PassedViaPipeBuild"
     },
-    "SIGNED_PACKAGES": {
+    "BUILD_FULL_PLATFORM_MANIFEST": {
+      "value": "true"
+    },
+    "PUBLISH_RID_AGNOSTIC_PACKAGES": {
       "value": "true"
```
dagood: What's the difference between these vars? It seems like `BUILD_FULL_PLATFORM_MANIFEST` would be to address what @ericstj said in https://github.com/dotnet/core-setup/issues/2136#issuecomment-296316740:

>  Tests and shared framework publishing rely on having identity packages produced in all legs.

If that were the case I would expect it to be true for all legs, though.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -309,21 +309,34 @@ public static BuildTargetResult PublishCoreHostPackages(BuildTargetContext c)
...
         }
 
+        private static bool IsRidAgnosticPackage(string packagePath)
+        {
+            var fileName = Path.GetFileName(packagePath);
+            return !fileName.StartsWith("runtime.", StringComparison.OrdinalIgnoreCase);
+        }
+
         [Target]
         public static BuildTargetResult PublishManagedPackages(BuildTargetContext c)
```
gkhanna79: Should we rename this function to something like "PublishRIDAgnosticPackages"?

buildpipeline/Core-Setup-Signing-Windows-x64.json
gkhanna79: I think this distinction helps in managing them independent of each other.

dagood: I think I see the scenario this prepares us for now: if we need another pipebuild leg that uses the Core-Setup-Signing-Windows-x64 definition, then we can enable `BUILD_FULL_PLATFORM_MANIFEST` for both, but only enable `PUBLISH_RID_AGNOSTIC_PACKAGES` for one. Thanks.

eerhardt:
Our latest build `2.0.0-preview2-002073-00` had this exact problem in it.  The last leg to finish was the `win-x86` leg.  Thus it overwrote the "good" identity package for Microsoft.NETCore.App and it doesn't have the full manifest file in it.

This causes breaks in the CLI - see https://github.com/dotnet/cli/pull/6402.  When the manifest file isn't created correctly, publishing an app will cause all the native assets (like `runtimes/fedora/System.Security.Native.OpenSsl.so`) to show up app local in the `runtimes` folder.  Later the debian installer breaks because you can't have fedora and rhel assemblies in an ubuntu .deb package.

build_projects/dotnet-host-build/PublishTargets.cs
eerhardt: No, this function only uploads the managed packages.

The nupkgs are split in our build output - everything that gets built from the `pkg` folder goes to `ROOT\artifacts\RID\corehost`.  While the packages that are built with `dotnet pack` go to `ROOT\artifacts\RID\packages`

PR #2154, opened 2017-04-24
Allow additional ICompilationAssemblyResolvers

### LINKED ISSUE #625: Ability to load dependency model from non-app assembly
Moved from https://github.com/dotnet/cli/issues/3253 on behalf of @pakrym.

----

To allow custom compilation assembly load locations.

@pranavkm @davidfowl

eerhardt:
This allows for callers to use their own compilation resolvers along with the default resolvers shipped in the DependencyModel.

Fix #625

@pakrym @ramarag

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/CompilationLibrary.cs
+++ b/src/Microsoft.Extensions.DependencyModel/CompilationLibrary.cs
@@ -51,8 +51,32 @@ namespace Microsoft.Extensions.DependencyModel
 
         public IEnumerable<string> ResolveReferencePaths()
         {
+            return ResolveReferencePaths(DefaultResolver);
+        }
+
+        public IEnumerable<string> ResolveReferencePaths(params ICompilationAssemblyResolver[] additionalResolvers)
+        {
+            ICompilationAssemblyResolver resolver;
+            if (additionalResolvers?.Length > 0)
+            {
+                var allResolvers = new ICompilationAssemblyResolver[additionalResolvers.Length + 1];
+                additionalResolvers.CopyTo(allResolvers, 0);
+                allResolvers[additionalResolvers.Length] = DefaultResolver;
+
+                resolver = new CompositeCompilationAssemblyResolver(allResolvers);
+            }
+            else
+            {
+                resolver = DefaultResolver;
+            }
+
+            return ResolveReferencePaths(resolver);
+        }
+
+        private IEnumerable<string> ResolveReferencePaths(ICompilationAssemblyResolver resolver)
+        {
             var assemblies = new List<string>();
-            if (!DefaultResolver.TryResolveAssemblyPaths(this, assemblies))
+            if (!resolver.TryResolveAssemblyPaths(this, assemblies))
             {
                 throw new InvalidOperationException($"Cannot find compilation library location for package '{Name}'");
             }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Extensions.DependencyModel/CompilationLibrary.cs
```diff
@@ -51,8 +51,32 @@ public class CompilationLibrary : Library
 
         public IEnumerable<string> ResolveReferencePaths()
         {
+            return ResolveReferencePaths(DefaultResolver);
+        }
+
+        public IEnumerable<string> ResolveReferencePaths(params ICompilationAssemblyResolver[] additionalResolvers)
```
pakrym: I'm a bit concerned by performance of doing this per `CompilationLibrary`. Also resolvers happen throw in some cases (https://github.com/dotnet/core-setup/blob/master/src/Microsoft.Extensions.DependencyModel/Resolution/AppBaseCompilationAssemblyResolver.cs#L102) so having additional resolvers last might not solve everything.

eerhardt: > so having additional resolvers last might not solve everything.

The additional resolvers are called first, then the default resolver.

> I'm a bit concerned by performance of doing this per CompilationLibrary.

There should be no difference in perf between what is happening today and this code. Why the concern now?

pakrym: > The additional resolvers are called first, then the default resolver.
My bad, read the code wrong. 
> There should be no difference in perf between what is happening today and this code. Why the concern now?
New array and `CompositeCompilationAssemblyResolver` per invocation.

src/Microsoft.Extensions.DependencyModel/CompilationLibrary.cs
```diff
@@ -51,31 +51,31 @@ public class CompilationLibrary : Library
...
         public IEnumerable<string> ResolveReferencePaths()
         {
-            return ResolveReferencePaths(DefaultResolver);
+            var assemblies = new List<string>();
+
+            return ResolveReferencePaths(DefaultResolver, assemblies);
         }
 
-        public IEnumerable<string> ResolveReferencePaths(params ICompilationAssemblyResolver[] additionalResolvers)
+        public IEnumerable<string> ResolveReferencePaths(params ICompilationAssemblyResolver[] customResolvers)
```
pakrym: What do you think about exposing API to create `ICompilationAssemblyResolver` that extends DefaultResolver with custom ones (`CompilationLibrary.CreateCustomReolver`?) and then have `ResolveReferencePaths` overload that takes a single `ICompilationAssemblyResolver`?

eerhardt: I'm not sure what the advantage of that approach is.  What problems can you solve with that API that you can't with this one?

eerhardt: I also don't like exposing DefaultResolver outside of the class (either explicitly or implicitly by passing references to it outside of the class).

pakrym: Duplicating loop from composite resolver in `ResolveReferencePaths`

pakrym:
Is is not worth spending too much time. Approved.

PR #2167, opened 2017-04-25
Fix casing of OfficialBuildRID

### LINKED ISSUE #2161: M.N.App Identity package does not have reference to Tizen and ArmEl (Identity?) packages
@weshaggard this is the issue we were looking into on Friday that @eerhardt broughtup. Can you please look into this as we should get this in for Preview1?

CC @ericstj

ericstj:
Working around https://github.com/Microsoft/msbuild/issues/1751

Fixes #2161


weshaggard:
Wow I didn't know about this case-sensitive bug.

PR #2171, opened 2017-04-25
Use checked-in Docker Containers for Linux build

wtgodbe:
Attempts to resolve https://github.com/dotnet/core-setup/issues/1719

CC @gkhanna79 

@ravimeda @dagood @MichaelSimons @crummel PTAL


wtgodbe:
Should I remove the files at https://github.com/dotnet/core-setup/tree/master/scripts/docker as well?

buildpipeline/pipeline.json
```diff
@@ -33,11 +31,7 @@
           "Name": "Core-Setup-Linux",
           "Parameters": {
             "PB_DockerOS": "ubuntu.14.04",
-            "REPO_ID": "562fbfe0b2d7d0e0a43780c4",
-            "REPO_USER": "dotnet",
-            "REPO_SERVER": "azure-apt-cat.cloudapp.net",
-            "CLI_NUGET_FEED_URL": "https://dotnet.myget.org/F/cli-deps/api/v2/package",
-            "CLI_NUGET_SYMBOLS_FEED_URL": "https://dotnet.myget.org/F/cli-deps/symbols/api/v2/package"
```
dagood: Are you planning on adding these to the PipeBuild root def?

buildpipeline/Core-Setup-Linux.json
```diff
@@ -88,8 +88,8 @@
         "definitionType": "task"
       },
       "inputs": {
-        "scriptPath": "build.sh",
-        "args": "$(BuildArguments)",
+        "scriptPath": "scripts/dockerrun-as-current-user.sh",
+        "args": "-t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB $(DockerImageName) /bin/bash -c \"HOME=/opt/code; git clean -X -d -f; ./build.sh $(BuildArguments)\"",
```
dagood: I don't think `-e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB` is enough for all builds, e.g. `REPO_ID`. I think it would make sense to change/add to the build scripts to work with a specified image name to avoid duplicating the environment variable names [in dockerrun](https://github.com/dotnet/core-setup/blob/53d6b3c1488ffd67888b86a5feda12e3a8a29c84/scripts/dockerrun.sh#L138-L154). Then the other build defs could consolidate their `-e` strings too.

buildpipeline/pipeline.json
wtgodbe: That was my plan, is it alright to have these be passed to every build, and not just the Linux ones?

wtgodbe:
@dagood I just updated the change to explicitly pass the environment variables to dockerrun-as-current-user.sh - can you PTAL?

MichaelSimons:
@wtgodbe, do the microsoft/dotnet-buildtools-prereqs images really contain all of the prereqs core-setup builds need?  Have you run test builds with the images?

MichaelSimons:
I would be in favor of removing the checked in Dockerfiles.  I think they will become obsolete if not used in builds and therefore should be cleaned up.

buildpipeline/pipeline.json
```diff
@@ -32,12 +33,14 @@
...
           "Parameters": {
-            "PB_DockerOS": "ubuntu.14.04",
             "REPO_ID": "562fbfe0b2d7d0e0a43780c4",
             "REPO_USER": "dotnet",
             "REPO_SERVER": "azure-apt-cat.cloudapp.net",
             "CLI_NUGET_FEED_URL": "https://dotnet.myget.org/F/cli-deps/api/v2/package",
-            "CLI_NUGET_SYMBOLS_FEED_URL": "https://dotnet.myget.org/F/cli-deps/symbols/api/v2/package"
+            "CLI_NUGET_SYMBOLS_FEED_URL": "https://dotnet.myget.org/F/cli-deps/symbols/api/v2/package",
+            "PB_EnvVars": "-e CLI_NUGET_FEED_URL -e CLI_NUGET_SYMBOLS_FEED_URL ",
+            "PB_DockerOS": "ubuntu.14.04",
```
MichaelSimons: Is there consumer of PB_DockerOS after this change?  Seems like it could be removed.

buildpipeline/pipeline.json
dagood: Yeah, the steps that use the values are already conditioned to only run on the correct platform, so passing them when they aren't used won't change any behavior.

buildpipeline/pipeline.json
wtgodbe: https://github.com/dotnet/core-setup/pull/2171/files#diff-5219404725e88f32bbcd0eabecc415eeR216 uses PB_DockerOS

wtgodbe:
@MichaelSimons I'm in the process of setting up a test pipebuild with these changes now (using my own clone of the Master Pipebuild definition, so it won't interfere with official pipebuilds)

buildpipeline/pipeline.json
MichaelSimons: What do you think about replaced it with PB_DockerTag?

wtgodbe:
Just updated to delete the dockerfiles.

MichaelSimons:
With the removal of the dockerfiles, is there a need for https://github.com/dotnet/core-setup/blob/master/scripts/dockerrun.sh anymore?

buildpipeline/Core-Setup-Linux.json
```diff
@@ -88,8 +88,8 @@
         "definitionType": "task"
       },
       "inputs": {
-        "scriptPath": "build.sh",
-        "args": "$(BuildArguments)",
+        "scriptPath": "scripts/dockerrun-as-current-user.sh",
+        "args": "-t --rm --sig-proxy=true --name $(DockerContainerName) -v $(Build.SourcesDirectory):/opt/code -w /opt/code -e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB -e REPO_ID -e REPO_USER -e REPO_SERVER $(PB_EnvVars)$(DockerImageName) /bin/bash -c \"HOME=/opt/code; git clean -X -d -f; ./build.sh $(BuildArguments)\"",
```
dagood: I would prefer refactoring this so `-e CONNECTION_STRING -e PUBLISH_TO_AZURE_BLOB -e REPO_ID -e REPO_USER -e REPO_SERVER` doesn't duplicate [dockerrun.sh#L138-L154](https://github.com/dotnet/core-setup/blob/53d6b3c1488ffd67888b86a5feda12e3a8a29c84/scripts/dockerrun.sh#L138-L154). There's no need to be specific when specifying these like what you have set up now with `PB_EnvVars` for extras.

To refactor, I think we can remove the existing `docker build` functionality of `dockerrun.sh` and have it take an image name, since we won't need the building feature after this change. `dockerrun-as-current-user.sh` could call `dockerrun.sh`, maybe.

wtgodbe:
@MichaelSimons the jobs are failing on the docker images with the following:

2017-04-25T22:11:25.5163430Z /opt/code/build_projects/dotnet-host-build/build.sh: line 140: curl: command not found
2017-04-25T22:11:25.5186830Z /opt/code/build_projects/dotnet-host-build/build.sh: line 146: dotnet: command not found

How can we go about getting these commands on the images?

Also, I'll take a look at modifying dockerrun/dockerrun-as-current-user.sh

dagood:
`dotnet` is what we're downloading with `curl`. :smile: Currently in CoreFX/CoreCLR we support both `wget` and `curl`, but I believe CLI-style repos assume you have `curl` installed.

It might make sense to standardize on `curl` and add it to our build images. I'm not sure if this might be a problem for some platform/distro or some devs--if it is, we'll need to work support for both into the scripts here, too.

MichaelSimons:
@wtgodbe - The Dockerfiles for the microsoft/dotnet-buildtools-prereqs images are located [here](https://devdiv.visualstudio.com/DevDiv/_git/DotNetCore?path=%2Fdockerfiles&version=GBmaster&_a=contents).  I have no objection to adding curl.  That being said, I am skeptical that there aren't additional components required by core-setup.  What I would suggest is that you define a Dockerfile that composes on top of the microsoft/dotnet-buildtools-prereqs images and figure out the bare minimum Dockerfile needed to build Core-Setup.  Once that is done, we can decide which components make sense to put into the base images.  The remaining components can go into a Core-Setup specific Dockerfile.  I can also help you get the official images built and pushed to microsoft/dotnet-buildtools-prereqs.  Feel free to ping me directly if you have questions or need help.

chcosta:
the core-setup transition to buildtools moved us to using the Microsoft/dotnet-buildtools-prereqs docker images.  I'm all for adding curl to the prereqs on the images, but as a temporary workaround I created a hacked version of dotnet-install.sh which would fall back to using wget.  Please undo this hack with your changes https://github.com/dotnet/core-setup/blob/master/src/test/bootstrap/bootstrap.sh#L197 and delete the checked in dotnet-install.sh file, https://github.com/dotnet/core-setup/blob/master/src/test/bootstrap/dotnet-install.sh.

PR #2177, opened 2017-04-25
Disable RIDs not built

gkhanna79:
Reverting to unblock Preview1 builds.

@weshaggard PTAL

CC @ericstj @eerhardt


eerhardt:
Can we get this in to unblock the builds?  CLI is looking to get updated with `core-setup/release/2.0.0`

eerhardt:
Merging to unblock the builds.

Issue #2176, opened 2017-04-25
AppDomain.CurrentDomain.BaseDirectory does not end with a slash on Desktop


ramarag:
@ViktorHofer commented on [Wed Apr 12 2017](https://github.com/dotnet/coreclr/issues/10920)

`AppDomain.CurrentDomain.BaseDirectory` does not contain a slash at the end. Failing test in corefx with targetgroup **netfx** here: https://github.com/dotnet/corefx/issues/18196

netfx:
`C:\Users\vikto\Documents\src\corefx\bin\AnyOS.AnyCPU.Debug\System.Configuration.ConfigurationManager.Tests\netstandard`

core:
`C:\Users\vikto\Documents\src\corefx\bin\AnyOS.AnyCPU.Debug\System.Configuration.ConfigurationManager.Tests\netstandard\`

The BaseDirectory property retrieves the path from the GetModuleFileName member from System.Runtime.InteropServices.RuntimeEnvironment in coreclr which itself is declared as extern:
```
[MethodImplAttribute(MethodImplOptions.InternalCall)]
internal static extern String GetModuleFileName();
```

Test executable from corefx project root (build -framework:netfx required before!):
`msbuild src/System.Configuration.ConfigurationManager/tests /t:BuildAndTest /p:targetgroup=netfx /p:xunitoptions="-method System.ConfigurationTests.UrlPathTests.GetDirectoryOrRootName_GettingDirectoryFromAFilePath"`

---

@danmosemsft commented on [Wed Apr 12 2017](https://github.com/dotnet/coreclr/issues/10920#issuecomment-293640981)

@gkhanna79 where would the fix be, on the managed side or native (fusion side)?

---

@gkhanna79 commented on [Wed Apr 12 2017](https://github.com/dotnet/coreclr/issues/10920#issuecomment-293647808)

There is no fusion in CoreCLR. Depending upon how the API is plumbed, it could be in either managed or native.

---

@AlexGhiondea commented on [Wed Apr 12 2017](https://github.com/dotnet/coreclr/issues/10920#issuecomment-293693590)

@gkhanna79 I thought that the host sets this value -- is that true?

---

@gkhanna79 commented on [Wed Apr 12 2017](https://github.com/dotnet/coreclr/issues/10920#issuecomment-293695036)

Feel free to move to Core-Setup and assign to @ramarag if you think the bug is there and not in AD setup code.

---

[...]

PR #2178, opened 2017-04-26
remove trailing DIR_SEPARATOR from APP_CONTEXT_BASE_DIRECTORY

ramarag:
fixes https://github.com/dotnet/core-setup/issues/2176

@gkhanna79 @eerhardt @danmosemsft PTAL

### DIFF
--- a/src/corehost/cli/hostpolicy.cpp
+++ b/src/corehost/cli/hostpolicy.cpp
@@ -80,7 +80,11 @@ int run(const arguments_t& args)
     // Note: these variables' lifetime should be longer than coreclr_initialize.
     std::vector<char> tpa_paths_cstr, app_base_cstr, native_dirs_cstr, resources_dirs_cstr, fx_deps, deps, clrjit_path_cstr, probe_directories;
     pal::pal_clrstring(probe_paths.tpa, &tpa_paths_cstr);
-    pal::pal_clrstring(args.app_dir, &app_base_cstr);
+
+    //removing trailing directory seperator for app compat reasons
+    pal::string_t trimmed_app_dir = trim_end(args.app_dir, DIR_SEPARATOR);
+    pal::pal_clrstring(trimmed_app_dir, &app_base_cstr);
+
     pal::pal_clrstring(probe_paths.native, &native_dirs_cstr);
     pal::pal_clrstring(probe_paths.resources, &resources_dirs_cstr);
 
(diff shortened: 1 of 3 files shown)


src/corehost/cli/hostpolicy.cpp
```diff
@@ -81,6 +81,13 @@ int run(const arguments_t& args)
     std::vector<char> tpa_paths_cstr, app_base_cstr, native_dirs_cstr, resources_dirs_cstr, fx_deps, deps, clrjit_path_cstr, probe_directories;
     pal::pal_clrstring(probe_paths.tpa, &tpa_paths_cstr);
     pal::pal_clrstring(args.app_dir, &app_base_cstr);
+
+    //removing trailing directory seperator for app compat reasons
+    while (app_base_cstr.end()[-2] == DIR_SEPARATOR)
```
eerhardt: Is it possible for the app base dir to be at the root of the file system?  I.e just a '/'?  Will this work in that case?

ramarag: Let us assume so, added a check

danmosemsft:
@viktorhofer

src/corehost/cli/hostpolicy.cpp
```diff
@@ -81,6 +81,13 @@ int run(const arguments_t& args)
     std::vector<char> tpa_paths_cstr, app_base_cstr, native_dirs_cstr, resources_dirs_cstr, fx_deps, deps, clrjit_path_cstr, probe_directories;
     pal::pal_clrstring(probe_paths.tpa, &tpa_paths_cstr);
     pal::pal_clrstring(args.app_dir, &app_base_cstr);
+
+    //removing trailing directory seperator for app compat reasons
+    while (app_base_cstr.size() > 2 && app_base_cstr.end()[-2] == DIR_SEPARATOR )
```
gkhanna79: A simpler way to do this would be to use std::string's back method for comparison and pop_back to remove the character.

Also, why are we doing this here and not in hostfxr that passes in this data so that we are working with a consistent value throughout the various components?

ramarag: you can't do a pop_back, as the last character in the vector is null,
we are just doing for appcompat

schellap: CoreCLR api uses UTF8 now, so your change should be fine, but it is better to manipulate a copy of the pal string than the buffer. There may be code in util already that does this. You can zero out if you are editing the vector (which assumes encoding). Better to put this in util for better flow.

ramarag: done

ramarag:

ramarag:
i am merging this, if there are concerns will follow it up later

PR #2185, opened 2017-04-26
Exclude debian and tizen from platform manifest

ericstj:
Debian and Tizen are not built by our build process, but we want to
include entries for them in our runtime.json.

This was interfering with another feature where we must include a list
of all files that are in all runtime packages in a manifest.  The
manifest is used when building a shared-framework app to understand
which files are part of the shared framework and should be considered
for conflict resolution.  If the application were to reference an old
package which had one of these files we want to trim it from the
application if it failed conflict resolution (was a lower version).

We cannot gather this information from debian and tizen packages since
we don't build them.  For now we will opt them out of this feature.
This is likely "good enough" for them since they won't have any unique
files that previously shipped in RID-less packages and require conflict
resolution.

A alternative solution for debian and tizen would be for them to ship
their own RID-less package with their own platform manifest.  They could
do so transparently by implicitly referencing their package for folks
that wanted this.

/cc @weshaggard @gkhanna79 @eerhardt

### DIFF
--- a/pkg/projects/netcoreappRIDs.props
+++ b/pkg/projects/netcoreappRIDs.props
@@ -16,9 +16,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <OfficialBuildRID Include="debian.8-armel">
-      <Platform>armel</Platform>
-    </OfficialBuildRID>
     <OfficialBuildRID Include="linux-x64" />
     <OfficialBuildRID Include="osx-x64" />
     <OfficialBuildRID Include="win-x86">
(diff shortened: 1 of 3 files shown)


pkg/projects/netcoreappRIDs.props
```diff
@@ -34,7 +31,16 @@
     <OfficialBuildRID Include="linux-arm">
       <Platform>arm</Platform>
     </OfficialBuildRID>
+
+    <!-- The following RIDs are not officically supported and are not 
+         built during official builds, however we wish to include them 
+         in our runtime.json to enable others to provide them.  -->
+    <OfficialBuildRID Include="debian.8-armel">
```
ericstj: Do you think it would be clearer if I renamed these items to something like "UnofficialBuildRID" or "UnsupportedBuildRID"?

ericstj: Renamed it to "UnofficialBuildRID", that feels cleaner to me.

gkhanna79: > Renamed it to "UnofficialBuildRID", that feels cleaner to me.

I agree.

PR #2186, opened 2017-04-26
[release/2.0.0] Exclude debian and tizen from platform manifest

ericstj:
Debian and Tizen are not built by our build process, but we want to
include entries for them in our runtime.json.

This was interfering with another feature where we must include a list
of all files that are in all runtime packages in a manifest.  The
manifest is used when building a shared-framework app to understand
which files are part of the shared framework and should be considered
for conflict resolution.  If the application were to reference an old
package which had one of these files we want to trim it from the
application if it failed conflict resolution (was a lower version).

We cannot gather this information from debian and tizen packages since
we don't build them.  For now we will opt them out of this feature.
This is likely "good enough" for them since they won't have any unique
files that previously shipped in RID-less packages and require conflict
resolution.

A alternative solution for debian and tizen would be for them to ship
their own RID-less package with their own platform manifest.  They could
do so transparently by implicitly referencing their package for folks
that wanted this.

/cc @weshaggard @gkhanna79 @eerhardt

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -65,7 +65,9 @@
     <MSBuild Projects="@(ProjectReference)"
              Condition="'%(ProjectReference.PackageTargetRuntime)' == '$(PackageRID)' OR 
                         '%(ProjectReference.BuidOnRID)' == '$(PackageRID)' OR 
-                        ('$(IncludeAllRuntimePackagesInPlatformManifest)' == 'true' AND '%(ProjectReference.PackageTargetRuntime)' != '')"
+                        ('$(IncludeAllRuntimePackagesInPlatformManifest)' == 'true' AND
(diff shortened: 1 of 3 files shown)


ericstj:
This is a revert of the workaround put into release/2.0.0, followed by cherry-pick of the change that went into master.

gkhanna79:
We should email shiproom once official master build confirms this works.

ericstj:
Looks like official build of core-setup can't resolve dotnet.myget.org.

gkhanna79:
@ericstj Do you want to email shiproom about this, as the DNS outage is orthogonal, to get this in?

ericstj:
Looks like a new build is running.  If that one succeeds we can send mail. https://mc.dot.net/#/product/netcore/master/source/official~2Fcore-setup~2Fmaster~2F/type/build~2Fproduct~2F/build/20170426.05/workItem/Orchestration/analysis/external/Visual%20Studio%20Build%20Information

gkhanna79:
Looks like that build was good. Do you want to send the email now @ericstj ?

PR #2209, opened 2017-04-27
Use manually-cloned repo for CrossBuild pkg dir

dagood:
This avoids interference from changes in the master branch.

@Petermarcu @gkhanna79

### DIFF
--- a/buildpipeline/Core-Setup-CrossBuild.json
+++ b/buildpipeline/Core-Setup-CrossBuild.json
@@ -66,7 +66,7 @@
         "definitionType": "task"
       },
       "inputs": {
-        "SourceFolder": "pkg",
+        "SourceFolder": "$(PB_RepoName)/pkg",
         "Contents": "**",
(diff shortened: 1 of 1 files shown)


buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -66,7 +66,7 @@
         "definitionType": "task"
       },
       "inputs": {
-        "SourceFolder": "pkg",
+        "SourceFolder": "$(PB_RepoName)/pkg",
```
dagood: I haven't tested this. I'm pretty confident it will work because it matches what we use for build.sh: https://github.com/dotnet/core-setup/blob/e01befc220e46313417e8ff226ab371930f94370/buildpipeline/Core-Setup-CrossBuild.json#L127

    "scriptPath": "$(PB_RepoName)/scripts/docker-as-current-user.sh",

PR #2211, opened 2017-04-27
Fix bad merge

chcosta:
Rebasing to latest caused a bad merge in publishing.  publishing is only exercised during official builds, so it wasn't caught.


chcosta:
change doesn't affect CI, merging

PR #2227, opened 2017-04-28
Convert dotnet-deb-tool to use csproj and netcoreapp2.0.

eerhardt:
Since this is the first .NET Core SDK .csproj being used, we need to install a new version of the CLI that supports MSBuild.  These two CLI versions will live side-by-side until the whole repo is off project.json.

I also fixed a few dotnet-deb-tool issues.

I also changed our CLI download location to azureedge.  This supports faster downloads when you don't live in Redmond.


DotnetCLISdkVersion.txt
```diff
@@ -0,0 +1 @@
+2.0.0-preview1-005899
```
chcosta: Did you create this file on Linux, or validate that line endings are right?

chcosta:
judging by successful CI, line endings are fine. =)

DotnetCLISdkVersion.txt
eerhardt: There actually isn't a line ending on this file. There isn't in the other one either.

Issue #2225, opened 2017-04-28
Compressed files not including -portable moniker


chcosta:
The dotnet-host-fxr-linux-x64-***.tar.gz, and other compressed files (zips / tarballs), used to include a "-portable" moniker, that moniker is missing form the latest zips.

Issue #2228, opened 2017-04-28
Symbol zips have wrong paths


chcosta:
We used to tar files using the "-C" option, this causes a problem and you get a "file content has changed" error when using this option.  I moved to just compressing an explicit filelist so that we wouldn't hit that error, but it looks like doing so may include additional root directory paths in the filenames which are compressed.  May have to bring back the '-C' option and just ignore any errors from the command. :/

PR #2229, opened 2017-04-29
Fix symbols tarballs, add portable moniker to tar / zips

chcosta:
Fixes https://github.com/dotnet/core-setup/issues/2225 and https://github.com/dotnet/core-setup/issues/2228

### DIFF
--- a/src/pkg/packaging/dir.props
+++ b/src/pkg/packaging/dir.props
@@ -14,22 +14,21 @@
 
   <PropertyGroup>
     <ProductMoniker>$(PackageTargetRid)-$(HostFullVersion)</ProductMoniker>
-    <PortableMoniker Condition="'$(PortableBuild)' == 'true'">-portable</PortableMoniker>
   </PropertyGroup>
 
   <PropertyGroup>
-    <CombinedCompressedFile>dotnet-$(ProductMoniker)$(PortableMoniker)$(CompressedFileExtension)</CombinedCompressedFile>
-    <HostFxrCompressedFile>dotnet-hostfxr-$(ProductMoniker)$(PortableMoniker)$(CompressedFileExtension)</HostFxrCompressedFile>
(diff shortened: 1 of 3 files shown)


src/pkg/packaging/dir.props
```diff
@@ -9,26 +9,27 @@
     <SharedHostPublishRoot>$(IntermediateOutputRootPath)sharedHost/</SharedHostPublishRoot>
     <HostFxrPublishRoot>$(IntermediateOutputRootPath)hostFxr/</HostFxrPublishRoot>
     <SharedFrameworkPublishRoot>$(IntermediateOutputRootPath)sharedFx/</SharedFrameworkPublishRoot>
-    <CombinedPublishRoot>$(IntermediateOutputRootPath)combined-framework-host/</CombinedPublishRoot>   
+    <CombinedPublishRoot>$(IntermediateOutputRootPath)combined-framework-host/</CombinedPublishRoot>
   </PropertyGroup>
 
   <PropertyGroup>
     <ProductMoniker>$(PackageTargetRid)-$(HostFullVersion)</ProductMoniker>
+    <PortableMoniker Condition="'$(PortableBuild)' == 'true'">-portable</PortableMoniker>
```
eerhardt: All of our builds are now "portable". I don't think we need this in our built assets anymore.

@gkhanna79 - thoughts?

chcosta: Today, we're still building Ubuntu and Debian as non portable. I presume just as validation.

eerhardt: Sorry, I should correct myself. All the builds that used to have conflicting assets are now only portable. So we don't need to differentiate the assets anymore.

You are right, we are building Ubuntu and Debian because they produce the Deb installers for those platforms.  But that will soon change. Rakesh is working on creating the Deb installers after the portable Linux build.

chcosta:
@dotnet-bot test Ubuntu16.04 x64 Release Build

Issue #2210, opened 2017-04-27
Re-implement 'publish identity package from a single build leg'


chcosta:
The core-setup on buildtools change completely changed publishing, and the recent work to ensure we publish the identity package from a single build leg work was lost.  Re-implement this https://github.com/dotnet/core-setup/commit/2bd171a3cf633244ba08b5944c2ace4191aea684 change

PR #2234, opened 2017-04-29
Reimplement only publish identity packages from a single build leg

chcosta:
Fixes https://github.com/dotnet/core-setup/issues/2210

Reimplement logic to only publish rid agnostic packages from a single build leg (windows x64)


buildpipeline/pipeline.json
```diff
@@ -110,7 +110,9 @@
           "Parameters": {
             "PB_DistroRid": "win7-x64",
             "PB_TargetArchitecture": "x64",
-            "PB_PortableBuild": "true"
+            "PB_PortableBuild": "true",
+            "PB_PublishRidAgnosticPackages": "true",
```
eerhardt: NOTE: This question is mostly for my understanding at our high-level direction.

If we just named these parameters the same as the MSBuild properties, then we wouldn't need to pass them into `/p` on the command-line.  They would just automatically be set for the entire build.

Why prefix these variables with `PB_` only to trim the `PB_` off and pass the value through?  This seems like an extra step that can easily cause type-o errors and/or missing setting it when the build is split across multiple steps.

chcosta: Yes, if these don't get prefixed with `PB_`, then they become environment variables. History (CoreFx) has shown that relying on environment variables can be problematic, lead to difficult to diagnose issues, and make repro's difficult.  Our preference is to eliminate environment variables entirely, in favor of explicit declarations.  This isn't possible in VSTS, so we add a `PB_` prefix which prevents us from having variables named after something actually implemented in the code.

Issue #2195, opened 2017-04-26
Extract / sign engine bundle / re-bundle


chcosta:
Our Windows official builds used to extract the engine bundle, sign it, and then re-bundle it.  This capability has been lost and needs to be added back for official builds.

PR #2235, opened 2017-04-30
Extract engine bundle, sign, reattach; also do vsts agent cleanup

chcosta:
Fixes https://github.com/dotnet/core-setup/issues/2195

This is a minimal change, it just creates targets we directly call in the build which wrap calling `insignia.exe` to extract the engine from the bundle, then we sign it, then another wrapper call to reattach the signed engine to the bundle.

The diff for the build definition looks larger because I'm deleting some steps that were disabled.


buildpipeline/Core-Setup-Signing-Windows-BT.json
```diff
@@ -535,41 +521,49 @@
...
         "versionSpec": "1.*",
         "definitionType": "task"
       },
       "inputs": {
-        "CopyRoot": "",
-        "Contents": "**\\*.log",
-        "ArtifactName": "Build Logs",
-        "ArtifactType": "Container",
-        "TargetPath": "\\\\my\\share\\$(Build.DefinitionName)\\$(Build.BuildNumber)"
+        "solution": "$(PB_SourcesDirectory)\\Tools\\scripts\\vstsagent\\cleanupagent.proj",
```
eerhardt: Why is this necessary?

buildpipeline/Core-Setup-Signing-Windows-BT.json
```diff
@@ -411,85 +373,109 @@
...
         "versionSpec": "1.*",
         "definitionType": "task"
       },
       "inputs": {
-        "solution": "build_projects\\signing\\sign.proj",
-        "platform": "",
+        "solution": "$(PB_SourcesDirectory)\\src\\pkg\\packaging\\dir.proj",
+        "platform": "$(PB_TargetArchitecture)",
         "configuration": "$(BuildConfiguration)",
-        "msbuildArguments": "/t:SignBundle $(MsbuildSigningArguments)",
```
eerhardt: If we don't sign the bundle here, where does it get signed?

eerhardt: In the old world the steps went:

Compile
Sign Assemblies
Test and Package (would create .msi)
Sign MSI and cab
Generate Bundle
Extract Engine from bundle
Sign Engine
Reattach Engine to Bundle
Sign Bundle
Publish


Do your new steps happen in the same order?

---
In reply to: [114195530](https://github.com/dotnet/core-setup/pull/2235#discussion_r114195530) [](ancestors = 114195530)

buildpipeline/Core-Setup-Signing-Windows-BT.json
chcosta: This is not specifically related to the engine extract / sign / reattach stuff.  But, we've found that VSTS tends to leave builds around and after ~6 weeks, machines will hit out of disk space issues.  So we have this tooling around cleaning up build agents.

eerhardt: Just for windows?

buildpipeline/Core-Setup-Signing-Windows-BT.json
chcosta: This diff is a bit confusing.  The build definition was cloned directly from the old definition, and then steps were disabled and replaced with the new steps in the new system.  This delete is just cleaning up the disabled steps from the definition `"enabled": "false"`.  The steps should be similar to the previous system:

Compile
Sign HostFxr
Build Nuget packages
Build managed / publish shared framework
Sign crossgened assemblies
sign assemblies *.dll
package
build and run tests
sign msi and cab
generate bundle
extract engine
sign engine
reattach engine
sign bundle
publish

eerhardt: That's not the order I see when looking at the file:  https://github.com/chcosta/core-setup/blob/6c6978960e732857364f7475aea8759a8f8e2432/buildpipeline/Core-Setup-Signing-Windows-BT.json

I see:

Build nuget packages
Build managed / publish sharedframework
Sign Shared Framework Crossgened Assemblies
Sign Assemblies: *.dll
Package
Build and run tests
Sign MSI and cab
**Extract engine from bundle**
**Sign Bundle**
**Reattach engine to bundle**
**Sign engine**
Publish


Notice:  

- No generate bundle
- And signing the bundle before reattaching the engine to it

chcosta: When I load that file into VSTS, here are the build steps, in order...

image

Generate bundle isn't an explicit step anymore because I noticed it was already happening as part of building installers - https://github.com/dotnet/core-setup/blob/master/src/pkg/packaging/dir.proj#L147

chcosta: If the order is wrong, I'm completely open to fixing it, I just want to make sure we're looking at the same thing.

eerhardt: I'm just looking at the file as it exists in the PR.  I don't know how the ordering happens...  I assumed the order of the file was the order they ran.

chcosta: The order in the file appears to be the same as what I am seeing in the image above.

eerhardt: Your image above has the wrong ordering as well.

image

You can't sign the bundle before you reattach the engine to it.  That makes the signing invalid.

Please see my comment above for the correct order of things.

chcosta: ah, now I see what you're referring to.  Yes, that makes perfect sense.

buildpipeline/Core-Setup-Signing-Windows-BT.json
chcosta: For all platforms.  I had previously already added this step to Linux and osx.

src/pkg/packaging/dir.props
```diff
@@ -25,6 +25,7 @@
 
   <PropertyGroup>
     <CombinedInstallerFile>dotnet-$(ProductMoniker)$(CombinedInstallerExtension)</CombinedInstallerFile>
+    <CombinedInstallerEngine>dotnet-$(ProductMoniker)-engine.exe</CombinedInstallerEngine>
```
eerhardt: I was looking before, and I don't see the below property `SharedFrameworkEngine` being used.  Can it be removed?

buildpipeline/Core-Setup-Signing-Windows-BT.json
```diff
@@ -489,7 +475,7 @@
         "logProjectEvents": "false",
         "createLogFile": "false",
         "msbuildLocationMethod": "version",
-        "msbuildVersion": "15.0",
+        "msbuildVersion": "latest",
```
eerhardt: How come this one is `latest` but others are `15.0`?

buildpipeline/Core-Setup-Signing-Windows-BT.json
```diff
@@ -411,64 +373,88 @@
...
+        "configuration": "$(BuildConfiguration)",
+        "msbuildArguments": "/t:ReattachEngineToBundle $(PB_CommonMSBuildArgs) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\reattachengine.log",
+        "clean": "false",
+        "maximumCpuCount": "false",
+        "restoreNugetPackages": "false",
+        "logProjectEvents": "false",
+        "createLogFile": "false",
+        "msbuildLocationMethod": "version",
+        "msbuildVersion": "15.0",
+        "msbuildArchitecture": "x86",
```
eerhardt: Why are some `x64` and others `x86`?

buildpipeline/Core-Setup-Signing-Windows-BT.json
chcosta: I've been moving them all to latest in this and other pr's.  specifying 15.0 causes a warning and then it just defaults to "latest".  This caused noise that bothered a bunch of people that look at results on mission control.

buildpipeline/Core-Setup-Signing-Windows-BT.json
chcosta: ditto as above, I've been moving them all to x64.  I'll load the build definition again and double check the values for all of the steps.

PR #2236, opened 2017-04-30
Change publishrid list for current active build legs

chcosta:
Changes publishrid list to reflect what we're currently actively building in core-setup official builds.  With this change, Finalization will actually occur during core-setup official builds.  This change disables checking for arm legs and removes non-portable rids we're no longer producing.

### DIFF
--- a/publish/dir.props
+++ b/publish/dir.props
@@ -17,28 +17,23 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PublishRid Include="win7-x64"/>
-    <PublishRid Include="win7-x86"/>
-    <PublishRid Include="win8-arm"/>
-    <PublishRid Include="win10-arm64"/>
-    <PublishRid Include="osx.10.12-x64"/>
     <PublishRid Include="ubuntu.14.04-x64"/>
     <PublishRid Include="ubuntu.16.04-x64"/>
-    <PublishRid Include="ubuntu.14.04-arm"/>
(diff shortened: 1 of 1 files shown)


publish/dir.props
```diff
@@ -17,28 +17,23 @@
...
-    <PublishRid Include="fedora.23-x64"/>
-    <PublishRid Include="fedora.24-x64"/>
-    <PublishRid Include="opensuse.42.1-x64"/>
-    <PublishRid Include="linux-x64"/>
-    <PublishRid Include="linux-arm"/>
-    <PublishRid Include="linux-arm64"/>
-    <PublishRid Include="win-x86"/>
-    <PublishRid Include="win-x64"/>
+    <PublishRid Include="linux-x64_portable"/>
+    <PublishRid Include="win-x86_portable"/>
```
gkhanna79: Why do we have "_portable" in the RID name?

chcosta: This publishrid itemgroup only gets used for checking the badge files. It's not actually a rid list like the others we use. https://github.com/dotnet/core-setup/blob/master/publish/dir.targets#L61 ID line to merge this Metadata in with the other rid lists, but that can happen later. If we eliminate the portable moniker, we can get rid of "_portable"

eerhardt: Let's drop the "_portable".  The RID already tells us it is "portable" - "win" is portable; "osx" is portable; "linux" is portable.

I also think we should drop "portable" from the moniker as well.

PR #2239, opened 2017-05-01
Remove NugetRestoreCommand, since it is not used in core-setup.

eerhardt:
@chcosta @weshaggard

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -53,11 +53,6 @@
     <Message Importance="High" Text="Restoring all packages..." />
     
     <Exec Command="$(DnuRestoreCommand) %(DnuRestoreDirs.AdditionalArgs) %(DnuRestoreDirs.Identity)" StandardOutputImportance="Low" CustomErrorRegularExpression="^Unable to locate .*" />
-
-    <ItemGroup>
-      <_allPackagesConfigs Include="$(MSBuildProjectDirectory)\src\**\packages.config"/>
-    </ItemGroup>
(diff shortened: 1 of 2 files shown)


weshaggard:
cc @ericstj as I believe he uses it for depproj's. However this is something that I recently moved into BuildTools with the 2.0 changes.

ericstj:
No I don't use that one.

eerhardt:
> However this is something that I recently moved into BuildTools with the 2.0 changes.

Are you saying BuildTools now depends on `mono` as well?  I think we need to not depend on 'mono' in any of our repos.

weshaggard:
No BuildTools doesn't depend on mono. I mistoke this for the DnuRestoreCommands, which I moved into BuildTools https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props#L59. Once we update this repo to use BuildTools 2.0 then those can also be deleted.

eerhardt:
Looks like it does:

https://github.com/dotnet/buildtools/blob/c3eb52fd221fe1dab172f5324f29712dbbacf8dc/dir.props#L81

```
    <NugetRestoreCommand Condition="'$(OsEnvironment)'=='Unix'">mono $(NuGetRestoreCommand)</NugetRestoreCommand>
```

weshaggard:
That is for the building of BuildTools as opposed to the consumption of BuildTools. I don't think we actually use that restore command for building of BuildTools any longer either so we can probably delete it there as well.

PR #2241, opened 2017-05-01
Allow host to access long files

ramarag:
Fixes long path issue in the host
The core logic has been ported from https://github.com/dotnet/coreclr/blob/54891e0650e69f08832f75a40dc102efc6115d38/src/utilcode/longfilepathwrappers.cpp

fixes https://github.com/dotnet/core-setup/issues/1489

### DIFF
--- a/src/corehost/cli/dll/CMakeLists.txt
+++ b/src/corehost/cli/dll/CMakeLists.txt
@@ -45,7 +45,9 @@ set(SOURCES
 
 
 if(WIN32)
-    list(APPEND SOURCES ../../common/pal.windows.cpp)
+    list(APPEND SOURCES 
+        ../../common/pal.windows.cpp
+        ../../common/longfile.windows.cpp)
 else()
     list(APPEND SOURCES ../../common/pal.unix.cpp)
 endif()
--- a/src/corehost/common/longfile.windows.cpp
+++ b/src/corehost/common/longfile.windows.cpp
@@ -1,0 +1,126 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+//The logic in this file was ported from https://github.com/dotnet/coreclr/blob/54891e0650e69f08832f75a40dc102efc6115d38/src/utilcode/longfilepathwrappers.cpp
+//Please reflect any change here into the above file too!
+#include "pal.h"
+#include "trace.h"
+#include "utils.h"
+#include "longfile.h"
+
+const pal::char_t LongFile::DirectorySeparatorChar = _X('\\');
+const pal::char_t LongFile::AltDirectorySeparatorChar = _X('/');
+const pal::char_t LongFile::VolumeSeparatorChar = _X(':');
+const pal::string_t LongFile::ExtendedPrefix = _X("\\\\?\\");
+const pal::string_t LongFile::DevicePathPrefix = _X("\\\\.\\");
+const pal::string_t LongFile::UNCExtendedPathPrefix = _X("\\\\?\\UNC\\");
+const pal::string_t LongFile::UNCPathPrefix = _X("\\\\");
+
+bool ShouldNormalizeWorker(const pal::string_t& path)
+{
+    if (path.empty() || LongFile::IsDevice(path) || LongFile::IsExtended(path) || LongFile::IsUNCExtended(path))
+        return false;
+
+    if (!LongFile::IsPathNotFullyQualified(path) && path.size() < MAX_PATH)
+        return false;
+
+    return true;
+}
+
+//For longpath names on windows, if the paths are normalized they are always prefixed with
+//extended syntax, Windows does not do any more normalizations on this string and uses it as is
+//So we should ensure that there are NO adjacent DirectorySeparatorChar 
+bool AssertRepeatingDirSeparator(const pal::string_t& path)
+{
+    if (path.empty())
+        return true;
+
+    pal::string_t path_to_check = path;
+    if (LongFile::IsDevice(path))
+    {
+        path_to_check.erase(0, LongFile::DevicePathPrefix.length());
+    }
+    else if (LongFile::IsExtended(path))
+    {
+        path_to_check.erase(0, LongFile::ExtendedPrefix.length());
+    }
+    else if (LongFile::IsUNCExtended(path))
+    {
+        path_to_check.erase(0, LongFile::UNCExtendedPathPrefix.length());
+    }
+    else if (path_to_check.compare(0, LongFile::UNCPathPrefix.length(), LongFile::UNCPathPrefix) == 0)
+    {
+        path_to_check.erase(0, LongFile::UNCPathPrefix.length());
+    }
+
+    pal::string_t dirSeparator;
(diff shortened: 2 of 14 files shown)


ramarag:
@gkhanna79 @janvorli @JeremyKuhne PTAL

src/corehost/cli/dll/CMakeLists.txt
```diff
@@ -45,7 +45,9 @@ set(SOURCES
 
 
 if(WIN32)
-    list(APPEND SOURCES ../../common/pal.windows.cpp)
+    list(APPEND SOURCES
```
gkhanna79: Why are we only including this for Windows?

src/corehost/common/pal.unix.cpp
```diff
@@ -442,12 +442,33 @@ bool pal::get_own_executable_path(pal::string_t* recv)
...
+
+        if (buffer == NULL)
+        {
+            return false;
+        }
+
+        error_code = sysctl(mib, 4, buffer, &len, NULL, 0);
+        if (error_code == 0)
+        {
+            recv->assign(buf);
```
gkhanna79: Shouldn't the buffer be freed once it has been assigned?

gkhanna79:
How did you validate the change and can we please add tests for this?

src/corehost/cli/dll/CMakeLists.txt
gkhanna79: Never mind - got it.

gkhanna79:
@Petermarcu I would think this would be a DCR. Do you agree?

src/corehost/common/pal.windows.cpp
```diff
@@ -347,14 +385,68 @@ bool pal::clr_palstring(const char* cstr, pal::string_t* out)
...
+        (prefixLen != prefix->length() - LongFile::UNCPathPrefix.length()))
+    {
+        //Remove the leading '\\' from the UNC path to be replaced with UNCExtendedPathPrefix
+        *path = LongFile::UNCExtendedPathPrefix;
+        path->append(str.data() + prefixLen + LongFile::UNCPathPrefix.length());
+        
+        return true;
+    }
+    
+    //wcscpy_s always termintes with NULL, so we are saving the character that will be overwriiten
```
janvorli: A nit- spelling: overwriiten -> overwritten.

src/corehost/common/pal.windows.cpp
```diff
@@ -347,14 +385,68 @@ bool pal::clr_palstring(const char* cstr, pal::string_t* out)
...
     }
-    path->assign(buf);
+
+    if (size < MAX_PATH)
+    {
+        path->assign(buf);
+        return true;
+    }
+
+    const string_t* prefix = &LongFile::ExtendedPrefix;
```
janvorli: This function looks overcomplicated and hard to read. I think that it would be better to use string_t instead of vector, just resize str to the size returned by the first call to `GetFullPathNameW`, then call the `GetFullPathNameW` with `str.data()` and finally prepend the prefix if needed using string_t::insert. Without all the prefix len computations.

src/corehost/common/longfile.windows.cpp
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+//The logic in this file was ported from https://github.com/dotnet/coreclr/blob/54891e0650e69f08832f75a40dc102efc6115d38/src/utilcode/longfilepathwrappers.cpp
```
ramarag: I wanted to capture the point in time when we actually ported the logic.
Also this ensures that if the file is renamed or moved, we can always find the original source

src/corehost/common/pal.windows.cpp
ramarag: doing that will diverge from the original coreclr code, i agree it would make it more readable

src/corehost/common/longfile.windows.cpp
janvorli: Ah, it makes sense then.

src/corehost/common/pal.unix.cpp
```diff
@@ -442,12 +442,34 @@ bool pal::get_own_executable_path(pal::string_t* recv)
...
+        if (buffer == NULL)
+        {
+            return false;
+        }
+
+        error_code = sysctl(mib, 4, buffer, &len, NULL, 0);
+        if (error_code == 0)
+        {
+            recv->assign(buffer);
+	    delete(buffer);
```
ramarag: done

src/corehost/common/pal.windows.cpp
```diff
@@ -4,12 +4,42 @@
...
+    pal::char_t program_path[MAX_PATH];
+    DWORD dwModuleFileName = ::GetModuleFileNameW(hModule, program_path, MAX_PATH);
+
+    if (dwModuleFileName != 0 && dwModuleFileName < MAX_PATH)
+    {
+        recv->assign(program_path);
+        return true;
+    }
+
+    std::vector<pal::char_t> str;
```
ramarag: done

ramarag:
@gkhanna79  for testing 

1. I reduced MAX_PATH to a single digit number and ran through all the tests
2. Ran the Apps from a long file path on windows

Adding tests is not going to be straight forward as RS2 which has a lot of the shell operations are not yet broadly deployed

@janvorli can i get a sign off ?

src/corehost/common/pal.windows.cpp
```diff
@@ -62,12 +84,27 @@ bool pal::getcwd(pal::string_t* recv)
     return false;
 }
 
-bool pal::load_library(const char_t* path, dll_t* dll)
+bool pal::load_library(const string_t* in_path, dll_t* dll)
```
janvorli: A nit - could you please revert the parameter type change now that it is not needed anymore?

src/corehost/common/pal.windows.cpp
```diff
@@ -347,14 +378,49 @@ bool pal::clr_palstring(const char* cstr, pal::string_t* out)
...
+    if (size == 0)
     {
         trace::error(_X("Error resolving full path [%s]"), path->c_str());
         return false;
     }
-    path->assign(buf);
+
+    const string_t* prefix = &LongFile::ExtendedPrefix;
+    //Check if the resolved path is a UNC. By default we assume relative path to resolve to disk 
+    if (str.find(LongFile::UNCPathPrefix, 0) != string_t::npos)
```
janvorli: This condition should be `== 0`

gkhanna79:
@Petermarcu This change enables host to support long file paths. Would you consider this as a feature for DCR?

gkhanna79:
@ramarag Please do not merge this until @Petermarcu has chimed in.

src/corehost/common/pal.windows.cpp
ramarag: i would like to keep the string_t type as it is more convenient to call methods on it

ramarag:
@gkhanna79  sure will wait on @Petermarcu

Petermarcu:
Its on the border but I think its fine to take before 5/10.

ramarag:

PR #2163, opened 2017-04-25
[ARM] Update rootfs for Tizen armel

hqueue:
*This PR should be merged only after all related PR in coreclr and corefx merged.*

This is required to build CoreCLR for Tizen armel as a part of .NET Core 2.0.0 for Tizen.
Existing rootfs for armel Tizen is out-of-dated and not suitable for .NET Core 2.0.0 for Tizen.

Two changes are made.
* Update rootfs for Tizen armel to recent version
* Updates CI for Tizen armel to make use of a prebuilt embedded rootfs as Ubuntu arm does.

Related PR : 
https://github.com/dotnet/coreclr/pull/11202
https://github.com/dotnet/corefx/pull/18916

PR #2243, opened 2017-05-01
Fix UpdateDependencies using VersionTools

karajas:
This is to bring core-setup update dependencies back online using the buildtools version tools.

/cc @chcosta @mellinoe @eerhardt @weshaggard

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -60,12 +60,18 @@
 
   <Target Name="ValidateExactRestore"
           Condition="'$(AllowInexactRestore)'!='true'">
-    <ValidateExactRestore ProjectLockJsons="@(ProjectJsonFiles->'%(RootDir)%(Directory)%(Filename).lock.json')" />
+    <ItemGroup>
+      <ProjectJsonTemplateFiles Include="$(MSBuildThisFileDirectory)**\project.json.template" />
+      <ProjectJsonsExcludingTemplateFiles Include="@(ProjectJsonFiles)" Exclude="@(ProjectJsonTemplateFiles)" />
+    </ItemGroup>
+    <ValidateExactRestore ProjectLockJsons="@(ProjectJsonsExcludingTemplateFiles->'%(RootDir)%(Directory)%(Filename).lock.json')" />
   </Target>
 
   <Import Project="dir.targets" />
   <Import Project="dir.traversal.targets" />
   <Import Project="$(ToolsDir)clean.targets" />
+  <Import Project="$(ToolsDir)VersionTools.targets" Condition="Exists('$(ToolsDir)VersionTools.targets')" />
+
 
   <!-- Override RestorePackages from dir.traversal.targets and do a batch restore -->
   <Target Name="RestorePackages" DependsOnTargets="BatchRestorePackages" />
--- a/dependencies.props
+++ b/dependencies.props
@@ -1,14 +1,59 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Source of truth for dependency tooling: the commit hash of the dotnet/versions master branch as of the last auto-upgrade. -->
+
+  <!-- For detailed explanation of the reference versions see:
+
+        https://github.com/dotnet/buildtools/blob/master/Documentation/project-nuget-dependencies.md
+        https://github.com/dotnet/buildtools/blob/master/Documentation/annotated-dependency-props.md
+
+       These ref versions are pulled from https://github.com/dotnet/versions.
+  -->
   <PropertyGroup>
-       <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
-       <JitVersion>2.0.0-preview2-25226-02</JitVersion>
+    <CoreFxCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</CoreFxCurrentRef>
+    <CoreClrCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</CoreClrCurrentRef>
+    <StandardCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</StandardCurrentRef>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <PlatformPackageVersion>2.0.0-preview2-25226-02</PlatformPackageVersion>
+    <MicrosoftNETCoreRuntimeCoreCLRPackageVersion>2.0.0-preview2-25226-02</MicrosoftNETCoreRuntimeCoreCLRPackageVersion>
+    <MicrosoftNETCoreRuntimeJitPackageVersion>$(MicrosoftNETCoreRuntimeCoreCLRPackageVersion)</MicrosoftNETCoreRuntimeJitPackageVersion>
+  </PropertyGroup>
+  
+  <!-- Package dependency verification/auto-upgrade configuration. -->
+  <PropertyGroup>
+    <BaseDotNetBuildInfo>build-info/dotnet/</BaseDotNetBuildInfo>
+    <DependencyBranch>master</DependencyBranch>
+    <CurrentRefXmlPath>$(MSBuildThisFileFullPath)</CurrentRefXmlPath>
   </PropertyGroup>
 
-  <!-- Create a collection of all project.json files for dependency updates. -->
   <ItemGroup>
-    <_ProjectJsonExcludes Include="$(SourceDir)test\**\project.json" />
-    <_ProjectJsonExcludes Include="$(SourceDir)pkg\**\project.json" />
-    <ProjectJsonFiles Include="$(SourceDir)**\project.json" Exclude="@(_ProjectJsonExcludes)" />
+    <RemoteDependencyBuildInfo Include="CoreFx">
+      <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/$(DependencyBranch)</BuildInfoPath>
+      <CurrentRef>$(CoreFxCurrentRef)</CurrentRef>
+    </RemoteDependencyBuildInfo>
+    <RemoteDependencyBuildInfo Include="CoreClr">
+      <BuildInfoPath>$(BaseDotNetBuildInfo)coreclr/$(DependencyBranch)</BuildInfoPath>
+      <CurrentRef>$(CoreClrCurrentRef)</CurrentRef>
+    </RemoteDependencyBuildInfo>
+    <RemoteDependencyBuildInfo Include="Standard">
+      <BuildInfoPath>$(BaseDotNetBuildInfo)standard/$(DependencyBranch)</BuildInfoPath>
+      <CurrentRef>$(StandardCurrentRef)</CurrentRef>
+    </RemoteDependencyBuildInfo>
+
+    <DependencyBuildInfo Include="@(RemoteDependencyBuildInfo)">
+      <RawVersionsBaseUrl>https://raw.githubusercontent.com/dotnet/versions</RawVersionsBaseUrl>
+    </DependencyBuildInfo>
+    
+    <XmlUpdateStep Include="CoreFx">
+      <Path>$(MSBuildThisFileFullPath)</Path>
+      <ElementName>PlatformPackageVersion</ElementName>
+      <PackageId>Microsoft.NETCore.Platforms</PackageId>
+    </XmlUpdateStep>
+    <XmlUpdateStep Include="CoreClr">
+      <Path>$(MSBuildThisFileFullPath)</Path>
+      <ElementName>MicrosoftNETCoreRuntimeCoreCLRPackageVersion</ElementName>
+      <PackageId>Microsoft.NETCore.Runtime.CoreCLR</PackageId>
+    </XmlUpdateStep>
   </ItemGroup>
 </Project>
--- a/dir.props
+++ b/dir.props
@@ -111,6 +111,8 @@
     <_DnuExcludeProjectJsonFiles Include="$(MSBuildThisFileDirectory)Bin\**\project.json" />
     
     <_ProjectJsonFiles Include="$(MSBuildThisFileDirectory)**\project.json" Exclude="@(_DnuExcludeProjectJsonFiles)" />
+    <ProjectJsonFiles Include="$(MSBuildThisFileDirectory)**\project.json.template" />
+    <ProjectJsonFiles Include="$(MSBuildThisFileDirectory)src\pkg\deps\project.json" />
     <DnuRestoreDirs Include="@(_ProjectJsonFiles->'%(RootDir)%(Directory)')">
       <AdditionalArgs>%(AdditionalArgs) --disable-parallel</AdditionalArgs>
     </DnuRestoreDirs>


dependencies.props
```diff
@@ -1,14 +1,71 @@
...
+    <CoreFxCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</CoreFxCurrentRef>
+    <CoreClrCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</CoreClrCurrentRef>
+    <StandardCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</StandardCurrentRef>
   </PropertyGroup>
-
-  <!-- Create a collection of all project.json files for dependency updates. -->
+  
+  <!-- Auto-upgraded properties for each build info dependency. -->
+  <PropertyGroup>
+	<CoreFxExpectedPrerelease>preview2-25301-02</CoreFxExpectedPrerelease>
```
eerhardt: (nit) can you clean up the whitespace in this file?  There is a mixture of tabs and spaces being used.  Use 2 spaces per indent (the default VS setting).

build.proj
```diff
@@ -71,6 +71,18 @@
   <Import Project="dir.targets" />
   <Import Project="dir.traversal.targets" />
   <Import Project="$(ToolsDir)clean.targets" />
+  
+  <Target Name="CollectTemplateFiles" BeforeTargets="UpdateDependencies" >
+    <ItemGroup>
+      <ProjectJsonTemplateFiles Include="$(MSBuildThisFileDirectory)/tools-local/**/project.json.template" />
+      <ProjectJsonTemplateFiles Include="$(MSBuildThisFileDirectory)/src/**/project.json.template" />
+      <ProjectJsonFiles Remove="*.json" />
```
weshaggard: Why do you need to remove "*.json" files? Are we all json.template files currently? Also where would the *.json files be getting added?

build.proj
```diff
@@ -71,6 +71,18 @@
   <Import Project="dir.targets" />
   <Import Project="dir.traversal.targets" />
   <Import Project="$(ToolsDir)clean.targets" />
+  
+  <Target Name="CollectTemplateFiles" BeforeTargets="UpdateDependencies" >
```
weshaggard: Is there a need to make this a target as opposed to just a static item group?

dependencies.props
```diff
@@ -1,14 +1,71 @@
...
+    <XmlUpdateStep Include="Standard">
+		<Path>$(MSBuildThisFileFullPath)</Path>
+		<ElementName>StandardExpectedPrerelease</ElementName>
+		<BuildInfoName>Standard</BuildInfoName>
+	</XmlUpdateStep>
+</ItemGroup>
+  
+  <!-- Copied Microsoft.NETCore.Runtime.CoreCLR and Microsoft.NETCore.Runtime.JIT versions -->
+  <PropertyGroup>
+       <CoreCLRVersion>2.0.0-$(CoreClrExpectedPrerelease)</CoreCLRVersion>
```
weshaggard: You should use an XmlUpdateStep to create these properties instead of relying on the fact they are always 2.0.0? See https://github.com/dotnet/corefx/blob/master/dependencies.props#L99 for example.

dependencies.props
```diff
@@ -1,14 +1,71 @@
...
+	<RemoteDependencyBuildInfo Include="Standard">
+		<BuildInfoPath>$(BaseDotNetBuildInfo)standard/$(DependencyBranch)</BuildInfoPath>
+		<CurrentRef>$(CoreClrCurrentRef)</CurrentRef>
+	</RemoteDependencyBuildInfo>
+
+	<DependencyBuildInfo Include="@(RemoteDependencyBuildInfo)">
+		<RawVersionsBaseUrl>https://raw.githubusercontent.com/dotnet/versions</RawVersionsBaseUrl>
+	</DependencyBuildInfo>
+	
+	<XmlUpdateStep Include="CoreFx">
```
eerhardt: The old update-dependencies system had 4 updaters:

[https://github.com/dotnet/core-setup/blob/release/2.0.0/build_projects/update-dependencies/Program.cs#L160-L168](https://github.com/dotnet/core-setup/blob/release/2.0.0/build_projects/update-dependencies/Program.cs#L160-L168)

1. Update 2 project.json.template files and a pkg/deps/project.json
2. Hard-coded CoreCLRVersion using `transport.Microsoft.NETCore.Runtime.CoreCLR`
3. Hard-coded JitVersion using `transport.Microsoft.NETCore.Jit`
4. Update pkg/dir.props and replace the version of `Microsoft.NETCore.Platforms` that is hard-coded there.

I don't see the pkg/deps/project.json file being updated
I don't see #2 and #3 happening like they used to (using the transport package versions)
I don't see #4 happening.

build.proj
karajas: They are getting added in dir.props but you're right we still use the project.jsons at least for tests, will add them back

build.proj
karajas: Now that I don't need Remove which only works for dynamic itemgroups I don't need a separate target.

dependencies.props
karajas: This was to take care of the cases @eerhardt  mentions below, I'll clean this up.

build.proj
chcosta: Another option, is to use `exclude` if you want to prevent some items from being included during static item group generation

dependencies.props
karajas: I can add the deps json,

2 and 3 I think I assumed we used the non transport packages will correct this

4 I don't think I tested, will fix

dependencies.props
```diff
@@ -1,14 +1,71 @@
...
+		<BuildInfoName>Standard</BuildInfoName>
+	</XmlUpdateStep>
+</ItemGroup>
+  
+  <!-- Copied Microsoft.NETCore.Runtime.CoreCLR and Microsoft.NETCore.Runtime.JIT versions -->
+  <PropertyGroup>
+       <CoreCLRVersion>2.0.0-$(CoreClrExpectedPrerelease)</CoreCLRVersion>
+       <JitVersion>2.0.0-$(CoreClrExpectedPrerelease)</JitVersion>
+  </PropertyGroup>
 </Project>
```
dagood: It would be slightly better for Build from Source if you moved the ProjectJsonFiles items into this file, like in CoreFX [dependencies.props#L177](https://github.com/dotnet/corefx/blob/dde2d618cc02fe1f226c7448b6bc9e309f0aede4/dependencies.props#L177).

Right now the workaround is to specify them in core-setup.props (like they are now in [corefx.props](https://github.com/dotnet/source-build/blob/3111bcb8c8307f29a7f9ffa4e29cc39647e4bd43/targets/corefx.props#L15), although that can be removed now).

dependencies.props
karajas: I've checked that all 4 scenarios work

dependencies.props
karajas: Moved into dependencies.props

dependencies.props
```diff
@@ -1,14 +1,71 @@
...
+  -->
+  <PropertyGroup>
+    <CoreFxCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</CoreFxCurrentRef>
+    <CoreClrCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</CoreClrCurrentRef>
+    <StandardCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</StandardCurrentRef>
+  </PropertyGroup>
+  
+  <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
-       <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
```
weshaggard: Don't you still want to define these?

karajas: yes we do.

karajas:
I don't think CI failures are mine. can you PTAL?

karajas:
msbuild build.proj /t:UpdateDependenciesAndSubmitPullRequest also works locally (seen here https://github.com/karajas/core-setup/tree/updateDependencies-UpdateDependencies). Thanks @dagood

chcosta:
I think this one is yours 
```
/mnt/resource/j/workspace/dotnet_core-setup/master/debug_debian8.2_x64_prtest/build.proj(63,5): error MSB4018: The "ValidateExactRestore" task failed unexpectedly.
```

eerhardt:
> I've checked that all 4 scenarios work

I don't see how 

```
  <PropertyGroup>
     <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
     <JitVersion>2.0.0-preview2-25226-02</JitVersion>
  </PropertyGroup>
```

Those are getting updated.

I also don't see how `4. Update pkg/dir.props and replace the version of Microsoft.NETCore.Platforms that is hard-coded there.` is getting updated.

dagood:
> /mnt/resource/j/workspace/dotnet_core-setup/master/debug_debian8.2_x64_prtest/build.proj(63,5): error MSB4018: The "ValidateExactRestore" task failed unexpectedly.

This means that when a project.json was restored, it ended up restoring a different package than it requested--this indicates that a version is in a project.json file that doesn't exist in MyGet, and the NuGet client is doing automatic dependency resolution. (It's unstable and we never want to depend on it.)

A way to get the specific package is to run the task locally and get the info logs: [ValidateExactRestore.cs#L81](https://github.com/dotnet/buildtools/blob/307a2e958f34e6892e72b3969c00fbc28ed58617/src/Microsoft.DotNet.Build.Tasks/ValidateExactRestore.cs#L81)

(I remembered the wrong thing in our IM, @eerhardt --I was thinking of [IsRestoreRequired](https://github.com/dotnet/buildtools/blob/78260239e4db68ad5c2df774a27d5f9a54fd92e5/src/Microsoft.DotNet.Build.Tasks/IsRestoreRequired.cs).)

dependencies.props
```diff
@@ -1,14 +1,76 @@
...
+    <XmlUpdateStep Include="Standard">
+        <Path>$(MSBuildThisFileFullPath)</Path>
+        <ElementName>StandardExpectedPrerelease</ElementName>
+        <BuildInfoName>Standard</BuildInfoName>
+    </XmlUpdateStep>
+  </ItemGroup>
+
+  <PropertyGroup>
+     <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
+     <JitVersion>2.0.0-preview2-25226-02</JitVersion>
```
dagood: Are these used somewhere? They aren't being upgraded, and we should probably remove these if not.

If they are being used, they should be tied to an exact package using an `XmlUpdateStep` with a `PackageId` metadata and probably given a more detailed name. For example, `MicrosoftNETCoreRuntimeCoreCLRPackageVersion` for `Microsoft.NETCore.Runtime.CoreCLR`.

dependencies.props
```diff
@@ -1,14 +1,76 @@
...
+    <CoreFxCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</CoreFxCurrentRef>
+    <CoreClrCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</CoreClrCurrentRef>
+    <StandardCurrentRef>824797963a7c7cfce78899f9c1af01e1011a1bd9</StandardCurrentRef>
+  </PropertyGroup>
+  
+  <!-- Auto-upgraded properties for each build info dependency. -->
+  <PropertyGroup>
+    <CoreFxExpectedPrerelease>preview2-25301-02</CoreFxExpectedPrerelease>
+    <CoreClrExpectedPrerelease>preview2-25301-02</CoreClrExpectedPrerelease>
+    <StandardExpectedPrerelease>preview2-25301-01</StandardExpectedPrerelease>
```
dagood: Removing these would be great, so that nobody down the line can try `2.0.0-$(CoreFxExpectedPrerelease)`. It makes the repo fragile to version revving down the road.

dependencies.props
karajas: https://github.com/dotnet/core-setup/blob/8030cdcdd47fb8db834fca10de3f9f5bfd63a6ac/src/managed/dir.proj#L120 seems to use it

karajas:
> I don't see how
> ```
>   <PropertyGroup>
>      <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
>      <JitVersion>2.0.0-preview2-25226-02</JitVersion>
>   </PropertyGroup>
> ```
> Those are getting updated.

since these are only used in [dir.props](https://github.com/dotnet/core-setup/blob/8030cdcdd47fb8db834fca10de3f9f5bfd63a6ac/src/managed/dir.proj#L120), can this use the prerelease tag then? @dagood 

> I also don't see how 4. Update pkg/dir.props and replace the version of Microsoft.NETCore.Platforms that is hard-coded there. is getting updated.

@eerhardt I don't see a hard-coded version in src/pkg/dir.props, let me check the pre buildtools update

As for the restore error, I think this is an ordering issue with the targets

karajas:
PlatformsVersion seems to be consumed here, https://github.com/dotnet/core-setup/blob/6887ab556bc8302390782711dcdd75b75e769cf5/src/pkg/dir.targets#L5 which is defaulted to 1.0.1 in buildtools 
```
<PlatformPackageVersion Condition="'$(PlatformPackageVersion)' == ''">1.0.1</PlatformPackageVersion>
```
should I add an xmlupdate step which overrides this?

karajas:
Seems like it's looking for a project.json.lock.json

dependencies.props
gkhanna79: Can you please update these to use the latest preview2 builds from CoreCLR and CoreFX since they bring updated Tizen toolset for building arm binaries?

karajas:
@gkhanna79 I just updated them

karajas:
The current error is because of the update PlatformsPackageVersion

dependencies.props
```diff
@@ -1,14 +1,59 @@
...
-       <JitVersion>2.0.0-preview2-25226-02</JitVersion>
+    <CoreFxCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</CoreFxCurrentRef>
+    <CoreClrCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</CoreClrCurrentRef>
+    <StandardCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</StandardCurrentRef>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <PlatformPackageVersion>2.0.0-preview2-25226-02</PlatformPackageVersion>
+    <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
+    <JitVersion>$(CoreCLRVersion)</JitVersion>
```
dagood: If CoreCLR does a servicing release that only includes one of these, they'll start to version separately and it would help to have them upgraded individually--but I don't think that ever happens, so this is probably fine. /cc @gkhanna79

karajas:
Can you PTAL?

dependencies.props
```diff
@@ -1,14 +1,59 @@
...
+    
+    <XmlUpdateStep Include="CoreFx">
+        <Path>$(MSBuildThisFileFullPath)</Path>
+        <ElementName>PlatformPackageVersion</ElementName>
+        <PackageId>Microsoft.NETCore.Platforms</PackageId>
+    </XmlUpdateStep>
+    <XmlUpdateStep Include="CoreClr">
+        <Path>$(MSBuildThisFileFullPath)</Path>
+        <ElementName>CoreCLRVersion</ElementName>
+        <PackageId>Microsoft.NETCore.Runtime.CoreCLR</PackageId>
```
dagood: I would slightly prefer a more descriptive `ElementName` than `CoreCLRVersion`, specifically something that mentions `Runtime`. (My suggestion earlier was the full id, `MicrosoftNETCoreRuntimeCoreCLRPackageVersion`.) Just a suggestion though.

dependencies.props
```diff
@@ -1,14 +1,59 @@
...
+    <PlatformPackageVersion>2.0.0-preview2-25226-02</PlatformPackageVersion>
+    <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
+    <JitVersion>$(CoreCLRVersion)</JitVersion>
+  </PropertyGroup>
+  
+  <!-- Package dependency verification/auto-upgrade configuration. -->
+   <PropertyGroup>
+      <BaseDotNetBuildInfo>build-info/dotnet/</BaseDotNetBuildInfo>
+      <DependencyBranch>master</DependencyBranch>
+      <CurrentRefXmlPath>$(MSBuildThisFileFullPath)</CurrentRefXmlPath>
```
dagood: Indentation needs to be fixed up (throughout the file).

build.proj
```diff
@@ -65,7 +65,23 @@
...
   <Import Project="dir.traversal.targets" />
-  <Import Project="$(ToolsDir)clean.targets" />
+  <Import Project="$(ToolsDir)clean.targets" />  
+ 
+  <!-- 
+    This is to initialize the ProjectJsonFiles ItemGroup with the template files during the update dependencies target
+    only to prevent ValidateExactRestore from failing on just the regular project.json files.
+  -->
+  <Target Name="IncludeOnlyTemplateFilesForUpdate" BeforeTargets="UpdateDependencies">
+   <ItemGroup>
```
dagood: Indentation is off.

dagood:
> > I don't see how
> > ```
> >   <PropertyGroup>
> >      <CoreCLRVersion>2.0.0-preview2-25226-02</CoreCLRVersion>
> >      <JitVersion>2.0.0-preview2-25226-02</JitVersion>
> >   </PropertyGroup>
> > ```
> > Those are getting updated.
> 
> [@karajas]
> since these are only used in [dir.props](https://github.com/dotnet/core-setup/blob/8030cdcdd47fb8db834fca10de3f9f5bfd63a6ac/src/managed/dir.proj#L120), can this use the prerelease tag then? @dagood 

In the PR now you're using `XmlUpdateStep`s to update to the exact package versions, which is the right way to go.

To maybe clarify, having the `preview2-25226-02` string *in* there is fine, but *only* having the `preview2-25226-02` string as a property is fragile. The full, unbroken `2.0.0-preview2-25226-02` is sturdier.

eerhardt:
> PlatformsVersion seems to be consumed here, https://github.com/dotnet/core-setup/blob/6887ab556bc8302390782711dcdd75b75e769cf5/src/pkg/dir.targets#L5 which is defaulted to 1.0.1 in buildtools
<PlatformPackageVersion Condition="'$(PlatformPackageVersion)' == ''">1.0.1</PlatformPackageVersion>
should I add an xmlupdate step which overrides this?

@ericstj - Can you comment here?  I'm not sure why we needed this before and I'm not sure if we need it now. But I'm fairly confident that if we do need it now, `1.0.1` is not going to be the correct version.

karajas:
It's weird that the linux runs are passing. They don't have a populated JitVersion

dependencies.props
```diff
@@ -1,14 +1,59 @@
...
+    </DependencyBuildInfo>
+    
+    <XmlUpdateStep Include="CoreFx">
+      <Path>$(MSBuildThisFileFullPath)</Path>
+      <ElementName>PlatformPackageVersion</ElementName>
+      <PackageId>Microsoft.NETCore.Platforms</PackageId>
+    </XmlUpdateStep>
+    <XmlUpdateStep Include="CoreClr">
+      <Path>$(MSBuildThisFileFullPath)</Path>
+      <ElementName>CoreCLRVersion</ElementName>
```
weshaggard: Shouldn't this be MicrosoftNETCoreRuntimeCoreCLRPackageVersion?

dependencies.props
```diff
@@ -1,14 +1,59 @@
...
+      <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/$(DependencyBranch)</BuildInfoPath>
+      <CurrentRef>$(CoreFxCurrentRef)</CurrentRef>
+    </RemoteDependencyBuildInfo>
+    <RemoteDependencyBuildInfo Include="CoreClr">
+      <BuildInfoPath>$(BaseDotNetBuildInfo)coreclr/$(DependencyBranch)</BuildInfoPath>
+      <CurrentRef>$(CoreClrCurrentRef)</CurrentRef>
+    </RemoteDependencyBuildInfo>
+    <RemoteDependencyBuildInfo Include="Standard">
+      <BuildInfoPath>$(BaseDotNetBuildInfo)standard/$(DependencyBranch)</BuildInfoPath>
+      <CurrentRef>$(CoreClrCurrentRef)</CurrentRef>
```
dagood: Just noticed--this should be `StandardCurrentRef`.

build.proj
```diff
@@ -65,7 +65,23 @@
...
   <Import Project="dir.targets" />
   <Import Project="dir.traversal.targets" />
-  <Import Project="$(ToolsDir)clean.targets" />
+  <Import Project="$(ToolsDir)clean.targets" />  
+ 
+  <!-- 
+    This is to initialize the ProjectJsonFiles ItemGroup with the template files during the update dependencies target
+    only to prevent ValidateExactRestore from failing on just the regular project.json files.
+  -->
+  <Target Name="IncludeOnlyTemplateFilesForUpdate" BeforeTargets="UpdateDependencies">
```
dagood: I think this needs to also be before `VerifyDependencies` and `UpdateDependenciesAndSubmitPullRequest`, because they're independent.

build.proj
```diff
@@ -65,7 +65,23 @@
...
+    <ItemGroup>
+      <ProjectJsonFiles Remove="*" />
```
dagood: Is this line doing anything? It doesn't seem like it would match anything (I'd expect `Remove="@(ProjectJsonFiles)"`) and I can't find where any `ProjectJsonFiles` is defined in the current iteration.

weshaggard: Yes remove * isn't going to work. If you are trying to clear the items do `<ProjectJsonFiles Remove="@(ProjectJsonFiles)" />`

weshaggard: Although I'm still not sure why we need to do this in a target? Why can't we control what we are putting into this and not need to all Remove?

karajas: @weshaggard I thought it might interfere with the update pr but I see that project.json are included in the gitignore. I'll remove the Remove statement. 

I still think the target may be needed unless you'd prefer a condition on the ItemGroup.

weshaggard: In general we should prefer to keep them in a static itemgroup. What condition would be needed?

dagood: Aha, I thought `ValidateExactRestore` was a target in BuildTools, but it's actually here in build.proj. I agree, no need for the target or for remove--the `ValidateExactRestore` target here should be modified to support the templated project.jsons.

karajas: Because the ProjectJsonFiles ItemGroup currently is used in both ValidateExactRestore (which requires actual project.json to look up the lockfile) and UpdateDependencies (which requires the template files to update). I thought a condition would be to separate the two cases

karajas: @dagood ok will do that instead.

karajas:
PTAL.

build.proj
```diff
@@ -60,12 +60,18 @@
 
   <Target Name="ValidateExactRestore"
           Condition="'$(AllowInexactRestore)'!='true'">
-    <ValidateExactRestore ProjectLockJsons="@(ProjectJsonFiles->'%(RootDir)%(Directory)%(Filename).lock.json')" />
+    <ItemGroup>
+      <ProjectJsonTemplateFiles Include="$(MSBuildThisFileDirectory)**\project.json.template" />
+      <ProjectJsonsExcludingTemplateFiles Include="@(ProjectJsonFiles)" Exclude="@(ProjectJsonTemplateFiles)" />
+    </ItemGroup>
+    <ValidateExactRestore ProjectLockJsons="@(ProjectJsonsExcludingTemplateFiles->'%(RootDir)%(Directory)%(Filename).lock.json')" />
```
dagood: Can this be changed to support project.json.template files rather than ignore them? (We have evaluated the template and restored them at this point, right?) There is value in this validation--protection against bad auto-upgrades and manual tweaks.

On the other hand, when we get rid of the `project.json.template` files, this validation won't work, so maybe it's not worth it. I don't know how far away that is.

weshaggard: Don't bother with it. This is OK for now.

karajas: I can follow up if needed

dagood:
LGTM too.

dependencies.props
```diff
@@ -1,14 +1,59 @@
...
+  -->
+  <PropertyGroup>
+    <CoreFxCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</CoreFxCurrentRef>
+    <CoreClrCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</CoreClrCurrentRef>
+    <StandardCurrentRef>f3d23289c93469d99c811fe911679d6c96ebfebf</StandardCurrentRef>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <PlatformPackageVersion>2.0.0-preview2-25226-02</PlatformPackageVersion>
+    <MicrosoftNETCoreRuntimeCoreCLRPackageVersion>2.0.0-preview2-25226-02</MicrosoftNETCoreRuntimeCoreCLRPackageVersion>
```
gkhanna79: Aren't these from last week?

karajas: Will update in a separate PR.

gkhanna79: Please do CC me on that PR as I would expect Arm CI to break.

gkhanna79: And I would need to merge https://github.com/dotnet/core-setup/pull/2163 to address that break - so, we should coordinate the merges.

PR #2244, opened 2017-05-01
CrossGen Microsoft.NETCore.App nuget packages

### LINKED ISSUE #2196: Crossgen binaries in flat NETCoreApp package
https://github.com/dotnet/corefx/issues/14595#issuecomment-297149207

This work was started by @ericstj, and needs to be completed in core-setup after the buildtools cut over

ericstj:
This moves crossgen to happen when building out the nuget packages for MS.NETCore.App rather than when building the shared framework.

I've refactored the build a little to have src build phase where all lose files that require signing are produced.

I've tried to reign in some of the refactoring to have a semi-minimal set of changes.  There is a bit more to do, but I was trying to keep the change semi-minimal.  Please review commit-by-commit.

Fixes #2196


ericstj:
Looking into the unix failures.  Probably something messed up with a path.

src/pkg/packages.builds
```diff
@@ -1,10 +1,18 @@
...
+  <PropertyGroup>
+    <!-- don't build project references from pkgprojs during the packages build phase,
+         they will be built during the source build phase -->
+    <AdditionalProperties>BuildPackageLibraryReferences=false</AdditionalProperties>
+  </PropertyGroup>
 
   <ItemGroup>
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)\projects\*\Microsoft.NETCore.UniversalWindowsPlatform.builds" Condition="'$(OSEnvironment)' != 'Windows_NT' or '$(Platform)' == 'arm64'" />
-    <Project Include="$(MSBuildThisFileDirectory)\projects\*\*.builds" Exclude="@(ProjectExclusions)" />
+    <Project Include="$(MSBuildThisFileDirectory)\projects\*\*.builds" Exclude="@(ProjectExclusions)">
```
weshaggard: I know this is already an existing change but how do we ensure that this *.builds wildcard only pulls packages?

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -24,17 +25,61 @@
...
+      <_fxSystemRuntime Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.FileName)%(ReferenceCopyLocalPaths.Extension)' == 'System.Runtime.dll'" />
     </ItemGroup>
-    <!-- start with all the files resolved -->
+
+    <PropertyGroup>
+      <!-- workaround https://github.com/dotnet/coreclr/issues/9884
+      <_crossDir Condition="'$(NuGetRuntimeIdentifier)' == 'win8-arm'">x86_arm</_crossDir>
+      <_crossDir Condition="'$(NuGetRuntimeIdentifier)' == 'win10-arm64'">x64_arm</_crossDir>
+       -->
+      <_crossDir Condition="'$(NuGetRuntimeIdentifier)' == 'win8-arm'">/x86_AnyCPU</_crossDir>
```
weshaggard: where is _crossdir getting set for the other architectures?

src/pkg/packages.builds
ericstj: Right now that remains true, purely based on convention.

I started to apply a different convention by putting these under a pkg folder then getting rid of the projects folder but it became too large of a change.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -92,6 +137,61 @@
     <Message Importance="low" Text="%(FilesToPackage.PackagePath) -> @(FilesToPackage)" />
   </Target>
 
+  <!-- Prepares all items for cross-gen and replaces package file items with their cross-gen'ed equivalents -->
+  <Target Name="PrepareForCrossGen" Condition="'$(NuGetRuntimeIdentifier)' != ''" DependsOnTargets="GetFilesFromPackages">
```
weshaggard: We should consider refactoring the crossgen targets into a common location as soon as we have another place we need to crossgen.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
ericstj: It's not needed on other architectures.  Here _crossdir implies "cross targeting directory" and is used when we build on one platform for another platform.

weshaggard:
cc @gkhanna79

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
ericstj: yep, I expect to do so once we get them working for UWP.

ericstj:
some of the x-plat failures are due to missing this hack from the depproj build:
https://github.com/dotnet/core-setup/blob/6887ab556bc8302390782711dcdd75b75e769cf5/src/pkg/dir.proj#L9-L13

I'm trying to remove that hack.  @chcosta let me know if you had a good reason for doing this...

chcosta:
No good reason.  it looked like a hack and there were no build issues with it removed.

chcosta:
no good reason, I didn't add that

ericstj:
Right now it is being used, but I'm trying to fix it the right way.  It looks like OSEnvironment is no longer being set to a value we expect so I have to find out some way to determine the OS we're running on..

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -164,7 +164,10 @@
     </ItemGroup>
     
     <PropertyGroup>
-      <_crossgenPlatformAssemblies>@(_crossgenPlatformDirectories->'%(Identity)', '%3B')</_crossgenPlatformAssemblies>
+      <!-- crossgen uses ':' (%3A) on unix, but ';' (%3B) on windows -->
```
ericstj: @gkhanna79 do you know if this is intentional?

ericstj:
Now it looks like I'm hitting some MSBuild difference between desktop (or at least Dev14 desktop) and core.  In core MSBuild is able to evaluate the following: `'$([MSBuild]::Escape($([System.IO.Path]::PathSeparator)))'` whereas on dev14/desktop this fails with `Invalid static method  invocation syntax: "[MSBuild]::Escape("$([System.IO.Path]::PathSeparator)")". Object of type 'System.Char' cannot be converted to type 'System.String'.`  I can workaround this by changing it to `'$([MSBuild]::Escape($([System.IO.Path]::PathSeparator.ToString())))'`.  I'll check dev15 desktop behavior and file a bug if its inconsistent with core.  /cc @rainersigwald

ericstj:
Monitoring https://mc.dot.net/#/product/netcore/master/source/official~2Fcore-setup~2Fmaster~2F/type/build~2Fproduct~2F/build/20170502.04/workItem/Orchestration/analysis/external/Visual%20Studio%20Build%20Information to ensure buildpipeline changes work as expected.

ericstj:
Looks like all xplat legs have completed successfully, and signed windows legs have made it successfully past the point of all of my changes:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=717613
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=717612

PR #2248, opened 2017-05-01
Use dotnet nuget to push to myget feed

chcosta:
Now that we have the updated cli in core-setup, use it to push to myget feed instead of a dependency on nuget.

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -106,8 +106,8 @@
     
     <Message Text="Pushing CoreHost packages to $(NuGetFeedUrl)" />
     <PropertyGroup>
-      <NuGetPushCommand>$(NuGetToolPath) push -Source $(NuGetFeedUrl) -ApiKey $(NuGetApiKey) -Timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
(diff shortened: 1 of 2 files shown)


publish/publish.proj
```diff
@@ -106,8 +106,8 @@
     
     <Message Text="Pushing CoreHost packages to $(NuGetFeedUrl)" />
     <PropertyGroup>
-      <NuGetPushCommand>$(NuGetToolPath) push -Source $(NuGetFeedUrl) -ApiKey $(NuGetApiKey) -Timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
```
eerhardt: I think `NuGetToolPath` can now be removed from dir.props as well

chcosta: I think you're right

chcosta: The CI failures appear to be because we had a build which published newer packages to myget.

chcosta: I'm going to test out setting buildnumberminor to 99 for non-official builds.  I'd expect that to prevent this type of failure until we get out of the "preview2-*" versioning in test projects

PR #2253, opened 2017-05-02
Specify content-type for .svg and .version files

chcosta:
Fix readme links, and specify the content-type for .svg and .version files so that they display correctly on the readme.md page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ Daily Builds
 
 [win-x64-badge-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/sharedfx_win-x64_Release_version_badge.svg
 [win-x64-version-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/dnvm/latest.sharedfx.win-x64.version
-[win-x64-installer-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/Latest/dotnet-win-x64.latest.exe
-[win-x64-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-win-x64.latest.zip
-[win-x64-symbols-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-sharedframework-symbols-win-x64.latest.zip
+[win-x64-installer-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/Latest/dotnet-win-x64-latest.exe
+[win-x64-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-win-x64-latest.zip
+[win-x64-symbols-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-sharedframework-symbols-win-x64-latest.zip
 
 [win-x64-badge-1.1.X]: https://dotnetcli.blob.core.windows.net/dotnet/release/1.1.0/Binaries/Latest/sharedfx_Windows_x64_Release_version_badge.svg
 [win-x64-version-1.1.X]: https://dotnetcli.blob.core.windows.net/dotnet/release/1.1.0/dnvm/latest.sharedfx.win.x64.version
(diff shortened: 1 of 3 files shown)


tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs
```diff
@@ -47,7 +47,8 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
             string filePath,
             string destinationBlob)
         {
-           
+            System.Console.WriteLine($"attach for {filePath}");
+            System.Console.ReadLine();
```
eerhardt: I assume both of these lines are unintentional.

tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs
```diff
@@ -142,6 +141,14 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
                     var req = new HttpRequestMessage(HttpMethod.Put, blockListUploadUrl);
                     req.Headers.Add(AzureHelper.DateHeaderString, dt1.ToString("R", CultureInfo.InvariantCulture));
                     req.Headers.Add(AzureHelper.VersionHeaderString, AzureHelper.StorageApiVersion);
+                    if(!string.IsNullOrEmpty(DetermineContentTypeBasedOnFileExtension(filePath)))
```
eerhardt: should probably put `DetermineContentTypeBasedOnFileExtension` in a variable and use it below, so we don't need two calls.

tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs
```diff
@@ -142,6 +141,14 @@ public string EncodeBlockIds(int numberOfBlocks, int lengthOfId)
...
+                    {
+                        req.Headers.Add(AzureHelper.ContentTypeString, DetermineContentTypeBasedOnFileExtension(filePath));
+                    }
+                    if (!string.IsNullOrEmpty(DetermineCacheControlBasedOnFileExtension(filePath)))
```
eerhardt: same as above for `DetermineCacheControlBasedOnFileExtension`

tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs
eerhardt: Actually, I would structure this like it was previously.  I think that approach makes more sense:

```
private void SetRequestHeadersBasedOnFileType(HttpRequestMessage req, string fileName)
{
  if(Path.GetExtension(filename) == ".svg")
  {
      req.Headers.Add(ContentTypeString, "image/svg+xml");
      req.Headers.Add(CacheControlString, "No-Cache");
  }
  else if (Path.GetExtension(fileName) == ".version")
  {
    req.Headers.Add(ContentTypeString, "text/plain");
  }
}

File: tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs (hunk 1 above)
Comment (chcosta) [inline tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs]: yes, this was for debugging.

File: tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs (hunk 2 above)
Comment (chcosta) [inline tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs]: I went this direction because it's more flexible, and I moved the values into methods so that they were more re-usable.

Comment (chcosta) [inline tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs]: It's a goal of mine, to get all of the tasks under Microsoft.DotNet.Build.Tasks.Local merged back into BuildTools where they can be consumed more general purpose by other repo's.  So, flexibility is slightly more important for these tasks than it would be for the tasks under `tools-local\tasks` which are more core-setup specific.

**Comment by chcosta** [comment]
@dotnet-bot test Fedora24 x64 Debug Build

**Comment by chcosta** [comment]
@dotnet-bot test Fedora24 x64 Debug Build

**Comment by chcosta** [comment]
appears to be an intermittent failure causing the ci failure related to removing a directory.  I don't think it's worth investigating, unless it's a persistent problem, until after mellinoe's test changes are merged.

File: tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs (hunk 2 above)
Comment (eerhardt) [inline tools-local/Microsoft.DotNet.Build.Tasks.Local/UploadClient.cs]: Well then back to my original comment:  don't call the method twice.

File: README.md [hunk 4]
```diff
@@ -59,10 +59,10 @@ Daily Builds
 
 
 [win-x64-badge-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/sharedfx_win-x64_Release_version_badge.svg
-[win-x64-version-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/dnvm/latest.sharedfx.win,x64.version
-[win-x64-installer-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/Latest/dotnet-win-x64.latest.exe
-[win-x64-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-win-x64.latest.zip
-[win-x64-symbols-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-sharedframework-symbols-win-x64.latest.zip
+[win-x64-version-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/dnvm/latest.sharedfx.win-x64.version
+[win-x64-installer-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/Latest/dotnet-win-x64-latest.exe
+[win-x64-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-win-x64-latest.zip
```
eerhardt: As discussed in the [CLI PR](https://github.com/dotnet/cli/pull/6493#discussion_r114471036), we can't be changing the format of these URLs.

chcosta: Yes, I'll be fixing these links / format later today in a separate PR

chcosta:
Merging this, and then fixing links.  I'm happy to address any additional feedback if necessary

PR #2254, opened 2017-05-02
Fix Auto update of versions repo

karajas:
This is to allow versions repo to be updated using VersionTools.
Tested here https://github.com/karajas/versions/commit/8c92965add0e3751e84df55a1877903f7ae71dd0 

/cc @dagood @weshaggard @eerhardt @chcosta

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -67,6 +67,12 @@
     <ValidateExactRestore ProjectLockJsons="@(ProjectJsonsExcludingTemplateFiles->'%(RootDir)%(Directory)%(Filename).lock.json')" />
   </Target>
 
+  <Target Name="InitializeShippedNuGetPackage" BeforeTargets="UpdatePublishedVersions" >
+    <ItemGroup>
+      <ShippedNuGetPackage Include="$(PackagesOutDir)/**/*.nupkg" />
+    </ItemGroup>
+  </Target>
+
   <Import Project="dir.targets" />
   <Import Project="dir.traversal.targets" />
   <Import Project="$(ToolsDir)clean.targets" />


publish/publish.proj
```diff
@@ -65,6 +65,14 @@
                    Condition="'@(_MissingBlobNames)' == ''" />
     <Message Importance="High" Condition="'@(_MissingBlobNames)' != ''" Text="Skipping finalization, missing these blobs: '@(_MissingBlobNames)'" />
   </Target>
+  
+  <Target Name="UpdateVersionsRepo"
```
weshaggard: Any reason we wouldn't do this in the publish build definition like in corefx? https://github.com/dotnet/corefx/blob/b5ac646f30e2d36729bdb95dac072ceb1e5b186d/buildpipeline/DotNet-Trusted-Publish.json#L177

weshaggard: We should just do that instead of adding this here then right?

publish/publish.proj
```diff
@@ -65,6 +65,14 @@
...
+          DependsOnTargets="FinalizeBuildInAzure"
```
eerhardt: This dependency is backwards.  You should only update the versions repo after everything is finished.  It should be the absolute last thing that is done.

publish/publish.proj
```diff
@@ -65,6 +65,14 @@
...
+          Condition="'$(IsOfficialBuild)'=='true'" >
+      <MSBuild Projects="$(MSBuildThisFileDirectory)../build.proj"  Targets="UpdatePublishedVersions"
+               Properties="ShippedNuGetPackage=@(ShippedNuGetPackage);VersionsRepoPath=$(VersionsRepoPath);GitHubAuthToken=$(GitHubAuthToken);GitHubUser=$(GitHubUser);GitHubEmail=$(GitHubEmail);VersionsRepo=$(VersionsRepo);VersionsRepoOwner=$(VersionsRepoOwner)" />
```
chcosta: Have you checked to ensure that these properties are getting passed by the builds?

publish/publish.proj
chcosta: I'd like to get something working to validate our builds are signaling maestro to schedule cli builds which consume core-setup.

dir.props
```diff
@@ -120,6 +120,8 @@
       <AdditionalArgs>%(AdditionalArgs) --disable-parallel --infer-runtimes</AdditionalArgs>
     </DnuRestoreDirs>
     <DnuRestoreDirs Include="$(MSBuildThisFileDirectory)src/pkg/deps" />
+    
+    <ShippedNuGetPackage Include="$(PackagesOutDir)/**/*.nupkg" />
```
eerhardt: How does this work?  Doesn't this get evaluated before the `nupkgs` are created?

Also, `dir.props` is a dumping ground.  Not every project needs this.  Most don't need it.  This line is only needed in a single project.  Why put this in a common place when only 1 place needs it?

weshaggard: Agreed.

karajas: You're right, I'll move it to the target.

publish/publish.proj
weshaggard: Then manually run this command temporarily. I would much prefer we spend our energy on getting this done right the first time instead of doing again later this week.

publish/publish.proj
karajas: I think the official build  should be setting these currently.

publish/publish.proj
karajas: Don't we want the versions to publish after all legs have succeeded? Is that not after finalize?

publish/publish.proj
chcosta: That's reasonable.  I'll ask @karajas to send me the manual command / tokens for this.

weshaggard:
Based on our conversation yesterday I'm assuming we can close this so I'm closing.

build.proj
```diff
@@ -67,6 +67,12 @@
     <ValidateExactRestore ProjectLockJsons="@(ProjectJsonsExcludingTemplateFiles->'%(RootDir)%(Directory)%(Filename).lock.json')" />
   </Target>
 
+  <Target Name="InitializeShippedNuGetPackage" BeforeTargets="UpdatePublishedVersions" >
+    <ItemGroup>
+      <ShippedNuGetPackage Include="$(PackagesOutDir)/**/*.nupkg" />
```
weshaggard: To handle this corefx just passed the ShippedNuGetPackageGlobPath property to https://github.com/dotnet/corefx/blob/b5ac646f30e2d36729bdb95dac072ceb1e5b186d/buildpipeline/DotNet-Trusted-Publish.json#L177

Issue #2257, opened 2017-05-02
Share source between depprojs that build framework packages


ericstj:
Microsoft.NETCore.App and Microsoft.NETCore.UniversalWindowsPlatform duplicate a lot of targets source.  We should unify this.

ericstj:
Moving this out as it isn't particularly interesting for UWP since its is on a different schedule than CoreApp.

dagood:
UWP is no longer built in core-setup, and tooling is unified in the Arcade SDK.

PR #2258, opened 2017-05-02
Add source to framework symbols packages

### LINKED ISSUE #2156: Include source in Core.App and UAP packages
We are flowing symbols and source from corefx but it looks like its not all making it into the meta-packages.

ericstj:
Fixes #2156 

/cc @dagood @chcosta @weshaggard @joperezr 

We flow the source in transport/private packages but it wasn't getting pulled in, so it was missing from the framework symbols packages.

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -134,6 +134,19 @@
     </ItemGroup>
     <Error Text="Unexpected System package(s) @(_secondarySystemPackages)" Condition="'@(_secondarySystemPackages)' != ''" />
 
+    <ItemGroup>
+      <!-- pick up any src our sources directory from packages contributing files -->
+      <_sourcePaths Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\sources')" />
+      <_sourcePaths Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\src')" />
+
+      <!-- include all source from those directories -->
+      <_sourceFile Condition="'%(_sourcePaths.Identity)' != ''" Include="%(_sourcePaths.Identity)\**\*.*" />
+      <FilesToPackage Include="@(_sourceFile)">
+        <TargetPath>src\%(RecursiveDir)%(FileName)%(Extension)</TargetPath>
+        <IsSourceCodeFile>true</IsSourceCodeFile>
+      </FilesToPackage>
+    </ItemGroup>
+
     <Message Importance="low" Text="%(FilesToPackage.PackagePath) -> @(FilesToPackage)" />
   </Target>
 
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -136,18 +136,30 @@
 
     <ItemGroup>
       <!-- pick up any src our sources directory from packages contributing files -->
-      <_sourcePaths Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\sources')" />
-      <_sourcePaths Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\src')" />
+      <_sourcePathCandidate Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\sources')" />
+      <_sourcePathCandidate Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\src')" />
+      <_sourcePath Include="@(_sourcePathCandidate)" Condition="Exists('%(Identity)')" />
+    </ItemGroup>
 
-      <!-- include all source from those directories -->
-      <_sourceFile Condition="'%(_sourcePaths.Identity)' != ''" Include="%(_sourcePaths.Identity)\**\*.*" />
-      <FilesToPackage Include="@(_sourceFile)">
-        <TargetPath>src\%(RecursiveDir)%(FileName)%(Extension)</TargetPath>
+    <!-- on windows workaround max-path -->
+    <!-- this can be removed once we build on core MSBuild which supports long paths. -->
+    <ItemGroup Condition="'$(OsEnvironment)'=='Windows_NT'">
+      <!-- choose a shorter path name, excluding package version -->
+      <_shortSourcePath Include="@(_sourcePath->'$(ObjDir.Replace('/', '\'))src-%(NuGetPackageId)')">
+        <Original>%(_sourcePath.FullPath)</Original>
+      </_shortSourcePath>
+      <_sourcePath Remove="@(_sourcePath)" />
+      <_sourcePath Include="@(_shortSourcePath)" />
+    </ItemGroup>
+    <RemoveDir Condition="'@(_shortSourcePath)' != '' AND Exists('%(_shortSourcePath.Identity)')" Directories="%(_shortSourcePath.Identity)" />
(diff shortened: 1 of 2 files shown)


ericstj:
I have opened https://github.com/dotnet/core-setup/issues/2257 to track de-duping these project files.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -134,6 +134,19 @@
...
+      <_sourcePaths Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\sources')" />
+      <_sourcePaths Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\src')" />
+
+      <!-- include all source from those directories -->
+      <_sourceFile Condition="'%(_sourcePaths.Identity)' != ''" Include="%(_sourcePaths.Identity)\**\*.*" />
+      <FilesToPackage Include="@(_sourceFile)">
+        <TargetPath>src\%(RecursiveDir)%(FileName)%(Extension)</TargetPath>
+        <IsSourceCodeFile>true</IsSourceCodeFile>
+      </FilesToPackage>
+    </ItemGroup>
```
dagood: Would it be reasonable to put this in a `dir.targets` somewhere along the line instead of duplicating? (Not very familiar with the `pkg` projects, not sure myself.)

ericstj: Yeah, this isn't the first place we're duplicating so I decided not to tackle that yet, but I did make sure the change was identical so that deduping later will be easy.

weshaggard:
Where are the pdb's being harvested? I took a quick look but didn't see that in these projects.

ericstj:
Looks like we're hitting MAX_PATH on windows: 
```
error MSB4023: Cannot evaluate the item metadata "%(FileName)". The item metadata "%(FileName)" cannot be applied to the path "D:\j\workspace\release_windo---0c55dee9\packages/\runtime.win7-x64.Microsoft.Private.CoreFx.NETCoreApp\4.4.0-preview2-25226-02\sources\bin\obj\AnyOS.AnyCPU.Release\System.Security.Cryptography.OpenSsl\netstandard\System.Security.Cryptography.OpenSsl.notsupported.cs". The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
```

ericstj:
> Where are the pdb's being harvested?

They actually come from NuGet automatically if in the package.

ericstj:
I'm trying to see if I can workaround the max-path bug.

ericstj:
I was able to workaround this in a couple places in my projects but after this we hit it later in the build process.  We need to either a) get core-setup windows builds running on core MSBuild which can handle long paths or b) shorten the path that jenkins clones the repo to.

ellismg:
> I was able to workaround this in a couple places in my projects but after this we hit it later in the build process. We need to either a) get core-setup windows builds running on core MSBuild which can handle long paths or b) shorten the path that jenkins clones the repo to.

Could we just `subst` as part of the build script?

weshaggard:
We have other folks hitting this in other places as well. See https://github.com/dotnet/coreclr/issues/11442.

We should see what paths we can cut down to help with this issue. @mellinoe had a good suggestion which would help which would be to remove TargetName from the file name at https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/notsupported.targets#L7

ericstj:
@weshaggard / @mellinoe  the notsupported files are only one case.  If I skip those then I hit places where the repo structure itself is too deep.  EG:
```
F:\dotnet\core-setup-break-max-path-xxx\src\pkg\projects\Microsoft.NETCore.App\src\Microsoft.NETCore.App.depproj(137, 5): error MSB4023: Cannot evaluate the item metadata "%(FileName)". The item metadata "%(FileName)" cannot be applied to the path "F:\dotnet\core-setup-break-max-path-xxx\packages/\runtime.win7-x64.Microsoft.Private.CoreFx.NETCoreApp\4.4.0-preview2-25226-02\sources\src\System.ComponentModel.TypeConverter\src\System\ComponentModel\Design\Serialization\DefaultSerializationProviderAttribute.cs". The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
```

@ellismg subst is an interesting suggestion but it makes me nervous that it bleeds into machine state.

I'll see if I can workaround by copying/moving stuff.

ericstj:
I tried using NuGet's globbing by letting it evaluate the wildcard, but it fails there with PathTooLong just as well (since it eventually calls our Path APIs on desktop).

Looking at a copy.

ericstj:
Fixed with a bit of a dirty hack...  I don't like it but it works.

ericstj:
@gkhanna79 did you break the arm CI?
@weshaggard can you review the long-path workaround?

ericstj:
@dotnet-bot test Ubuntu arm Release Build

ericstj:
@ellismg is looking at the CI issue, thanks!  /cc @eerhardt

ericstj:
@dotnet-bot test Ubuntu arm Release Build

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -134,7 +134,32 @@
...
+    <ItemGroup>
+      <!-- pick up any src our sources directory from packages contributing files -->
+      <_sourcePathCandidate Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\sources')" />
+      <_sourcePathCandidate Include="@(FilesToPackage->'$(PackagesDir)\%(NuGetPackageId)\%(NuGetPackageVersion)\src')" />
+      <_sourcePath Include="@(_sourcePathCandidate)" Condition="Exists('%(Identity)')" />
+    </ItemGroup>
+
+    <!-- on windows workaround max-path -->
+    <!-- this can be removed once we build on core MSBuild which supports long paths. -->
+    <ItemGroup Condition="'$(OsEnvironment)'=='Windows_NT'">
```
weshaggard: Should we just make the condition based on MSBuildRuntimeType != core?

weshaggard:
>@weshaggard can you review the long-path workaround?

Looks like an OK hack but this really seems like it is going to start causing us more problems and we need to come up with an all up strategy.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
ericstj: I considered it but it felt wrong since mklink is windows only.  I suppose I could make it `'$(MSBuildRuntimeType)' != 'core' and '$(OsEnvironment)'=='Windows_NT'` if you think that is cleaner.

ericstj:
@weshaggard All up strategy would be build with .NET Core which supports long paths, and/or update desktop version pre-req of build machines to one that supports long paths.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
weshaggard: No I'm good with this for now.

weshaggard:
>@weshaggard All up strategy would be build with .NET Core which supports long paths, and/or update desktop version pre-req of build machines to one that supports long paths.

I don't know if depending on long path support is the right approach as there will likely always be folks using windows non-core tools to read our packages which will break.

ericstj:
@weshaggard these are only the symbols packages, and the packages themselves don't exceed long-path; just the path once extracted to a somewhat deep packages folder.  If we want to instead enforce a path maximum in our repo for source files that's an alternative.  We could assume the maximum length based on `MAX_PATH - (<maximumPackageFolderLength>+1+<maximumIdLength>+1+<maximumVersionLength>+1)`.  This could still cause problems when folks decide to clone our repo in a deep path and it exceeds `<maximumPackageFolderLength>`

weshaggard:
Yeah I was only referenceing https://github.com/dotnet/coreclr/issues/11442 which has starting to cause some issues. We can make some tweaks to improve but not necessary eliminate the problem completely.

Issue #2127, opened 2017-04-20
Tests should restore a specific version of Microsoft.NETCore.DotNetAppHost, and not the highest version available on MyGet


ravimeda:
## Steps to reproduce
Restore `Microsoft.NETCore.DotNetAppHost` package, which is needed for tests located at `/opt/code/TestAssets/TestProjects`
This can be by running `build.sh` or `build.cmd`.

## Expected  behavior
Version matching should not be `*`. Rather should match a specific version. For example, `2.0.0-preview1-002042-00`

## Actual behavior
Restores the highest version matching `2.0.0-preview1-*`
Refer to https://github.com/dotnet/core-setup/blob/master/TestAssets/TestProjects/LightupClient/project.json

## Environment data

ravimeda:
FYI @chcosta @weshaggard

chcosta:
I think the issue is restoring "Microsoft.NETCore.App", not DotNetAppHost specifically.  The tests reference Microsoft.NETCore.App and then that package pulls in its dependencies , which include DotNetAppHost

gkhanna79:
The * enables referencing the version to be restored that built during the build. Thus, it cannot be the specific version.

gkhanna79:
CC @eerhardt @ramarag

weshaggard:
Yes but once we move to PackageReference vs project.json we should directly reference the specific version we are building to eliminate the trouble we have with the * versioning.

eerhardt:
Yes, the main issue here is that the tests *need* to reference the version of the product that was just built.  This was not as feasible in project.json world since you couldn't use parameters.

I agree with @weshaggard that moving to MSBuild will help this issue.

mellinoe:
Thanks for letting me know about this. I was banging my head against the wall trying to figure out some problems related to this.

PR #2259, opened 2017-05-02
Build managed projects and tests with csproj's

mellinoe:
* Replace all test projects and test assets with csproj files.
  * Update some test code to use only properties that the 2.0 CLI understands.
  * Pass through additional properties in the test code in order to restore and build projects for the correct RID and version of Microsoft.NETCore.App (Fixes https://github.com/dotnet/core-setup/issues/2127)
* Replace src/managed/*project.json with csproj files.
  * Update `GetBuildArgsByFrameworks` so it understands how to read and parse csproj files instead of project.json files.
* Update dir.proj's to correctly restore and build the csproj files (using `DotnetSdkToolCommand` instead of `DotnetToolCommand`).

@eerhardt @weshaggard


mellinoe:
I pushed a couple of commits which should fix some Linux problems. I'm going to need some help understanding the remaining test issues -- they don't make a lot of sense to me.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -0,0 +1,44 @@
...
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
+    <Reference Include="System" />
+    <Reference Include="Microsoft.CSharp" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
```
weshaggard: nit: should these last 2 be the same item group?

src/managed/dir.proj
```diff
@@ -24,7 +24,7 @@
       <Output ItemName="buildCmdArgs" TaskParameter="BuildArgs" />
     </GetBuildArgsByFrameworks>
 
-    <Exec Command="$(DotnetToolCommand) build $(Args) %(buildCmdArgs.Identity)" 
+    <Exec Command="$(DotnetSdkToolCommand) build $(Args) %(buildCmdArgs.Identity)"
```
weshaggard: potential future clean-up we should just let msbuild handle the building of these projects instead of shelling out to dotnet build.

src/pkg/packaging/dir.proj
```diff
@@ -160,18 +160,19 @@
 
   <Target Name="GenerateNugetPackages" DependsOnTargets="InitPackage">
     <ItemGroup>
-      <PackageProjects Include="$(ProjectDir)src\managed\Microsoft.DotNet.PlatformAbstractions\project.json" />
-      <PackageProjects Include="$(ProjectDir)src\managed\Microsoft.Extensions.DependencyModel\project.json" />
+      <PackageProjects Condition="'$(OS)' == 'Windows'" Include="$(ProjectDir)src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
```
weshaggard: Why only on windows?

weshaggard: If it is to make it unique across build legs don't we also need to take into consideration arch and configurationgroup?

src/test/Assets/TestProjects/LightupLib/LightupLib.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AssemblyName>LightupLib</AssemblyName>
+    <OutputType>Library</OutputType>
+    <PackageId>LightupLib</PackageId>
+    <RuntimeFrameworkVersion>$(MNAVersion)</RuntimeFrameworkVersion>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
```
weshaggard: Why the condition?

src/test/TestUtils/RepoDirectoriesProvider.cs
```diff
@@ -39,16 +41,23 @@ public class RepoDirectoriesProvider
...
             string dotnetSdk = null)
         {
-            _repoRoot = repoRoot ?? Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.FullName;
+            _repoRoot = repoRoot ?? Directory.GetParent(Directory.GetCurrentDirectory()).Parent.Parent.Parent.Parent.Parent.FullName;
 
             string baseArtifactsFolder = artifacts ?? Path.Combine(_repoRoot, "Bin");
             
             _targetRID = Environment.GetEnvironmentVariable("TEST_TARGETRID");
             _buildRID = Environment.GetEnvironmentVariable("BUILDRID");
+            _mnaVersion = Environment.GetEnvironmentVariable("MNA_VERSION");
```
weshaggard: Boo... don't like seeing these environment variables.

src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -310,15 +317,10 @@ public TestProjectFixture EnsureRestored(params string[] fallbackSources)
...
-            var sourceProjectJson = Path.Combine(_testProject.ProjectDirectory, "project.json.template");
-            var targetProjectJson = Path.Combine(_testProject.ProjectDirectory, "project.json");
-
-            // apply RID to template
-            File.WriteAllText(targetProjectJson, File.ReadAllText(sourceProjectJson).Replace("{RID}", rid));
-
             if ( ! _testProject.IsRestored())
             {
-                RestoreProject(fallbackSources);
+                string extraMSBuildProperties = $"/p:RuntimeIdentifiers={rid} /p:RuntimeIdentifier={rid}";
```
weshaggard: Do we really need to set both of these?

src/test/dir.proj
```diff
@@ -102,10 +106,9 @@
       <IsCrossArch Condition="'$(TargetArchitecture)' == 'arm' or '$(TargetArchitecture)' == 'arm64' or '$(TargetArchitecture)' == 'armel'">true</IsCrossArch>
       <TestResultsXml>$(TestsOutputDir)$(TestProjectFilename)-testResults.xml</TestResultsXml>
     </PropertyGroup>
-
-    <Exec Command="$(DotnetTestToolCommand) test $(TestArgs) -xml $(TestResultsXml) -notrait category=failing"
-          WorkingDirectory="$(TestProject)"
-          EnvironmentVariables="NUGET_PACKAGES=$(PackagesDir);TEST_ARTIFACTS=$(SystemPathTestsOutputDir);TEST_TARGETRID=$(TestTargetRid);BUILDRID=$(TargetRid)"
+    <Exec Command="$(DotnetTestToolCommand) test $(TestArgs) --logger &quot;trx;LogFileName=$(TestResultsXml)&quot;"
+          WorkingDirectory="$(TestWorkingDirectory)"
+          EnvironmentVariables="NUGET_PACKAGES=$(PackagesDir);TEST_ARTIFACTS=$(SystemPathTestsOutputDir);TEST_TARGETRID=$(TestTargetRid);BUILDRID=$(TargetRid);MNA_VERSION=$(ProductVersion)"
```
weshaggard: Do msbuild properties correctly get propagated through to test projects now that we are using msbuild for test?

src/pkg/packaging/dir.proj
mellinoe: It is because these projects cannot be built outside of Windows -- they rely on the .NET Framework targeting pack. Previously, we were skipping the "net45" portion of the build, but still running `dotnet pack --no-build` on them in this phase. With the 2.0 SDK, the `dotnet pack` command fails because the .NET Framework targeting pack can't be found, even though we've passed `--no-build`.

src/test/Assets/TestProjects/LightupLib/LightupLib.csproj
mellinoe: Hrm -- just cruft from `dotnet migrate`. I will clean that up.

src/test/TestUtils/TestProjectFixture.cs
mellinoe: We can just make it a single property and then make the two passed in here reference that one.

src/test/dir.proj
mellinoe: Hm, could you clarify what you mean?

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
mellinoe: Yeah, I think so. I can also probably condense the first two ItemGroups above. Some of this is just cruftiness from the `dotnet migrate` output, I think.

mellinoe:
With the two latest commits, I have a clean build on my Ubuntu machine. Hopefully we will see some green in CI now.

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -50,6 +50,12 @@ public void Running_Build_Output_Standalone_EXE_with_DepsJson_and_RuntimeConfig_
 
             var appExe = fixture.TestProject.AppExe;
 
+            // TODO: Use FS.Chmod when build utility project is converted to csproj.
```
mellinoe: @eerhardt This is how I'm handling the executable permission flags problem for now. We can probably just remove this workaround when we pick up an SDK that handles this.

src/test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
```diff
@@ -17,7 +17,8 @@ public class PackageResolverTest
     {
         private static string PackagesPath = Path.Combine("package", "directory", "location");
 
-        [Fact]
+        // This test does not work on .NET Core 2.0.
```
mellinoe: @eerhardt @ramarag I've commented these out to see if I can get a green build without them. We should figure out what we want to do about these, though.

tools-local/tasks/GetBuildableFrameworks.cs
```diff
@@ -31,21 +32,20 @@ public override bool Execute()
             List<string> args = new List<string>();
             foreach (var projectJsonPath in ProjectJsonPaths)
             {
-                using (TextReader projectFileReader = File.OpenText(projectJsonPath.ItemSpec))
+                string dir = Path.GetDirectoryName(projectJsonPath.ItemSpec);
```
weshaggard: This entire task seems like it is going to cause us trouble. Do we have a better way to get what we need here?

mellinoe: This all falls out of the hacky way we are building the stuff under src/managed. If those projects were properly hooked into the build system (perhaps with a CoreFX-like configuration system), then we wouldn't need to be doing this craziness.

src/test/dir.proj
weshaggard: As in could we just pass these values as properties to the test command and then get them passed as arguments to xunit instead of relying on env variables.

mellinoe:
I think we will need to modify the netci.groovy script, because we are not producing xunit test results anymore. Rather, we are producing "TRX" logs.

src/test/TestUtils/TestProjectFixture.cs
mellinoe: Actually -- I had already made that change part of the way in the project file. I'm just passing one property through here now.

src/test/dir.proj
mellinoe: A couple things:

* The test runner doesn't allow you to pass variables through to Xunit.
* These values are actually used within the test code, and I'm not aware of a way to get Xunit to pass that sort of thing through, anyways.

I agree that these environment variables are awful, though 😞

src/managed/dir.proj
mellinoe: Yeah, 100% agreed there. This is the only purpose for the `GetBuildArgsByFrameworks` task, by the way. If these projects were hooked into the build regularly, we wouldn't need it.

weshaggard:
I"m good with merging this if you have confirmed it works and follow-up after your netci changes are in to ensure everything is green.

mellinoe:
The only thing I need to fix is the x86 run. I think I understand why it failed, though.

mellinoe:
I think the latest commit should fix the last issue. Once all the jobs finish and I see that the only problem is the testResults.xml files not being Xunit logs, I will merge.

mellinoe:
@dotnet-bot Test CI please

dir.props
```diff
@@ -129,11 +132,13 @@
     <DotnetSdkToolCommand Condition="'$(DotnetSdkToolCommand)'=='' and '$(OsEnvironment)'=='Unix'">$(DotnetSdkPath)dotnet</DotnetSdkToolCommand>
     <OverrideToolHost Condition="'$(MSBuildRuntimeType)' == 'Core'">$(DotnetToolCommand)</OverrideToolHost>
     <DnuRestoreSource>@(DnuSourceList -> '--source %(Identity)', ' ')</DnuRestoreSource>
-
-    <DnuRestoreCommand>"$(DotnetToolCommand)"</DnuRestoreCommand>
+    <DnuRestorePrefix Condition="'$(OsEnvironment)' != 'Windows_NT'">DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1</DnuRestorePrefix>
```
eerhardt: I think this should/needs to be set for the whole build.  And also on both Windows and non-Windows - so the OS's are consistent.

run.sh
```diff
@@ -17,5 +17,5 @@ __dotnet=$__toolsLocalPath/dotnetcli/dotnet
 
 cp -fR $__scriptpath/tools-override/* $__toolsLocalPath 
 
-$__dotnet $__toolsLocalPath/run.exe $*
-exit $?
\ No newline at end of file
+DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 $__dotnet $__toolsLocalPath/run.exe $*
```
eerhardt: I think this should also happen for Windows.

Also, why do we need this twice?  I saw it in build.proj as well?

src/managed/Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <AssemblyOriginatorKeyFile>../../../tools-local/setuptools/Key.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <PackageId>Microsoft.DotNet.PlatformAbstractions</PackageId>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>git://github.com/dotnet/core-setup</RepositoryUrl>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
+    <Reference Include="System" />
```
eerhardt: Is this reference auto-included when targeting `net45`?  Did the `dotnet migrate` put this here?

src/managed/Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj
```diff
@@ -0,0 +1,34 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Abstractions for making code that uses file system and environment testable.</Description>
+    <VersionPrefix>2.0.0</VersionPrefix>
+    <TargetFrameworks>net45;netstandard1.3</TargetFrameworks>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <AssemblyName>Microsoft.DotNet.PlatformAbstractions</AssemblyName>
```
eerhardt: Shouldn't need this.  It gets defaulted.

src/managed/Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <AssemblyOriginatorKeyFile>../../../tools-local/setuptools/Key.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <PackageId>Microsoft.DotNet.PlatformAbstractions</PackageId>
```
eerhardt: Shouldn't need this.  It gets defaulted.

src/managed/Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj
eerhardt: Here are the implicit references for net framework:  [https://github.com/dotnet/sdk/blob/4a7e4717d3ac152ca82099c324fb19b610b6053a/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.BeforeCommon.targets#L78-L103](https://github.com/dotnet/sdk/blob/4a7e4717d3ac152ca82099c324fb19b610b6053a/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.BeforeCommon.targets#L78-L103)

---
In reply to: [114665264](https://github.com/dotnet/core-setup/pull/2259#discussion_r114665264) [](ancestors = 114665264)

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Abstractions for reading `.deps` files.</Description>
+    <VersionPrefix>2.0.0</VersionPrefix>
+    <TargetFrameworks>net451;netstandard1.3;netstandard1.6</TargetFrameworks>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <AssemblyName>Microsoft.Extensions.DependencyModel</AssemblyName>
```
eerhardt: Same as the other one, this and `PackageId` are defaulted.  This can be removed.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <AssemblyOriginatorKeyFile>../../../tools-local/setuptools/Key.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <PackageId>Microsoft.Extensions.DependencyModel</PackageId>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>git://github.com/dotnet/core-setup</RepositoryUrl>
+    <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'netstandard1.6' ">$(PackageTargetFallback);portable-net45+wp80+win8+wpa81+dnxcore50</PackageTargetFallback>
```
eerhardt: Can you do me a favor and try it without this line?  I know it was in the existing project.json, but I think this isn't necessary.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -0,0 +1,37 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
+    <Reference Include="System" />
```
eerhardt: Can be removed.  The SDK implicitly references this.

src/managed/dir.proj
eerhardt: This doesn't work because of the way we are running msbuild.  On Windows, we are using the desktop msbuild in VS 2015, which doesn't support the new "SDK" projects.  On non-Windows, we are using the msbuild that comes with build-tools, which also doesn't support the new "SDK' projects.

eerhardt:
>       ProjectJsonPaths="@(PackageProjects)">

Should rename `ProjectJsonPaths`

---
Refers to: src/managed/dir.proj:23 in bd5f1ce. [](commit_id = bd5f1ce852d98db0e8d34d9e7fa53387437f7f9f, deletion_comment = False)

run.sh
mellinoe: I added it to the Windows scripts, and to init-tools as well. There are multiple ways to build things (e.g. through the scripts or using MSBuild directly), so I wanted to make sure the first-time experience never ran, regardless of what command was used..

src/managed/Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj
mellinoe: Yeah, this is from `dotnet migrate`.

src/pkg/packaging/dir.proj
eerhardt: We should see if we can fix this.  It works in the CLI repo to "build" for netstandard, and then later come back and say `dotnet pack --no-build` on non-Windows.  I'm assuming there is more going on here, since this works in other repos.

src/pkg/packaging/dir.proj
```diff
@@ -160,18 +160,19 @@
...
+      <PackageProjects Condition="'$(OS)' == 'Windows'" Include="$(ProjectDir)src\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj" />
     </ItemGroup>
 
     <PropertyGroup>
-      <BuildBasePathArg>--build-base-path $(IntermediateOutputForPackaging)</BuildBasePathArg>
+      <BuildBasePathArg></BuildBasePathArg>
```
eerhardt: This can be removed.

src/test/Assets/TestProjects/LightupClient/LightupClient.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AssemblyName>LightupClient</AssemblyName>
```
eerhardt: Like before, AssemblyName and PackageId can be removed.  Repeated for the rest of the TestAssets projects.

src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -277,16 +277,23 @@ private void ValidateRequiredDirectories(RepoDirectoriesProvider repoDirectories
...
             var restoreArgs = new List<string>();
             foreach (var fallbackSource in fallbackSources)
             {
-                restoreArgs.Add("-f");
+                restoreArgs.Add("--source");
                 restoreArgs.Add(fallbackSource);
             }
             restoreArgs.Add("--disable-parallel");
 
+            restoreArgs.Add($"/p:MNAVersion={_repoDirectoriesProvider.MicrosoftNETCoreAppVersion}");
```
eerhardt: How about instead of making up a new property for this, instead just set the property that is already used for it - `$(RuntimeFrameworkVersion)`?

src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -310,15 +317,10 @@ public TestProjectFixture EnsureRestored(params string[] fallbackSources)
...
-            var sourceProjectJson = Path.Combine(_testProject.ProjectDirectory, "project.json.template");
-            var targetProjectJson = Path.Combine(_testProject.ProjectDirectory, "project.json");
-
-            // apply RID to template
-            File.WriteAllText(targetProjectJson, File.ReadAllText(sourceProjectJson).Replace("{RID}", rid));
-
             if ( ! _testProject.IsRestored())
             {
-                RestoreProject(fallbackSources);
+                string extraMSBuildProperties = $"/p:TestTargetRid={rid}";
```
eerhardt: Same thing here as the `RuntimeFrameworkVersion`.  Let's just use the regular property - `RuntimeIdentifier`.  There's no need to give a different name to this property.  It just adds confusion for new people coming in.

src/test/Assets/TestProjects/PortableTestApp/PortableTestApp.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AssemblyName>PortableTestApp</AssemblyName>
+    <PackageId>PortableTestApp</PackageId>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <PackageTargetFallback>$(PackageTargetFallback);dotnet5.4;portable-net451+win8</PackageTargetFallback>
```
eerhardt: Do we know why this `PackageTargetFallback` is necessary?  I don't see it on others

src/test/Assets/TestProjects/StandaloneApp/StandaloneApp.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AssemblyName>StandaloneApp</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <PackageId>StandaloneApp</PackageId>
+    <RuntimeIdentifiers>$(TestTargetRid)</RuntimeIdentifiers>
```
eerhardt: If you are setting `RuntimeIdentifier`, you don't need to set `RuntimeIdentifiers`.  Just like TargetFrameworks, the plural and the singular shouldn't be used at the same time.

src/test/Assets/TestProjects/StandaloneTestApp/StandaloneTestApp.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AssemblyName>StandaloneTestApp</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <PackageId>StandaloneTestApp</PackageId>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <PackageTargetFallback>$(PackageTargetFallback);dotnet5.4;portable-net451+win8</PackageTargetFallback>
+    <RuntimeIdentifiers>$(TestTargetRid)</RuntimeIdentifiers>
```
eerhardt: Same as before - only need to set `RuntimeIdentifier`.

src/test/bootstrap/.cliversion
```diff
@@ -1 +1 @@
-1.0.0-preview3-003886
+2.0.0-preview1-005899
```
eerhardt: I thought we can get rid of this "boostrap" thing now?  Is this a follow up item?

src/test/build/shared-build-targets-utils/Utils/DotNetCli.cs
```diff
@@ -96,7 +96,7 @@ private void ComputeSharedFxPaths()
             GreatestVersionSharedFxPath = sharedFxVersionDirectories
                 .OrderByDescending(p => p.ToLower())
                 .First();
-
```
eerhardt: Looks like an unnecessary change.  Please revert.

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -50,6 +50,12 @@ public void Running_Build_Output_Standalone_EXE_with_DepsJson_and_RuntimeConfig_
...
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                Command.Create("chmod", "+x", appExe).Execute().EnsureSuccessful();
```
eerhardt: I think just `chmod u+x` will suffice.  The whole world won't need to execute this file.  Just the owner.

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
eerhardt: Can you add a link to [https://github.com/NuGet/Home/issues/4424](https://github.com/NuGet/Home/issues/4424)?  That way we know what issue it is tracking.

---
In reply to: [114636019](https://github.com/dotnet/core-setup/pull/2259#discussion_r114636019) [](ancestors = 114636019)

src/test/HostActivationTests/HostActivationTests.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <AssemblyName>HostActivationTests</AssemblyName>
+    <PackageId>HostActivationTests</PackageId>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
```
eerhardt: Why is this necessary?  I don't think it should be.

eerhardt: A blank test project created by 'dotnet new' looks like:

```

  <PropertyGroup>
    <TargetFramework>netcoreapp1.1</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
    <PackageReference Include="xunit" Version="2.2.0" />
    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
  </ItemGroup>

</Project>
```

---
In reply to: [114668486](https://github.com/dotnet/core-setup/pull/2259#discussion_r114668486) [](ancestors = 114668486)

src/test/Microsoft.Extensions.DependencyModel.Tests/Microsoft.Extensions.DependencyModel.Tests.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <PackageReference Include="Microsoft.DotNet.InternalAbstractions" Version="1.0.0" />
+    <PackageReference Include="FluentAssertions" Version="4.0.0" />
+    <PackageReference Include="moq.netcore" Version="4.4.0-beta8" />
+    <PackageReference Include="xunit" Version="2.2.0-beta5-build3474" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="NuGet.Versioning" Version="4.0.0" />
+    <ProjectReference Include="..\..\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
+    <ProjectReference Include="..\..\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj" />
+
+    <!-- TODO: Should this dependency be removed? -->
```
eerhardt: This will need a major re-working of these tests.  If you feel this should be removed, please log an issue for it.  Else, just take the TODO out, as I don't think anyone is going to do this work.

src/test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
eerhardt: my vote would be to keep them and assert that the result doesn't contain it.  Also adding a comment for why.

---
In reply to: [114636097](https://github.com/dotnet/core-setup/pull/2259#discussion_r114636097) [](ancestors = 114636097)

src/test/TestUtils/TestProject.cs
```diff
@@ -50,8 +50,8 @@ public class TestProject
             _sharedLibraryExtension = sharedLibraryExtension;
             _sharedLibraryPrefix = sharedLibraryPrefix;
             _projectName = Path.GetFileName(_projectDirectory);
-            _projectJson = Path.Combine(_projectDirectory, "project.json");
-            _projectLockJson = Path.Combine(_projectDirectory, "project.lock.json");
+            _projectJson = Path.Combine(_projectDirectory, $"{_projectName}.csproj");
```
eerhardt: Can we rename `_projectJson` and `_projectLockJson`?

src/test/TestUtils/TestUtils.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
eerhardt: I'm surprised this needs unsafe code...

netci.groovy
```diff
@@ -84,7 +84,7 @@ platformList.each { platform ->
     Utilities.standardJobSetup(newJob, project, isPR, "*/${branch}")
 
     if (!(os == 'Windows_NT' && architecture == 'arm')) {
-        Utilities.addXUnitDotNETResults(newJob, '**/*-testResults.xml')
+        Utilities.addMSTestResults(newJob, '**/*-testResults.xml')
```
eerhardt: should be `Utilities.addMSTestResults(newJob, '**/*.trx')`

tools-local/tasks/GetBuildableFrameworks.cs
```diff
@@ -31,21 +32,20 @@ public override bool Execute()
...
-                    var projectJsonReader = new JsonTextReader(projectFileReader);
-                    var serializer = new JsonSerializer();
-                    var project = serializer.Deserialize<JObject>(projectJsonReader);
-                    var dir = Path.GetDirectoryName(projectJsonPath.ItemSpec);
-                    var frameworksSection = project.Value<JObject>("frameworks");
-                    foreach (var framework in frameworksSection.Properties())
+                    string[] tfms = match.Groups[1].Value.Split(';');
+                    foreach (string framework in tfms)
                     {
+                        Console.WriteLine("OSGroup: " + OSGroup);
```
eerhardt: Instead of Console.Writeline, how about using the MSBuild logger?

tools-local/tasks/GetBuildableFrameworks.cs
eerhardt: This variable doesn't appear to be used.

tools-local/tasks/GetBuildableFrameworks.cs
```diff
@@ -31,21 +32,20 @@ public override bool Execute()
...
                         if (OSGroup == "Windows_NT"
-                            || framework.Name.StartsWith("netstandard")
-                            || framework.Name.StartsWith("netcoreapp"))
+                            || framework.StartsWith("netstandard")
+                            || framework.StartsWith("netcoreapp"))
                         {
-                            args.Add($"--framework {framework.Name} {dir}");
+                            args.Add($"--framework {framework} {projectJsonPath}");
```
eerhardt: `projectJsonPath` should be renamed.

src/test/TestUtils/TestProjectFixture.cs
mellinoe: I have a slight preference for this because it doesn't force the project to define `RuntimeFrameworkVersion` -- it can be opt-in based on this property.

src/test/TestUtils/TestProjectFixture.cs
mellinoe: As above, I have a preference for this method. Most of the test projects do not define `RuntimeIdentifiers` or `RuntimeIdentifier`. Those that do can define it to be based on this auxiliary property.

tools-local/tasks/GetBuildableFrameworks.cs
mellinoe: This was just debug code; removing.

src/test/bootstrap/.cliversion
mellinoe: This will have to be a follow-up item.

src/test/Microsoft.Extensions.DependencyModel.Tests/Microsoft.Extensions.DependencyModel.Tests.csproj
mellinoe: Removing the TODO.

src/test/HostActivationTests/HostActivationTests.csproj
mellinoe: This is just what dotnet migrate spits out. Ditto for a lot of other comments. I'll try to clean up these things where I see them.

src/test/TestUtils/TestProjectFixture.cs
eerhardt: I don't get it.  The project doesn't need to define `RuntimeFrameworkVersion` at all with my proposal.

When you are saying "a slight preference for this", does "this" mean what I am proposing?  or what the code is doing now?

src/test/Microsoft.Extensions.DependencyModel.Tests/PackageResolverTest.cs
mellinoe: Inverting the assertion sounds like a good idea for now. I'll do that.

src/test/TestUtils/TestProjectFixture.cs
mellinoe: There's no (effective) difference in this case, so I'll just pass in RuntimeFrameworkVersion here. RuntimeFrameworkVersion is defined in all of the test projects,.

mellinoe: Actually, I'm going to keep it as-is (unless you still disagree). I'll try to explain the rationale more clearly:

1) These projects are batch-restored on the command line, and it would be nice to pass the same parameters to them all
2) The two "real" test projects (HostActivationTests and DependencyModel) do not want this property to be set, while the rest of the test assets do
3) If we pass it in as a global property, we can't "undefine" it in the two projects that we don't want it in.

mellinoe:
@dotnet-bot Test Windows_NT x64 Release Build

mellinoe:
I think this is at an okay state to go in now. All of the CI "failures" are because of the mismatched test results files, and shouldn't show up after the update to netci.groovy gets consumed.

PR #2271, opened 2017-05-03
Add CoreLib to TPA explicitly

gkhanna79:
With https://github.com/dotnet/coreclr/pull/11359, CoreLib is always next to CoreCLR.dll, which is under runtime-native assets category.

Since runtime-native assets are not on TPA, but we need CoreLib to be on TPA (for folks who maybe consuming TPA explicitly for their own use - even if they should not do so), we explicitly add CoreLib to be on TPA.

@jkotas PTAL


gkhanna79:
Something messed up in the commit list above. Closing this PR.

PR #2272, opened 2017-05-03
Add CoreLib explicitly to TPA list

gkhanna79:
With dotnet/coreclr#11359, CoreLib is always next to CoreCLR.dll, which is under runtime-native assets category.

Since runtime-native assets are not on TPA, but we need CoreLib to be on TPA (for folks who maybe consuming TPA explicitly for their own use - even if they should not do so), we explicitly add CoreLib to be on TPA.

@jkotas PTAL


gkhanna79:
@dotnet-bot test PortableLinux x64 Release Build please

Issue #2198, opened 2017-04-26
Infer-runtimes cleanup


chcosta:
There was some work to remove "infer-runtimes" from core-setup builds (it was re-enabled), but that work has likely regressed further after the buildtools change.  

https://github.com/dotnet/core-setup/pull/2059

ellismg:
Since I caused the debt, I'll try to clean this up.

eerhardt:
We are completely off any project.json files now in the repo.  Thus there are no `--infer-runtimes` usages since that is not a valid option with the MSBuild CLI commands.

Closing.

PR #2275, opened 2017-05-04
Move shared framework publish to MSBuild and off project.json

eerhardt:
@chcosta @mellinoe @weshaggard @ericstj


tools-local/tasks/ProcessSharedFrameworkDeps.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            using (var depsStream = File.Create(DepsFilePath))
+            {
+                new DependencyContextWriter().Write(context, depsStream);
+            }
+
+            return true;
+        }
+
+        private static string[] ToStringArray(ITaskItem[] items)
```
ellismg: I think you can just change the types of the parameters to the task to `string[]` instead of `ITaskItem[]` and MSBuild will do the conversion for you.  Doesn't look like there's a need to be able to access any Metadata on the items, so this should be safe.

ellismg:
This will also address #2198

src/sharedFramework/sharedFramework.proj
```diff
@@ -97,37 +100,25 @@
...
     </PropertyGroup>
 
     <ItemGroup>
       <TrimPkgsFromDeps Include="runtime.$(PackageTargetRid).microsoft.netcore.dotnetapphost" />
       <TrimPkgsFromDeps Include="runtime.$(PackageTargetRid).microsoft.netcore.dotnethostresolver" />
       <TrimPkgsFromDeps Include="microsoft.netcore.dotnetapphost" />
       <TrimPkgsFromDeps Include="microsoft.netcore.dotnethostresolver" />
-      <TrimPkgsWithCmdPrefix Include="@(TrimPkgsFromDeps ->'--remove %(Identity)')" />
+    
+      <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='Windows_NT'" Include="win" />
```
ellismg: Should this be a property instead?  Do we really want to allow multiple values?  I understand that DepsProcessor would allow you to pass multiple runtimes but we never really did that before.

tools-local/tasks/RuntimeGraphManager.cs
```diff
@@ -1,29 +1,35 @@
...
 
-namespace Microsoft.DotNet.ProjectModel
+namespace Microsoft.DotNet.Build.Tasks
 {
     public class RuntimeGraphManager
     {
         private const string RuntimeJsonFileName = "runtime.json";
 
-        public NuGet.RuntimeModel.RuntimeGraph Collect(IEnumerable<LibraryExport> exports)
+        public RuntimeGraph Collect(LockFile lockFile)
```
ellismg: Thoughts on making these helper classes internal?

build.proj
```diff
@@ -39,9 +39,15 @@
       <MSBuild Projects="tools-local/Microsoft.DotNet.Build.Tasks.Local/Microsoft.DotNet.Build.Tasks.Local.builds" />
       <MSBuild Projects="tools-local/tasks/core-setup.tasks.builds" />
 
+    <PropertyGroup>
+      <DependencyModelTFM>netstandard1.3</DependencyModelTFM>
+      <DependencyModelTFM Condition="'$(BuildToolsTargets45)' == 'true' OR '$(MSBuildRuntimeType)' != 'Core'">net451</DependencyModelTFM>
```
weshaggard: We have generally been moving away from BuildToolsTargets45 and instead just using the MSBuildRuntimeType check.

src/sharedFramework/sharedFramework.proj
```diff
@@ -97,37 +100,25 @@
 
   <Target Name="GenerateRuntimeGraph">
     <PropertyGroup>
-      <runtimeGraphGeneratorName>DepsProcessor</runtimeGraphGeneratorName>
-      <runtimeGraphGeneratorProject>$(ProjectDir)tools-local/setuptools/independent/$(runtimeGraphGeneratorName)</runtimeGraphGeneratorProject>
-      <runtimeGraphGeneratorOutput>$(IntermediateOutputRootPath)setuptools/independent/$(runtimeGraphGeneratorName)</runtimeGraphGeneratorOutput>
-      <runtimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='Windows_NT'">win</runtimeGraphGeneratorRuntime>
-      <runtimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='OSX'">osx</runtimeGraphGeneratorRuntime>
-      <runtimeGraphGeneratorRuntime Condition="'$(runtimeGraphGeneratorRuntime)'==''">linux</runtimeGraphGeneratorRuntime>      
+      <SharedFrameworkAssetsFile>$(SharedFrameworkSourceRoot)/obj/project.assets.json</SharedFrameworkAssetsFile>
```
weshaggard: Ideally the obj folder would be under the bin directory instead of in the source tree.

tools-local/tasks/project.json
```diff
@@ -22,7 +24,7 @@
         "portable-net45+win8+wpa81"
       ]
     },
-    "net45": {
+    "net451": {
```
mellinoe: Should this reference `Microsoft.TargetingPack.NETFramework.v4.5.1`? It doesn't really matter given that it is building as-is, but it would be more consistent.

src/sharedFramework/sharedFramework.proj
eerhardt: Yes, we only use a single value.  If we ever need multiple values, we can change it to support multiple.  I'm switching this to a property.

tools-local/tasks/RuntimeGraphManager.cs
eerhardt: Good catch.  done.

tools-local/tasks/ProcessSharedFrameworkDeps.cs
eerhardt: Done.

build.proj
eerhardt: Fixed.  And I fixed the other 2 places in our code with this check.

tools-local/tasks/project.json
eerhardt: Changed.  This was also wrong in `tools-local\Microsoft.DotNet.Build.Tasks.Local\project.json`.

src/sharedFramework/sharedFramework.proj
eerhardt: Agreed it ideally would. There are other places with this same issue already.  We should address this holistically.  https://github.com/dotnet/core-setup/issues/2322

PR #2179, opened 2017-04-26
Update Branding in Mac and Windows Runtime installers

rakeshsinghranchi:
Fix for Issue : https://github.com/dotnet/core-setup/issues/2155

PR #2276, opened 2017-05-04
Update branding for Installer packages

rakeshsinghranchi:
Update the branding in installers to reflect Release Branding :

https://github.com/dotnet/core-setup/pull/2179


src/pkg/packaging/dir.props
```diff
@@ -33,8 +33,11 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <SharedHostBrandName>'Microsoft .NET Core $(ProductionVersion) - Host'</SharedHostBrandName>
-    <HostFxrBrandName>'Microsoft .NET Core $(ProductionVersion) - Host FX Resolver'</HostFxrBrandName>
-    <SharedFrameworkBrandName>'Microsoft .NET Core $(ProductionVersion) - Runtime'</SharedFrameworkBrandName>
+    <SharedHostBrandName>'Microsoft .NET Core $(ProductionVersion) Host'</SharedHostBrandName>
+    <SharedHostBrandName Condition="'$(ReleaseBrandSuffix)'!=''">'Microsoft .NET Core $(ProductionVersion) $(ReleaseBrandSuffix) Host'</SharedHostBrandName>
```
gkhanna79: Instead of repeating this, can you please define a new variable that carries either the value of ProductionVersion or both ProductionVersion and ReleaseBrandSuffix and use that variable in all the locations. This will also make it simpler to modify the value in a single place.

rakeshsinghranchi:
@gkhanna79  PTAL

PR #2286, opened 2017-05-04
Fix combinedinstallerfile path

chcosta:
https://github.com/dotnet/core-setup/commit/77179251dd0c4954ff812e30e645c26cbb2a4506 made the combinedinstallerfile and combinedinstallerengine properties fully rooted paths.  Removing the `PackagesOutDir` paths from those property calls to insignia for engine extract / attach to align with that change.


eerhardt:
I thought I got all these, but I guess I didn't.  Thanks for fixing.

Issue #1871, opened 2017-03-28
Remove version info from apphost.exe


ramarag:
cc @gkhanna79 @eerhardt

gkhanna79:
We should also add an explicit resource (on Windows) that carries the original filename and explicitly set major/minor versions that can be updated when the shared host (i.e. dotnet.exe) is updated.

gkhanna79:
@ramarag Are we using this as a tracker inspite of your comment at https://github.com/dotnet/core-setup/pull/2090#issuecomment-297218769?

PR #2287, opened 2017-05-05
remove version info from apphost.exe

ramarag:
fixes https://github.com/dotnet/core-setup/issues/1871

@gkhanna79 PTAL

### DIFF
--- a/src/corehost/build.proj
+++ b/src/corehost/build.proj
@@ -36,7 +36,7 @@
 
     <!-- Generate Version files -->
     <ItemGroup>
-      <HostFiles Include="apphost;dotnet;hostfxr;hostpolicy" />
+      <HostFiles Include="dotnet;hostfxr;hostpolicy" />
     </ItemGroup>
     <MSBuild Projects="$(MSBuildProjectFullPath)"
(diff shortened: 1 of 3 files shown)


gkhanna79:
Overall, looks fine. What does the Details tab look like after this change?

ramarag:
oops forgot to add
image

gkhanna79:
LGTM.

@leecow @morganbr Are you fine with this as well (since apphost.exe is supposed to be used for standalone scenarios representing external apps)?

morganbr:
Does publishing the app fill in any of those details? I agree that they shouldn't say this is apphost.exe by Microsoft, but it should ideally match the app in some way.

ramarag:
The default[or any] publishing mechanism in cli does not populate any of the above fields
@dsplaisted for confirmation

gkhanna79:
No, there is no CLI support for it. It can be looked into for the next release, if deemed necessary. FWIW, it never did in the past as well (when dotnet.exe was used as the host).

morganbr:
Doing it in the future sounds fine. It just occurs to me that this doesn't really leave a way (other than postprocessing) for apps to differentiate themselves. That would be more of a problem if administrators want to block a particular standalone .NET Core app without blocking all of them.

gkhanna79:
> this doesn't really leave a way (other than postprocessing) for apps to differentiate themselves. 

While I agree, this has not changed from prior releases. @ramarag Can you please file a CLI (or SDK) future issue to get this information added?

leecow:
I'm fine with the change and looking into enabling support in in the CLI.

PR #2290, opened 2017-05-05
Fix publish build break, remove static creation of debinstallerfile

chcosta:
/cc @eerhardt

### DIFF
--- a/publish/dir.props
+++ b/publish/dir.props
@@ -8,7 +8,6 @@
   </PropertyGroup>
     
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


chcosta:
Note that none of the debian publishing code path has been tested, so this is all new territory.  We got past the debian build step though so i'm hopeful that will be enough to unblock other work, including testing CLI ingestion.  I'll keep an eye on the next build after merging this.

eerhardt:
Looks good. Thanks for fixing this.

PR #2292, opened 2017-05-05
Add publish step to build definition

karajas:
Adding a separate publish step to the builds which will finalize and publish packages only after all the legs have succeeded and update the versions repo.

/cc @dagood @chcosta @eerhardt @weshaggard


karajas:
Some of the variables may be in correct, want to send this out to get feedback.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,413 @@
...
+      "task": {
+        "id": "c6c4c611-aa2e-4a33-b606-5eaba2196824",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "solution": "$(PB_SourcesDirectory)\\publish\\publish.proj",
+        "platform": "$(PB_TargetArchitecture)",
+        "configuration": "$(BuildConfiguration)",
+        "msbuildArguments": "/t:PublishFinalOutput /p:Configuration=$(BuildConfiguration) $(PB_CommonMSBuildArgs) /p:NuGetFeedUrl=$(NUGET_FEED_URL) /p:NuGetSymbolsFeedUrl=$(NUGET_SYMBOLS_FEED_URL) /p:NuGetApiKey=$(NUGET_API_KEY) /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:ContainerName=$(STORAGE_CONTAINER) /p:GitHubUser=$(GITHUB_USER) /p:GitHubEmail=$(GITHUB_EMAIL) /p:GitHubAuthToken=$(GITHUB_AUTH_TOKEN) /p:VersionsRepoOwner=$(PB_VersionsRepoOwner) /p:VersionsRepo=$(PB_VersionsRepo) /p:VersionsRepoPath=build-info/dotnet/$(PB_VersionsRepo)/$(PB_Branch) /p:ShippedNuGetPackageGlobPath=$(PB_SourcesDirectory)\\Bin\\publishpackages\\$(PB_FilesToPublish)",
```
dagood: `/p:VersionsRepo=$(PB_VersionsRepo)` should have `PB_VersionsRepo` being `versions` (this is meant to specify the name of the versions repo), but it's `core-setup` below.

Then `build-info/dotnet/$(PB_VersionsRepo)/$(PB_Branch)` needs to use some other variable corresponding to the Core-Setup repo name, since it corresponds to https://github.com/dotnet/versions/tree/master/build-info/dotnet/core-setup/master.

eerhardt:
@karajas - please hold off on merging this change until we can get the current builds "green".

publish/publish.proj
```diff
@@ -6,9 +6,22 @@
...
   <Target Name="Build"
-          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;FinalizeBuildInAzure;PublishCoreHostPackagesToFeed" />
+          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;" />
+  
+  <Import Project="$(ToolsDir)VersionTools.targets" Condition="Exists('$(ToolsDir)VersionTools.targets')" />
+  <Target Name="PublishFinalOutput"
+          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+    <RemoveDir Directories="$(BinDir)/publishpackages" />
```
eerhardt: I don't like how we have `$(BinDir)/publishpackages` duped between here and in the Core-Setup-Publish.json` file.  Can we refactor so this folder is only declared in a single spot?

publish/publish.proj
```diff
@@ -6,9 +6,22 @@
...
+  <Target Name="PublishFinalOutput"
+          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+    <RemoveDir Directories="$(BinDir)/publishpackages" />  
+    <ItemGroup>
+      <PackagesToShip Include="$(BinDir)/ForPublishing/*.nupkg" Exclude="$(BinDir)/ForPublishing/*.symbols.nupkg" />
+      <PackagesToShip Remove="%(PackagesToShip.Identity)" Condition="$([System.String]::Copy('%(PackagesToShip.Identity)').Contains('latest'))" />
+    </ItemGroup>
+    <Copy SourceFiles="@(PackagesToShip)"
```
eerhardt: >  [](start = 9, length = 1)

Instead of doing all this, why don't we just set the `@(ShippedNuGetPackage)` item to what we want the UpdatePublishedVersions target to use?

publish/publish.proj
```diff
@@ -6,9 +6,22 @@
...
+          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;" />
+  
+  <Import Project="$(ToolsDir)VersionTools.targets" Condition="Exists('$(ToolsDir)VersionTools.targets')" />
+  <Target Name="PublishFinalOutput"
+          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+    <RemoveDir Directories="$(BinDir)/publishpackages" />  
+    <ItemGroup>
+      <PackagesToShip Include="$(BinDir)/ForPublishing/*.nupkg" Exclude="$(BinDir)/ForPublishing/*.symbols.nupkg" />
```
eerhardt: Can we refactor so the `ForPublishing` directory is only defined in a single spot?  I see it 3 times in this file.

publish/publish.proj
```diff
@@ -6,9 +6,22 @@
   <UsingTask TaskName="GetAzureBlobList" AssemblyFile="$(LocalBuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.Local.dll" />
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(LocalBuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.Local.dll" />
 
-  <!-- ToDo: Update versions repo -->
   <Target Name="Build"
-          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;FinalizeBuildInAzure;PublishCoreHostPackagesToFeed" />
+          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;" />
+  
+  <Import Project="$(ToolsDir)VersionTools.targets" Condition="Exists('$(ToolsDir)VersionTools.targets')" />
```
eerhardt: It feels really random to just `Import` a targets file in the middle of the build file.  Can this go at the top or bottom of the document?

buildpipeline/pipeline.json
```diff
@@ -136,6 +136,33 @@
...
+        "TreatWarningsAsErrors": "false"
+      },
+      "BuildParameters": {
+        "BuildConfiguration": "Release"
+      },
+      "Definitions": [
+        {
+          "Name": "Core-Setup-Publish",
+          "Parameters": {
+            "PB_DistroRid": "win7-x86",
```
eerhardt: There's some usages of `x86` here and some `x64`.  I think they should be consistent in this build def.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,413 @@
...
+    "STORAGE_CONTAINER": {
+      "value": "core-setup"
+    },
+    "CONNECTION_STRING": {
+      "value": "PassedViaPipeBuild"
+    },
+    "PUBLISH_TO_AZURE_BLOB": {
+      "value": "true"
+    },
+    "NUGET_FEED_URL": {
```
eerhardt: We can probably remove a bunch of these in the other builds defs now.  The "finalize build"-only variables should no longer be needed in the individual legs.  Only this leg will need them.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,413 @@
...
+    "BuildConfiguration": {
+      "value": "Release",
+      "allowOverride": true
+    },
+    "COREHOST_TRACE": {
+      "value": "0",
+      "allowOverride": true
+    },
+    "STORAGE_ACCOUNT": {
+      "value": "karajascli"
```
rakeshsinghranchi: personal account

eerhardt:
Removing "no merge" now that we have a green build. Feel free to merge when approved.

buildpipeline/pipeline.json
weshaggard: Why do we even need a rid for the publish step?

publish/publish.proj
weshaggard: Why do we need to copy the packages around? We should try and avoid all the unnecessary copying and just get them into the right location from the beginning and publish them from there.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,413 @@
...
+        "modifyEnvironment": "false",
+        "workingFolder": "$(PB_SourcesDirectory)",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Initialize tooling",
```
weshaggard: Why is this "Initialize tooling"? Isn't that the previous step?

weshaggard: Nevermind I see you are only calling the Sync and build custom tasks targets.

publish/publish.proj
karajas: I had added this because from the drop I tested it seemed to bring down both the version generated and a latest set of packages. So I needed a way to isolate only the version we last built to publish.

karajas: I had to seperate them out into a separate ItemGroup so that the ShippedNuGetPackageGlobPath picked the right set of packages

buildpipeline/pipeline.json
karajas: We don't I'll remove this.

buildpipeline/Core-Setup-Publish.json
karajas: Yup I'll remove the extra publish specific variables.

buildpipeline/pipeline.json
```diff
@@ -136,6 +136,33 @@
...
+      },
+      "BuildParameters": {
+        "BuildConfiguration": "Release"
+      },
+      "Definitions": [
+        {
+          "Name": "Core-Setup-Publish",
+          "Parameters": {
+            "PB_TargetArchitecture": "x64",
+            "PB_PublishLocation": "Bin\\publishpackages",
```
karajas: will remove

eerhardt:
Can we get rid of the "race condition" checking now as well?

See https://github.com/dotnet/core-setup/blob/e2876222982f409678ebbf53041557eff092c9a5/tools-local/tasks/FinalizeBuild.cs#L57

eerhardt:
> Can we get rid of the "race condition" checking now as well?

Maybe not, if 2 separate builds are finalizing at the same time....?

karajas:
@eerhardt But if two builds are finalizing at the same time, shouldn't they be using entirely different relative blob paths?

karajas:
Also for the debian packages, is there supposed to be a separate feed where they are published? I see variables like CLI_NUGET_FEED_URL used to be used in addition to NUGET_FEED_URL.

eerhardt:
It's the "Latest" folder and the `dnvm/sharedfx.version` files that are the problem.

Each leg uploads its individual assets to the current "version" path in azure storage.  Then the "finalize" leg comes through and moves them all into the "Latest" folder, so we can have constant links to the "latest" installers (like in the repo's README).

eerhardt:
> Also for the debian packages, is there supposed to be a separate feed where they are published?

Yes, see https://github.com/dotnet/core-setup/blob/282c8d3e278b675410e249eda17a5fade5e9e391/publish/publish.proj#L41-L45 for where the deb packages should be published.

And the "dotnet-deb-tool" should be published to the `cli-deps` feed.  See this preview1 code for how it worked:  https://github.com/dotnet/core-setup/blob/release/2.0.0/build_projects/dotnet-host-build/PublishTargets.cs#L70-L86

karajas:
So is there a chance of 2 official builds running in parallel?

eerhardt:
> So is there a chance of 2 official builds running in parallel?

It's unlikely, but it is possible for 2 people to kick off a build at the same time.

rakeshsinghranchi:
@karajas 
CLI_NUGET_FEED is used for publishing DotnetDebToolPackage to CLI FEED for their consumption.

karajas:
>And the "dotnet-deb-tool" should be published to the cli-deps feed. See this preview1 code for how it worked: https://github.com/dotnet/core-setup/blob/release/2.0.0/build_projects/dotnet-host-build/PublishTargets.cs#L70-L86

I don't think we do that currently, will add this back.

eerhardt:
> I don't think we do that currently, will add this back.

You don't need to block this PR on it though.

publish/publish.proj
eerhardt: You don't need to rely on `$(ShippedNuGetPackageGlobPath)`.  You can just set `@(ShippedNuGetPackage)` directly, and it will work fine.

See https://github.com/dotnet/buildtools/blob/8c000f1c9d79d4d3ba711585a7212619dd86749d/src/Microsoft.DotNet.Build.Tasks/PackageFiles/VersionTools.targets#L20-L25

If you set the globpath, build-tools will use it to populate `@(ShippedNuGetPackage)`, and then pass that item into the `UpdatePublishedVersions` task.  I'm saying, skip the glob path all together, as it doesn't bring us any value.  Instead just directly set `@(ShippedNuGetPackage)` yourself, and that will get passed into the `UpdatePublishedVersions` task

/cc @dagood

karajas: Yeah think I had updated this to do that in the latest iteration.

karajas:
Rebased on latest. Can you PTAL?

buildpipeline/Core-Setup-CrossBuild.json
```diff
@@ -125,7 +125,7 @@
       },
       "inputs": {
         "scriptPath": "$(PB_RepoName)/scripts/docker-as-current-user.sh",
-        "args": "run --name $(PB_DockerContainerName) --rm --privileged -v $(System.DefaultWorkingDirectory)/$(PB_RepoName):$(PB_GitDirectory) -w $(PB_GitDirectory) -e HOME -e ROOTFS_DIR -e PUBLISH_TO_AZURE_BLOB -e CONNECTION_STRING -e CLI_NUGET_FEED_URL -e CLI_NUGET_SYMBOLS_FEED_URL -e CLI_NUGET_API_KEY -e NUGET_FEED_URL -e NUGET_SYMBOLS_FEED_URL -e NUGET_API_KEY -e GITHUB_PASSWORD $(PB_DockerImageName) ./build.sh $(CommonArguments) $(BuildArguments) $(PB_PortableBuild) --env-vars \"DISABLE_CROSSGEN=1,TARGETPLATFORM=$(PB_Architecture),TARGETRID=$(TargetRid),CROSS=1\"",
```
chcosta: This definition isn't used.  you can delete it.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,376 @@
...
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "inlineScript",
+        "scriptName": "",
+        "arguments": "-path $(PB_SourcesDirectory)",
+        "workingFolder": "",
+        "inlineScript": "param($path)\n\nif (Test-Path $path){\n    # this will print out an error each time a file can't be deleted.\n    Remove-Item -Recurse -Force $path\n }",
```
chcosta: You may hit an issue with long paths and need to do something like https://devdiv.visualstudio.com/DevDiv/_Build/explorer?_a=simple-process&definitionId=6256 does

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,376 @@
...
+        "platform": "$(PB_TargetArchitecture)",
+        "configuration": "$(BuildConfiguration)",
+        "msbuildArguments": "/t:BatchRestorePackages;BuildCustomTasks",
+        "clean": "false",
+        "maximumCpuCount": "false",
+        "restoreNugetPackages": "false",
+        "logProjectEvents": "false",
+        "createLogFile": "false",
+        "msbuildLocationMethod": "version",
+        "msbuildVersion": "15.0",
```
chcosta: set msbuildversion to `Latest` and `x64` for all of these so we don't hit warnings during the build about not being able to find "v15.0"

buildpipeline/Core-Setup-Publish.json
```diff
@@ -0,0 +1,376 @@
...
+      "task": {
+        "id": "c6c4c611-aa2e-4a33-b606-5eaba2196824",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "solution": "$(PB_SourcesDirectory)\\publish\\publish.proj",
+        "platform": "$(PB_TargetArchitecture)",
+        "configuration": "$(BuildConfiguration)",
+        "msbuildArguments": "/t:PublishFinalOutput /p:Configuration=$(BuildConfiguration) $(PB_CommonMSBuildArgs) /p:NuGetFeedUrl=$(NUGET_FEED_URL) /p:NuGetSymbolsFeedUrl=$(NUGET_SYMBOLS_FEED_URL) /p:NuGetApiKey=$(NUGET_API_KEY) /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:ContainerName=$(STORAGE_CONTAINER) /p:GitHubUser=$(GITHUB_USER) /p:GitHubEmail=$(GITHUB_EMAIL) /p:GitHubAuthToken=$(GITHUB_AUTH_TOKEN) /p:VersionsRepoOwner=$(PB_VersionsRepoOwner) /p:VersionsRepo=$(PB_VersionsRepo) /p:VersionsRepoPath=build-info/dotnet/$(PB_RepoName)/$(PB_Branch)",
```
chcosta: It'd be nice if, instead of `/t:PublishFinalOutput`, it was a property that was specified like `/p:Finalize=true`

chcosta: the output from the publish step isn't super descriptive, adding a `/flp:v=detailed;LogFile=$(PB_SourcesDirectory)\publish.log` and ensuring that gets propagated to VSTS would help diagnose issues.

buildpipeline/Core-Setup-Signing-Windows-BT.json
```diff
@@ -441,7 +441,7 @@
         "solution": "$(PB_SourcesDirectory)\\publish\\publish.proj",
         "platform": "$(PB_TargetArchitecture)",
         "configuration": "$(BuildConfiguration)",
-        "msbuildArguments": "$(PB_CommonMSBuildArgs)  /p:NuGetFeedUrl=$(NUGET_FEED_URL) /p:NuGetSymbolsFeedUrl=$(NUGET_SYMBOLS_FEED_URL) /p:NuGetApiKey=$(NUGET_API_KEY) /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:PublishRidAgnosticPackages=$(PB_PublishRidAgnosticPackages) /p:BuildFullPlatformManifest=$(PB_BuildFullPlatformManifest) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish.log",
+        "msbuildArguments": "$(PB_CommonMSBuildArgs)  /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:PublishRidAgnosticPackages=$(PB_PublishRidAgnosticPackages) /p:BuildFullPlatformManifest=$(PB_BuildFullPlatformManifest) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish.log",
```
chcosta: nit: I probably caused it, but there's an extra `\` in `/flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish.log"`

buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -89,7 +89,7 @@
         "solution": "$(PB_SourcesDirectory)\\publish\\publish.proj",
         "platform": "$(PB_TargetArchitecture)",
         "configuration": "$(BuildConfiguration)",
-        "msbuildArguments": "$(PB_BuildArguments) /p:NuGetFeedUrl=$(NUGET_FEED_URL) /p:NuGetSymbolsFeedUrl=$(NUGET_SYMBOLS_FEED_URL) /p:NuGetApiKey=$(NUGET_API_KEY) /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken)",
+        "msbuildArguments": "/p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:DistroRid=$(PB_DistroRid) /p:TargetArchitecture=$(PB_TargetArchitecture) /p:ConfigurationGroup=$(BuildConfiguration)",
```
chcosta: You shouldn't need to pass `/p:DistroRid=$(PB_DistroRid)`,  I think  you'll see when you rebase, that property has been removed.

publish/dir.targets
```diff
@@ -62,8 +62,10 @@
     <ItemGroup>
       <_FoundBlobNames Include="%(_BlobNames.FileName)%(_BlobNames.Extension)" Condition="'%(_BlobNames.Extension)' == '.svg'" />
       <_MissingBlobNames Include="@(PublishRid->'sharedfx_%(Identity)_$(ConfigurationGroup)_version_badge.svg')" 
-                          Exclude="@(_FoundBlobNames)" />
+                         Exclude="@(_FoundBlobNames)" />
     </ItemGroup>
+    <Error Text="Missing publish output from all legs of the build. Missing blobs are '@(_MissingBlobNames)'"
```
chcosta: This is a duplicate of the message in publish.proj.  Having it here is fine, but make it `Importance=High`, and remove it from publish.proj

publish/publish.proj
```diff
@@ -6,9 +6,20 @@
...
-  <!-- ToDo: Update versions repo -->
   <Target Name="Build"
-          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;FinalizeBuildInAzure;PublishCoreHostPackagesToFeed" />
+          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;" />
+  
+  <Target Name="PublishFinalOutput"
+          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+    <RemoveDir Directories="$(CurrentVersionPublishOutput)" />
```
chcosta: Where is `CurrentVersionPublishOutput` defined?

publish/publish.proj
```diff
@@ -6,9 +6,20 @@
...
+          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;" />
+  
+  <Target Name="PublishFinalOutput"
+          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+    <RemoveDir Directories="$(CurrentVersionPublishOutput)" />  
+    <ItemGroup>
+      <PackagesToShip Include="$(BinDir)/ForPublishing/*.nupkg" Exclude="$(BinDir)/ForPublishing/*.symbols.nupkg" />
+      <PackagesToShip Remove="%(PackagesToShip.Identity)" Condition="$([System.String]::Copy('%(PackagesToShip.Identity)').Contains('latest'))" />
```
chcosta: I wouldn't expect these packages to contain 'latest'.  Is this necessary?  If it's not necessary, perhaps you can just use `PackagesToShip` instead of `ShippedNuGetPackages` below?

karajas:
This was the last working build, https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=728537&_a=summary

publish/publish.proj
```diff
@@ -51,7 +52,7 @@
 
   <Target Name="PublishDebFilesToDebianRepo"
           DependsOnTargets="GenerateDebRepoFiles"
-          Condition="'@(DebInstallerFile)' != ''">
```
chcosta: I think we only want to publish deb files to debian repo during finalization.

karajas: @eerhardt @rakeshsinghranchi Is this the case?

karajas:
PTAL.

karajas:
I'll merge this in and monitor official builds.

chcosta:
Thanks Karthik

publish/publish.proj
```diff
@@ -6,11 +6,23 @@
...
+          DependsOnTargets="PublishToAzure;PublishFinalOutput" />
+  
+  <Target Name="PublishFinalOutput"
+          Condition="'$(Finalize)' == 'true'"
+          DependsOnTargets="FinalizeBuildInAzure;PublishDebFilesToDebianRepo;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+    <ItemGroup>
+      <PackagesToShip Include="$(BinDir)/ForPublishing/*.nupkg" Exclude="$(BinDir)/ForPublishing/*.symbols.nupkg" />
+      <PackagesToShip Remove="%(PackagesToShip.Identity)" Condition="$([System.String]::Copy('%(PackagesToShip.Identity)').Contains('latest'))" />
```
weshaggard: For my understanding why are we excluding things that contain "latest" in the name?

karajas: This was because in my testing the download from azure step during finalization pulled down the packages marked "latest"along with the last built version

weshaggard: We should understand why that is and see if we can avoid those.

rakeshsinghranchi:
cc @vivmishra

buildpipeline/pipeline.json
```diff
@@ -180,6 +180,32 @@
...
+      "Definitions": [
+        {
+          "Name": "Core-Setup-Publish",
+          "Parameters": {
+            "PB_TargetArchitecture": "x64",
+          },
+          "ReportingParameters": {
+            "OperatingSystem": "Windows",
+            "SubType": "Publish",
+            "Type": "build/product/",
```
weshaggard: From a reporting MC standpoint shouldn't this be `"Type": "build/publish/",`?

karajas: Noticed this as well, will update this.

PR #2293, opened 2017-05-05
Ensure DOTNET_SKIP_FIRST_TIME_EXPERIENCE is on no matter how the tests are executed.

eerhardt:
We are now using the 2.0 SDK in our tests.  The 2.0 SDK's first time experience now changes the AppData/Roaming/NuGet/NuGet.config file for the current user.  We turn the first time experience off in `run.cmd`, which ensures this doesn't happen during builds.

However, when we run the tests in VSO official builds, we are not using `run.cmd`.  Instead we are just invoking msbuild directly.

This fix ensures that the first time experience is always off, no matter how we run the tests.

@chcosta @mellinoe @weshaggard 

Note: I'm working with @chcosta on thinking about a "more global" setting for this that works across dev machine, Jenkins, and VSO.  This setting should be 100% off all the time in our builds, everywhere.

### DIFF
--- a/src/test/build/shared-build-targets-utils/Utils/DotNetCli.cs
+++ b/src/test/build/shared-build-targets-utils/Utils/DotNetCli.cs
@@ -32,7 +32,8 @@ namespace Microsoft.DotNet.Cli.Build
                 newArgs.Insert(0, "-v");
             }
 
-            return Command.Create(Path.Combine(BinPath, $"dotnet{Constants.ExeSuffix}"), newArgs);
+            return Command.Create(Path.Combine(BinPath, $"dotnet{Constants.ExeSuffix}"), newArgs)
+                .EnvironmentVariable("DOTNET_SKIP_FIRST_TIME_EXPERIENCE", "1");
         }
 
         public Command Restore(params string[] args) => Exec("restore", args);


src/test/build/shared-build-targets-utils/Utils/DotNetCli.cs
```diff
@@ -32,7 +32,8 @@ public Command Exec(string command, params string[] args)
                 newArgs.Insert(0, "-v");
             }
 
-            return Command.Create(Path.Combine(BinPath, $"dotnet{Constants.ExeSuffix}"), newArgs);
+            return Command.Create(Path.Combine(BinPath, $"dotnet{Constants.ExeSuffix}"), newArgs)
+                .EnvironmentVariable("DOTNET_SKIP_FIRST_TIME_EXPERIENCE", "1");
```
weshaggard: Is environment variable the only way to avoid this?

weshaggard: It seems like we should avoid this in any cases where we are using an xcopy CLI.

eerhardt: Yes, the environment variable is the only way to control this in the CLI.  There's been asks since before 1.0.0-preview1 to have a CLI .config file, which would control things like this.  But that feature's priority has never bubbled up high enough to gain attention.

/cc @livarcocc 

> It seems like we should avoid this in any cases where we are using an xcopy CLI.

For sure we should avoid it in any CI system.  This is yet another reason to build a "minimal .NET Core SDK" that doesn't have things like this in it. @Eilon and I were just discussing that.

I've asked @chcosta to think about a more "global" way of setting this env var in our build, so it is always on, no matter if the build is invoked on dev machine, Jenkins or VSO.

PR #2295, opened 2017-05-05
Fix finalization, use sharedframeworknugetversion and sharedhostnugetversion

chcosta:
Finalization was busted because the regex was trying to match the version and would match

Linux-x**64.2.0**.0-prerelease2-23505-01.tar.gz

instead of

Linux-x64.**2.0.0-prerelease2-23505-01**.tar.gz

Fix this by following similar semantics previously used, and explicitly pass the sharedframework version as well as host version to the finalization.

I validated this locally and was able to publish the latest drop with the correct filenames, links on core-setup\readme.md now work (finally)

/cc @eerhardt @karajas

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -59,6 +59,8 @@
                    SemaphoreBlob="$(Channel)/Binaries/sharedFxPublishSemaphore"
                    Channel="$(Channel)"
                    Version="$(SharedFrameworkNugetVersion)"
+                   SharedFrameworkNugetVersion="$(SharedFrameworkNugetVersion)"
+                   SharedHostNuGetVersion="$(HostVersion)"
                    PublishRids="@(PublishRid)"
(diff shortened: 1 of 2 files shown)


publish/publish.proj
```diff
@@ -59,6 +59,8 @@
                    SemaphoreBlob="$(Channel)/Binaries/sharedFxPublishSemaphore"
                    Channel="$(Channel)"
                    Version="$(SharedFrameworkNugetVersion)"
+                   SharedFrameworkNugetVersion="$(SharedFrameworkNugetVersion)"
```
eerhardt: What's the difference between Version and SharedFxNugetVersion?

chcosta: In this case, they're the same. I didn't see any insurance where they were different, but it brings parity with the previous implementation, so I carried it over

eerhardt: Sounds good. Thanks

PR #2296, opened 2017-05-05
Fixing 3 problems with our publish assets:

eerhardt:
1. we are using a `-` when we used to use a `.`
2. We are not putting the commit hash in the version file
3. We are not setting the ContentType to 'text/plain' on the .version file.

Fix https://github.com/dotnet/cli/issues/6517

@chcosta 

/cc @JosephTremoulet


chcosta:
Lgtm

PR #2297, opened 2017-05-06
Fix publish by importing the targets file that has GetLatestCommitHash.

eerhardt:
The official build is broken with my last change with error:

`error MSB4057: The target "GetLatestCommitHash" does not exist in the project.`

Fixing this by importing the right .targets file.

@chcosta


eerhardt:
Merging to unblock the official build.

PR #2298, opened 2017-05-06
Build PlatformAbstractions and DependencyModel nupkgs

chcosta:
Fix to build PlatformAbstractions and DependencyModel nupkgs, also prevent first time experience from executing during the publish step on Unix platforms.

/cc @eerhardt as an FYI

I'm going to merge this after a clean CI so that I can schedule a build for validation tomorrow

### DIFF
--- a/src/src.builds
+++ b/src/src.builds
@@ -2,6 +2,11 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
+  <!-- required to build the projects in their specified order -->
+  <PropertyGroup>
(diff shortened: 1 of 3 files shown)


src/src.builds
```diff
@@ -2,6 +2,11 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
 
+  <!-- required to build the projects in their specified order -->
+  <PropertyGroup>
+    <SerializeProjects>true</SerializeProjects>
```
eerhardt: can we do this at a lower level?  Maybe just so only DependencyModel and PlatformAbstractions are serialized?

This is a really good case for using the CLI's MSBuild instead of VS 2015's MSBuild. Then we wouldn't have to shell out to build SDK projects, and MSBuild would have a chance to do Dependency management itself.

PR #2304, opened 2017-05-07
Fix monikers

chcosta:
Fix product monikers for ubuntu.14.04 and debian.8, they should be `ubuntu` and `debian` respectively.


eerhardt:
Let's discuss this on Monday. I want to get away from these specialized monikers where we can and just use the RID directly. I think this might be a place where we can take a step in the right direction.

If this is only for the .deb installer links, I think we can just change the CLI to point to the new location - Ubuntu.14.04.  The dotnet-install script doesn't depend on the .deb files.

chcosta:
Good.  I'm all for moving that direction

chcosta:
Spoke with @eerhardt.  We're going to accept this break in format (moniker -> rid) and move in the direction of using RID where possible instead of a moniker.  For now, that means that Eric is going to make a fix in CLI for this, rather than continuing to propagate the moniker here.  Closing this issue.

Issue #2265, opened 2017-05-03
Remove /t:GenerateVersionSourceFile from our official builds


eerhardt:
In all our official builds we have a step like the following:

```
"filename": "$(PB_SourcesDirectory)/Tools/msbuild.sh",
"arguments": "$(PB_SourcesDirectory)/build.proj /t:GenerateVersionSourceFile /p:GenerateVersionSourceFile=true /p:OfficialBuildId=$(OfficialBuildId)",
```
See https://github.com/dotnet/core-setup/blob/571fdf07301ed6367890fac3812cd87c0998c8d3/buildpipeline/Core-Setup-OSX-BT.json#L88

Looking into what this target does:

https://github.com/dotnet/buildtools/blob/03988127f489bcd1cf58d71131ac628620d53de9/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L181-L186

1. The comment in the code says it is for non-Windows, yet we have it for Windows builds.
2. core-setup doesn't use the outputted `$(NativeVersionSourceFile)`.

This is a waste and clutters our build definitions.  We should remove these steps. We shouldn't have them just because other repos need them, or because "we might need them in the future". If we need it in the future, we will can add it then.

And with removing this, I think we can remove the "init-tools" calls before it as well.

chcosta:
We should probably change the step to call the `CreateVersionFileDuringBuild` target.  I think that's the real intent.  We need the version file to be created so that we produce the `BuildVersion.props` file which defines the buildtools versioning `BuildVersionMajor` and `BuildVersionMinor`.  I don't think it's hurting anything as-is, but i'll look into changing this after some of the other stabilization work has quieted.

chcosta:
I see that the "CreateVersionFile" target is now a build dependency in build.proj.  That would indeed make this target irrelevant.

PR #2311, opened 2017-05-08
Sign arm binaries in pipeline

wtgodbe:
Attempts to resolve https://github.com/dotnet/core-setup/issues/1977

@chcosta @eerhardt @gkhanna79 PTAL

@chcosta, how did you test your changes to the core-setup pipeline when you did your refactor? I'm trying to set up a clone of Core-Setup's pipebuild to build against this branch, but I'm running into problems because the legs of the build want to build against the VSO repo, while the pipebuild builds with the github repo.

### DIFF
--- a/buildpipeline/Core-Setup-Windows-BT.json
+++ b/buildpipeline/Core-Setup-Windows-BT.json
@@ -60,7 +60,44 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Build",
+      "displayName": "Install Signing Plugin",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "30666190-6959-11e5-9f96-f56098202fef",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "signType": "$(PB_SignType)",
+        "zipSources": "false",
+        "version": "",
+        "feedSource": "https://devdiv.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Run init-tools.cmd",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "bfc8bf76-e7ac-4a8c-9a55-a944a9f632fd",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "$(PB_SourcesDirectory)\\init-tools.cmd",
+        "arguments": "",
+        "modifyEnvironment": "false",
+        "workingFolder": "$(PB_SourcesDirectory)",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Generate version assets",
       "timeoutInMinutes": 0,
       "task": {
         "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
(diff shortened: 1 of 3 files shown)


buildpipeline/pipeline.json
```diff
@@ -126,6 +126,7 @@
           "Parameters": {
             "PB_AdditionalBuildArguments": "/p:PortableBuild=true /p:SkipTests=true",
             "PB_TargetArchitecture": "arm",
+            "PB_DistroRid": "win8-arm",
```
wtgodbe: @chcosta @gkhanna79 are these the right DistroRids? See below as well

wtgodbe:
It may be worth it later on to refactor these definitions so that all Windows jobs go through the 'signing' definition, but since there are 4-5 steps in there that shouldn't be run on arm/arm64, it makes more sense to get it working this way for 2.0.0.

buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -78,6 +133,141 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
+      "displayName": "Build binaries",
```
gkhanna79: Were we not already building for Arm and Arm64 on Windows?

buildpipeline/pipeline.json
gkhanna79: If Portable build is being triggered, then "win-arm" and "win-arm64". Otherwise, "win8-arm" and "win10-arm64".

buildpipeline/Core-Setup-Windows-BT.json
wtgodbe: We were, but we were doing it by just running build.cmd as a monolith - https://github.com/dotnet/core-setup/blob/master/buildpipeline/Core-Setup-Windows-BT.json#L60-L76. I'm copying the pattern from the signing job for this PR, where we run build.cmd to generate version assets, then explicitly build src.builds to build binaries.

buildpipeline/pipeline.json
eerhardt: When are we ever building non-portable for windows in our official builds?  I think this should always be `win-arm`.

buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -60,7 +60,62 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Build",
+      "displayName": "Install Signing Plugin",
```
eerhardt: The names of these .json files are now super confusing.

`Core-Setup-Signing-Windows-BT.json` - does signing
`Core-Setup-Windows-BT.json` - also does signing

buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -60,7 +60,62 @@
...
+      "displayName": "Generate version assets",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "$(PB_SourcesDirectory)\\build.cmd",
+        "arguments": "-- /t:GenerateVersionSourceFile /p:GenerateVersionSourceFile=true /p:OfficialBuildId=$(OfficialBuildId) $(PB_CommonMSBuildArgs)",
```
eerhardt: Do we need this step?  @chcosta - this is why I think we should remove it in the other build, because people look at it and think it is necessary, so it gets copied around.

buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -78,6 +133,141 @@
...
+      "task": {
+        "id": "c6c4c611-aa2e-4a33-b606-5eaba2196824",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "solution": "$(PB_SourcesDirectory)\\src\\pkg\\packaging\\dir.proj",
+        "platform": "$(PB_TargetArchitecture)",
+        "configuration": "$(BuildConfiguration)",
+        "msbuildArguments": "$(PB_CommonMSBuildArgs) /p:BuildFullPlatformManifest=$(PB_BuildFullPlatformManifest) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\packaging.log",
```
eerhardt: `/p:BuildFullPlatformManifest=$(PB_BuildFullPlatformManifest)`

I just fixed a bug in this area, please see https://github.com/dotnet/core-setup/pull/2316

buildpipeline/Core-Setup-Windows-BT.json
chcosta: The copy around problem is how it got in here in the first place.  If you just build from root, it runs the "GenerateVersionFile" target as one of its dependencies, and this step is not needed.  We have cleanup to remove this from other definitions as well.  https://github.com/dotnet/core-setup/issues/2265

wtgodbe: @chcosta so should I remove this step?

buildpipeline/Core-Setup-Windows-BT.json
wtgodbe: Just updated this, Core-Setup-Windows-BT is now the x86/x64 job, Core-Setup-Windows-Arm-BT is the arm/arm64 job.

buildpipeline/Core-Setup-Windows-BT.json
wtgodbe: Should be up-to-date now.

buildpipeline/Core-Setup-Windows-BT.json
eerhardt: Thanks! looks much better now.

eerhardt: although this makes it hard to review because the build defs got switched around.  Oh well.

chcosta:
Everything should refer to the VSO repo

Issue #2076, opened 2017-04-15
Follow up of DependencyModelResolution


ramarag:
followup https://github.com/dotnet/core-setup/pull/2039 with end to end test
and see if all probing paths should be exposed or not

PR #2312, opened 2017-05-08
Lookup dir

ramarag:
Limiting the  probe directories which are made visible to DependencyModel.Resolution with only those that have a nuget layout

Fixes https://github.com/dotnet/core-setup/issues/2076

@gkhanna79 @eerhardt  PTAL

### DIFF
--- a/src/corehost/cli/args.h
+++ b/src/corehost/cli/args.h
@@ -42,6 +42,14 @@ struct probe_config_t
     {
     }
 
+    bool is_lookup() const
+    {
+        return (probe_deps_json == nullptr) &&
+            !only_runtime_assets &&
+            !only_serviceable_assets &&
+            !probe_publish_dir;
+    }
+
     static probe_config_t svc_ni(const pal::string_t& dir)
     {
         return probe_config_t(dir, nullptr, true, true, false);
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -164,13 +164,16 @@ void deps_resolver_t::setup_shared_store_probes(
     }
 }
 
-pal::string_t deps_resolver_t::get_probe_directories()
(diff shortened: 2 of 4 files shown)


ramarag:
@chcosta are you only producing nuget packages for Microsoft.Extensions.DependencyModel only on windows ?

src/corehost/cli/args.h
```diff
@@ -42,6 +42,14 @@ struct probe_config_t
     {
     }
 
+    bool is_lookup() const
```
gkhanna79: Can you please help me understand the rationale behind this and how it maps to "does not have nuget layout"?

ramarag: These are the conditions that are set when nuget layout probe_path is set
look in https://github.com/dotnet/core-setup/blob/bee5e1f9718192b495016fb99097831943434984/src/corehost/cli/deps_resolver.cpp#L214

ramarag:
Adding tests for this feature requires some infra work.
Will open a separate PR once i figure it out

chcosta:
Yes, only producing those nupkgs on Windows.  If they're required somewhere else, we'll need to figure out some model to prevent collisions during publishing.

eerhardt:
> Yes, only producing those nupkgs on Windows. If they're required somewhere else, we'll need to figure out some model to prevent collisions during publishing.

I don't think this is the correct model.  We should ALWAYS be creating the nupkgs.  It's only whether we publish them or not should be decided separately.  That's why we have the `$(PublishRidAgnosticPackages)` property in our build.

That way things like build from source (RHEL, Tizen or others) can still produce these packages and do with them what they want.

chcosta:
That sounds valid.  I don't think the packages have been being built since I took over the conversion project, which was before publishing was a thing.  Someone will need to investigate.  Filed issue https://github.com/dotnet/core-setup/issues/2319

PR #2318, opened 2017-05-09
Add PortableBuild to PB_CommonMSBuildArgs

karajas:
/cc @eerhardt @chcosta

### DIFF
--- a/buildpipeline/Core-Setup-Windows-BT.json
+++ b/buildpipeline/Core-Setup-Windows-BT.json
@@ -69,7 +69,7 @@
       },
       "inputs": {
         "filename": "$(PB_SourcesDirectory)\\build.cmd",
-        "arguments": "-- $(PB_BuildArguments)",
+        "arguments": "-- $(PB_CommonMSBuildArgs)",
         "workingFolder": "$(PB_SourcesDirectory)",
         "failOnStandardError": "false"
       }
@@ -89,7 +89,7 @@
         "solution": "$(PB_SourcesDirectory)\\publish\\publish.proj",
         "platform": "$(PB_TargetArchitecture)",
         "configuration": "$(BuildConfiguration)",
-        "msbuildArguments": "/p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:TargetArchitecture=$(PB_TargetArchitecture) /p:ConfigurationGroup=$(BuildConfiguration)",
+        "msbuildArguments": "$(PB_CommonMSBuildArgs) /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken))",
         "clean": "false",
         "maximumCpuCount": "false",
         "restoreNugetPackages": "false",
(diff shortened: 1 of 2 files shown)


buildpipeline/pipeline.json
```diff
@@ -141,8 +140,7 @@
           "Parameters": {
             "PB_AdditionalBuildArguments": "/p:PortableBuild=true /p:SkipTests=true /p:NativeToolSetDir=C:\\tools\\clr",
             "PB_TargetArchitecture": "arm64",
-            "PB_PortableBuild": "true",
-            "PB_DistroRid": "win-arm64"
+            "PB_PortableBuild": "true"
```
eerhardt: I don't believe this is the correct fix - it is not consistent with how we are doing non-ARM builds.

Instead, we need to be setting `$(PortableBuild)` to `true` in the ARM builds.

chcosta: Distro is only a Linux concept, and that should be cleaned up in the future, but for now, passing the same parameters to `publish` that are passed to `build`, is the right thing to do. 

I think that the better fix, is to make `publish` a part of `build`, but only run it on official builds.  Then we can remove the extra publish step from these build definitions.

karajas: Agreed. Should I combine the build and publish as a part of this change?

karajas:
PTAL.

buildpipeline/pipeline.json
chcosta: It would simplify a lot of things.

karajas: Talked with @eerhardt , I think @wtgodbe has a signing change that needs to have separate steps for build and publish.

chcosta: We would still need some way to explicitly call publish (for the Windows signed builds which are `build` broken into individual steps), but that should still be feasible with a property.  Also, if a build definition is not just calling build (like the Windows signing steps), then publish wouldn't automatically be included.

That being said, it may be too many moving pieces at the moment.  I could buy that argument.  There's a fine line to walk at the moment between expedience and improvements.

buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -195,8 +195,8 @@
     "BuildConfiguration": {
       "value": "Release"
     },
-    "PB_BuildArguments": {
-      "value": "/p:TargetArchitecture=$(PB_TargetArchitecture) /p:ConfigurationGroup=$(BuildConfiguration) $(PB_AdditionalBuildArguments)"
+    "PB_CommonMSBuildArgs": {
```
chcosta: You should add `PB_PortableBuild=false` to the definition so that there is an obvious default value if someone tries to run the build definition, or it's not passed from pipeline.json

eerhardt: Since everything is turning into portable, what if we flipped this?  The default should be portable=true.  And only in places where that isn't true, set it to false.

Maybe, let's merge as is to get the build unblocked.  Then come back through and turn all build defs to portable=true by default.

thoughts?
/cc @gkhanna79

karajas:
merging this in.

buildpipeline/Core-Setup-Windows-BT.json
gkhanna79: Sounds good to me.

PR #2324, opened 2017-05-09
fix portablelinux in ci

chcosta:
PortableLinux was clashing with other Linux distros because the logic was incorrectly setting the GitHub trigger to "PortableLinux" for all non-arm builds (and incorrectly setting them to portable).


chcosta:
@dotnet-bot test ci please

Issue #2268, opened 2017-05-03
Investigate usage of $(HostVersion) properties and other version properties


eerhardt:
We have a bunch of places that are using $(HostVersion) incorrectly.

Here's an example https://github.com/dotnet/core-setup/blob/c9154ce462025e2300ec2a685ace08a8b698df34/src/pkg/packaging/osx/package.props#L6

` <HostFxrComponentId>com.microsoft.dotnet.hostfxr.$(HostVersion).component.osx.x64</HostFxrComponentId>`

Here we are building the hostfxr component, but we are using the `$(HostVersion)` property.  This is wrong because these components can and will version independently.

We need to investigate all the version usages - especially HostVersion as it seems to be the popular one to use - and correct the code to use the right version property.

gkhanna79:
@rakeshsinghranchi PTAL.

eerhardt:
FYI - I'm fixing the ones in the shared framework logic now.

gkhanna79:
Will you endup fixing this as well, as part of it?

eerhardt:
No, not the whole thing.  I'm only concentrating on shared framework logic for now.  I wrote this bug so it is addressed wholistically.

Issue #2245, opened 2017-05-01
validate servicing flow


chcosta:
Validate version numbers are properly updated, packages are properly built / etc.. for servicing releases in core-setup on buildtools

gkhanna79:
Closed via https://github.com/dotnet/core-setup/pull/2328

PR #2328, opened 2017-05-09
Versioning Cleanup

gkhanna79:
This PR cleans up various versioning variables overall and those that are required for 2.0.0 servicing eventually.

I have also fixed an issue with picking up crossgen from the correct location for cross builds.

Fixes https://github.com/dotnet/core-setup/issues/2268 and https://github.com/dotnet/core-setup/issues/2245.

@eerhardt @weshaggard PTAL

@rakeshsinghranchi PTAL at the setup changes.

CC @vivmishra

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -9,8 +9,7 @@
     <MajorVersion>2</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <StabilizePackageVersion>false</StabilizePackageVersion>
-    <LockHostVersion>false</LockHostVersion>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
     <PreReleaseLabel>preview2</PreReleaseLabel>
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
     <ReleaseBrandSuffix>Preview 2</ReleaseBrandSuffix>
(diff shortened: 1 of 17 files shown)


dir.props
```diff
@@ -71,24 +70,50 @@
...
+      Dotnet/AppHost. Thus, the "UseShipped*" properties (below) are always set to false.
+      
+      However, there are scenarios when some of these components will not change (e.g. during
+      servicing, we may only change HostPolicy but not HostFXR and Dotnet/AppHost). In such cases,
+      set the appropriate "UseShipped*" property below to true so that we will use the last shipped
+      version of the package. 
+    -->
+
+    <!-- The host/apphost package versions are only updated whenever there is a change in the components -->
+    <UseShippedHostPackage>false</UseShippedHostPackage>
```
weshaggard: Why do we even need these extra properties? Why wouldn't we just have the version and then in servicing change it from SharedFrameworkNugetVersion to whatever hard-coded version we need? It seems kind of redundant to need to set both this bool and the version.

dir.props
```diff
@@ -10,7 +10,6 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <StabilizePackageVersion>false</StabilizePackageVersion>
```
weshaggard: looks like StabilizePackageVersion is set twice in this file.

dir.props
gkhanna79: Initially, I was planning to just have the version. However, in my experience, while changing that is straightforward for those who have context, making it more explicit is simpler for someone who is reading this file at any point in time.

Also, these UseShipped* property enable extensibility (e.g. we could choose to use them to skip building the corresponding host packages).

dir.props
gkhanna79: Fixed.

dir.props
```diff
@@ -68,27 +67,52 @@
...
+      
+      However, there are scenarios when some of these components will not change (e.g. during
+      servicing, we may only change HostPolicy but not HostFXR and Dotnet/AppHost). In such cases,
+      set the appropriate "UseShipped*" property below to true so that we will use the last shipped
+      version of the package. 
+    -->
+
+    <!-- The host/apphost package versions are only updated whenever there is a change in the components -->
+    <UseShippedHostPackage>false</UseShippedHostPackage>
+    <HostVersion Condition="'$(UseShippedHostPackage)' != 'true'">$(SharedFrameworkNugetVersion)</HostVersion>
```
eerhardt: Instead of setting this to `$(SharedFrameworkNugetVersion)`, I think it would make more sense to set it to `$(ProductVersion)`

src/corehost/build.proj
```diff
@@ -42,7 +44,7 @@
              Properties="GenerateNativeVersionInfo=true;AssemblyName=%(HostFiles.Identity);NativeVersionFileDir=$(IntermediateOutputRootPath)hostResourceFiles\%(HostFiles.Identity);NativeVersionHeaderFile=$(IntermediateOutputRootPath)hostResourceFiles\%(HostFiles.Identity)\version_info.h"
              Targets="GenerateVersionHeader" />
     <PropertyGroup>
-      <BuildArgs>$(ConfigurationGroup) $(TargetArchitecture) version $(HostVersion) commit $(LatestCommit) rid $(TargetRid)</BuildArgs>
+      <BuildArgs>$(ConfigurationGroup) $(TargetArchitecture) version $(SharedFrameworkNugetVersion) commit $(LatestCommit) rid $(TargetRid)</BuildArgs>
```
eerhardt: Should be `$(ProductVersion)` IMO

src/pkg/packaging/osx/package.targets
```diff
@@ -39,7 +39,7 @@
 
     <PropertyGroup>
         <InstallLocation>/usr/local/share/dotnet</InstallLocation>
-        <CommonArgs>--version $(HostVersion) --install-location $(InstallLocation)</CommonArgs>
+        <CommonArgs>--version $(SharedFrameworkNugetVersion) --install-location $(InstallLocation)</CommonArgs>
```
eerhardt: `$(ProductVersion)`?

src/pkg/packaging/osx/package.targets
```diff
@@ -72,7 +72,7 @@
         ReplacementPatterns="@(ConfigPattern)"
         ReplacementStrings="@(ConfigReplace)" />
     
-    <Exec Command="productbuild --version $(HostVersion) --identifier $(SharedPackageId) --package-path $(PackagesIntermediateDir) --resources $(ResourcePath) --distribution $(DistributionFile) $(OutFilePath)" />
+    <Exec Command="productbuild --version $(SharedFrameworkNugetVersion) --identifier $(SharedPackageId) --package-path $(PackagesIntermediateDir) --resources $(ResourcePath) --distribution $(DistributionFile) $(OutFilePath)" />
```
eerhardt: `$(ProductVersion)`?

src/pkg/packaging/dir.proj
```diff
@@ -83,7 +83,7 @@
 
     <WriteLinesToFile
       File="$(OutputVersionBadge)"
-      Lines="$([System.IO.File]::ReadAllText('$(templateSvg)').Replace('ver_number', '$(HostVersion)'))"
+      Lines="$([System.IO.File]::ReadAllText('$(templateSvg)').Replace('ver_number', '$(SharedFrameworkNugetVersion)'))"
```
eerhardt: `$(ProductVersion)`?

src/corehost/build.proj
gkhanna79: This should actually have been split, like it is for non-Windows. I have fixed it to be in sync with that.

src/pkg/packaging/osx/package.targets
gkhanna79: Install location is the version of SharedFX - hence this.

src/pkg/packaging/osx/package.targets
gkhanna79: Fixed.

src/pkg/packaging/dir.proj
gkhanna79: Fixed (even though, today, they are the same).

dir.props
gkhanna79: Agreed - fixed.

gkhanna79:
@eerhardt Is this good to merge?

eerhardt:
Yep.  I approved earlier.

PR #2329, opened 2017-05-09
Fixing the build break by putting our custom task NuGet dependencies in the right spot.

eerhardt:
@chcosta @weshaggard @karajas @mellinoe 

hopefully when we move the custom task to use the 2.0 SDK, we can just "publish" it, and this kind of stuff won't be necessary.

### DIFF
--- a/tools-local/tasks/net45/ProcessSharedFrameworkDeps.net45.cs
+++ b/tools-local/tasks/net45/ProcessSharedFrameworkDeps.net45.cs
@@ -1,0 +1,25 @@
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.Extensions.DependencyModel;
+using NuGet.Common;
+using NuGet.ProjectModel;
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public partial class ProcessSharedFrameworkDeps
+    {
+        partial void EnsureInitialized(string buildToolsTaskDir)
+        {
+            // ensure the build tools AssemblyResolver is enabled, so we get the correct assembly unification
+            // even if the build tools assembly hasn't been loaded yet.
+            string buildTasksPath = Path.Combine(buildToolsTaskDir, "Microsoft.DotNet.Build.Tasks.dll");
+            Assembly buildTasksAssembly = Assembly.Load(AssemblyName.GetAssemblyName(buildTasksPath));
+            Type assemblyResolver = buildTasksAssembly.GetType("Microsoft.DotNet.Build.Common.Desktop.AssemblyResolver");
+            assemblyResolver.GetMethod("Enable").Invoke(null, new object[] { });
+        }
+    }
+}
--- a/src/sharedFramework/sharedFramework.proj
+++ b/src/sharedFramework/sharedFramework.proj
@@ -117,7 +117,8 @@
     <ProcessSharedFrameworkDeps AssetsFilePath="$(SharedFrameworkAssetsFile)"
                                 DepsFilePath="$(SharedFrameworkDepsFile)"
                                 PackagesToRemove="@(TrimPkgsFromDeps)"
-                                Runtime="$(RuntimeGraphGeneratorRuntime)" />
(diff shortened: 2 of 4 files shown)


tools-local/tasks/project.json
```diff
@@ -2,7 +2,7 @@
   "dependencies": {
     "System.Reflection.Metadata": "1.3.0",
     "Newtonsoft.Json": "9.0.1",
-    "NuGet.ProjectModel": "3.5.0",
+    "NuGet.ProjectModel": "4.3.0-preview1-4051",
```
eerhardt: FYI - I needed to move to the latest NuGet here because 4.3.0 depends on newtonsoft 9 for all TFMs.  Earlier versions of NuGet had `net45` depending on newtonsoft 6, which caused me some problems.

eerhardt:
I changed the way I fixed this error by ensuring that the build tools `AssemblyResolver` is always enabled.  It will ensure that the correct NuGet and Newtonsoft assemblies are loaded/unified correctly.

@weshaggard and @ericstj - can you take a look if this is an acceptable approach?  I couldn't get the right assemblies copied to the right spots because I was either mis-matching NuGet on non-Windows or Newtonsoft on Windows.

The error we were getting was

```
System.IO.FileNotFoundException: Could not load file or assembly 'NuGet.Common, Version=3.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.
File name: 'NuGet.Common, Version=3.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
   at Microsoft.DotNet.Build.Tasks.ProcessSharedFrameworkDeps.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
```

During Windows builds because the windows builds are split apart.  Half of the build runs, and then we sign, and then some more of the build runs again.  When the sharedframework gets pubilshed in its own step, no one is loading build tools, and thus enabling the AssemblyResolver.  So we get assembly load errors during official builds but not on dev-boxes or in Jenkins.

eerhardt:
I'm going to merge to unblock the official build. Please let me know if there is a better way to fix this issue.  I'll make any necessary changes in another PR.

tools-local/tasks/net45/ProcessSharedFrameworkDeps.net45.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public partial class ProcessSharedFrameworkDeps
+    {
+        partial void EnsureInitialized(string buildToolsTaskDir)
+        {
+            // ensure the build tools AssemblyResolver is enabled, so we get the correct assembly unification
+            // even if the build tools assembly hasn't been loaded yet.
+            string buildTasksPath = Path.Combine(buildToolsTaskDir, "Microsoft.DotNet.Build.Tasks.dll");
+            Assembly buildTasksAssembly = Assembly.Load(AssemblyName.GetAssemblyName(buildTasksPath));
+            Type assemblyResolver = buildTasksAssembly.GetType("Microsoft.DotNet.Build.Common.Desktop.AssemblyResolver");
```
weshaggard: Do you prefer this approach over just including a copy of https://github.com/dotnet/buildtools/blob/master/src/common/AssemblyResolver.cs in this task library?

eerhardt: I thought of copying the file into core-setup, but I wasn't crazy about it for obvious reasons (duplication of code, duplication of logic running during our build - since there would be 2 AssemblyResolvers that were "enabled").

I think the approach I would have preferred is for this to be exposed in a NuGet package that this tasks library can reference...  But I think our plan is to pull these tasks into build-tools anyway, right?  When that happens, this can just call AssemblyResolver.Enable directly.

weshaggard: > I think our plan is to pull these tasks into build-tools anyway, right? 

Yes that is the longer term plan but will likely not happen for 2.0. 

I can live with this approach but as with most reflection stuff it might be a little fragile.

PR #2330, opened 2017-05-09
Add a versions file to the Microsoft.NETCore.App packages with repo hashes

joperezr:
PTAL: @weshaggard @dagood @ericstj 

Given that core-setup is not using the latest buildtools yet, then PR https://github.com/dotnet/buildtools/pull/1482 won't be able to be ingested until that happens. For now, adding the task that we need into the local task library since we need this new file in the package that will add the hashes of corefx and coreclr used to produce the package.


src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -80,6 +80,43 @@
...
+
+  <Target Name="GenerateHashVersionsFile" BeforeTargets="GetPackageFiles">
+
+    <PropertyGroup>
+      <_SystemRuntimePath Condition="$([System.String]::new('%(File.Identity)').ToLowerInvariant().Contains('system.runtime.dll'))">%(File.Identity)</_SystemRuntimePath>
+      <_SystemPrivateCoreLibPath Condition="$([System.String]::new('%(File.Identity)').ToLowerInvariant().Contains('system.private.corelib.dll'))">%(File.Identity)</_SystemPrivateCoreLibPath>
+    </PropertyGroup>
+
+    <GetVersionFromAssembly PathToAssembly="$(_SystemRuntimePath)">
+      <Output TaskParameter="AssemblyHash" PropertyName="CoreFxHash" />
```
dagood: (Just a comment) CoreFX packages do have a version.txt file inside with the commit hash, but I don't know if it's reasonable to read that file from here. (CoreCLR doesn't have version.txt, anyway.)

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -80,6 +80,43 @@
...
+
+    <Message Importance="Low" Text="The hash of corefx is $(CoreFxHash) taken from $(_SystemRuntimePath)" />
+    <Message Importance="Low" Text="The hash of coreclr is $(CoreClrHash) taken from $(_SystemPrivateCoreLibPath)" />
+
+    <ItemGroup>
+      <_VersionsFileLines Include="CoreSetup $(LatestCommit)" />
+      <_VersionsFileLines Include="CoreFx $(CoreFxHash)" />
+      <_VersionsFileLines Include="dotnet/versions/corefx $(CoreFxCurrentRef)" />
+      <_VersionsFileLines Include="CoreClr $(CoreClrHash)" />
+      <_VersionsFileLines Include="dotnet/versions/coreclr $(CoreClrCurrentRef)" />
```
dagood: Is this the best format for consumers of the new file? Is the consumer an msbuild script, or something else?

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -80,6 +80,43 @@
...
+    </ItemGroup>
+
+    <WriteLinesToFile Lines="@(_VersionsFileLines)"
+                      File="$(ObjDir)\Microsoft.NETCore.App.versions.txt"
+                      Overwrite="true"/>
+
+    <ItemGroup>
+      <File Include="$(ObjDir)\Microsoft.NETCore.App.versions.txt" />
```
dagood: Centralize path into a property? (Also in `WriteLinesToFile`.)

tools-local/Microsoft.DotNet.Build.Tasks.Local/GetVersionFromAssembly.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Microsoft.Build.Framework;
```
dagood: Copyright header missing.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -80,6 +80,42 @@
...
+    </GetVersionFromAssembly>
+    <GetVersionFromAssembly PathToAssembly="$(_SystemPrivateCoreLibPath)" Condition="'$(_SkipCoreClrHash)' != 'true'" >
+      <Output TaskParameter="AssemblyHash" PropertyName="CoreClrHash" />
+    </GetVersionFromAssembly>
+
+    <Message Importance="Low" Text="The hash of corefx is $(CoreFxHash) taken from $(_SystemRuntimePath)" />
+    <Message Importance="Low" Text="The hash of coreclr is $(CoreClrHash) taken from $(_SystemPrivateCoreLibPath)" />
+
+    <ItemGroup>
+      <_VersionsFileLines Include="CoreSetup $(LatestCommit)" />
```
weshaggard: nit but I would make the names for these commits match the repo names so lowercase and a '-' for core-setup.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -80,6 +80,42 @@
                               PreferredPackages="$(Id);@(RuntimeDependency)" />
   </Target>
 
+  <UsingTask TaskName="GetVersionFromAssembly" AssemblyFile="$(LocalBuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.Local.dll"/>
+
+  <Target Name="GenerateHashVersionsFile" BeforeTargets="GetPackageFiles">
+    <PropertyGroup>
+      <_SystemRuntimePath Condition="$([System.String]::new('%(File.Identity)').ToLowerInvariant().Contains('system.runtime.dll'))">%(File.Identity)</_SystemRuntimePath>
+      <_SystemPrivateCoreLibPath Condition="$([System.String]::new('%(File.Identity)').ToLowerInvariant().Contains('system.private.corelib.dll'))">%(File.Identity)</_SystemPrivateCoreLibPath>
+      <_SkipCoreClrHash Condition="'$(_SystemPrivateCoreLibPath)' == ''">true</_SkipCoreClrHash>
```
weshaggard: In what scenarios do you thin system.private.corelib will not exist?

joperezr: When building the identity package, coreclr binaries won't be in the `File` item.

weshaggard: Fair.

joperezr:
By debugging in OSX, looks like the commit hash is not embedded in any of the binaries that are built outside Windows. I can make this task to WarnAndContinue and don't have it in all packages as a workaround, but I will debug a bit more into corefx build to try to find out why is it that we are not adding the hash in non-Windows builds.

weshaggard:
>By debugging in OSX, looks like the commit hash is not embedded in any of the binaries that are built outside Windows. I can make this task to WarnAndContinue and don't have it in all packages as a workaround, but I will debug a bit more into corefx build to try to find out why is it that we are not adding the hash in non-Windows builds.

Yes feel free to ignore it for now but we should file and issue to get that fixed.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
ericstj: Yeah, we could just flow the data in the transport packages, rather than having anything fancy here.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
ericstj: You could avoid a lot of duplication by doing this in the depproj where we already identify all of these files: https://github.com/dotnet/core-setup/blob/f0fed403e0f455ebaf37626a9561b7de5fdbb160/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj#L30

joperezr:
ok so with my last commit d342c6b I have added what @dagood suggested originally which is read the version.txt file comming from the transport packages instead of reading the assembly info. This in combination with [the recent change I merged in coreclr](https://github.com/dotnet/coreclr/pull/11497) that adds the version.txt to the coreclr transport package will add the right hashes to the netcoreapp package.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -74,6 +74,37 @@
...
+    <ItemGroup>
+      <_VersionsFileLines Include="core-setup $(LatestCommit)" />
+      <_VersionsFileLines Include="corefx $(CoreFxHash)" Condition="'$(CoreFxHash)'!=''" />
+      <_VersionsFileLines Include="dotnet/versions/corefx $(CoreFxCurrentRef)" Condition="'$(CoreFxHash)'!=''" />
+      <_VersionsFileLines Include="coreclr $(CoreClrHash)" Condition="'$(CoreClrHash)'!=''" />
+      <_VersionsFileLines Include="dotnet/versions/coreclr $(CoreClrCurrentRef)" Condition="'$(CoreClrHash)'!=''" />
+    </ItemGroup>
+
+    <WriteLinesToFile Lines="@(_VersionsFileLines)"
+                       File="$(ObjDir)\Microsoft.NETCore.App.versions.txt"
```
ericstj: I think this should be IntermediateOutputPath so that it doesn't clash.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -74,6 +74,37 @@
...
+    <ReadLinesFromFile File="$(CoreFxVersionFile)" Condition="Exists('$(CoreFxVersionFile)')">
+      <Output TaskParameter="Lines" PropertyName="CoreFxHash" />
+    </ReadLinesFromFile>
+    <ReadLinesFromFile File="$(CoreClrVersionFile)" Condition="Exists('$(CoreClrVersionFile)')">
+      <Output TaskParameter="Lines" PropertyName="CoreClrHash" />
+    </ReadLinesFromFile>
+
+    <ItemGroup>
+      <_VersionsFileLines Include="core-setup $(LatestCommit)" />
+      <_VersionsFileLines Include="corefx $(CoreFxHash)" Condition="'$(CoreFxHash)'!=''" />
```
weshaggard: Instead of conditioning these can you make this fail if they aren't found? I think that would be better then accidentally breaking this in the future.

joperezr: will do.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -74,6 +74,44 @@
...
+      <Output TaskParameter="Lines" PropertyName="CoreClrHash" />
+    </ReadLinesFromFile>
+
+    <Error Condition="!Exists('$(CoreFxVersionFile)')" Text="Failed to find the corefx version.txt file at $(CoreFxVersionFile)" />
+
+    <ItemGroup>
+      <_VersionsFileLines Include="core-setup $(LatestCommit)" />
+      <_VersionsFileLines Include="corefx $(CoreFxHash)" />
+      <_VersionsFileLines Include="dotnet/versions/corefx $(CoreFxCurrentRef)" />
+      <_VersionsFileLines Include="coreclr $(CoreClrHash)" Condition="'$(CoreClrHash)'!=''" />
```
weshaggard: I would also remove the condition here once you get a coreclr package with the version.

joperezr: It won't always be set though, like for example when we are building the identity package, in that case the path to version.txt won't be set.

ericstj: Once we switch to PackageReference we can just pull it from the identity package and we'll have the id/version handy.  Until then we'll be missing this part in the text file.  Consider just putting a TODO above the propertygroup that defines the paths to the version files.

joperezr: As discussed offline, instead of a ToDo I'll just log an issue about fixing this once we move to packageReference, as well as add one more hash for standard repo.

joperezr:
Thanks for the review guys.

Issue #637, opened 2016-11-12
DependencyModel lacks Copyright assembly attribute


piotrpMSFT:
Moved from https://github.com/dotnet/cli/issues/3829 on behalf of @dagood.

----

## Steps to reproduce

Look at properties for `Microsoft.Extensions.DependencyModel.dll` (1.0.0).
## Expected  behavior

Properties are like these (`System.Collections.dll` shown):

```
Product name = Microsoft® .NET Framework
Product version = 1.0.24212.01. Commit Hash: 9688ddbb62c04189cac4c4a06e31e93377dccd41
Copyright = © Microsoft Corporation.  All rights reserved.
```
## Actual behavior

Properties are these (`Microsoft.Extensions.DependencyModel.dll` shown):

```
Product name =
Product version = 1.0.0
Copyright =
```

(In CoreFX, the attributes that provide this data are [generated by Buildtools](https://github.com/dotnet/buildtools/blob/197567b/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L57).)

gkhanna79:
@schellap This is likely tied to moving Core-Setup rep to use BuildTools.

gkhanna79:
@AlfredoMS Assigning this to you since you are working on the transition to build tools.

eerhardt:
We also need to add the license file to both DependencyModel and PlatformAbstractions.

From a customer:

> Hello, The Microsoft.Extensions.DependencyModel package is missing the associated Licence. Could you add it please?

@Petermarcu @gkhanna79 - should this be fixed for 2.0?

gkhanna79:
If we are doing the same thing for other binaries/packages, we should fix it for 2.0.

PR #2331, opened 2017-05-09
Create symbols packages, add serviceable attribute, and update assembly info

chcosta:
For Microsoft.DotNet.PlatformAbstractions, and Microsoft.Extensions.DependencyModel...

- Creates symbols packages, which will get uploaded with the rest of thy symbols packages to myget
- Ensure serviceable attribute is set in the package
- Align assembly info with values we define on the rest of our netcore assemblies (copyright, informational string, etc...)

Open question.  Do we want to version the assemblies as 2.0.0?  That's the version that they were previously stamped with.

Fixes https://github.com/dotnet/core-setup/issues/637


tools-override/versioning.targets
```diff
@@ -0,0 +1,389 @@
...
+  <!-- #################################### -->
+  <!-- Generate Assembly Info -->
+  <!-- #################################### -->
+  <PropertyGroup>
+    <AssemblyVersion Condition="'$(AssemblyVersion)'==''">1.0.0.0</AssemblyVersion>
+    <CLSCompliant Condition="'$(CLSCompliant)'==''">false</CLSCompliant>
+    <AssemblyFileVersion Condition="'$(AssemblyFileVersion)'==''">$(MajorVersion).$(MinorVersion).$(BuildNumberMajor).$(BuildNumberMinor)</AssemblyFileVersion>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)versioning.props" />
```
chcosta: This is an added import

tools-override/versioning.targets
```diff
@@ -0,0 +1,389 @@
...
+
+  <Target Name="GenerateAssemblyInfo"
+      Inputs="$(MSBuildProjectFile)"
+      Outputs="$(AssemblyInfoFile)"
+      Condition="'$(GenerateAssemblyInfo)'=='true'"
+      DependsOnTargets="CreateVersionFileDuringBuild">
+
+    <Error Condition="!Exists('$(IntermediateOutputPath)')" Text="GenerateAssemblyInfo failed because IntermediateOutputPath isn't set to a valid directory" />
+
+    <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
```
chcosta: These sections have changed to utilize the values provided in version.props instead of hard-coding them here.

tools-override/versioning.targets
```diff
@@ -0,0 +1,389 @@
...
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyProduct(&quot;$(Product)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyCopyright(&quot;$(Copyright)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyVersion(&quot;$(AssemblyVersion)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyFileVersion(&quot;$(FileVersion)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyInformationalVersion(&quot;$(FileVersion)&quot;)&gt;" />
+      <AssemblyInfoLines Condition="'$(CLSCompliant)'=='true'" Include="&lt;Assembly:CLSCompliant(True)&gt;" />
+      <AssemblyInfoLines Condition="'$(AssemblyComVisible)'!=''" Include="&lt;Assembly:System.Runtime.InteropServices.ComVisible($(AssemblyComVisible))&gt;" />
+      <AssemblyInfoLines Condition="'$(SkipFrameworkAssemblyMetadata)' != 'true'"
+          Include="&lt;assembly:System.Reflection.AssemblyMetadata(&quot;%(AssemblyMetadata.Identity)&quot;, &quot;%(AssemblyMetadata.Value)&quot;)&gt;" />
+    </ItemGroup>
```
chcosta: no changes below this line.  This diff is kind of a pain to examine. :(

chcosta:
Before merging, I'll need to see if any changes to versioning.targets came in with the recent (or soon to be) BuildTools update

src/pkg/packaging/dir.proj
```diff
@@ -158,7 +158,9 @@
   </PropertyGroup>
   <Target Name="GenerateInstallers" DependsOnTargets="$(InstallerDependsOn)" />
 
-  <Target Name="GenerateNugetPackages" DependsOnTargets="InitPackage">
+  <Import Project="$(ProjectDir)src/managed/CommonManaged.props" />
+  <Target Name="GenerateNugetPackages"
```
eerhardt: This whole file doesn't need to be changed.  Setting `IncludeSymbols` and `Serviceable` in the CommonManaged.props file is enough to get `dotnet pack` to do the right thing.

src/managed/dir.proj
```diff
@@ -5,12 +5,16 @@
...
+  <Target Name="Build"
+          DependsOnTargets="GetLatestCommitHash">
     
     <MakeDir  Condition="!Exists('$(pOutDir)')" Directories="$(pOutDir)" /> 
 
     <PropertyGroup>
       <Args>/p:Configuration=$(ConfigurationGroup)</Args>
+      <_EnvironmentVariables>NUGET_PACKAGES=$(PackagesDir)</_EnvironmentVariables>
+      <_EnvironmentVariables Condition="'$(LatestCommit)' != ''">$(_EnvironmentVariables);LatestCommit=$(LatestCommit)</_EnvironmentVariables>
+      <_EnvironmentVariables Condition="'$(BuiltByString)' != ''">$(_EnvironmentVariables);BuiltByString=$(BuiltByString)</_EnvironmentVariables>
```
eerhardt: Where does `$(BuiltByString)` get set?  I'm not seeing it set anywhere.

tools-override/versioning.targets
```diff
@@ -0,0 +1,389 @@
...
+      <AssemblyInfoUsings Include="Imports System.Reflection" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyTitle(&quot;$(AssemblyName)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyDescription(&quot;$(Description)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyDefaultAlias(&quot;$(AssemblyName)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyCompany(&quot;$(Company)n&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyProduct(&quot;$(Product)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyCopyright(&quot;$(Copyright)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyVersion(&quot;$(AssemblyVersion)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyFileVersion(&quot;$(FileVersion)&quot;)&gt;" />
+      <AssemblyInfoLines Include="&lt;Assembly:AssemblyInformationalVersion(&quot;$(FileVersion)&quot;)&gt;" />
```
eerhardt: If we are going this route, this line should use `$(InformationalVersion)` instead of `$(FileVersion)`

src/managed/CommonManaged.props
```diff
@@ -26,4 +32,5 @@
     <Reference Include="System.Core" />
   </ItemGroup>
 
+  <Import Condition="Exists('$(RepoRoot)Tools/versioning.targets')" Project="$(RepoRoot)Tools/versioning.targets" />
```
eerhardt: I don't think we need to use this.  The .NET Core SDK has this functionality built into it.  You just need to set the correct properties in the project.

The added benefit is that not only are they set in your assemblies, but they are also set correctly in the output .nupkg.

See [https://github.com/dotnet/sdk/pull/81](https://github.com/dotnet/sdk/pull/81)

src/managed/dir.proj
chcosta: It's in the same buildversion.props file which gets generated and includes `BuildNumberMajor` and `BuildNumberMinor`

src/managed/CommonManaged.props
chcosta: The intent is to have similar property values defined between this, and other repo's.  ie, copyright info specified for a CoreFx binary would be the same as copyright info for a core-setup binary.  BuildTools is the common infrastructure used to define this.  However, your point is valid, all we really need to import is versioning.props, not all of versioning.targets.  Versioning.props will be imported by versioning.targets (which is used by other repos) and, also imported for the `dotnet build` here.

src/managed/CommonManaged.props
```diff
@@ -6,8 +6,14 @@
...
     <VersionPrefix>2.0.0</VersionPrefix>
+    <!-- This is consistent with previously packaged assemblies, but it's not clear if it's what we want
+         going forward -->
+    <AssemblyFileVersion>2.0.0</AssemblyFileVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+    <IncludeSymbols>true</IncludeSymbols>
+    <Serviceable>true</Serviceable>
+    <SkipVersionGeneration>true</SkipVersionGeneration>
   </PropertyGroup>
```
weshaggard: Given that it appears like you want to skip our generation of assembly version info don't you also want to set GenerateAssemblyInfo?

chcosta:
Ping

src/managed/dir.proj
```diff
@@ -3,11 +3,15 @@
...
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
 
-  <Target Name="Build">
+  <Target Name="Build"
+          DependsOnTargets="GetLatestCommitHash">
     <MakeDir  Condition="!Exists('$(pOutDir)')" Directories="$(pOutDir)" />
 
     <PropertyGroup>
       <Args>/p:Configuration=$(ConfigurationGroup)</Args>
+      <_EnvironmentVariables>NUGET_PACKAGES=$(PackagesDir)</_EnvironmentVariables>
```
weshaggard: We really need to get out of the this business of relying on environment variables. It makes things a lot harder on ourselves.

weshaggard: Is there any reason we cannot pass these parameters as args?

As for the PackagesDir I think we should be switching to using a nuget.config file to define the package folder instead of relying on the NUGET_PACKAGES env variable.

eerhardt: Agreed.

1. I think the NUGET_PACKAGES env var can go away now.  NuGet will write the restore feeds to the assets file, so build will pick it up from there.  No need to set NUGET_PACKAGES as an env var anymore.
2. The other properties can just be passed as normal MSBUild properties `/p:`

---
In reply to: [117293682](https://github.com/dotnet/core-setup/pull/2331#discussion_r117293682) [](ancestors = 117293682)

src/pkg/packaging/dir.proj
```diff
@@ -154,6 +154,8 @@
   </PropertyGroup>
   <Target Name="GenerateInstallers" DependsOnTargets="InitPackage;$(InstallerDependsOn)" />
 
+  <Import Project="$(ProjectDir)src/managed/CommonManaged.props" />
```
eerhardt: Is this necessary?  I don't see what needs this.

chcosta: CommonManaged.props sets the `Serviceable` and `IncludeSymbols` properties used for packaging

src/managed/CommonManaged.props
```diff
@@ -6,8 +6,14 @@
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
     <VersionPrefix>2.0.0</VersionPrefix>
+    <!-- This is consistent with previously packaged assemblies, but it's not clear if it's what we want
+         going forward -->
+    <AssemblyFileVersion>2.0.0</AssemblyFileVersion>
```
eerhardt: I think the AssemblyFileVersion should have the full build number in it, right?  That way it changes for every build.

Also, it would be nice if we could use the `$(VersionPrefix)` here, so we don't have to change 2 numbers.

src/managed/CommonManaged.props
```diff
@@ -6,8 +6,14 @@
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
     <VersionPrefix>2.0.0</VersionPrefix>
```
eerhardt: Gaurav's new 2.1 change points out a problem with this property.  We should be sharing the root dir.props values.

This points out one of the problems with dumping everything into dir.props files.  Here we just want to share the repo's "branch info".  We can't use all the other settings in the dir.props because it doesn't jive with the .NET Core SDK.

I've logged [https://github.com/dotnet/core-setup/issues/2435](https://github.com/dotnet/core-setup/issues/2435) for this.

eerhardt:
test Windows_NT arm Debug Build

src/managed/CommonManaged.props
```diff
@@ -6,8 +6,12 @@
...
+    <AssemblyFileVersion>$(VersionPrefix)</AssemblyFileVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+    <IncludeSymbols>true</IncludeSymbols>
+    <Serviceable>true</Serviceable>
+    <SkipVersionGeneration>true</SkipVersionGeneration>
```
eerhardt: NIT: This is unnecessary, since we aren't including the versioning.targets file anymore.

eerhardt:
It looks like the last thing that is needed is the license.  But that can be added separately.  Or if you want to log a new bug for it.  See https://github.com/dotnet/core-setup/issues/637#issuecomment-297765512

src/managed/CommonManaged.props
weshaggard: ciorrect.

chcosta:
@eerhardt , I've added the license files

src/managed/CommonManaged.props
```diff
@@ -11,6 +11,8 @@
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
     <IncludeSymbols>true</IncludeSymbols>
     <Serviceable>true</Serviceable>
+    <PackageLicenseUrl>https://github.com/dotnet/core-setup/blob/master/LICENSE.TXT</PackageLicenseUrl>
```
eerhardt: Microsoft.NETCore.App's license url is:

`<licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>`.

We should probably use the same fwlink.

eerhardt: Looks like it comes from https://github.com/dotnet/buildtools/blob/3309085641d2fe6a6193915dd89ffbb159dad9fb/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L40-L46

chcosta: Then, should we also align this? 

https://github.com/dotnet/core-setup/blob/master/src/pkg/dir.props#L19

eerhardt: yea - I just found that.  Nevermind.  What you have here is good.

I was looking at what we shipped for 2.0.0-preview1.  Looks like that has changed since.

eerhardt:
@chcosta - I assume we will take this to shiproom for 2.0?

chcosta:
Yes, this and https://github.com/dotnet/core-setup/pull/2430 will need to go to shiproom.  I need to sync up with @gkhanna79 to figure out the process there, and it'd be nice to see a clean build out of the 2.0.0 branch

gkhanna79:
Core-Setup master is locked down for now, so email shiproom prior to any merges. Ofcourse, do the same for any ports to 2.0.0 as well.

PR #2342, opened 2017-05-10
Use run.sh from build.sh "the right way"

### LINKED ISSUE #2313: build.sh should just call run.sh directly
Today build.sh does some parsing and then calls run.sh with a different set of arguments.  This is different from build.cmd which just forwards its arguments to run.cmd.

Because of how this is done, you can't use the -- trick to pass additional properties to MSBuild on *nix.

We should either adopt `run.sh` fully or remove it and go back to parsing in shell scripts.  Landing on run.sh is probably the most expedient thing.

Long term, we need to ensure the native code is still build-able without managed code.

ellismg:
Previously, we did some argument parsing in build.sh to construct a
command line to pass to run.sh which would do additional parsing and
then actually invoke MSBuild. This change cleans things up so build.sh
works the same way as build.cmd, a simple wrapper around run.sh

As part of this work, we need to update our build configurations since
the arguments we pass to build.sh have now changed.

Fixes #2313


ellismg:
If anyone has good ideas on how to stage this such that I can get a CI read and an official build read, that's great.  Otherwise, I guess I can just respond to any breaks after things go in.

buildpipeline/pipeline.json
```diff
@@ -98,7 +98,7 @@
             "PB_DistroRid": "ubuntu.14.04-arm", 
             "PB_DockerTag": "ubuntu-14.04-cross-0cd4667-20172211042239", 
             "PB_TargetArchitecture": "arm", 
-            "PB_AdditionalBuildArguments":"arm cross disablecrossgen -portable skiptests", 
+            "PB_AdditionalBuildArguments":"-TargetArchitecture=arm -DistroRid=linux-arm -DisableCrossgen=true -PortableBuild=true -SkipTests=true",
```
ellismg: I looked at how Jenkins was actually invoking the build (based on what build.sh was computing) and moved it here.

config.json
```diff
@@ -34,7 +34,7 @@
       "description": "Specifies the distro rid for Unix OS.",
       "valueType": "property",
       "values": [],
-      "defaultValue": "${OSRid}"
+      "defaultValue": "${OSRid}-${CPUArch}"
```
ellismg: The changes here mean that /p:DistroRid will be set to something like 'osx.10-11-x64' OSX (previously we were passing an empty value), but it looks like that will be OK?  @eerhardt or @gkhanna79 do you have a sense here?

build.sh
```diff
@@ -64,179 +20,4 @@ if [ -z "$HOME" ] || [ ! -d "$HOME" ]; then
...
-# Configure environment if we are doing a verbose build
-if [ $__VerboseBuild == 1 ]; then
-    export VERBOSE=1
-	__RunArgs="$__RunArgs -verbose"
-fi
-
-echo "$__RunArgs"
-$DIR/run.sh build $__RunArgs
\ No newline at end of file
+$DIR/run.sh build "$@"
```
karajas: Looking at the current config.json, does this mean the default here is to not restore, skip tests and always build portable?

ellismg:
@dotnet-bot test ci please

build.sh
ellismg: Maybe I don't understand config.json very well, but I think no:

```js
    "PortableBuild": {
      "description": "Indicates if this is a portable build.",
      "valueType": "property",
      "values": [ "True", "False"],
      "defaultValue": "False"
    },
    "RestoreDuringBuild": {
      "description": "Enables/Disables the restoration of packages.",
      "valueType": "property",
      "values": ["True", "False"],
      "defaultValue": true
    },
    "SkipTests": {
      "description": "Skip running tests",
      "valueType": "property",
      "values": ["true", "false"],
      "defaultValue":false
    },
```
From building locally, restore and testing are happening and the build is not portable by default.  Am I confused about something?

karajas: I thought it would be using the settings specified in https://github.com/dotnet/core-setup/blob/master/config.json#L173-L248

ellismg: I'm not a run/config.json expert by any means, so this could all be incorrect, but I think that section defines two things:

1) Aliases so you can write something like -Portable (or maybe it's just portable, without the dash) and have it mean -PortableBuild=True
2) Defining what parameters are always passed to MSBuild.

I think the default values are taken from the sections I showed above, and can be overwritten either by doing -ParameterName=Value or using one of the aliases.

ellismg: (Confusion like this why I dislike run/config.json since I think the value it provides isn't outweighed by having to learn yet another system, but that's a debate for a different day.)

build.sh
```diff
@@ -4,50 +4,6 @@
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
 
-# Set OFFLINE environment variable to build offline
-
-set -e
```
eerhardt: I think you still want this, right?

ellismg: set -e or OFFLINE?  OFFLINE wasn't used anywhere, but I can add back set -e.

build.sh
karajas: Ah I didn't read the alias brackets correctly you're right.

ellismg:
I used the generator to build some jobs and ran my PR through them, things looked fine:

- [OSX](https://ci.dot.net/job/dotnet_core-setup/job/master/job/GenPRTest/job/release_osx10.12_x64_prtest/1/console)
- [Debian](https://ci.dot.net/job/dotnet_core-setup/job/master/job/GenPRTest/job/debug_debian8.2_x64_prtest/)
- [PortableLinux](https://ci.dot.net/job/dotnet_core-setup/job/master/job/GenPRTest/job/release_portablelinux_x64_prtest/1/console)
- [Arm](https://ci.dot.net/job/dotnet_core-setup/job/master/job/GenPRTest/job/release_ubuntu_arm_prtest/1/)

config.json
eerhardt: Who sets OSRid?  Where does that value come from?

---
In reply to: [115813333](https://github.com/dotnet/core-setup/pull/2342#discussion_r115813333) [](ancestors = 115813333)

build.sh
eerhardt: `set -e` was my concern

---
In reply to: [115821558](https://github.com/dotnet/core-setup/pull/2342#discussion_r115821558) [](ancestors = 115821558)

config.json
ellismg: It's an intrinsic that run understands:

https://github.com/dotnet/buildtools/blob/78260239e4db68ad5c2df774a27d5f9a54fd92e5/src/Run/SettingValueProvider.cs#L72
https://github.com/dotnet/buildtools/blob/78260239e4db68ad5c2df774a27d5f9a54fd92e5/src/Run/Interop.cs#L12

Ideally it would be using M.N.PA but here we are.

build.sh
```diff
@@ -3,51 +3,8 @@
 # Copyright (c) .NET Foundation and contributors. All rights reserved.
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
-
-# Set OFFLINE environment variable to build offline
-
 set -e
```
weshaggard: Is this this interesting? If so should we keep the comment?

config.json
weshaggard: At some point I want to switch away from this DistroRID and use something like RuntimeOS (https://github.com/dotnet/corefx/blob/master/config.json#L33) like we did in corefx. That gets generated as a configuration when doing the build so that we have a consistent way of getting the OS without crazy parsing in various targets/projects.

netci.groovy
```diff
@@ -46,26 +44,25 @@ platformList.each { platform ->
         else if (os == 'Ubuntu16.04') {
             dockerContainer = "ubuntu-16.04-cross-ef0ac75-20175511035548"
         }
-        portableArgs = " -portable cross skiptests disablecrossgen"
         dockerCommand = "docker run --name ${dockerContainer} --rm -v \${WORKSPACE}:${dockerWorkingDirectory} -w=${dockerWorkingDirectory} ${dockerRepository}:${dockerContainer}"
-        buildCommand = "${dockerCommand} ./build.sh ${configuration} ${architecture}${portableArgs}"
+        buildCommand = "${dockerCommand} ./build.sh -ConfigurationGroup=${configuration} -TargetArchitecture=${architecture} -PortableBuild=true -DistroRid=linux-arm -SkipTests=true -DisableCrossgen=true"
```
weshaggard: Should this be `-DistroRid=linux-${architecture}`?

build.sh
ellismg: OFFLINE is not used anywhere else in the repository, so the comment is misleading :-(.

config.json
gkhanna79: @ellismg Have you confirmed this gets the expected value for both cross and non-cross + portable/non-portable builds?

netci.groovy
ellismg: I think that makes sense.

build.sh
weshaggard: I guess I was assuming `set -e` put it into offline mode because I don't know that that does.

config.json
ellismg: When the portable build is run, DistroRid is getting set to rhel.7-x64 (see https://ci.dot.net/job/dotnet_core-setup/job/master/job/GenPRTest/job/release_portablelinux_x64_prtest/1/consoleText) but the packages are being built for the linux-x64 RID (see the output from the packaging task later).

I don't know how in run to say if portable is set, override DistroRid based on that, unfortunately.  I can try to hack that into build.sh but I'm not sure how clean it will be.

Do you know if this is going to be an issue?

build.sh
ellismg: > I guess I was assuming set -e put it into offline mode because I don't know that that does.

Ahh!  What `set -e` means is "if any command in the script file fails, exit immediately instead of continuing."

A nice hygienic thing to stop things early if something you expected would work fails.

config.json
gkhanna79: https://github.com/dotnet/core-setup/pull/2342/files#diff-0b83f9dedf40d7356e5ca147a077acb4L40 is how DistroRid was forced to be portable RID when doing portable build. 

Also, DistroRid appears to be used for both portable and non-portable scenarios, so it will be good to have it in sync with what is expected.

ellismg: > is how DistroRid was forced to be portable RID when doing portable build.

Okay, I can try to think about how I can put that into build.sh while still not going whole hog parsing arguments there.

ellismg:
@gkhanna79 please take a look at 3f68deef131e57c1efebe4b6114d25ceaa10e5a7 and see if that addresses your issue.  That was the cleanest way I could thing to do things.

gkhanna79:
Portable RID computation is done at https://github.com/dotnet/core-setup/blob/master/dir.props#L266. Could you leverage that to reset DistroRid?

ellismg:
> Portable RID computation is done at https://github.com/dotnet/core-setup/blob/master/dir.props#L266. Could you leverage that to reset DistroRid?

We would have to do additional work to reset DistroRid in the case where it was passed in (since it's passed in as a global property which you can not override).

It looks like DistroRid is only ever used in the case where you are not building with portable=true (which is why the the jobs I ran initially passed and generated the correct packages).

So we have two options:

1) Keep this change is for clarity so DistroRid is explicitly set on the command line when calling run.sh
2) Remove this commit.  DistroRid will always be the RID of the host machine, and if you want to use a different rid for some reason (to take portable into account) do what the build system does where you set some other property.

My preference is (2), concretely this would mean we fold this back to a single commit.

gkhanna79:
> Remove this commit.

Are you referring to https://github.com/dotnet/core-setup/commit/3f68deef131e57c1efebe4b6114d25ceaa10e5a7?

ellismg:
Correct.

In answer to your other question I will back at a PC around 9:30, so I can respond to feedback and merge then.

gkhanna79:
Sounds good.

ellismg:
@dotnet-bot test ci please

ellismg:
The windows and ci generation jobs passed, which is the best we can expect from CI until this is merged and the new jobs are generated.

I'm going to merge this now, if it puts the official build on the floor, @gkhanna79, I can try to figure out how to fix it.

Issue #2280, opened 2017-05-04
Ubuntu Arm Builds


olegsavelos:
There was an Ubuntu arm builds available previously.
Anyone aware of where did they go and when we can have them back ?

JensNordenbro:
Also generic linux-arm is missing

chcosta:
@gkhanna79 

Our build system underwent some major changes recently.  We are working on bringing back arm support in the next few days.

JensNordenbro:
generic linux-arm is back. Do you use the same build @olegsavelos ?

olegsavelos:
Generic linux-arm did not work for me , not sure if they even should.
@chcosta any news on when ubuntu builds will be back ?

gkhanna79:
@olegsavelos Where did you pickup the linux-arm build from and why didn't it work for you? Can you please share the failure details?

olegsavelos:
I have tried "Linux (arm) (for glibc based OS)" on ubuntu 16.04 (32bit)
What i get is "cannot execute binary file: Exec format error"
To be frank i am not sure if it should have worked anyway?
I was using previously available builds and they worked but i was still getting "Segmentation fault" when trying ASPNetCore projects.

gkhanna79:
@olegsavelos I presume you are running on Ubuntu Arm32 bit, right?

olegsavelos:
@gkhanna79 Armv71 on  Ubuntu 16.04.2 LTS .

gkhanna79:
We do build with armv7 HF ABI support (see https://github.com/dotnet/core-setup/blob/786a3f40a17b88eecbfb22e504dd37c607856ffa/cross/arm/toolchain.cmake#L10). What  is the ABI of the binaries on the Ubuntu build? Does it use Soft ABI?

@hqueue @hseok-oh @jyoungyun incase they have anything to add as well.

olegsavelos:
its armhf in my case.
This is the output from readelf on dotnet executable 

`root@opiz:/usr/src# readelf -h dotnet
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x40a30b
[...]

PR #2348, opened 2017-05-10
Fix RIDPlat computation

gkhanna79:
While investigating https://github.com/dotnet/core-setup/issues/2280, I noticed that for portable builds, RID computed to be used for HostPolicy servicing was still distro specific (e.g win7-x64) and not portable (e.g. win-x64) that would resulted in inability to lookup serviced hostpolicy binary.

This change fixes it.

@eerhardt PTAL.

### DIFF
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -35,7 +35,7 @@ init_rid_plat()
     fi
 
     if [ "$(uname -s)" == "Darwin" ]; then
-        __rid_plat=osx.10.12
+        export __rid_plat=osx.10.12
     fi
 
     if [ $__linkPortable == 1 ]; then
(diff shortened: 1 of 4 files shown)


src/corehost/build.sh
```diff
@@ -34,9 +34,13 @@ init_rid_plat()
         fi
     fi
 
+    if [ "$(uname -s)" == "Darwin" ]; then
+        __rid_plat=osx.10.12
```
eerhardt: Does this need to be `export`?

gkhanna79: Fixed.

gkhanna79:
@dotnet-bot test PortableLinux x64 Release Build please

PR #2351, opened 2017-05-10
Remove project.json usage to restore locked host and use PackageReference instead.

eerhardt:
@mellinoe @gkhanna79 @weshaggard

### DIFF
--- a/src/sharedFramework/sharedFramework.proj
+++ b/src/sharedFramework/sharedFramework.proj
@@ -23,7 +23,7 @@
     
     <PropertyGroup>
       <SharedFrameworkSourceRoot>$(MSBuildThisFileDirectory)framework</SharedFrameworkSourceRoot>
-      <CommonSharedFrameworkArgs>/p:TargetFramework=$(Framework) /p:RuntimeIdentifier=$(TargetRid) /p:RuntimeFrameworkVersion=$(SharedFrameworkNugetVersion)</CommonSharedFrameworkArgs>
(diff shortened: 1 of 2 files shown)


src/sharedFramework/sharedFramework.proj
```diff
@@ -23,7 +23,7 @@
     
     <PropertyGroup>
       <SharedFrameworkSourceRoot>$(MSBuildThisFileDirectory)framework</SharedFrameworkSourceRoot>
-      <CommonSharedFrameworkArgs>/p:TargetFramework=$(Framework) /p:RuntimeIdentifier=$(TargetRid) /p:RuntimeFrameworkVersion=$(SharedFrameworkNugetVersion)</CommonSharedFrameworkArgs>
+      <CommonSharedFrameworkArgs>/p:TargetFramework=$(Framework) /p:RuntimeIdentifier=$(PackageTargetRid) /p:RuntimeFrameworkVersion=$(SharedFrameworkNugetVersion)</CommonSharedFrameworkArgs>
```
eerhardt: I wasn't 100% sure on which property should be used.  These properties are really close, and I don't understand the difference.  My assumption is `PackageTargetRid` is correct because that's the RID that the packages will be built for.

mellinoe: That sounds correct to me, based on our discussion.

mellinoe:
LGTM, let's see what happens in CI. I should have a change for the rest of the project.json's up shortly after this.

eerhardt:
Uggh - the failures are because of https://github.com/NuGet/Home/issues/4424.  The bane of my existence for the last few months.

I'll make a hack fix.

PR #2355, opened 2017-05-11
Add OfficialBuildId to config.json

karajas:
Breaking change to official builds
skip ci please

/cc @ellismg


ellismg:
Thanks for fixing this!

PR #2365, opened 2017-05-11
Add doc files to framework packages

ericstj:
I've also removed the de-duping in netcoreapp since it's no longer needed.

/cc @joperezr @weshaggard


ericstj:
@joperezr we don't need de-duping in the NETCoreApp package as it has no duplicates.  UWP still has duplicates so I've left that logic there.

joperezr:
I see, makes sense.

ericstj:
FYI, this grows the MS.NETCore.App package by 4x.  Previously it was 2.5 MB, with this change it is 11MB

PR #2368, opened 2017-05-11
Add checksum generation

karajas:
@eerhardt @chcosta 

This change will introduce checksum generation for core-setup.

### DIFF
--- a/buildpipeline/Core-Setup-Linux-BT.json
+++ b/buildpipeline/Core-Setup-Linux-BT.json
@@ -177,7 +177,7 @@
       },
       "inputs": {
         "filename": "docker",
-        "arguments": "run --privileged --rm $(DockerCommonRunArgs) $(PB_GitDirectory)/Tools/msbuild.sh $(PB_GitDirectory)/publish/publish.proj /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:DistroRid=$(PB_DistroRid) /p:TargetArchitecture=$(PB_TargetArchitecture) /p:ConfigurationGroup=$(BuildConfiguration) /p:PortableBuild=$(PB_PortableBuild) /p:OSGroup=Linux /p:DebRepoUser=$(PB_DebRepoUser) /p:DebRepoServer=$(PB_DebRepoServer) /p:DebRepoPass=$(DEB_REPO_PASSWORD) /p:DebianId_ubuntu1404-x64=$(PB_DebianId_ubuntu1404-x64) /p:DebianId_debian8-x64=$(PB_DebianId_debian8-x64) /p:DebianId_ubuntu1604-x64=$(PB_DebianId_ubuntu1604-x64) /p:DebianId_ubuntu1610-x64=$(PB_DebianId_ubuntu1610-x64)",
+        "arguments": "run --privileged --rm $(DockerCommonRunArgs) $(PB_GitDirectory)/Tools/msbuild.sh $(PB_GitDirectory)/publish/publish.proj /p:AzureAccountName=$(PB_AzureAccountName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:ChecksumAzureAccountName=$(PB_ChecksumAzureAccountName) /p:ChecksumAzureAccessToken=$(PB_ChecksumAzureAccessToken) /p:DistroRid=$(PB_DistroRid) /p:TargetArchitecture=$(PB_TargetArchitecture) /p:ConfigurationGroup=$(BuildConfiguration) /p:PortableBuild=$(PB_PortableBuild) /p:OSGroup=Linux /p:DebRepoUser=$(PB_DebRepoUser) /p:DebRepoServer=$(PB_DebRepoServer) /p:DebRepoPass=$(DEB_REPO_PASSWORD) /p:DebianId_ubuntu1404-x64=$(PB_DebianId_ubuntu1404-x64) /p:DebianId_debian8-x64=$(PB_DebianId_debian8-x64) /p:DebianId_ubuntu1604-x64=$(PB_DebianId_ubuntu1604-x64) /p:DebianId_ubuntu1610-x64=$(PB_DebianId_ubuntu1610-x64)",
         "workingFolder": "$(PB_SourcesDirectory)",
         "failOnStandardError": "false"
       }
@@ -310,6 +310,13 @@
       "isSecret": true
     },
     "PB_DebianId_ubuntu1610-x64": {
+      "value": null,
+      "isSecret": true
+    },
+    "PB_ChecksumAzureAccountName": {
+      "value": "dotnetclichecksums"
+    },
+    "PB_ChecksumAzureAccessToken": {
       "value": null,
       "isSecret": true
     },
--- a/publish/dir.props
+++ b/publish/dir.props
@@ -4,8 +4,13 @@
 
   <PropertyGroup>
     <ContainerName Condition="'$(ContainerName)' == ''">dotnet</ContainerName>
+    <ChecksumContainerName Condition="'$(ChecksumContainerName)' == ''">dotnet</ChecksumContainerName>
     <BaseUrl Condition="'$(BaseUrl)' == ''">https://dotnetcli.blob.core.windows.net/</BaseUrl>
     <NuGetPushTimeoutSeconds Condition="'$(NuGetPushTimeoutSeconds)' == ''">3600</NuGetPushTimeoutSeconds>
+    <BinariesRelativePath>$(Channel)/Binaries/$(SharedFrameworkNugetVersion)</BinariesRelativePath>
+    <InstallersRelativePath>$(Channel)/Installers/$(SharedFrameworkNugetVersion)</InstallersRelativePath>
+    <ChecksumsRelativePath>Runtime/Checksums/$(ProductVersion)</ChecksumsRelativePath>
+    <ChecksumExtension>.sha512<ChecksumExtension>
   </PropertyGroup>
   
   <PropertyGroup>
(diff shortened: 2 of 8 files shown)


tools-local/Microsoft.DotNet.Build.Tasks.Local/GenerateChecksums.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                    Log.LogMessage(
+                        MessageImportance.High,
+                        "Generating checksum for '{0}' into '{1}'...",
+                        item.ItemSpec,
+                        destinationPath);
+
+                    using (FileStream stream = File.OpenRead(item.ItemSpec))
+                    {
+                        HashAlgorithm hashAlgorithm = SHA512.Create();
```
ellismg: HashAglorithm's are Dispoasable, so this should be wrapped in a using.

publish/dir.targets
```diff
@@ -1,33 +1,52 @@
...
+        <AdditionalProperties>%(Project.AdditionalProperties);ItemsToPublish=@(ChecksumItems);RelativeBlobPath=$(ChecksumsRelativePath)</AdditionalProperties>
+      </Project>
+    </ItemGroup>
+  </Target>
+  
+  <Target Name="CreateChecksums">
+    <ItemGroup>
+      <ItemsToChecksum Include="@(CompressedFile)" />
+      <ItemsToChecksum Include="@(RuntimePackageFile)" />
+      <ItemsToChecksum Include="@(IdentityPackageFile)" />
```
ellismg: I am changing the name of this property in a pending PR, so we need to be careful about how this lands.

publish/dir.targets
```diff
@@ -1,33 +1,52 @@
...
+      <ItemsToChecksum Include="@(InstallerFile)" />
+      <ArtifactsForGeneratingChecksums Include="@(ItemsToChecksum)">
+        <DestinationPath>%(ItemsToChecksum.FullPath).sha</DestinationPath>
```
ellismg: Should we have this be .sha512 so it's clear we are using sha512?

publish/dir.targets
ellismg: I think we will want to make this conditional so it only happens if we are going publish the Identity/RidAgnostic packages.  Otherwsie we will end up publishing SHAs for all the *nix versions of the stuff that's getting built in #2370

tools-local/tasks/FinalizeBuild.cs
```diff
@@ -97,6 +97,8 @@ public override bool Execute()
                     CopyBlobs($"{Channel}/Installers/{SharedFrameworkNugetVersion}", $"{Channel}/Installers/Latest/");
 
                     CopyBlobs($"{Channel}/Installers/{SharedHostNugetVersion}", $"{Channel}/Installers/Latest/");
+                    
+                    CopyBlobs($"{Channel}/Checksums/{SharedFrameworkNugetVersion}", $"{Channel}/Checksums/Latest/");
```
eerhardt: The checksums have to go to a separate blob storage.  It doesn't make sense to put the checksums and the bits in the same place.  If that one place gets compromised, the checksums are no good.

publish/dir.props
```diff
@@ -5,6 +5,9 @@
   <PropertyGroup>
     <ContainerName Condition="'$(ContainerName)' == ''">dotnet</ContainerName>
     <NuGetPushTimeoutSeconds Condition="'$(NuGetPushTimeoutSeconds)' == ''">3600</NuGetPushTimeoutSeconds>
+    <BinariesRelativePath>$(Channel)/Binaries/$(SharedFrameworkNugetVersion)</BinariesRelativePath>
```
eerhardt: This should all be `$(ProductVersion)`.

I'm still not sure why we have both, but when we are talking generally about the build number/version, use `$(ProductVersion)`.  When you are specifically talking about the SharedFramework or MS.NC.App, use `$(SharedFrameworkNuGetVersion)`

tools-local/tasks/FinalizeBuild.cs
eerhardt: Actually, checksums should not be necessary in 'Finalize' at all.  This should be removed.

---
In reply to: [116110189](https://github.com/dotnet/core-setup/pull/2368#discussion_r116110189) [](ancestors = 116110189)

publish/dir.targets
```diff
@@ -1,33 +1,52 @@
...
+        <AdditionalProperties>%(Project.AdditionalProperties);ItemsToPublish=@(InstallerFile);RelativeBlobPath=$(InstallersRelativePath)</AdditionalProperties>
       </Project>
     </ItemGroup>
   </Target>
+  
+  <Target Name="GatherPublishToAzureChecksums" DependsOnTargets="CreateChecksums" >
+    <ItemGroup>
+      <ChecksumItems Include="$(PackagesOutDir)**/*.sha" />
+      <Project Include="publish.proj">
+        <PublishTarget>UploadToAzure</PublishTarget>
```
eerhardt: As stated in another place, the checksums need to go into a separate storage account.  See in the CLI where and how it is done there.

publish/dir.props
```diff
@@ -5,6 +5,9 @@
...
+    <InstallersRelativePath>$(Channel)/Installers/$(SharedFrameworkNugetVersion)</InstallersRelativePath>
+    <ChecksumsRelativePath>$(Channel)/Checksums/$(SharedFrameworkNugetVersion)</ChecksumsRelativePath>
```
eerhardt: @mellinoe will be working next week on changing our storage structure.  I think we can start making that change with the checksums now.  Sync with Eric M and me about it.  that way we don't need to redo this work next week.

tools-local/tasks/FinalizeBuild.cs
karajas: Yeah I was wondering whether we should have an explicit checksum storage account, will ping you

publish/dir.targets
karajas: Will update

publish/dir.targets
karajas: Noted.

karajas: Agreed.

karajas:
PTAL

karajas:
PTAL @ellismg @eerhardt

ellismg:
Looks good to me, but I'd let Eric take a look as well.

buildpipeline/Core-Setup-Linux-BT.json
```diff
@@ -313,6 +313,14 @@
       "value": null,
       "isSecret": true
     },
+    "PB_ChecksumAzureAccountName": {
+      "value": null,
+      "isSecret": true
```
eerhardt: Not that it really matters, but the account name isn't really secret, right?

publish/dir.props
```diff
@@ -4,8 +4,13 @@
 
   <PropertyGroup>
     <ContainerName Condition="'$(ContainerName)' == ''">dotnet</ContainerName>
+    <ChecksumContainerName Condition="'$(ChecksumContainerName)' == ''">dotnet</ChecksumContainerName>
     <BaseUrl Condition="'$(BaseUrl)' == ''">https://dotnetcli.blob.core.windows.net/</BaseUrl>
     <NuGetPushTimeoutSeconds Condition="'$(NuGetPushTimeoutSeconds)' == ''">3600</NuGetPushTimeoutSeconds>
+    <BinariesRelativePath>$(Channel)/Binaries/$(ProductVersion)</BinariesRelativePath>
+    <InstallersRelativePath>$(Channel)/Installers/$(ProductVersion)</InstallersRelativePath>
+    <ChecksumsRelativePath>$(Channel)/Checksums/$(ProductVersion)</ChecksumsRelativePath>
```
eerhardt: Were we going to put the checksum layout in the new structure to begin with?  That way @mellinoe doesn't need to change it on the next checkin.

publish/dir.targets
```diff
@@ -1,33 +1,52 @@
...
+        <AdditionalProperties>%(Project.AdditionalProperties);ItemsToPublish=@(ChecksumItems);RelativeBlobPath=$(ChecksumsRelativePath)</AdditionalProperties>
       </Project>
     </ItemGroup>
   </Target>
+  
+  <Target Name="CreateChecksums">
+    <ItemGroup>
+      <ItemsToChecksum Include="@(CompressedFile)" />
+      <ItemsToChecksum Include="@(RuntimePackageFile)" />
+      <ItemsToChecksum Include="@(IdentityPackageFile)" Condition="'$(PublishRidAgnosticPackages)' == 'true'" />
```
eerhardt: I don't think we need the nupkg's checksum'd.  No one downloads them from our storage account.  The Sdk doesn't have checksums for their nukgs either.

We just need archives and installers.

publish/publish.proj
```diff
@@ -23,7 +23,7 @@
 
   <Target Name="PublishToAzure"
           Condition="'$(Finalize)' != 'true'"
-          DependsOnTargets="GatherPublishToAzureBinaries;GatherPublishToAzureInstallers">
+          DependsOnTargets="GatherPublishToAzureBinaries;GatherPublishToAzureInstallers;GatherPublishToAzureChecksums">
```
eerhardt: I know this pattern is already existing, but I find it quiet odd and confusing that publish.proj depends on this target in dir.targets, which then calls back into publish.proj.  It just seems too circular and over-engineered for what needs to be done.

You don't need to fix it in this PR, but I just wanted to state my opinion on this pattern.

ellismg:
@karajas I just merged my change in, so please rebase and merge the property name updates.

publish/publish.proj
karajas: The only benefit I can see from this was reuse of the same parameters but we don't need that anymore.

karajas:
PTAL @eerhardt I removed the extra files being included for checksums

publish/dir.props
```diff
@@ -14,13 +19,27 @@
...
+    <CompressedFile Include="$(PackagesOutDir)**/*$(CompressedFileExtension)">
+      <RelativeBlobPath>$(BinariesRelativePath)</RelativeBlobPath>
+    </CompressedFile>
+    <RuntimePackageFile Include="$(PackagesOutDir)**/runtime.*.nupkg" >
+      <RelativeBlobPath>$(BinariesRelativePath)</RelativeBlobPath>
+    </RuntimePackageFile>
+    <RuntimePackageFile Include="$(PackagesOutDir)**/runtime.*.symbols.nupkg" >
+      <RelativeBlobPath>$(BinariesRelativePath)</RelativeBlobPath>
+    </RuntimePackageFile>
+    <RidAgnosticPackageFile Condition="'$(PublishRidAgnosticPackages)' == 'true'" Include="$(PackagesOutDir)**/*.nupkg" Exclude="@(RuntimePackageFile)" >
```
ellismg: This being conditional on PublishRidAgnosticPackages being true seems weird.  Instead, I would always add stuff to the RidAgnosticsPackageFile ItemGroup but only consume it when PublishRidAgnosticPackages is set to true.

publish/dir.props
```diff
@@ -4,8 +4,13 @@
 
   <PropertyGroup>
     <ContainerName Condition="'$(ContainerName)' == ''">dotnet</ContainerName>
+    <ChecksumContainerName Condition="'$(ChecksumContainerName)' == ''">dotnet</ChecksumContainerName>
     <BaseUrl Condition="'$(BaseUrl)' == ''">https://dotnetcli.blob.core.windows.net/</BaseUrl>
     <NuGetPushTimeoutSeconds Condition="'$(NuGetPushTimeoutSeconds)' == ''">3600</NuGetPushTimeoutSeconds>
+    <BinariesRelativePath>Runtime/$(ProductVersion)/$(Channel)</BinariesRelativePath>
```
eerhardt: The "channel" should not be in the path.

Also, don't change the binaries and the installer paths here.  @mellinoe is doing that.  I was just talking about the checksum paths being in the new structure.

publish/dir.props
karajas: Removed.

publish/dir.props
karajas: Changed it back.

karajas:
PTAL

ellismg:
I'm happy with this, if @eerhardt is.

publish/dir.props
```diff
@@ -4,8 +4,13 @@
 
   <PropertyGroup>
     <ContainerName Condition="'$(ContainerName)' == ''">dotnet</ContainerName>
+    <ChecksumContainerName Condition="'$(ChecksumContainerName)' == ''">dotnet</ChecksumContainerName>
     <BaseUrl Condition="'$(BaseUrl)' == ''">https://dotnetcli.blob.core.windows.net/</BaseUrl>
     <NuGetPushTimeoutSeconds Condition="'$(NuGetPushTimeoutSeconds)' == ''">3600</NuGetPushTimeoutSeconds>
+    <BinariesRelativePath>$(Channel)/Binaries/$(SharedFrameworkNugetVersion)</BinariesRelativePath>
+    <InstallersRelativePath>$(Channel)/Installers/$(SharedFrameworkNugetVersion)</InstallersRelativePath>
+    <ChecksumsRelativePath>Runtime/Checksums/$(ProductVersion)</ChecksumsRelativePath>
```
eerhardt: The Sdk checksums don't have a "Checksums" folder.  Let's keep it consistent between the Runtime and the SDK.  This should just be 

`<ChecksumsRelativePath>Runtime/$(ProductVersion)</ChecksumsRelativePath>`

publish/dir.targets
```diff
@@ -1,33 +1,50 @@
...
     </ItemGroup>
   </Target>
+  
+  <Target Name="CreateChecksums">
+    <ItemGroup>
+      <ItemsToChecksum Include="@(CompressedFile)" />
+      <ItemsToChecksum Include="@(InstallerFile)" />
+      <ArtifactsForGeneratingChecksums Include="@(ItemsToChecksum)">
+        <DestinationPath>%(ItemsToChecksum.FullPath)$(ChecksumExtension)</DestinationPath>
+        <RelativeBlobPath>%(ItemsToChecksum.RelativeBlobPath)$(ChecksumExtension)</RelativeBlobPath>
```
eerhardt: Is this metadata used at all?  I don't see where.

publish/publish.proj
```diff
@@ -49,6 +49,26 @@
                    Items="@(ItemsToPublish)" 
                    Overwrite="false" />
   </Target>
+  
+  <Target Name="UploadChecksumsToAzure"
```
eerhardt: The SDK also updates the "channel" during finalize build so we can always have a correct "latest" link.  However, given the current state of things, this would probably be easiest done after @mellinoe's change.  So I'm fine with the first check-in for this feature to not account for "finalize build".  But once #2385 is merged, we should update the checksums to be 'finalized' as well.

karajas: I agree, I will follow up on this

PR #2369, opened 2017-05-11
Add target to validate runtime-specific packages

### LINKED ISSUE #1755: Add validation in NETCore.App that RID-specific assets are present
Nothing in NuGet nor MSBuild makes sure a RID-specific asset is present.  We used to be able to rely on the matching REF/LIB check (AKA: gaurdrails) but that has been removed in NuGet 4.

We should add our own check in our targets for .NETCore.App to make sure the RID specific runtime bits are present.

/cc @weshaggard @Petermarcu

ericstj:
Fixes #1755

We no longer have guardrails to make sure runtime matches compile time.  This implements a smaller check to ensure that MS.NETCore.App has some runtime-specific package brought in.

I've made the check suppressible.

Note that I have not made the change to UWP since it is still on project.json-based tools. 

/cc @Petermarcu @joperezr

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
+++ b/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="EnsureNETCoreAppRuntime" 
+          Condition="'$(RuntimeIdentifier)' != '' AND '$(EnsureNETCoreAppRuntime)' != 'false'" 
+          AfterTargets="RunResolvePackageDependencies">
+    <ItemGroup>
+      <_netCoreAppPackageDependencies 
+          Include="@(PackageDependencies->WithMetadataValue('ParentPackage', '$(MSBuildThisFileName)/$(_Microsoft_NETCore_App_Version)'))" />
+      <_activeRIDNetCoreAppPackageDependencies 
+          Include="@(_netCoreAppPackageDependencies->WithMetadataValue('ParentTarget', '$(NuGetTargetMoniker)/$(RuntimeIdentifier)'))" />
+      <_activeTFMNetCoreAppPackageDependencies 
+          Include="@(_netCoreAppPackageDependencies->WithMetadataValue('ParentTarget', '$(NuGetTargetMoniker)'))" />
+      <_ridSpecificNetCoreAppPackageDependencies 
+          Include="@(_activeRIDNetCoreAppPackageDependencies)" 
+          Exclude="@(_activeTFMNetCoreAppPackageDependencies)" />
+    </ItemGroup>
+    
+    <Error Condition="'@(_ridSpecificNetCoreAppPackageDependencies)' == ''"
+           Text="Project is targeting runtime '$(RuntimeIdentifier)' but did not resolve any runtime-specific packages for the '$(MSBuildThisFileName)' package.  This runtime may not be supported by .NET Core." />
+  </Target>
+</Project>
--- a/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -53,6 +53,9 @@
     <File Include="$(PlatformManifestFile)">
       <TargetPath>build/$(NETCoreAppFramework)</TargetPath>
     </File>
+    <File Include="$(MSBuildProjectName).targets">
+      <TargetPath>build/$(NETCoreAppFramework)</TargetPath>
+    </File>
   </ItemGroup>
 
   <!-- Redistributed package content from other nuget packages-->
@@ -75,6 +78,7 @@
     
     <GenerateFileVersionProps Files="@(SharedFrameworkRuntimeFiles)"
                               PackageId="$(Id)"
+                              PackageVersion="$(Version)"
                               PlatformManifestFile="$(PlatformManifestFile)"
                               PropsFile="$(PropsFile)"
                               PreferredPackages="$(Id);@(RuntimeDependency)" />
--- a/src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
+++ b/src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
@@ -61,6 +61,7 @@
     
     <GenerateFileVersionProps Files="@(UapRuntimeFiles)"
                               PackageId="$(Id)"
+                              PackageVersion="$(Version)"
                               PlatformManifestFile="$(PlatformManifestFile)"
                               PropsFile="$(PropsFile)"
                               PreferredPackages="$(Id);@(RuntimeDependency)" />
--- a/tools-local/tasks/GenerateFileVersionProps.cs
+++ b/tools-local/tasks/GenerateFileVersionProps.cs
@@ -22,6 +22,9 @@ namespace Microsoft.DotNet.Build.Tasks
         public string PackageId { get; set; }
 
         [Required]
+        public string PackageVersion { get; set; }
+
+        [Required]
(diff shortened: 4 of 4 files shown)


src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="EnsureNETCoreAppRuntime" 
+          Condition="'$(RuntimeIdentifier)' != '' AND '$(EnsureNETCoreAppRuntime)' != 'false'" 
+          AfterTargets="RunResolvePackageDependencies">
```
ericstj: Runs after RunResolvePackageDependencies since that has produces PackageDependencies item which I'm reading and has the right condition to avoid running in design time.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
```diff
@@ -0,0 +1,22 @@
...
+    <ItemGroup>
+      <_netCoreAppPackageDependencies 
+          Include="@(PackageDependencies->WithMetadataValue('ParentPackage', '$(MSBuildThisFileName)/$(_Microsoft_NETCore_App_Version)'))" />
+      <_activeRIDNetCoreAppPackageDependencies
```
ericstj: I've created my own items and avoided using all items/properties that do similar things in Microsoft.PackageDependencyResolution.targets since those are all private ('_').

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
```diff
@@ -0,0 +1,22 @@
...
+          Include="@(_netCoreAppPackageDependencies->WithMetadataValue('ParentTarget', '$(NuGetTargetMoniker)/$(RuntimeIdentifier)'))" />
+      <_activeTFMNetCoreAppPackageDependencies 
+          Include="@(_netCoreAppPackageDependencies->WithMetadataValue('ParentTarget', '$(NuGetTargetMoniker)'))" />
+      <_ridSpecificNetCoreAppPackageDependencies 
+          Include="@(_activeRIDNetCoreAppPackageDependencies)" 
+          Exclude="@(_activeTFMNetCoreAppPackageDependencies)" />
+    </ItemGroup>
+    
+    <Error Condition="'@(_ridSpecificNetCoreAppPackageDependencies)' == ''"
+           Text="Project is targeting runtime '$(RuntimeIdentifier)' but did not resolve any runtime-specific packages for the '$(MSBuildThisFileName)' package.  This runtime may not be supported by .NET Core." />
```
ericstj: Please review this text.

ericstj:
Here's and example of the way this fails:
```
F:\standaloneTest>dotnet build /p:RuntimeIdentifier=foo-x86
Microsoft (R) Build Engine version 15.3.117.23532
Copyright (C) Microsoft Corporation. All rights reserved.

C:\Users\ericstj\.nuget\packages\microsoft.netcore.app\2.0.0-preview2-25302-9\build\netcoreapp2.0\Microsoft.NETCore.App.targets(19,5): error : Project is targeting runtime 'foo-x86' but did not resolve any runtime-specific packages for the 'Microsoft.NETCore.App' package.  This runtime may not be supported by .NET Core. [F:\standaloneTest\standaloneTest.csproj]

Build FAILED.
```

joperezr:
> Note that I have not made the change to UWP since it is still on project.json-based tools.

not sure what you mean by that, isn't netcoreapp also using the project.json.template for the depproj?

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
joperezr: Why did you choose an error instead of  just a warning? Is it because that was the behaviour in the project.json world?

ericstj:
> not sure what you mean by that, isn't netcoreapp also using the project.json.template for the depproj

I meant that folks consuming UWP aren't using PackageReference and I'm not even sure we have new templates that'd do that.  Once the UWP v.Next projects and tooling starts lining up we can see if they need a similar target.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
ericstj: IMO if we know 100% something is going to fail in a blocking way later but can provide a better blocking error early that is preferable.

Folks always focus on the blocking error and will often miss a warning which occurred earlier.

If we can come up with any scenario where someone would want to build with a RID and not care that they are missing the runtime package then I'd relax this to a warning but I can't come up with that.  I considered the case of RID-specific non-selfcontained, but even there we'll get the RID-specific package and trim it out later.

joperezr:
makes sense.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="EnsureNETCoreAppRuntime" 
+          Condition="'$(RuntimeIdentifier)' != '' AND '$(EnsureNETCoreAppRuntime)' != 'false'" 
+          AfterTargets="RunResolvePackageDependencies">
+    <ItemGroup>
+      <_netCoreAppPackageDependencies 
+          Include="@(PackageDependencies->WithMetadataValue('ParentPackage', '$(MSBuildThisFileName)/$(_Microsoft_NETCore_App_Version)'))" />
```
weshaggard: It isn't clear to me where this gets _Microsoft_NETCore_App_Version substituted. I assume it isn't included in the SDK targets.

weshaggard: nevermind I think I figured it out. YOu are putting it into your props file which you include in this package.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
```diff
@@ -0,0 +1,22 @@
...
+      <_activeRIDNetCoreAppPackageDependencies 
+          Include="@(_netCoreAppPackageDependencies->WithMetadataValue('ParentTarget', '$(NuGetTargetMoniker)/$(RuntimeIdentifier)'))" />
+      <_activeTFMNetCoreAppPackageDependencies 
+          Include="@(_netCoreAppPackageDependencies->WithMetadataValue('ParentTarget', '$(NuGetTargetMoniker)'))" />
+      <_ridSpecificNetCoreAppPackageDependencies 
+          Include="@(_activeRIDNetCoreAppPackageDependencies)" 
+          Exclude="@(_activeTFMNetCoreAppPackageDependencies)" />
+    </ItemGroup>
+    
+    <Error Condition="'@(_ridSpecificNetCoreAppPackageDependencies)' == ''"
```
weshaggard: What about cases like Tizen that might not follow the same naming conventions? Would their ParentPackage be correctly set to the identity package?

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
ericstj: That is written to the props file for this package.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.targets
ericstj: This target file only gets included for MS.NETCore.App.  For cases like Tizen I'd expect them to parent that package in their runtime.json, even if they provide runtime.json in a separate package.  Otherwise they can't replace our runtime mappings.

weshaggard: OK I wasn't sure if ParentPackage could be controlled by an external runtime.json.

ericstj: Yeah, there are no constraints on runtime.json.

PR #2373, opened 2017-05-11
Update to Buildtools 2.0, Remove remaining project.json's

mellinoe:
This updates buildtools to version 2.0, and removes the 1.x CLI from all parts of the build. The build is now run on .NET Core 2.0. I have replaced the contents of init-tools.cmd/sh with the versions from corefx. I have deleted everything in Tools-Override, as that stuff should be part of buildtools now.

All project.json's should now be deleted from the repository. Projects previously using those now use regular csproj's with `<PackageReference>` items.

A few kinds of changes were needed in some targets/project files:
* PackageReferences are based on properties from dependencies.props. I added some extra properties there which should get auto-updated.
* All packages are now restored to folders with lower-case names. This required adding some `.ToLowerInvariant` calls.
* MSBuild on .NET Core 2.0 has some issue with calls to `TrimEnd(char[])`. We had to make similar changes in other repos using 2.0 buildtools.
* Another call to `chmod` is needed in our crossgen step; same reason as the others.

~~@eerhardt There is still some issue with the NuGet fallback folder. I want to see what happens in CI, but perhaps you have some idea about why that folder is getting populated again, given we should be disabling the first-time experience everywhere.~~ Nevermind, this was just a simple mistake.


mellinoe:
@dotnet-bot Test PortableLinux x64 Release Build

mellinoe:
@eerhardt I fixed the remaining issue, everything should be good to go now.

mellinoe:
@dotnet-bot Test PortableLinux x64 Release Build

dependencies.props
```diff
@@ -47,17 +50,18 @@
     
     <XmlUpdateStep Include="CoreFx">
       <Path>$(MSBuildThisFileFullPath)</Path>
-      <ElementName>PlatformPackageVersion</ElementName>
-      <PackageId>Microsoft.NETCore.Platforms</PackageId>
+      <ElementName>CoreFxExpectedPrerelease</ElementName>
```
ellismg: Can you explain to me what this part of the diff does?  Is it no longer looked at Latest_Packages.txt and instead looking at something else (I ask because I depend on this working and if this triggers it looking at another file I need to augment my local tasks in the build from source meta repository to handle that.

init-tools.cmd
```diff
@@ -34,21 +29,15 @@ if exist "%BUILD_TOOLS_SEMAPHORE%" (
...
-:: core-setup repo has a global.json which causes a conflict with the init-tools publish step
-if NOT exist "%PACKAGES_DIR%" mkdir %PACKAGES_DIR% > nul 2>&1
-echo { "projects": [] } > %DUMMY_GLOBAL_JSON_PATH%
+if exist "%DOTNET_CMD%" goto :afterdotnetrestore
 
 echo Installing dotnet cli...
 if NOT exist "%DOTNET_PATH%" mkdir "%DOTNET_PATH%"
 set DOTNET_ZIP_NAME=dotnet-dev-win-x64.%DOTNET_VERSION%.zip
-set DOTNET_REMOTE_PATH=https://dotnetcli.azureedge.net/dotnet/preview/Binaries/%DOTNET_VERSION%/%DOTNET_ZIP_NAME%
+set DOTNET_REMOTE_PATH=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/%DOTNET_VERSION%/%DOTNET_ZIP_NAME%
```
ellismg: Was the change in hostname intentional?  I think the old one was to use the CDN which was faster for remote folks?

mellinoe: This is what corefx uses, but I can revert that specific part.

ellismg: If the answer is "I just copied this file from corefx, maybe we can upstream the fix there?"

init-tools.sh
```diff
@@ -123,7 +61,7 @@ if [ ! -e $__INIT_TOOLS_DONE_MARKER ]; then
             cp -r $DOTNET_TOOL_DIR/* $__DOTNET_PATH
         else
             echo "Installing dotnet cli..."
-            __DOTNET_LOCATION="https://dotnetcli.azureedge.net/dotnet/preview/Binaries/${__DOTNET_TOOLS_VERSION}/${__DOTNET_PKG}.${__DOTNET_TOOLS_VERSION}.tar.gz"
+            __DOTNET_LOCATION="https://dotnetcli.blob.core.windows.net/dotnet/Sdk/${__DOTNET_TOOLS_VERSION}/${__DOTNET_PKG}.${__DOTNET_TOOLS_VERSION}.tar.gz"
```
ellismg: Same feedback as above.

init-tools.cmd
ellismg: Let me just try to submit a fix there instead :-)

dependencies.props
mellinoe: This should just cause the "CoreFxExpectedPrerelease" property to be updated, instead of the "PlatformPackageVersions" property. The latter is now just based on the former.

ellismg: How is `BuildInfoName` used?

src/pkg/deps/deps.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dependencies.props))\dependencies.props" />
+
+  <PropertyGroup>
+    <TargetFramework>dnxcore50</TargetFramework>
```
ellismg: Dumb question, I'm sure, but why is this stuff dnxcore50 with this crazy fallback?  Can we not just say the targetframework is netstandard1.0?

dependencies.props
mellinoe: Not sure -- I copied that part from corefx. @dagood ?

src/pkg/deps/deps.csproj
mellinoe: We probably can -- this was just part of the `dotnet migrate` output.

ellismg: Ahh, might be nice to just clean up if possible.

dir.props
```diff
@@ -131,39 +127,25 @@
...
     <DotnetToolCommand Condition="'$(DotnetToolCommand)'=='' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotnetToolCommand>
-    <DotnetSdkToolCommand Condition="'$(DotnetSdkToolCommand)'=='' and '$(OsEnvironment)'!='Unix'">$(DotnetSdkPath)dotnet.exe</DotnetSdkToolCommand>
-    <DotnetSdkToolCommand Condition="'$(DotnetSdkToolCommand)'=='' and '$(OsEnvironment)'=='Unix'">$(DotnetSdkPath)dotnet</DotnetSdkToolCommand>
     <OverrideToolHost Condition="'$(MSBuildRuntimeType)' == 'Core'">$(DotnetToolCommand)</OverrideToolHost>
     <DnuRestoreSource>@(DnuSourceList -> '--source %(Identity)', ' ')</DnuRestoreSource>
     <DnuRestorePrefix Condition="'$(OsEnvironment)' != 'Windows_NT'">DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1</DnuRestorePrefix>
     <DnuRestoreCommand>$(DnuRestorePrefix) "$(DotnetToolCommand)"</DnuRestoreCommand>
     <DnuRestoreCommand>$(DnuRestoreCommand) restore</DnuRestoreCommand>
-    <DnuRestoreCommand>$(DnuRestoreCommand) --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))" $(DnuRestoreSource)</DnuRestoreCommand>
+    <DnuRestoreCommand>$(DnuRestoreCommand) --packages "$(PackagesDir.TrimEnd('/').TrimEnd('\'))" $(DnuRestoreSource)</DnuRestoreCommand>
```
eerhardt: Can we now take the opportunity to rename these properties to not use "Dnu"?  That product has been dead for over a year and a half.

src/sharedFramework/sharedFramework.proj
```diff
@@ -28,11 +28,11 @@
 
     <RemoveDir Directories="$(SharedFrameworkNameAndVersionRoot)" />
 
-    <Exec Command="$(SdkRestoreCommand) --source $(PackagesOutDir) $(CommonSharedFrameworkArgs)"
+    <Exec Command="$(DnuRestoreCommand) --source $(PackagesOutDir) $(CommonSharedFrameworkArgs)"
```
eerhardt: Again, let's not propagate "Dnu" all over our code.

init-tools.cmd
eerhardt: I specifically changed this in core-setup to use `azureedge`.  We need to use that URL because it is a CDN that is cached.  Thus, if you don't live on the west coast of the US, your download times speed up greatly when using `azureedge`.

Yes, this should propagate to all other repos who are doing it wrong.

In all honesty, this code (and all other repos) should be using the `dotnet-install` script, which already uses `azureedge`.

dependencies.props
eerhardt: Let's not use pre-release properties.  They make it hard to handle servicing and RTM packages.  See @dagood's last paragraph in https://github.com/dotnet/core-setup/pull/2243#issuecomment-298700379.

eerhardt: Instead, just have two separate properties which get updated appropriately:

`<CoreFxVersion>4.4.0-preview2-25310-01</CoreFxVersion>`
`<PlatformPackageVersion>2.0.0-preview2-25310-01</PlatformPackageVersion>`

These two things (especially in servicing releases) can and will have different "pre-release" versions.  Say when we just change the CoreFx package, but keep the PlatformPackage unchanged.

init-tools.cmd
```diff
@@ -57,29 +46,12 @@ if NOT exist "%DOTNET_LOCAL_PATH%" (
...
-set DOTNET_LOCAL_PATH=%DOTNET_SDK_PATH%%DOTNET_ZIP_NAME%
-echo Installing '%DOTNET_REMOTE_PATH%' to '%DOTNET_LOCAL_PATH%' >> "%INIT_TOOLS_LOG%"
-powershell -NoProfile -ExecutionPolicy unrestricted -Command "$retryCount = 0; $success = $false; do { try { (New-Object Net.WebClient).DownloadFile('%DOTNET_REMOTE_PATH%', '%DOTNET_LOCAL_PATH%'); $success = $true; } catch { if ($retryCount -ge 6) { throw; } else { $retryCount++; Start-Sleep -Seconds (5 * $retryCount); } } } while ($success -eq $false); Add-Type -Assembly 'System.IO.Compression.FileSystem' -ErrorVariable AddTypeErrors; if ($AddTypeErrors.Count -eq 0) { [System.IO.Compression.ZipFile]::ExtractToDirectory('%DOTNET_LOCAL_PATH%', '%DOTNET_SDK_PATH%') } else { (New-Object -com shell.application).namespace('%DOTNET_SDK_PATH%').CopyHere((new-object -com shell.application).namespace('%DOTNET_LOCAL_PATH%').Items(),16) }" >> "%INIT_TOOLS_LOG%"
-if NOT exist "%DOTNET_LOCAL_PATH%" (
-  echo ERROR: Could not install dotnet cli correctly. See '%INIT_TOOLS_LOG%' for more details. 1>&2
-  exit /b 1
-)
-
-:afterdotnetsdkinstall
+:afterdotnetrestore
```
eerhardt: This name doesn't really make sense to me.  We aren't calling "dotnet restore" before this.  I think it should be left as it is: `afterdotnetinstall`.

init-tools.sh
```diff
@@ -6,74 +6,18 @@ __PACKAGES_DIR=$__scriptpath/packages
 __TOOLRUNTIME_DIR=$__scriptpath/Tools
 __DOTNET_PATH=$__TOOLRUNTIME_DIR/dotnetcli
 __DOTNET_CMD=$__DOTNET_PATH/dotnet
-__DOTNET_SDK_PATH=$__TOOLRUNTIME_DIR/dotnetclisdk
-__DOTNET_SDK_CMD=$__DOTNET_SDK_PATH/dotnet
 if [ -z "$__BUILDTOOLS_SOURCE" ]; then __BUILDTOOLS_SOURCE=https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json; fi
+export __BUILDTOOLS_USE_CSPROJ=true
```
eerhardt: ?? What's this do?  And why is it an environment variable?

init-tools.sh
```diff
@@ -105,12 +49,6 @@ if [ ! -e $__INIT_TOOLS_DONE_MARKER ]; then
     if [ -e $__TOOLRUNTIME_DIR ]; then rm -rf -- $__TOOLRUNTIME_DIR; fi
     echo "Running: $__scriptpath/init-tools.sh" > $__init_tools_log
 
-    # core-setup repo has a global.json which causes a conflict with the init-tools publish step
```
eerhardt: Can you delete that `global.json` file as well?  It is no longer relevant.

init-tools.sh
```diff
@@ -168,40 +87,33 @@ if [ ! -e $__INIT_TOOLS_DONE_MARKER ]; then
...
-    if [ -e $__scriptpath/tools-override/crossgen.sh ]; then 
-        $__scriptpath/tools-override/crossgen.sh $__scriptpath/Tools
-    fi
+    echo "Making all .sh files executable under Tools."
+    # Executables restored with .NET Core 2.0 do not have executable permission flags. https://github.com/NuGet/Home/issues/4424
+    ls $__scriptpath/Tools/*.sh | xargs chmod +x
+    ls $__scriptpath/Tools/scripts/docker/*.sh | xargs chmod +x
 
-    cp -fR $__scriptpath/tools-override/* $__TOOLRUNTIME_DIR 
+    Tools/crossgen.sh $__scriptpath/Tools
```
eerhardt: Why do we crossgen in the unix version of init-tools, but not in the Windows version?

src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/src/Microsoft.NETCore.UniversalWindowsPlatform.depproj
```diff
@@ -8,12 +8,50 @@
...
+      <Version>$(MicrosoftNETCoreRuntimeCoreCLRPackageVersion)</Version>
+    </PackageReference>
+    <PackageReference Include="Microsoft.NETCore.Platforms">
+      <Version>2.0.0-$(CoreFxExpectedPrerelease)</Version>
+    </PackageReference>
+    <PackageReference Include="NETStandard.Library">
+      <Version>$(NETStandardVersion)</Version>
+    </PackageReference>
+    <PackageReference Include="System.ServiceModel.Duplex">
+      <Version>4.4.0-beta-25205-01</Version>
```
eerhardt: These versions should all be in the dependencies.props file, so they can be auto-updated as necessary.

tools-local/Microsoft.DotNet.Build.Tasks.Local/Microsoft.DotNet.Build.Tasks.Local.csproj
```diff
@@ -4,19 +4,41 @@
...
+    <NuGetTargetMonikerShort Condition="'$(TargetGroup)' == 'net45'">net451</NuGetTargetMonikerShort>
     <CLSCompliant>false</CLSCompliant>
     <ProjectGuid>{B507BED2-0A28-456A-99C6-AD7AA148E35F}</ProjectGuid>
     <OutputPath Condition="'$(TargetGroup)' != 'net45'">$(LocalToolRuntimePath)</OutputPath>
-    <IntermediateOutputPath Condition="'$(TargetGroup)' == 'net45'">$(IntermediateOutputPath)\net45</IntermediateOutputPath>
     <OutputPath Condition="'$(TargetGroup)' == 'net45'">$(LocalToolRuntimePath)\net46</OutputPath>
+    <IntermediateOutputPath>$(BaseIntermediateOutputPath)tools-local/$(MSBuildProjectName)/$(TargetGroup)</IntermediateOutputPath>
+    <PackageTargetFallback>dnxcore50;portable-net45+win8+wpa81</PackageTargetFallback>
+    <ContainsPackageReferences>true</ContainsPackageReferences>
+    <TargetFrameworkIdentifier>.NETStandard</TargetFrameworkIdentifier>
```
eerhardt: What about when `TargetGroup` == `net45`?  Shouldn't the TargetFrameworkIdentifier be `.NETFramework`?

init-tools.sh
ellismg: Because on Windows we still use Dekstop based MSBuild and such and the performance has never been bad enough for us to bother

init-tools.sh
mellinoe: It controls some initialization stuff in buildtools. This actually should be auto-detected but that is broken ATM

init-tools.sh
mellinoe: Will do.

tools-local/Microsoft.DotNet.Build.Tasks.Local/Microsoft.DotNet.Build.Tasks.Local.csproj
mellinoe: This is a little bit deceiving because this doesn't actually control what framework we build for. I just set it to this to make sure the portable targets get imported in frameworktargeting.targets. We build the netfx flavor against nuget package references just like the netstandard flavor.

src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/src/Microsoft.NETCore.UniversalWindowsPlatform.depproj
mellinoe: Will do.

init-tools.cmd
mellinoe: Yeah, I don't really disagree. However, this file is pretty much identical to what we have in other repos (with the exception of the first-time-experience env var), so I think I'd rather just keep it in sync with the others.

dependencies.props
mellinoe: Sounds good. I haven't worked with this stuff too often so I'm not up-to-date with the best practices. I'll do what you're suggesting.

dir.props
mellinoe: `DnuRestoreCommand` is baked into buildtools -- we will need to change it there first, but I agree we should do it.

init-tools.cmd
eerhardt: Nothing wrong with changing the others to keep them in sync....

dir.props
eerhardt: Let's do this for now -

In this repo, change all these properties to not say "Dnu".  Then set the single property that build tools needs.

Example:

```
<DotnetRestorePrefix> ...
<DotnetRestoreCommand>....
<DotnetRestoreCommand>....

<DnuRestoreCommand>$(DotnetRestoreCommand)</DnuRestoreCommand>
```

That way 99% of this code makes sense, and only 1% is there because build tools isn't updated.

eerhardt: Actually, looking into buildtools, you don't need to keep this Dnu property around:

https://github.com/dotnet/buildtools/blob/03988127f489bcd1cf58d71131ac628620d53de9/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props#L66-L67

You can just set `$(DotnetRestoreCommand)` as appropriate and buildtools will set `DnuRestoreCommand` to it's value.

Let's not make this one more place that needs to be cleaned up for "legacy" reasons.  We are already in a place to eliminate this.

dependencies.props
```diff
@@ -40,13 +45,22 @@
...
+    </RemoteDependencyBuildInfo>
 
     <DependencyBuildInfo Include="@(RemoteDependencyBuildInfo)">
       <RawVersionsBaseUrl>https://raw.githubusercontent.com/dotnet/versions</RawVersionsBaseUrl>
     </DependencyBuildInfo>
     
     <XmlUpdateStep Include="CoreFx">
       <Path>$(MSBuildThisFileFullPath)</Path>
+      <ElementName>CoreFxVersion</ElementName>
+      <PackageId>Microsoft.CSharp</PackageId>
```
eerhardt: This should be `Microsoft.Private.CoreFx.NETCoreApp`.  The way this auto-update stuff works is it gets the new version from [https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt#L4](https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt#L4).  If the package hasn't been updated, then the version isn't updated.

So in servicing releases, where we only change `System.Net.Http`, we won't get a new `Microsoft.CSharp` package, but the private/transport package will be updated.  We want the version property to be updated whenever a new `Microsoft.Private.CoreFx.NETCoreApp` package is created.

dir.props
dagood: Nice catch, @eerhardt. Here's the core-eng issue tracking this, by the way: https://github.com/dotnet/core-eng/issues/108

weshaggard: Yes as part of the BuildTools 2.0 work I moved these into BuildTools so you shouldn't define them in the repo's any more. Only add the necessary additional arguments as necessary.

src/pkg/packaging/deb/package.targets
```diff
@@ -12,8 +12,8 @@
   </PropertyGroup>
 
   <Target Name="InitializeDotnetDebTool">
-    <Exec Command="$(SdkRestoreCommand) $(dotnetDebToolSource)"/>
-    <Exec Command="$(DotnetSdkToolCommand) pack $(dotnetDebToolSource) --output $(PackagesOutDir)intermediate/ --version-suffix $(VersionSuffix)"/>
+    <Exec Command="$(DnuRestoreCommand) $(dotnetDebToolSource)"/>
```
weshaggard: This can be DotnetRestoreCommand now with BT2.

mellinoe:
* Replaced DnuRestoreCommand with DotnetRestoreCommand, and removed most of the initial definitions for it, since they are in buildtools now. I kept the DOTNET_SKIP_FIRST_TIME_EXPERIENCE env-var on Unix, though.
* Link CoreFxVersion to Microsoft.Private.CoreFx.NETCoreApp in dependencies.props.

dir.props
```diff
@@ -116,54 +115,30 @@
...
-    <OverrideToolHost Condition="'$(MSBuildRuntimeType)' == 'Core'">$(DotnetToolCommand)</OverrideToolHost>
-    <DnuRestoreSource>@(DnuSourceList -> '--source %(Identity)', ' ')</DnuRestoreSource>
-    <DnuRestorePrefix Condition="'$(OsEnvironment)' != 'Windows_NT'">DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1</DnuRestorePrefix>
-    <DnuRestoreCommand>$(DnuRestorePrefix) "$(DotnetToolCommand)"</DnuRestoreCommand>
-    <DnuRestoreCommand>$(DnuRestoreCommand) restore</DnuRestoreCommand>
-    <DnuRestoreCommand>$(DnuRestoreCommand) --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))" $(DnuRestoreSource)</DnuRestoreCommand>
-    <DnuRestoreCommand Condition="'$(LockDependencies)' == 'true'">$(DnuRestoreCommand) --lock</DnuRestoreCommand>
-
-    <SdkRestoreCommand>$(DotnetSdkToolCommand) restore --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))" $(DnuRestoreSource)</SdkRestoreCommand>
+    <DotnetRestorePrefix Condition="'$(OsEnvironment)' != 'Windows_NT'">DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 </DotnetRestorePrefix>
```
weshaggard: I suspect there is a larger clean-up effort lurking but in general this condition should be `'$(RunningOnUnix)' == 'true'`

src/test/dir.proj
```diff
@@ -46,13 +46,13 @@
     <PropertyGroup>
       <RestoreSourceArg>@(RestoreTestSource->'--source %(Identity)', ' ')</RestoreSourceArg>
       <RestoreSourceArg>$(RestoreSourceArg) @(RestoreTestFallbackSource->'--source %(Identity)', ' ')</RestoreSourceArg>
-      <RestoreArgs>--disable-parallel --packages "$(PackagesDir.TrimEnd('/\'.ToCharArray()))" $(RestoreSourceArg)</RestoreArgs>
+      <RestoreArgs>--disable-parallel --packages "$(PackagesDir.TrimEnd('/').TrimEnd('\'))" $(RestoreSourceArg)</RestoreArgs>
       <RestoreArgs>$(RestoreArgs) /p:MNAVersion=$(ProductVersion) /p:TestTargetRid=$(TestTargetRid)</RestoreArgs>
       <RestorePrefix Condition="'$(OsEnvironment)' != 'Windows_NT'">DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1</RestorePrefix>
```
weshaggard: Any reason we cannot use the DotnetRestoreCommand with additional options?

dir.props
mellinoe: I'll fix this one since I added it, at least.

src/test/dir.proj
mellinoe: I'm not sure. I think we wanted the flexibility to use different package sources from the default here, but ultimately it looks like we are kind of using the same options, more-or-less, plus some extras.

mellinoe:
I've addressed all the top-priority feedback, and have clean local builds. Once I see that CI is green, I'll merge this.

PR #2378, opened 2017-05-12
Fix branding quotes in installer

karajas:
This is to fix the quotes around the branding on the Windows installer.

/cc @rakeshsinghranchi


rakeshsinghranchi:
LGTM

PR #2379, opened 2017-05-12
Fix Microsoft.NETCore.App package version file location for RID specific packages

joperezr:
cc: @ericstj @weshaggard 

A recent change caused the Microsoft.NETCore.App.versions.txt file to be moved now to `runtimes/$(NuGetRuntimeIdentifier)/lib/$(PackageTargetFramework)` for all RID specific packages instead of the root of the package as it should be. With this small change this gets fixed by moving it again to the root of the package.

FYI: @gkhanna79

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -106,12 +106,6 @@
                        Overwrite="true"/>
   </Target>
 
-  <ItemGroup>
-    <FilesToPackage Include="$(IntermediateOutputPath)\Microsoft.NETCore.App.versions.txt">
-      <TargetPath></TargetPath>
-    </FilesToPackage>
-  </ItemGroup>
-
   <!-- Fetches all the file items from the packages that we want to redist -->
   <Target Name="GetFilesFromPackages" DependsOnTargets="GetPackagePaths">
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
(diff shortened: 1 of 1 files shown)


src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -152,6 +146,13 @@
       <_sourcePath Include="@(_sourcePathCandidate)" Condition="Exists('%(Identity)')" />
     </ItemGroup>
 
+    <!-- Add versions file with the hashes of the repos we consume -->
+    <ItemGroup>
+      <FilesToPackage Include="$(IntermediateOutputPath)\Microsoft.NETCore.App.versions.txt">
+        <TargetPath></TargetPath>
```
weshaggard: How does this fix it? Is there something that is changing empty TargetPaths to something else?

joperezr: This same target [is basically now getting all of the elements that currently exist in `FilesToPackage` and setting the TargetPath](https://github.com/dotnet/core-setup/blob/master/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj#L140-L155) to that runtimes path whenever RuntimeID is not empty. Given that the way I was adding this text file was staticly (outside a target), then it was always being added before this task ran which now started setting the targetPath for it. With this change, I'm adding the file to the Item AFTER the TargetPath is set, therefore it won't override my empty one anymore.

joperezr: It didn't happen before because @ericstj used to have a temp item for these files where TargetPath was set `_FilesToPackage`, but in a recent PR he moved to use the one without the underscore which made it such that my version file was impacted as well.

PR #2385, opened 2017-05-12
[WIP] Publish artifacts to new structure in Azure storage.

mellinoe:
I've tackled the basic parts of this here. I tested a single leg of the "publish" step (win-x64), and it successfully published to a test account I set up, and in the new folder structure.

I've made changes to handle the "Finalize" step, but I'm not sure how to test that. The logic looks sound to me, but I'm not sure if there are additional things I am missing.

@eerhardt @johnbeisner @karajas


publish/dir.targets
```diff
@@ -89,7 +89,6 @@
                          Exclude="@(_FoundBlobNames)" />
     </ItemGroup>
     <Error Text="Missing publish output from all legs of the build. Missing blobs are '@(_MissingBlobNames)'"
-           Importance="High"
```
mellinoe: This is just an existing (unrelated) problem. There is no "Importance" property on Error.

publish/dir.targets
```diff
@@ -4,7 +4,7 @@
 
   <Target Name="GatherPublishToAzureBinaries">
     <PropertyGroup>
-      <RelativeBlobPath>$(Channel)/Binaries/$(SharedFrameworkNugetVersion)</RelativeBlobPath>
+      <RelativeBlobPath>Runtime/$(SharedFrameworkNugetVersion)</RelativeBlobPath>
```
eerhardt: FYI - this will conflict with @karajas's change since he is moving these properties around.

publish/publish.proj
```diff
@@ -69,14 +69,14 @@
     <FinalizeBuild AccountName="$(AzureAccountName)"
                    AccountKey="$(AzureAccessToken)"
                    ContainerName="$(ContainerName)"
-                   SemaphoreBlob="$(Channel)/Binaries/sharedFxPublishSemaphore"
+                   SemaphoreBlob="Runtime/sharedFxPublishSemaphore"
```
eerhardt: The semaphore should be inside the channel.  So `Runtime/$(Channel)/sharedFxPublishSemaphore`.

If 2 builds from separate channels (branches) are publishing at the same time, there is no need to control the concurrency between them.  But if 2 builds in the same channel are publishing at the same time, we need to control the concurrency so they both don't write to the channel folder at the same time.

tools-local/tasks/FinalizeBuild.cs
```diff
@@ -92,20 +92,13 @@ public override bool Execute()
...
                 {
-                    CopyBlobs($"{Channel}/Binaries/{SharedFrameworkNugetVersion}", $"{Channel}/Binaries/Latest/");
+                    CopyBlobs($"Runtime/{SharedFrameworkNugetVersion}", $"Runtime/Latest/");
 
-                    CopyBlobs($"{Channel}/Installers/{SharedFrameworkNugetVersion}", $"{Channel}/Installers/Latest/");
-
-                    CopyBlobs($"{Channel}/Installers/{SharedHostNugetVersion}", $"{Channel}/Installers/Latest/");
-
-                    // Generate the Sharedfx Version text files
-                    List<string> versionFiles = PublishRids.Select(p => $"{p.GetMetadata("VersionFileName")}.version").ToList();
```
eerhardt: Do we still need to pass in `PublishRids` to this task anymore?

eerhardt: Similarly, is the `VersionFileName` metadata still used on that item after this change?  If no, can it be removed?

eerhardt:
Please update the `README.md` links as well.  Unless you want to wait to get a build out, so you can verify the links before changing them.  Either way, the README needs to be updated.

tools-local/tasks/FinalizeBuild.cs
```diff
@@ -92,20 +92,13 @@ public override bool Execute()
...
+                    CopyBlobs($"Runtime/{SharedHostNugetVersion}", $"Runtime/Latest/");
```
eerhardt: We have a lot of version usages problems in our build, and this really shows the problem.  I'll try to explain:

In a single build, we only ever build one version of things.  In our build code we call it `ProductVersion`.  It is the uber "I-don't-care-about-a-single-component" version.  Then each component can have its own version property because in servicing builds, the components are serviced separately.  We may only build a new SharedFramework, but not build a new host dotnet.exe.

So we have `SharedFrameworkNugetVersion` (the version of the Shared Framework we are building), `HostResolverVersion` (the version of the hostfxr.dll that should be used), `HostVersion` (the version of the dotnet.exe that should be used).

Let's say in the first servicing release, we just need to build a new Shared Framework.  So ProductVersion and SharedFrameworkVersion will be `2.0.1`, but HostVersion will still be `2.0.0` because we still want to use the same host that was shipped with 2.0 RTM.  Here you are going to be copying all the `2.0.0` blobs into `Runtime/Latest`, thus blowing away all the `2.0.1` blobs that were just copied above.

So 2 things should change here:

1. Don't use a single component's version during publish/finalize unless you are strictly referencing that single component.  Instead use `ProductVersion`.
1. Don't CopyBlobs twice.  Just do it once for ProductVersion.

tools-local/tasks/FinalizeBuild.cs
```diff
@@ -92,20 +92,13 @@ public override bool Execute()
 
                 try
                 {
-                    CopyBlobs($"{Channel}/Binaries/{SharedFrameworkNugetVersion}", $"{Channel}/Binaries/Latest/");
+                    CopyBlobs($"Runtime/{SharedFrameworkNugetVersion}", $"Runtime/Latest/");
```
eerhardt: This shouldn't be `Runtime/Latest`.  It should be `Runtime/{Channel}`.

tools-local/tasks/FinalizeBuild.cs
```diff
@@ -92,20 +92,13 @@ public override bool Execute()
...
-                    List<string> versionFiles = PublishRids.Select(p => $"{p.GetMetadata("VersionFileName")}.version").ToList();
+                    CopyBlobs($"Runtime/{SharedHostNugetVersion}", $"Runtime/Latest/");
 
+                    // Generate the latest version text file
                     string sfxVersion = GetSharedFrameworkVersionFileContent();
-                    foreach(string version in versionFiles)
-                    {
-                        PublishStringToBlob(ContainerName, $"{Channel}/dnvm/latest.sharedfx.{version}", sfxVersion, "text/plain");
-                    }
+                    PublishStringToBlob(ContainerName, $"Runtime/latest.sharedfx", sfxVersion, "text/plain");
```
eerhardt: This should be `Runtime/{Channel}/latest.version`.  It needs to go into a channel, and there isn't a reason to name it "sharedfx" since we are only dealing with "Runtime" stuff in this new structure.  (It was named "latest.sharedfx" before because the Sdk and the Runtime used to be publishing assets to the same place.)

publish/publish.proj
mellinoe: Makes sense -- fixing.

tools-local/tasks/FinalizeBuild.cs
mellinoe: I think they can both be removed.

publish/dir.targets
mellinoe: Just chatted with him -- I'm going to rebase my stuff once he merges his Checksum change in. It sounded like that change was very close to landing already.

mellinoe:
I updated links in README.md.

mellinoe:
@eerhardt Could you take a look at the latest version?

mellinoe:
Rebased again after #2399

README.md
```diff
@@ -61,11 +61,11 @@ Daily Builds
 [build-1.0.X]: https://devdiv.visualstudio.com/DevDiv/_build/index?definitionId=4187&_a=completed
 
 
-[win-x64-badge-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/sharedfx_win-x64_Release_version_badge.svg
-[win-x64-version-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/dnvm/latest.sharedfx.win.x64.version
-[win-x64-installer-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/Latest/dotnet-win-x64.latest.exe
-[win-x64-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-win-x64.latest.zip
-[win-x64-symbols-zip-master]: https://dotnetcli.blob.core.windows.net/dotnet/master/Binaries/Latest/dotnet-sharedframework-symbols-win-x64.latest.zip
+[win-x64-badge-master]: https://dotnetcli.blob.core.windows.net/dotnet/Runtime/master/sharedfx_win-x64_Release_version_badge.svg
+[win-x64-version-master]: https://dotnetcli.blob.core.windows.net/dotnet/Runtime/master/latest.sharedfx.win.x64.version
```
eerhardt: We removed these RID-specific version files and just have a single "latest.version" file now, right?

eerhardt:
>                                         .Replace(SharedHostNugetVersion, "latest");

This can now be removed, since we never upload blobs specifically to the `SharedHostNugetVersion`.  We only put them to `ProductVersion`.

---
Refers to: tools-local/tasks/FinalizeBuild.cs:125 in 14010d9. [](commit_id = 14010d91c11d3af147479e6a3c706eef402d675a, deletion_comment = False)

PR #2388, opened 2017-05-15
Set default ROOTFS_DIR path when it does not set

jyoungyun:
When ROOTFS_DIR does not set, this patch will assign the default ROOTFS_DIR path to the environment value.

/cc @gkhanna79


src/corehost/build.sh
```diff
@@ -155,6 +156,13 @@ case $__build_arch in
 esac
 __cmake_defines="${__cmake_defines} ${__arch_define}"
 
+# Configure environment if we are doing a cross compile.
```
gkhanna79: @jyoungyun What is the scenario when this is not set? Is it just the CI?

jyoungyun: @gkhanna79 If ROOTFS_DIR is not set in coreclr or corefx, the default path is set in cross/rootfs/${build_arch}. This is a patch to build core-setup without having to set ROOTFS_DIR separately when user creates rootfs by build-rootfs.sh script.

gkhanna79: Sounds good - can I merge this?

jyoungyun: @gkhanna79 Yes, sure. Thank you.

PR #2390, opened 2017-05-15
Fix deb package naming

karajas:
/cc @rakeshsinghranchi

### DIFF
--- a/src/pkg/packaging/deb/package.props
+++ b/src/pkg/packaging/deb/package.props
@@ -10,10 +10,10 @@
     <SharedHostDebPkgName>dotnet-host</SharedHostDebPkgName>
     <SharedHostDebPkgName>$(SharedHostDebPkgName.ToLower())</SharedHostDebPkgName>
     
-    <HostFxrDebPkgName>dotnet-hostfxr-$(HostResolverVersion)</HostFxrDebPkgName>
+    <HostFxrDebPkgName>dotnet-hostfxr</HostFxrDebPkgName>
     <HostFxrDebPkgName>$(HostFxrDebPkgName.ToLower())</HostFxrDebPkgName>
     
-    <SharedFxDebPkgName>dotnet-sharedframework-$(SharedFrameworkName)-$(SharedFrameworkNugetVersion)</SharedFxDebPkgName>
+    <SharedFxDebPkgName>dotnet-sharedframework</SharedFxDebPkgName>
     <SharedFxDebPkgName>$(SharedFxDebPkgName.ToLower())</SharedFxDebPkgName>
   </PropertyGroup>
 </Project>


karajas:
@dotnet-bot test Windows_NT x86 Debug Build please

eerhardt:
Are we sure we want to make this change?  Did we need to?  What was broken before?  We've been building the .deb packages this way since before v1.0.

This is breaking the CLI from consuming the latest core-setup builds.  See the failure in Ubuntu x64 https://github.com/dotnet/cli/pull/6582.  I can change the CLI's build to expect the new names, but I'm not certain this change should have been made in the first place.

/cc @gkhanna79 @livarcocc @ellismg

gkhanna79:
@rakeshsinghranchi @karajas Can you share more context on this?

@leecow Was this naming influenced by the naming process you were driving for the platform installers?

karajas:
I made this change after checking the errors for publishing on the debian feed.

```

    {
        "dataFileName": "dotnet-hostfxr-ubuntu.16.10-x64.2.0.0-preview2-25312-01.deb",
        "details": "Package name or version does not match. (dotnet-hostfxr\t2.0.0-preview2-25312-01-1:dotnet-hostfxr-2.0.0-preview2-25312-01\t2.0.0-preview2-25312-01-1)",
        "id": "5915ffd041e38e177a99ae7e",
        "name": "dotnet-hostfxr",
        "owner": "dotnet",
        "repositoryId": "580a6ecca71906770595e108",
        "sourceUrl": "https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/2.0.0-preview2-25312-01/dotnet-hostfxr-ubuntu.16.10-x64.2.0.0-preview2-25312-01.deb",
        "status": "validationError",
        "version": "2.0.0-preview2-25312-01-1"
    },
    {
        "dataFileName": "dotnet-sharedframework-ubuntu.16.10-x64.2.0.0-preview2-25312-01.deb",
        "details": "Package name or version does not match. (dotnet-sharedframework\t2.0.0-preview2-25312-01-1:dotnet-sharedframework-microsoft.netcore.app-2.0.0-preview2-25312-01\t2.0.0-preview2-25312-01-1)",
        "id": "5915ffd041e38e177a99ae7f",
        "name": "dotnet-sharedframework",
        "owner": "dotnet",
        "repositoryId": "580a6ecca71906770595e108",
        "sourceUrl": "https://dotnetcli.blob.core.windows.net/dotnet/master/Installers/2.0.0-preview2-25312-01/dotnet-sharedframework-ubuntu.16.10-x64.2.0.0-preview2-25312-01.deb",
        "status": "validationError",
        "version": "2.0.0-preview2-25312-01-1"
    }
```

karajas:
To me this seemed like we wanted to update the packages to be like dotnet-host and publish a version less package name. I did not follow up with changing the cli subsequently. https://github.com/dotnet/cli/blob/6ecc1fc6c3f58bde065413cc869fb7f97d296143/build/package/Installer.DEB.targets#L32-L37 

I guess my question is was there a specific reason to include the $(SharedFrameworkName)-$(SharedFxDebianPackageVersion) in the shared framework package name or the -$(HostFxrDebianPackageVersion) version in the hostfxr package name instead of just the version?

gkhanna79:
I would think that we would need version information (of M.N.App) to distinguish the installer from a release to another.

eerhardt:
That's correct.  Both hostfxr and the shared framework packages are side-by-side installable.  The host isn't - so that's why it doesn't have a version in its name.

I'm going to revert this change to unblock taking new runtimes in the SDK.  We should investigate a better fix for the validation error you mention above.  As I've mentioned, we've been building the deb packages like this since before 1.0.

leecow:
You're getting the validation error because dataFileName and name must be an exact match when you publish.

PR #2392, opened 2017-05-15
Remove un-needed target

### LINKED ISSUE #2347: Remove un-needed target when upgrading to v2 BuildTools
We have done some change in the PR #2333 to change the license Url for packaging. because core-setup repo not using v2 of the buildtools we had to hack this by adding the temporary target OverrideLicenseUrl in the src/pkg/projects/dir.targets file. This target should be removed after upgrading to v2 of the buildtools and need to ensure when removing it the license Url in the packages nuspec files is correct

tarekgh:
Fixes #2347

### DIFF
--- a/src/pkg/projects/dir.targets
+++ b/src/pkg/projects/dir.targets
@@ -71,11 +71,4 @@
     </MSBuild>
   </Target>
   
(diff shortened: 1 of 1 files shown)


tarekgh:
@weshaggard

Issue #2336, opened 2017-05-10
multilevel sdk lookup policy does not consider different versions across more than one level


nguerrera:
Once any sdk is found at any level, none of the SDKs at different levels are considered. 

For example, if there's an sdk version at the local location, it will be used irrespective of global.json or if there are newer versions available globally.

## Steps to reproduce

* Install a version of the CLI that you don't already have in the global location to the local location. For example:

Global installs
```
D:\repro>dir /b "c:\Program Files\dotnet\sdk"
1.0.0
1.0.0-preview2-003121
1.0.0-preview2-003131
1.0.4
2.0.0-preview1-005957
```

Local installs
```
D:\repro>dir /b %USERPROFILE%\.dotnet\x64\sdk
1.0.3
```

* Use dotnet --version to check which sdk is used (expecting 2.0.0-preview1-005957)
```
D:\repro>dotnet --version
1.0.3
```

* Put a global.json specifying an sdk version that is in the global location, but not the local location
```
D:\repro>type global.json
{
  "sdk": {
    "version": "1.0.0"
  }
}
```

* Use dotnet --version to check which sdk is used (expecting 1.0.0)
```
D:\repro>dotnet --version
1.0.3
```

## Expected  behavior
Without global.json, it uses the latest version in the global location that is later than the latest version in the local location
With global.json, it uses the version specified that can be found in the global location, but not the local location

## Actual behavior
It uses the latest local version

## Environment data
```
D:\>dotnet --info
.NET Command Line Tools (1.0.3)

Product Information:
 Version:            1.0.3
 Commit SHA-1 hash:  37224c9917

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Users\nicholg\.dotnet\x64\sdk\1.0.3

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview1-002111-00
  Build    : 1ff021936263d492539399688f46fd3827169983
```

@eerhardt @gkhanna79 @livarcocc @dsplaisted @johnbeisner

gkhanna79:
@livarcocc Please triage for milestone and assign to appropriate individual.

nguerrera:
When there is no global.json involved, the latest version selected should be the latest version across all levels as well.

johnbeisner:
[...]

Issue #2334, opened 2017-05-09
If sdk version specified in global.json is not found, latest sdk is used silently


nguerrera:
## Steps to reproduce

* Use a global.json specifying an SDK that is not installed. 
* Check `dotnet --version` (or more generally, use any cli verb and see which sdk it uses)

```
D:\repro> dir /b "c:\Program Files\dotnet\sdk"
1.0.0
1.0.0-preview2-003121
1.0.0-preview2-003131
1.0.4
2.0.0-preview1-005957

D:\repro> type global.json
{
  "sdk": {
    "version": "1.2.0"
  }
}

D:\repro> dotnet --version
2.0.0-preview1-005957
```

## Expected  behavior
An error is reported indicating that the version specified in global.json is not present

## Actual behavior
It just uses latest SDK instead (effectively ignoring global.json)

## Scenarios
* Typos in global.json can get ignored.
* Teams can't impose a specific sdk version because on machines where that SDK is not installed, it will just use whatever is latest

## Environment data
```
.NET Command Line Tools (2.0.0-preview1-005957)

Product Information:
 Version:            2.0.0-preview1-005957
 Commit SHA-1 hash:  056ac0e9cd

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0-preview1-005957\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview1-002111-00
  Build    : 1ff021936263d492539399688f46fd3827169983
```

@eerhardt @gkhanna79 @livarcocc @johnbeisner @dsplaisted

gkhanna79:
@livarcocc Please triage for milestone and assign to appropriate individual.

nguerrera:
This is higher priority than #2336

karelz:
This is blocking .NET Core 2.0 ZBB on 5/10.
If it is non-trivial I am fine moving it to Post-ZBB, given that it came quite late and close to the ZBB data.

@nguerrera @gkhanna79 any preferences?

gkhanna79:
Livar should chime in.




Thanks,
Gaurav


-------- Original message --------

nguerrera:
It is not trivial enough to be filed on May 9 @ 5 PM and fixed by May 10. ;) Track it however appropriate, but it needs to be fixed. @johnbeisner is working on it.

[...]

PR #2393, opened 2017-05-15
If user specifies a CLI version in the global.json, honor it

ghost:
If user specifies a CLI version in the global.json, honor it - error out if not found.
And search the multilevel sdk lookup [if enabled] to find that specified version.

Addresses: 
https://github.com/dotnet/core-setup/issues/2334
https://github.com/dotnet/core-setup/issues/2336

fx_muxer.cpp:
1. do not resolve a compatible CLI version if we are using a global.json value
2. multilevel sdk lookup uses the latest version across all levels

GivenThatICareAboutMultilevelSDKLookup.cs:
Testing the error condition [specified version can't be found] & multilevel lookup.

run.cmd:
trivial change to URL

@dotnet/dotnet-cli

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -589,10 +589,10 @@ pal::string_t resolve_sdk_version(pal::string_t sdk_path)
     trace::verbose(_X("Checking if resolved SDK dir [%s] exists"), sdk_path.c_str());
     if (pal::directory_exists(sdk_path))
     {
-        retval = sdk_path;
+        retval = max_ver_str;
     }
 
-    trace::verbose(_X("Resolved SDK dir is [%s]"), retval.c_str());
+    trace::verbose(_X("Resolved SDK dir is [%s]"), sdk_path.c_str());
     return retval;
 }
 
@@ -661,25 +661,23 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
 
     pal::string_t retval;
     bool cli_version_specified = false;
-    bool cli_version_found = false;
     pal::string_t cli_version;
 
     for (pal::string_t dir : hive_dir)
     {
         trace::verbose(_X("Searching SDK directory in [%s]"), dir.c_str());
+        pal::string_t sdk_path = dir;
+        append_path(&sdk_path, _X("sdk"));
         if (!global.empty())
         {
             cli_version = resolve_cli_version(global);
             if (!cli_version.empty())
             {
                 cli_version_specified = true;
-                pal::string_t sdk_path = dir;
-                append_path(&sdk_path, _X("sdk"));
                 append_path(&sdk_path, cli_version.c_str());
 
                 if (pal::directory_exists(sdk_path))
                 {
-                    cli_version_found = true;
                     trace::verbose(_X("CLI directory [%s] from global.json exists"), sdk_path.c_str());
                     retval = sdk_path;
                 }
@@ -691,9 +689,9 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
         }
         if (retval.empty() && !cli_version_specified)
         {
-            pal::string_t sdk_path = dir;
-            append_path(&sdk_path, _X("sdk"));
-            retval = resolve_sdk_version(sdk_path);
+            cli_version = resolve_sdk_version(sdk_path);
+            append_path(&sdk_path, cli_version.c_str());
+            retval = sdk_path;
         }
         if (!retval.empty())
         {
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -660,20 +660,26 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
     }
 
     pal::string_t retval;
+    bool cli_version_specified = false;
+	bool cli_version_found = false;
+	pal::string_t cli_version;
+
     for (pal::string_t dir : hive_dir)
     {
         trace::verbose(_X("Searching SDK directory in [%s]"), dir.c_str());
         if (!global.empty())
         {
-            pal::string_t cli_version = resolve_cli_version(global);
+            cli_version = resolve_cli_version(global);
             if (!cli_version.empty())
             {
+			    cli_version_specified = true;
                 pal::string_t sdk_path = dir;
                 append_path(&sdk_path, _X("sdk"));
                 append_path(&sdk_path, cli_version.c_str());
 
                 if (pal::directory_exists(sdk_path))
                 {
+				    cli_version_found = true;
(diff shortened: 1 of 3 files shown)


src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -660,20 +660,26 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
     }
 
     pal::string_t retval;
+    bool cli_version_specified = false;
+	bool cli_version_found = false;
+	pal::string_t cli_version;
```
nguerrera: There's whitespace/indentation inconsistency here and elsewhere. (I won't comment everywhere).

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -698,6 +704,10 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
     }
     
     trace::verbose(_X("It was not possible to find any SDK version"));
```
nguerrera: This does not necessarily apply when we're looking for a specific version, so we should log either this or new message, but not both.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -698,6 +704,10 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
...
+	if (cli_version_specified && !cli_version_found)
+	{
+	    trace::error(_X("The specified SDK version [%s] from global.json [%s] doesn't exist; install specified SDK version [%s]"), cli_version.c_str(), global.c_str(), cli_version.c_str());
```
nguerrera: I think we should say the SDK was not "not found" rather than that it "doesn't exist".

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -683,7 +689,7 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
                 }
             }
         }
-        if (retval.empty())
+        if (retval.empty() && !cli_version_specified)
```
nguerrera: This fixes the repro steps I had listed in #2336, but we also need to make sure that when we pick the latest version when there's no global.json, it's the latest across all hive_dirs here. 

You can leave #2336 open and tackle that part separately if you want.

src/corehost/cli/fxr/fx_muxer.cpp
johnbeisner: done...

src/corehost/cli/fxr/fx_muxer.cpp
johnbeisner: done...

src/corehost/cli/fxr/fx_muxer.cpp
johnbeisner: done...

livarcocc (approved):

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -659,45 +677,71 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
...
+    pal::string_t cli_version;
+    pal::string_t sdk_path;
+    pal::string_t global_cli_version;
+
+    if (!global.empty())
+    {
+        global_cli_version = resolve_cli_version(global);
+        if (!global_cli_version.empty())
+        {
+            cli_version_specified = true;
```
nguerrera: Is this bool needed or is it invariantly `global_cli_version.empty()`. I'd eliminate the extra state if possible.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -659,45 +677,71 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
...
+        cli_sdk->assign(sdk_path);
+        trace::verbose(_X("Found CLI SDK in: %s"), cli_sdk->c_str());
+        return true;
+    }
+
+    if (cli_version_specified)
+    {
+        trace::error(_X("The specified SDK version [%s] from global.json [%s] not found; install specified SDK version"), cli_version.c_str(), global.c_str());
+    }
+        else
```
nguerrera: whitespace/indentation is off

nguerrera (approved):

eerhardt:
>         if (pal::get_local_dotnet_dir(&local_dir))

I thought we were going to remove USERPROFILE from the search path.

---
Refers to: src/corehost/cli/fxr/fx_muxer.cpp:664 in 7fdc239. [](commit_id = 7fdc23923d2f4352eac6e32d92be0e109e9d18aa, deletion_comment = False)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -609,6 +609,24 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, pal::stri
...
+{
+    bool retval = false;
+    fx_ver_t ver(-1, -1, -1);
+    fx_ver_t new_ver(-1, -1, -1);
+
+    if (fx_ver_t::parse(new_version, &new_ver, parse_only_production))
+    {
+       if (!fx_ver_t::parse(*version, &ver, parse_only_production) || (new_ver > ver))
+       {
+           version->assign(new_version);
```
eerhardt: I think it is a little odd this method changes state on the parameters.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -563,7 +563,7 @@ pal::string_t fx_muxer_t::resolve_cli_version(const pal::string_t& global_json)
     return retval;
 }
 
-pal::string_t resolve_sdk_version(pal::string_t sdk_path)
+pal::string_t resolve_sdk_version(pal::string_t sdk_path, bool parse_only_production)
```
eerhardt: I only see 'false' getting passed into this method.  Can we revert the method?

johnbeisner: "parse_only_production" is being passed into 2 methods; it must be the same value to get consistent behavior...
lines:720-722
            bool parse_only_production = false;  // false -- implies both production and prerelease.
            pal::string_t new_cli_version = resolve_sdk_version(current_sdk_path, parse_only_production);
            if (higher_sdk_version(new_cli_version, &cli_version, parse_only_production))

src/corehost/cli/fxr/fx_muxer.cpp
johnbeisner: fixed...

src/corehost/cli/fxr/fx_muxer.cpp
johnbeisner: I see that now; changed...

src/corehost/cli/fxr/fx_muxer.cpp
johnbeisner: Opened an issue for this work: https://github.com/dotnet/core-setup/issues/2451

PR #2394, opened 2017-05-15
Re-enable publishing of deb tool to feed

karajas:
/cc @eerhardt @chcosta

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -7,7 +7,7 @@
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(LocalBuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.Local.dll" />
 
   <Target Name="Build"
-          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;PublishFinalOutput" />
+          DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;PublishDebToolPackageToFeed;PublishFinalOutput" />
   
   <Target Name="PublishFinalOutput"
           Condition="'$(Finalize)' == 'true'"
@@ -141,8 +141,8 @@
     
     <Message Text="Pushing CoreHost packages to $(NuGetFeedUrl)" />
     <PropertyGroup>
-      <NuGetPushCommand>$(DnuRestorePrefix)$(DotnetToolCommand) nuget push --source $(NuGetFeedUrl) --api-key $(NuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
-      <NuGetPushSymbolsCommand>$(DnuRestorePrefix)$(DotnetToolCommand) nuget push --source $(NuGetSymbolsFeedUrl) --api-key $(NuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushSymbolsCommand>
+      <NuGetPushCommand>$(DotnetToolCommand) nuget push --source $(NuGetFeedUrl) --api-key $(NuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
+      <NuGetPushSymbolsCommand>$(DotnetToolCommand) nuget push --source $(NuGetSymbolsFeedUrl) --api-key $(NuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushSymbolsCommand>
     </PropertyGroup>
 
       <!-- ToDo: look at moving this to an msbuild task so that we can include retry logic, parallelize it, and hide the api key -->
@@ -150,6 +150,19 @@
 
     <Message Condition="'@(_DownloadedSymbolsPackages)' != ''" Text="Pushing CoreHost symbols packages to $(NuGetSymbolsFeedUrl)" />
     <Exec Condition="'@(_DownloadedSymbolsPackages)' != ''" Command="$(NuGetPushSymbolsCommand) %(_DownloadedSymbolsPackages.Identity)" />
+  </Target>
+  
+  <Target Name="PublishDebToolPackageToFeed"
+          Condition="'$(PublishDebToolToFeed)' == 'true'">
+    <Error Condition="'$(CliNuGetFeedUrl)' ==''" Text="Missing required property CliNuGetFeedUrl" />
+    <Error Condition="'$(CliNuGetApiKey)' == ''" Text="Missing required property CliNuGetApiKey" />
+    <ItemGroup>
+      <DebToolPackages Include="$(PackagesOutDir)/dotnet-deb-tool.*.nupkg" />
+    </ItemGroup>
+    <PropertyGroup>
+      <NuGetPushCommand>$(DotnetToolCommand) nuget push --source $(CliNuGetFeedUrl) --api-key $(CliNuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
+    </PropertyGroup>
+    <Exec Command="$(NuGetPushCommand) %(DebToolPackages.Identity)" />
   </Target>
   
   <Import Project="$(ToolsDir)VersionTools.targets" Condition="Exists('$(ToolsDir)VersionTools.targets')" />
--- a/buildpipeline/Core-Setup-Linux-BT.json
+++ b/buildpipeline/Core-Setup-Linux-BT.json
@@ -293,7 +293,7 @@
       },
       "inputs": {
         "filename": "docker",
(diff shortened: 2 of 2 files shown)


publish/publish.proj
```diff
@@ -132,6 +132,24 @@
     <Exec Condition="'@(_DownloadedSymbolsPackages)' != ''" Command="$(NuGetPushSymbolsCommand) %(_DownloadedSymbolsPackages.Identity)" />
   </Target>
   
+  <Target Name="PublishDebToolPackagesToFeed"
+          DependsOnTargets="CheckIfAllBuildsHavePublished"
```
chcosta: Chatted with @karajas about this.  He's going to look into how to enable this for a specific leg instead of a part of finalization.

karajas:
Added PublishDebToolToFeed to the condition so that we can configure which build leg pushes it's deb tool package. Historically it was Ubuntu.14.04-x64 https://github.com/dotnet/core-setup/blob/release/2.0.0/build_projects/dotnet-host-build/PublishTargets.cs#L71-L72

publish/publish.proj
karajas: Added PublishDebToolToFeed to the condition statement.

chcosta: I don't think this should depend on `CheckIfAllBuildsHavePublished`

karajas: You're right. This has to be done once that leg is done.

chcosta: You'll want to update the pipeline.json to pass `PublishDebToolToFeed` for the Ubuntu 14.04 leg

publish/publish.proj
```diff
@@ -11,7 +11,7 @@
   
   <Target Name="PublishFinalOutput"
           Condition="'$(Finalize)' == 'true'"
-          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;UpdatePublishedVersions" />
+          DependsOnTargets="FinalizeBuildInAzure;PublishCoreHostPackagesToFeed;PublishDebToolPackagesToFeed;UpdatePublishedVersions" />
```
chcosta: I don't think `PublishFinalOutput` should be dependent on `PublishDebToolPackagesToFeed` anymore

publish/publish.proj
```diff
@@ -132,6 +132,23 @@
...
+          Condition="'$(PublishDebToolToFeed)' == 'true'">
+    <Error Condition="'$(CliNuGetFeedUrl)' ==''" Text="Missing required property CliNuGetFeedUrl" />
+    <Error Condition="'$(CliNuGetApiKey)' == ''" Text="Missing required property CliNuGetApiKey" />
+    
+    <Message Text="Pushing Deb Tool package to $(CliNuGetFeedUrl)" />
+    <ItemGroup>
+      <DebToolPackages Include="$(PackagesOutDir)intermediate/*.nupkg" />
+    </ItemGroup>
+    <PropertyGroup>
+      <NuGetPushCommand>$(DnuRestorePrefix)$(DotnetToolCommand) nuget push --source $(CliNuGetFeedUrl) --api-key $(CliNuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
```
chcosta: NuGetPushCommand could be moved to publish/dir.props

chcosta:
It would be better if this, and all the deb package stuff, pushed to somewhere in azure during the build, then downloaded and re-pushed to its final location during finalization (just like the other artifacts / packages).  This would prevent us from publishing stuff that wasn't fully validated.  I don't think this is a major issue at the moment, but it could lead to inconsistent bits.  Can you file an issue for this?

karajas:
@chcosta I think this is actually ok for the deb tool (which rarely changes) getting published, the deb packages themselves less so.

publish/publish.proj
```diff
@@ -152,6 +152,19 @@
...
   
+  <Target Name="PublishDebToolPackageToFeed"
+          Condition="'$(PublishDebToolToFeed)' == 'true'">
+    <Error Condition="'$(CliNuGetFeedUrl)' ==''" Text="Missing required property CliNuGetFeedUrl" />
+    <Error Condition="'$(CliNuGetApiKey)' == ''" Text="Missing required property CliNuGetApiKey" />
+    <ItemGroup>
+      <DebToolPackages Include="$(PackagesOutDir)intermediate/*.nupkg" />
+    </ItemGroup>
+    <PropertyGroup>
+      <NuGetPushCommand>$(DnuRestorePrefix)$(DotnetToolCommand) nuget push --source $(CliNuGetFeedUrl) --api-key $(CliNuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
```
eerhardt: There are no more properties named "Dnu" anything in our build.

publish/publish.proj
```diff
@@ -152,6 +152,19 @@
     <Exec Condition="'@(_DownloadedSymbolsPackages)' != ''" Command="$(NuGetPushSymbolsCommand) %(_DownloadedSymbolsPackages.Identity)" />
   </Target>
   
+  <Target Name="PublishDebToolPackageToFeed"
+          Condition="'$(PublishDebToolToFeed)' == 'true'">
```
eerhardt: Who sets this property?

eerhardt:
> It would be better if this, and all the deb package stuff, pushed to somewhere in azure during the build, then downloaded and re-pushed to its final location during finalization (just like the other artifacts / packages).

The long term direction for the dotnet-deb-tool is that we move it into buildtools.  It is a common utility that both core-setup and cli need to use to create .deb packages.

Chatting with @weshaggard on this, we agreed that we should keep the main logic of this tool as-is in a `.sh` script.  However how the logic gets invoked will probably change when we move it into buildtools. Instead of it being a CLI Tool, we will probably opt for putting the .sh scripts inside a NuGet package and then having a `build` folder with a `.targets` file that contains an MSBuild `Target` which can be called that will invoke the .sh script.  This is more in-line with our current tooling direction of putting build logic in MSBuild instead of in CLI Tools.

/cc @livarcocc - for FYI since this will affect the CLI build.

publish/publish.proj
karajas: pipeline.json

publish/publish.proj
karajas: Will remove

buildpipeline/pipeline.json
```diff
@@ -38,7 +38,8 @@
             "PB_TargetArchitecture": "arm", 
             "PB_AdditionalBuildArguments":"-TargetArchitecture=arm -DistroRid=linux-arm -DisableCrossgen=true -PortableBuild=true -SkipTests=true -CrossBuild=true",
             "PB_CrossBuildArgs": "-e ROOTFS_DIR ",
-            "PB_PortableBuild": "true" 
+            "PB_PortableBuild": "true",
+            "PublishDebToolToFeed": "true"
```
eerhardt: We are publishing this during the `linux-arm` build?  That seems like an odd place.

karajas: I'll switch this back to the ubuntu 1404 leg

chcosta: I thought this previously was in the Ubuntu 14.04 build.

publish/publish.proj
```diff
@@ -152,6 +152,19 @@
...
+    <Error Condition="'$(CliNuGetFeedUrl)' ==''" Text="Missing required property CliNuGetFeedUrl" />
+    <Error Condition="'$(CliNuGetApiKey)' == ''" Text="Missing required property CliNuGetApiKey" />
+    <ItemGroup>
+      <DebToolPackages Include="$(PackagesOutDir)intermediate/*.nupkg" />
```
eerhardt: Using the `intermediate` folder here seems less than ideal.  I don't really like that we are producing the deb-tool nupkg here and then copying it out.

What if instead, we said:

`<DebToolPackages Include="$(PackagesOutDir)/dotnet-deb-tool.*.nupkg" />`

This is similar logic to what we are doing here:

[https://github.com/dotnet/core-setup/blob/master/src/pkg/packaging/deb/package.targets#L18-L23](https://github.com/dotnet/core-setup/blob/master/src/pkg/packaging/deb/package.targets#L18-L23)

buildpipeline/pipeline.json
karajas: fixed

publish/publish.proj
karajas: ok sounds good

karajas:
PTAL

publish/publish.proj
```diff
@@ -152,6 +152,19 @@
     <Exec Condition="'@(_DownloadedSymbolsPackages)' != ''" Command="$(NuGetPushSymbolsCommand) %(_DownloadedSymbolsPackages.Identity)" />
   </Target>
   
+  <Target Name="PublishDebToolPackageToFeed"
+          Condition="'$(PublishDebToolToFeed)' == 'true'">
+    <Error Condition="'$(CliNuGetFeedUrl)' ==''" Text="Missing required property CliNuGetFeedUrl" />
```
eerhardt: I'm not seeing where these 2 properties are being set.  Are they being set?

karajas: set them in Core-Setup-Linux-BT.json above

karajas:
Pointing to the right feed now

Issue #2144, opened 2017-04-21
Generate Native (crossgen) PDBs for Profiling for M.N.App


gkhanna79:
These are required for enabling profiling scenarios. This is for post 5/10.

@eerhardt Can you please file the corresponding issue against @ramarag to enable support for the same in store command as well?

CC @vancem @lt72 @brianrob @Petermarcu

eerhardt:
> @eerhardt Can you please file the corresponding issue against @ramarag to enable support for the same in store command as well?

We already have a work item for it https://github.com/dotnet/sdk/issues/1126.  It is assigned to me and I'm working on it.

eerhardt:
Fixed by #2397.

PR #2397, opened 2017-05-15
Generate crossgen symbols for the shared framework.

eerhardt:
Enables creating symbols (.ni.pdb on Windows and perf map .map on Linux) for our crossgen'd assemblies.  This allows profilers to work with crossgen'd/ready-to-run assemblies.

Note that OSX doesn't have symbols/perfmaps for crossgen, so we can't create them for OSX.

### DIFF
--- a/src/dir.props
+++ b/src/dir.props
@@ -27,5 +27,10 @@
     <PackageTargetRid Condition="'$(TargetRid)' == 'rhel.7.3-x64'">rhel.7-x64</PackageTargetRid>
     <PackageTargetRid Condition="'$(TargetRid)' == 'rhel.7.2-x64'">rhel.7-x64</PackageTargetRid>    
   </PropertyGroup>
-    
+
+  <PropertyGroup>
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'Windows_NT'">.ni.pdb</CrossGenSymbolExtension>
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'OSX'"></CrossGenSymbolExtension> <!-- OSX doesn't have crossgen symbols, yet -->
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' != 'Windows_NT' and '$(OSGroup)' != 'OSX'">.map</CrossGenSymbolExtension>
+  </PropertyGroup>
 </Project>
--- a/src/dir.props
+++ b/src/dir.props
@@ -29,8 +29,9 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <CrossGenSymbolExtension>.map</CrossGenSymbolExtension>
     <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'Windows_NT'">.ni.pdb</CrossGenSymbolExtension>
-    <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'OSX'"></CrossGenSymbolExtension> <!-- OSX doesn't have crossgen symbols, yet -->
-    <CrossGenSymbolExtension Condition="'$(OSGroup)' != 'Windows_NT' and '$(OSGroup)' != 'OSX'">.map</CrossGenSymbolExtension>
+    <!-- OSX doesn't have crossgen symbols, yet -->
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'OSX'"></CrossGenSymbolExtension>
   </PropertyGroup>
 </Project>
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -18,6 +18,7 @@
     <IntermediateOutputPath>$(IntermediateOutputPath)$(NuGetRuntimeIdentifier)</IntermediateOutputPath>
     <RestoreOutputPath>$(IntermediateOutputPath)</RestoreOutputPath>
     <AdditionalRestoreArgs>/p:HasRuntimePackages=false /p:IntermediateOutputPath=$(IntermediateOutputPath)</AdditionalRestoreArgs>
+    <CrossGenSymbolsOutputPath>$(PackageSymbolsBinDir)/$(MSBuildProjectName)</CrossGenSymbolsOutputPath>
   </PropertyGroup>
 
   <ItemGroup>
@@ -208,7 +209,13 @@
     <ItemGroup>
       <_filesToCrossGen Include="@(FilesToPackage)" 
                         Condition="'%(FilesToPackage.IsNative)' != 'true' AND '%(FileName)' != 'System.Private.CoreLib' AND '%(FileName)' != 'mscorlib' AND '%(Extension)' == '.dll'">
+        <CrossGenedDirectory>$(CrossGenOutputPath)/%(TargetPath)/</CrossGenedDirectory>
         <CrossGenedPath>$(CrossGenOutputPath)/%(TargetPath)/%(FileName)%(Extension)</CrossGenedPath>
+        
+        <!-- Note on non-Windows this path isn't accurate because crossgen inserts the hash of the crossgen'd image
+             into the symbol's file name. So incremental build won't work correctly on non-Windows.
+        -->
+        <CrossGenSymbolPath>$(CrossGenSymbolsOutputPath)/%(TargetPath)/%(FileName)$(CrossGenSymbolExtension)</CrossGenSymbolPath>
       </_filesToCrossGen>
 
       <FilesToPackage Remove="@(_filesToCrossGen)" />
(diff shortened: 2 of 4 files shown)


eerhardt:
Fix #2144

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -256,7 +266,47 @@
...
+    </ItemGroup>
+
+    <WriteLinesToFile File="$(_crossGenSymbolsResponseFile)" Lines="@(_crossGenSymbolsArgs)" Overwrite="true" />
+    
+    <MakeDir Directories="$(_crossGenSymbolsOutputDirectory)" />
+    
+    <Exec Command="$(_crossGenPath) @$(_crossGenSymbolsResponseFile)" WorkingDirectory="$(_clrDirectory)" EnvironmentVariables="COMPlus_PartialNGen=0" />
+  </Target>
+
+  <Target Name="GetCrossGenSymbolsFiles"
```
ericstj: Why didn't you do this in PrepareForCrossGen?

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -256,7 +266,47 @@
...
+          Condition="'$(CrossGenSymbolExtension)' != ''"
+          DependsOnTargets="CreateCrossGenSymbols">
+    <ItemGroup>
+      <FilesToPackage Include="$(CrossGenSymbolsOutputPath)/**/*$(CrossGenSymbolExtension)">
```
ericstj: Do we really need a wildcard?  I'd prefer a transform.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -208,7 +209,13 @@
     <ItemGroup>
       <_filesToCrossGen Include="@(FilesToPackage)" 
                         Condition="'%(FilesToPackage.IsNative)' != 'true' AND '%(FileName)' != 'System.Private.CoreLib' AND '%(FileName)' != 'mscorlib' AND '%(Extension)' == '.dll'">
+        <CrossGenedDirectory>$(CrossGenOutputPath)/%(TargetPath)/</CrossGenedDirectory>
         <CrossGenedPath>$(CrossGenOutputPath)/%(TargetPath)/%(FileName)%(Extension)</CrossGenedPath>
+        
+        <!-- Note on non-Windows this path isn't accurate because crossgen inserts the hash of the crossgen'd image
+             into the symbol's file name. So incremental build won't work correctly on non-Windows.
```
ericstj: You can fix that by finding the file after crossgen produces it, then moving/copying it to a predictable path.

eerhardt: I use this in the `Outputs` of the Target that produces that produces it.

ericstj: I understand, I'm just saying that you can fix the part about the hash by finding the file that crossgen produced and then copying it to the expected filename.  So if crossgen produced foo-1234abc.pdb, find foo-*.pdb and move/copy to foo.pdb.  That way even though crossgen doesn't have a predictable path you make your target have one.  Barring that, use a semaphore file that you touch as the output.

eerhardt: But we would never use that predictable path outside of the incremental build check right?

It's my understanding/assumption that these perf maps need to have the hash in their file name.  @brianrob can confirm if that is true or not.  If it is true, we have to be packing up the perfmap files with the hash name in it.  So these predictable path files would only be used for incremental builds.

ericstj: In that case use an empty semaphore created by Touch task as the output.

brianrob: @eerhardt, you're correct, the file names need to remain as they are set with GUIDs.  The GUID is used to map back to the native image so that we can validate that we have the right symbols.

ericstj: Ok, as mentioned if you need to preserve the non-deterministic path then create something else that's deterministic to use as an output.  EG: an empty semaphore file in the intermediate directory.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
ericstj: I guess you do need a wildcard because of the non-deterministic paths.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
eerhardt: 👍 working on that now.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
ericstj: You can't do that if you keep the wildcard, disregard this.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -256,7 +266,47 @@
...
+    <ItemGroup>
+      <FilesToPackage Include="$(CrossGenSymbolsOutputPath)/**/*$(CrossGenSymbolExtension)">
+        <IsSymbolFile>true</IsSymbolFile>
+        <TargetPath>runtimes/$(NuGetRuntimeIdentifier)/lib/$(PackageTargetFramework)</TargetPath>
+      </FilesToPackage>
+    </ItemGroup>
+  </Target>
+  
+  <Target Name="GetFilesToPackage"
+          DependsOnTargets="ResolveNuGetPackages;GetFilesFromPackages;PrepareForCrossGen;GetCrossGenSymbolsFiles"
```
ericstj: This target should not build anything since it will run during packaging.  Now you're going to cause crossgen to run due to the target dependencies.  I think you'll need to omit the dependency GetCrossGenSymbolsFiles > CreateCrossGenSymbols and put a comment in GetCrossGenSymbolsFiles that it assumes that the project has already been built.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
eerhardt: Updated.  PTAL.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
eerhardt: Fixed.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -68,11 +68,11 @@
       <_runtimePackageId>%(_runtimeCLR.NuGetPackageId)</_runtimePackageId>
       <_runtimePackageVersion>%(_runtimeCLR.NuGetPackageVersion)</_runtimePackageVersion>
       <_runtimePackagePath>$(PackagesDir)/$(_runtimePackageId.ToLowerInvariant())/$(_runtimePackageVersion)</_runtimePackagePath>
-      
+
```
eerhardt: Uggg, I'll rebase these changes out.

src/dir.props
```diff
@@ -27,5 +27,10 @@
     <PackageTargetRid Condition="'$(TargetRid)' == 'rhel.7.3-x64'">rhel.7-x64</PackageTargetRid>
     <PackageTargetRid Condition="'$(TargetRid)' == 'rhel.7.2-x64'">rhel.7-x64</PackageTargetRid>    
   </PropertyGroup>
-    
+
+  <PropertyGroup>
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'Windows_NT'">.ni.pdb</CrossGenSymbolExtension>
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' == 'OSX'"></CrossGenSymbolExtension> <!-- OSX doesn't have crossgen symbols, yet -->
+    <CrossGenSymbolExtension Condition="'$(OSGroup)' != 'Windows_NT' and '$(OSGroup)' != 'OSX'">.map</CrossGenSymbolExtension>
```
dagood: Nit: it seems like it would be clearer to have the catch-all setting conditionless and first, or `Condition="'$(CrossGenSymbolExtension)' == ''"` and last.

eerhardt: Can't use `Condition="'$(CrossGenSymbolExtension)' == ''" and last.` since OSX explicitly needs it to be set to blank.

I'll move this line to be first, and then OSX will blank it out.

Issue #2376, opened 2017-05-12
dotnet.exe does not handle absolute paths with mixed slashes


natemcmaster:
## Steps to reproduce
`dotnet.exe C:\path/to\some/assembly.dll`

## Expected  behavior
Loads assembly.dll for execution

## Actual behavior
> A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in (working directory)

## Environment data
Using: 2.0.0-preview2-25309-07 as bundled in .NET Core CLI 2.0.0-preview2-006065
Windows 10 x64

## Workaround
Convert all forward slashes to backslashes
`dotnet.exe C:\path\to\some\assembly.dll`

gkhanna79:
@eerhardt confirmed that this does not repro with build 2.0.0-preview2-25307-01.  Possible regression from https://github.com/dotnet/core-setup/commit/6186962c2546713d521dcee2afeb481fa589ecd0#diff-29fc8a28403b84f9a93634c7c93329d0 or https://github.com/dotnet/core-setup/commit/3b8111f53a59e846ddcba363c3c5d0c0acfb7a2a#diff-29fc8a28403b84f9a93634c7c93329d0.

@ramarag PTAL.

gkhanna79:
@ramarag PTAL at this when you are back as this appears to be high priority issue.

PR #2400, opened 2017-05-16
Account for ALT_DIR_SEPARATOR

ramarag:
fixes https://github.com/dotnet/core-setup/issues/2376

@gkhanna79 @eerhardt PTAL

### DIFF
--- a/src/corehost/common/longfile.windows.cpp
+++ b/src/corehost/common/longfile.windows.cpp
@@ -16,9 +16,13 @@ const pal::string_t LongFile::DevicePathPrefix = _X("\\\\.\\");
 const pal::string_t LongFile::UNCExtendedPathPrefix = _X("\\\\?\\UNC\\");
 const pal::string_t LongFile::UNCPathPrefix = _X("\\\\");
 
+bool LongFile::IsNormalized(const pal::string_t& path)
+{
+    return path.empty() || LongFile::IsDevice(path) || LongFile::IsExtended(path) || LongFile::IsUNCExtended(path);
+}
 bool ShouldNormalizeWorker(const pal::string_t& path)
 {
-    if (path.empty() || LongFile::IsDevice(path) || LongFile::IsExtended(path) || LongFile::IsUNCExtended(path))
+    if (LongFile::IsNormalized(path))
         return false;
 
     if (!LongFile::IsPathNotFullyQualified(path) && path.size() < MAX_PATH)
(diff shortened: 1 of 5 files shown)


src/corehost/cli/args.cpp
```diff
@@ -136,6 +136,7 @@ bool parse_arguments(
 
         if (!app_base.empty() && app_base.back() != DIR_SEPARATOR)
         {
+            assert(app_base.back() != ALT_DIR_SEPARATOR);
```
gkhanna79: Can we ever be in a case when this assert will fire?

src/corehost/common/longfile.windows.cpp
```diff
@@ -65,6 +65,18 @@ bool AssertRepeatingDirSeparator(const pal::string_t& path)
 
     assert(path_to_check.find(altDirSeparator) == pal::string_t::npos);
 
+    pal::string_t combDirSeparator1;
+    combDirSeparator1.push_back(LongFile::DirectorySeparatorChar);
+    combDirSeparator1.push_back(LongFile::AltDirectorySeparatorChar);
+
+    assert(path_to_check.find(combDirSeparator1) == pal::string_t::npos);
+
+    pal::string_t combDirSeparator2;
```
gkhanna79: I dont follow this - you are repeating the same check as above?

src/corehost/common/utils.cpp
```diff
@@ -118,7 +119,7 @@ pal::string_t get_filename(const pal::string_t& path)
         return path;
     }
 
-    auto name_pos = path.find_last_of(DIR_SEPARATOR);
+    auto name_pos = path.find_last_of(_X("/\\"));
```
gkhanna79: Why are we searching for "/\"?

ramarag: that is the root cause of the bug

ramarag:
pushed a better fix and reverted the path processing back to the original way

src/corehost/common/longfile.windows.cpp
ramarag: fixed it

eerhardt:
> //Please reflect any change here into the above file too!

Are you following this comment?  Are these changes going to be reflected in coreclr?

If not, we might as well remove this comment.

---
Refers to: src/corehost/common/longfile.windows.cpp:5 in 0f4cec1. [](commit_id = 0f4cec165f8681ffb1ef4020baca27748e59a093, deletion_comment = False)

eerhardt:
What are your thoughts about testing this change?

src/corehost/common/longfile.windows.cpp
```diff
@@ -65,6 +69,19 @@ bool AssertRepeatingDirSeparator(const pal::string_t& path)
 
     assert(path_to_check.find(altDirSeparator) == pal::string_t::npos);
 
+    pal::string_t combDirSeparator1;
+    combDirSeparator1.push_back(LongFile::DirectorySeparatorChar);
+    combDirSeparator1.push_back(LongFile::AltDirectorySeparatorChar);
+
+    assert(path_to_check.find(combDirSeparator1) == pal::string_t::npos);
```
gkhanna79: Is this an invariant that will always hold? Can there be a case if this happen because if it does, then the assert will not help.

Same comment about the change below.

gkhanna79:
And yes, I think we should add a test for this to identify such breaks.

src/corehost/common/longfile.windows.cpp
ramarag: The above asserts are to ensure bad path manipulations are not done. These will largely work for short paths, but if the extended prefix is added then they will not work

gkhanna79: Do we not want to support path specification why the extended prefix? Also, what does it mean that this will "largely" work for short paths - what cases will not work?

ramarag: For  example C:\users\user\..//new will work fine
But \\?\C:\users\user\..// will not work.

These asserts will prevent appending such special characters to a path. As with the extended syntax such appends will not work

gkhanna79: Got it - thanks for the explanation.

PR #2401, opened 2017-05-16
Pass required checksum properties during publish step for Linux legs

rakeshsinghranchi:
Newly added feature to publish checksum files requires mandatory properties for publishing .  Official build will still fail as ChecksumAzureAccessToken value is not getting set from pipebuild.


karajas:
LGTM

PR #2411, opened 2017-05-16
OfficialBuildId property is required to generate version prop

rakeshsinghranchi:
This should unblock building the official builds for Linux leg. OfficialBuildId property is required to generate the version prop file.


karajas:
LGTM.

karajas:
This can be merged - CI doesn't validate this

PR #2414, opened 2017-05-16
Fixing Property typo issue causing incomplete tar package push.

rakeshsinghranchi:


rakeshsinghranchi:
cc @chcosta , PTAL

buildpipeline/Core-Setup-Linux-BT.json
```diff
@@ -814,7 +814,7 @@
       "value": "azure-apt-cat.cloudapp.net"
     },
     "AdditionalMSBuildProperties": {
-      "value": "/p:UsePrebuiltPortableBinaries=true /p:SharedFrameworkPublishDir=/root/sharedFrameworkPublish/"
+      "value": "/p:UsingPrebuiltPortableBinariesForInstallers=true /p:SharedFrameworkPublishDir=/root/sharedFrameworkPublish/"
```
chcosta: Can we make it `UsePrebuiltPortableBinariesForInstallers`?  I think of the property as a directive, so syntactically it makes more sense to say "use" instead of "using", though I'd be ok with `UsePrebuiltPortableBinaries".

PR #2416, opened 2017-05-16
Update BuildTools to 01616-05

karajas:
This is to update buildtools.

This change should include the fix for clean up tasks /cc @chcosta


chcosta:
Let's wait on this until after we get a clean build and fork for release.

karajas:
Should we bump this to latest?

PR #2422, opened 2017-05-17
Move Core-Setup to 2.1 Preview1

gkhanna79:
@weshaggard @eerhardt PTAL

CC @leecow

### DIFF
--- a/branchinfo.txt
+++ b/branchinfo.txt
@@ -2,10 +2,10 @@
 # Any line that is not blank and does not start with '#' is interpreted as a variable to set.
 # Each line is expected to be in the format "[Name]=[Value]".
 MAJOR_VERSION=2
-MINOR_VERSION=0
+MINOR_VERSION=1
 PATCH_VERSION=0
 STABILIZE_PACKAGE_VERSION=false
 LOCK_HOST_VERSION=false
-RELEASE_SUFFIX=preview2
+RELEASE_SUFFIX=preview1
 CHANNEL=master
 BRANCH_NAME=master
--- a/src/managed/CommonManaged.props
+++ b/src/managed/CommonManaged.props
@@ -5,7 +5,7 @@
 
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
-    <VersionPrefix>2.0.0</VersionPrefix>
+    <VersionPrefix>2.1.0</VersionPrefix>
     <AssemblyFileVersion>$(VersionPrefix)</AssemblyFileVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
--- a/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
@@ -6,7 +6,10 @@
 
   <PropertyGroup>
     <Version>$(SharedFrameworkNugetVersion)</Version>
-    <NETCoreAppFramework>netcoreapp$([System.Version]::Parse('$(ProductionVersion)').ToString(2))</NETCoreAppFramework>
+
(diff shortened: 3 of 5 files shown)


branchinfo.txt
```diff
@@ -2,10 +2,10 @@
 # Any line that is not blank and does not start with '#' is interpreted as a variable to set.
 # Each line is expected to be in the format "[Name]=[Value]".
 MAJOR_VERSION=2
-MINOR_VERSION=0
+MINOR_VERSION=1
```
eerhardt: @chcosta - we should look at what is still using the branchinfo.txt file and switch it over to use the dir.props file.

I'm fairly certain this file should be deleted by now.

weshaggard: I agree.

chcosta: i'm not aware of anything still using branchinfo.txt.  Part of the intention of the buildtools effort was to make everything use dir.props.

src/managed/CommonManaged.props
```diff
@@ -5,7 +5,7 @@
 
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
-    <VersionPrefix>2.0.0</VersionPrefix>
+    <VersionPrefix>$(VersionPrefix)</VersionPrefix>
```
eerhardt: This doesn't work.  It is setting a property to itself, which is not yet defined.  Just hard-code this to `2.1.0`.  We have a bug to fix this for real, which will cause more work:  https://github.com/dotnet/core-setup/issues/2435

src/managed/dir.proj
```diff
@@ -7,14 +7,16 @@
...
-      <Args>/p:Configuration=$(ConfigurationGroup)</Args>
+      <Args>/p:Configuration=$(ConfigurationGroup) /p:VersionPrefix=$(ProductVersion)</Args>
     </PropertyGroup>
 
     <ItemGroup>
       <PackageProjects Include="$(MSBuildThisFileDirectory)Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj" />
       <PackageProjects Include="$(MSBuildThisFileDirectory)Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj" />
     </ItemGroup>
 
+    <Message Text="Buildargs: $(Args)"/>
```
eerhardt: Is this necessary?

src/managed/CommonManaged.props
eerhardt: Also, I believe this change is going to cause the suffix to be appended twice.  Just keep this hard-coded for now.  The referenced bug will fix this for real.

chcosta: hmm... that's not what I have on my machine.  I have 

```
    <VersionPrefix>2.0.0</VersionPrefix>
    <AssemblyFileVersion>$(VersionPrefix)</AssemblyFileVersion>
```
I'll try to figure out why this diff is weird

chcosta: oh, this is a different PR.  ignore me

gkhanna79:
all changes in place - @eerhardt PTAL.

src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -6,7 +6,10 @@
 
   <PropertyGroup>
     <Version>$(SharedFrameworkNugetVersion)</Version>
```
weshaggard: While not exactly related should this also be ProductVersion instead of SharedFrameworkNugetVersion?

eerhardt: Actually, I think this is one of the few correct usages of this property.  `Microsoft.NETCore.App`'s version is `$(SharedFrameworkNugetVersion)`. It just so happens they always align today.

If we ever were servicing something like UWP and wanted to keep `Microsoft.NETCore.App` static, like we do for Host and HostFxr, this should be ProductVersion.

All that being said, I don't know if we'd ever want/need that.  My guidance is:  if you are generally talking about the version, not tied to any specific component, use `ProductVersion`.  If you are specifically referencing a component, and it has its own version property, use it.

PR #2426, opened 2017-05-17
Revert "Fix deb package naming"

eerhardt:
Reverts dotnet/core-setup#2390.

This change was not correct.  We need the hostfxr and shared framework packages to be side-by-side installs.  That way you can install 2.0 and 2.1, and have them both installed at the same time.


eerhardt:
@gkhanna79 @rakeshsinghranchi @karajas

Issue #186, opened 2016-06-17
Strip symbols for x-plat assemblies and publish as NuPkgs


schellap:
## Steps to reproduce

`objdump -t libhostpolicy.so`
## Expected  behavior

Similar to coreclr,
`objdump -t libcoreclr.so`
and a separate file for the symbol information.

This separate file needs to be archived in symbol nupkgs like Windows .pdbs
## Actual behavior

`objdump -t libhostpolicy.so`
contains the full symbol table.
## Environment data

`dotnet --info` output:
Microsoft.NETCore.App 1.0.0
## Note

libhostpolicy.so is the highest sized binary, if you strip symbols you get 1.1 MB down to 0.78 MB

gkhanna79:
This is not blocking release and is best done once we have moved our build/packaging system to build tools, to prevent non-trivial amount of throwaway work.

gkhanna79:
@AlfredoMS Can you PTAL at this as part of your transition to use build tools? Essentially, we need to generate symbol packages for the host binaries.

dagood:
The CMake parts of enabling symbol stripping are [CMakeLists.txt#L139-L230](https://github.com/dotnet/corefx/blob/eab37f9a596583b55e67568036dc1987c1319447/src/Native/Unix/CMakeLists.txt#L139-L230). The `STRIP_SYMBOLS` arg is passed in by [build-native.sh#L216](https://github.com/dotnet/corefx/blob/eab37f9a596583b55e67568036dc1987c1319447/src/Native/build-native.sh#L216).

After that, to start putting the stripped symbols into symbol packages, I suggest a file like [src/Native/pkg/dir.targets](https://github.com/dotnet/corefx/blob/eab37f9a596583b55e67568036dc1987c1319447/src/Native/pkg/dir.targets). (For some context, https://github.com/dotnet/coreclr/pull/9664 might be useful.)

@AlfredoMS Let me know when you're getting to a point when you can work on this and let's talk in more depth.

danmosemsft:
@dagood is this still on track for 5/10 ZBB?
/cc @markwilkie @dleeapho

dagood:
@gkhanna79 Is this needed for 2.0.0? I haven't had it on my radar.

I think it would be most efficient if someone working on the BuildTools Core-Setup effort picked it up, especially if they are familiar with packaging in the new world.

gkhanna79:
@chcosta Can you help look into this?

chcosta:
[...]

PR #2421, opened 2017-05-17
Do not add MSTestResults in armel builds

jyoungyun:
Fixed Tizen armel Release CI failed issue

/cc @gkhanna79

PR #2430, opened 2017-05-17
Strip xplat symbols and repackage

chcosta:
Fixes https://github.com/dotnet/core-setup/issues/186

Strip symbols from xplat binaries, and package them in the nuget packages.

Symbol stripping is opt-in by specifying the `-strip-symbols` property when using the run tool or `/p:StripSymbols=true` for MSBuild.

@dagood @gkhanna79 @ericstj @eerhardt


src/corehost/cli/dll/CMakeLists.txt
```diff
@@ -67,3 +67,4 @@ if (WIN32 AND CLI_CMAKE_PLATFORM_ARCH_ARM)
     target_link_libraries(hostpolicy shell32.lib)
 endif()
 
+install_library_and_symbols (hostpolicy)
```
dagood: Keep the newlines at the ends of files for xplat?

src/pkg/projects/dir.targets
```diff
@@ -6,6 +6,43 @@
...
+      </File>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <NeedsPlaceholderPdb Condition="'@(ExistingNonWindowsSymbolFile)' != '' and '@(ExistingWindowsSymbolFile)' == ''">true</NeedsPlaceholderPdb>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <File Include="$(ToolsDir)\_.pdb"
+            Condition="'$(NeedsPlaceholderPdb)' == 'true' AND '$(PackageTargetRuntime)' != ''">
```
dagood: What does the `'$(PackageTargetRuntime)' != ''` check prevent?

config.json
```diff
@@ -72,6 +72,12 @@
       "values": [],
       "defaultValue": "/flp:v=normal"
     },
+    "StripSymbols": {
```
dagood: Are you going to enable this for the official builds?

src/pkg/projects/dir.targets
chcosta: Identity packages

config.json
chcosta: That was my intention.  If you're asking why I didn't include that change in this PR then... I don't know why I was mentally separating the two.  I can include it.

src/corehost/cli/dll/CMakeLists.txt
chcosta: This was copied from the other cmakelists.txt files.  I can remove the newlines if it's annoying.

src/pkg/projects/dir.targets
dagood: Do any identity packages currently end up with `ExistingNonWindowsSymbolFile` items (setting `NeedsPlaceholderPdb` to true) or is this just for safety?

src/corehost/cli/dll/CMakeLists.txt
dagood: Oh, sorry, the red "no newline at end of file" marker only shows up in the files diff view, so you probably didn't see it. I was just suggesting adding a newline after this line to make linux happy. The empty line before this line looks good to me.

src/pkg/projects/dir.targets
chcosta: If any identity packages included native binaries / symbols, that would be weird.  This code, however, was copied from corefx and seemed safe - https://github.com/dotnet/corefx/blob/master/src/Native/pkg/dir.targets#L54

dagood: Hmm, fair enough... I need to ask myself, apparently. :smile:

buildpipeline/pipeline.json
```diff
@@ -20,7 +20,7 @@
           "Parameters": {
             "PB_DistroRid": "rhel.7.2-x64",
             "PB_DockerTag": "rhel7_prereqs_2",
-            "PB_AdditionalBuildArguments":"-PortableBuild=true",
+            "PB_AdditionalBuildArguments":"-PortableBuild=true -strip-symbols",
```
eerhardt: Why is `-strip-symbols` opt-in?  Shouldn't it be the default since all of our builds opt-in to it?

src/corehost/CMakeLists.txt
```diff
@@ -1,2 +1,132 @@
...
+        endif()
+        if (OBJCOPY STREQUAL "OBJCOPY-NOTFOUND" AND NOT CMAKE_SYSTEM_PROCESSOR STREQUAL i686)
+            message(FATAL_ERROR "objcopy not found")
+        endif()
+    endif (CMAKE_SYSTEM_NAME STREQUAL Darwin)
+endif ()
+
+function(strip_symbols targetName outputFilename)
+    if(CLR_CMAKE_PLATFORM_UNIX STREQUAL 1)
+        message ("chcosta CLR_CMAKE_PLATFORM_UNIX = 1")
```
eerhardt: Do we really want your alias in here?

src/corehost/CMakeLists.txt
```diff
@@ -1,2 +1,132 @@
...
+    else(CLR_CMAKE_PLATFORM_UNIX STREQUAL 1)
+        message ("chcosta CLR_CMAKE_PLATFORM_UNIX is not 1")
```
eerhardt: Same as above

src/corehost/CMakeLists.txt
chcosta: haha, good catch

buildpipeline/pipeline.json
chcosta: Hunting around, it appeared that symbol stripping applied only to Release builds.  The alternative would be to have different default behavior between Debug / Release.  I modeled the behavior after how this is being handled in coreclr

src/dir.props
```diff
@@ -9,11 +9,18 @@
...
     <LibPrefix Condition="'$(OSGroup)' != 'Windows_NT'">lib</LibPrefix>
+    <LibSuffix>.so</LibSuffix> 
     <LibSuffix Condition="'$(OSGroup)' == 'Windows_NT'">.dll</LibSuffix>
     <LibSuffix Condition="'$(OSGroup)' == 'OSX'">.dylib</LibSuffix>
-    <LibSuffix Condition="'$(OSGroup)' != 'Windows_NT' and '$(OSGroup)' != 'OSX'">.so</LibSuffix>  
+    <SymbolFileExtension>.dbg</SymbolFileExtension>
+    <SymbolFileExtension Condition="'$(OSGroup)' == 'OSX'">.dwarf</SymbolFileExtension>
+    <SymbolFileExtension Condition="'$(OSGroup)' == 'Windows_NT'">.pdb</SymbolFileExtension> 
+
+    <CMakeBuildDir>$(IntermediateOutputRootPath)corehost\cmake\</CMakeBuildDir>
```
eerhardt: Do these path properties need to be put in the .props file?  they are only used in a single proj file.

IMO - it is better to keep these defines closer to where they are used.  It makes the code easier to reason about.  I only pull properties into .props files if they are being used across multiple targets/projs.

buildpipeline/pipeline.json
dagood: Symbol stripping needs to be independently switchable for build from source, and I set the default to "don't strip" because stripping is fairly specific to our builds: https://github.com/dotnet/corefx/pull/15440#issuecomment-275202555.

src/dir.props
chcosta: The BaseName properties were already defined here (because HostPolicyBaseName is used both in corehost project and sharedframework.proj), and it made more sense to me to define the paths in the same place in case someone wanted to make an update and wants to easily see what they are updating.  I tend to err towards defining things in the props file just because when they are statically defined, it's much easier to track down the value in an msbuild diag log.

chcosta: That was my reasoning anyways.  I am not opposed to moving these

chcosta:
Tizen CI failure tracked with https://github.com/dotnet/core-setup/issues/2432

gkhanna79:
Tizen CI failure will be fixed when https://github.com/dotnet/core-setup/pull/2421 is merged.

buildpipeline/pipeline.json
eerhardt: sounds good to me.  Thanks

src/dir.props
eerhardt: No, I don't think they need to move, if you want it this way.

FYI - You can still statically define properties in .targets and .proj files - so they are easy to track down in diag logs.

The reason I like to keep values defined close to where they are used is for the new person trying to read the code.  So they don't need to go searching all over for "what is the property's value?" when they are reading the code.

PR #2431, opened 2017-05-17
Add fragile pdbs again

eerhardt:
This includes reverting the original revert.  And then adding the support for the official builds to use crossgen -CreatePDB by initializing a VS developer command prompt through run.cmd.

Probably only need to review the last commit, since the first one was already reviewed in #2397.

@chcosta @mellinoe


jyoungyun:
If you rebase the netci.groovy file, the Tizen CI build failure is resolved.

PR #2442, opened 2017-05-19
Update channel and branch for release/2.0.0

eerhardt:
@gkhanna79 @weshaggard @chcosta @karajas 

This one is for the right branch.

PR #2445, opened 2017-05-19
Fix Channel to be used for Azure publishing

gkhanna79:
@eerhardt @dagood @chcosta PTAL

CC @weshaggard


eerhardt:
What's the difference between this and https://github.com/dotnet/core-setup/pull/2442?

gkhanna79:
No difference - just forgot you had the PR out :)

PR #2456, opened 2017-05-19
Finalize checksums

karajas:
skip ci please 

@eerhardt Finalize checksums to add links to README


publish/publish.proj
```diff
@@ -99,6 +99,22 @@
                    FinalizeContainer="Runtime/$(SharedFrameworkNugetVersion)"
                    ForcePublish="true"                  
                    Condition="'@(_MissingBlobNames)' == ''" />
+    <Error Condition="'$(ChecksumAzureAccessToken)' == ''" Text="Missing required property 'ChecksumAzureAccessToken'" />
+    <Error Condition="'$(ChecksumAzureAccountName)' == ''" Text="Missing required property 'ChecksumAzureAccountName'" />
+    <Error Condition="'$(ChecksumContainerName)' == ''" Text="Missing required property 'ChecksumContainerName'" />
+    <Message Importance="High" Text="Finalizing Checksums" />
+    <FinalizeBuild AccountName="$(ChecksumAzureAccountName)"
```
eerhardt: I think this should go into the FinalizeBuild task instead.  We should only call it once.

Isn't this just going to publish our assets to the other azure storage account?

karajas: Yup this is going to the checksums account. Reusing the task seemed ok to me since it essentially ensures that the checksum account will always have the layout as the actual publish output.

eerhardt: Ah, ok.  I get it.

As long as the FinalizeBuild task doesn't do something that should only be done for the real assets account, then this approach should work.

karajas:
@chcosta @eerhardt PTAL. There was an issue with updating universalwindowsplatform packages in the latest folder of the storage accounts.

@dotnet-bot test this please

Issue #2214, opened 2017-04-27
Generate RPM packages for Portable Linux binaries


vivmishra:
This work item tracks the work for generating RPM packages for distros (RHEL/Centos/Fedora) supporting RPM.

vivmishra:
Merged

PR #2462, opened 2017-05-20
Issue #2214: Generate RPM packages for Runtime Portable Linux binaries (2.1)

vivmishra:
This PR introduces following changes for Issue #2214  

- New msbuild targets/props; config json; scripts; templates for handling RPM package creation for Runtime (host, hostfxr and sharedFX).
        #	new file:   src/pkg/packaging/rpm/package.targets
        #	new file:   src/pkg/packaging/rpm/package.props
        #	new file:   src/pkg/packaging/rpm/dotnet-hostfxr-rpm_config.json
        #	new file:   src/pkg/packaging/rpm/dotnet-sharedframework-rpm_config.json
        #	new file:   src/pkg/packaging/rpm/dotnet-sharedhost-rpm_config.json
        #	new file:   src/pkg/packaging/rpm/scripts/after_install_host.sh
        #	new file:   src/pkg/packaging/rpm/scripts/after_remove_host.sh
        #	new file:   src/pkg/packaging/rpm/templates/changelog
        #	new file:   src/pkg/packaging/rpm/templates/copyright

- A new Task is added to handle the work of building the FPM tool parameters and updating the tokens in copyright and changelog files.
        #	new file:   tools-local/tasks/BuildFPMToolPreReqs .cs

- New GenerateRpms target is wired in packaging. 
- There is some refactoring in dir.props to detect InstallerExtension for .deb and .rpm
- New RIDs for rhel and centos.
        #	modified:   src/pkg/packaging/dir.proj
        #	modified:   dir.props
        #	modified:   publish/dir.props
      
- The Core-Setup-Linux-BT build definition has been updated to insert the new RPM package build steps.
        #	modified:   buildpipeline/Core-Setup-Linux-BT.json

- Readme is updated to wire the RPM packages under RHEL7.2. Eventhough, the packages install fine on Centos, the wiring for it has NOT been added yet. 
        #	modified:   README.md
     
- To support this packaging work a new Docker image was created earlier with fpm tool installed -- microsoft/dotnet-buildtools-prereqs:rhel-7-rpmpkg-c982313-20174116044113
- Packages currently only publish to Azure blob and setting up the feed part is still TBD.


src/pkg/packaging/rpm/dotnet-sharedframework-rpm_config.json
```diff
@@ -32,6 +32,14 @@
   "rpm_dependencies": [{
       "package_name": "%HOSTFXR_RPM_PACKAGE_NAME%",
       "package_version": ""
+    },
+    {
+      "package_name": "libssl1.0.0",
```
ellismg: I am not sure these are the correct package names.  On my RHEL machine, it looks like the packages we want to depend on are:

- openssl-libs
- libcurl

This also seems to match what Fedora is doing (no surprise).

vivmishra: Updated

src/pkg/packaging/rpm/package.targets
```diff
@@ -0,0 +1,271 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Import Project='package.props' />
+
+  <UsingTask TaskName="ReplaceFileContents" AssemblyFile="$(LocalBuildToolsTaskDir)core-setup.tasks.dll"/>
+  <UsingTask TaskName="BuildFPMToolPreReqs" AssemblyFile="$(LocalBuildToolsTaskDir)core-setup.tasks.dll"/>
+
+  <PropertyGroup>
+    <IsRPMBasedDistro Condition="$(PackageTargetRid.StartsWith('rhel')) or
```
ellismg: I'm not sure if the plan is to generate RPMs on OpenSuse as well, but that is also RPM based.

src/pkg/packaging/rpm/package.targets
```diff
@@ -0,0 +1,271 @@
...
+    <!-- Call the task to build the pre-reqs (parameters, copyright, changelog) for calling the FPM tool -->
+    <BuildFPMToolPreReqs  InputDir="$(rpmLayoutDirectory)"
+                          OutputDir="$(RpmIntermediatesDir)"
+                          PackageVersion="$(RpmPackageVersion)"
+                          ConfigJsonFile="$(rpmLayoutDirectory)$(rpmConfigJsonName)">
+                          <Output TaskParameter="FPMParameters" PropertyName="FPMCmdParameters" />
+    </BuildFPMToolPreReqs>
+
+    <!-- Build the RPM package by calling the FPM tool and passing the parameter list -->
+    <Exec Command="fpm $(FPMCmdParameters)" WorkingDirectory="$(RpmIntermediatesDir)" />
```
ellismg: The deb targets do a check to see if debuild is installed and if it is not, they don't build the targets.  Maybe we should do something similar for fpm?

ellismg:
For 2.1 do we plan to stop using dotnet-deb-tool and instead just use fpm everywhere?

src/pkg/packaging/rpm/scripts/after_install_host.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
ellismg: My gut tells me for these preinstall and post install scripts we should instead be using `#!/bin/sh` so we don't have a dependency on bash.

@omajid is that shebang reasonable or should we be using `#!/usr/bin/env sh`

src/pkg/packaging/rpm/scripts/after_remove_host.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+echo "Removing dotnet host symbolic link"
+sudo unlink /usr/bin/dotnet
```
ellismg: I'm very surprised that sudo was needed here.  Why is that?  Shouldn't these remove script be running as root anyway?

vivmishra: I tested the package without sudo and #!/bin/sh and it worked. I will update for these.

src/pkg/packaging/rpm/package.targets
vivmishra: The  docker image is built with fpm so it is expected to have the same. But I can make changes and retest.

ellismg: Yeah, my concern is more around a dev building the repository on some CentOS or RHEL machine where fpm is not installed (similar to how when building locally you don't need to have debuild installed).

src/pkg/packaging/rpm/scripts/after_remove_host.sh
ellismg: Awesome.

src/pkg/packaging/rpm/package.targets
vivmishra: For this round we are keeping this scoped to RHEL and Centos. In next round we can expand to Fedora and Opensuse once we have done more testing. 

We still do not have a generic file name for the rpms -- so that is another thing to resolve before we extend.

src/pkg/packaging/rpm/scripts/after_remove_host.sh
wfurt: also note that in many cases /bin/sh is bash. When /bin/sh is used, there may be surprises on distros where /bin/sh is more exotic - dash or original korn shell for example.

src/pkg/packaging/rpm/package.targets
vivmishra: The change was done and the build passed successfully.

Issue #2398, opened 2017-05-15
Various fixes for multi-level lookup


gkhanna79:
We need to do the following:

1) Disable looking up user-profile and only lookup dotnet EXE location and global location.
2) For patch roll forward for SharedFX, pickup the highest version from across all locations.

CC @davidfowl @nguerrera @ramarag @piotrpMSFT @blowdart @morganbr

gkhanna79:
CC @eerhardt

gkhanna79:
@ramarag Do you think we can get this change in today? I would like to have it bake in before Escrow next week.

natemcmaster:
Can I ask what's driving this change? Unfortunately, we updated ASP.NET Core's build system about a month ago to use %USERPROFILE% to keep our command-line builds and VS experience in sync. IIUC we can't do that anymore once this is in as scripts don't have write-permissions to global and VS doesn't look at local.

gkhanna79:
Please follow up offline with @davidfowl to understand the motivation and how the scenario will continue working.

gkhanna79:
Closing since fixed.

Issue #1553, opened 2017-02-15
Containerize the test execution


ramarag:
The dotnet hostpolicy will increasingly start to look in global locations for some artifacts, like servicing directory , assembly caches etc

These policies might interfere with the way tests execute, causing test failures on dev machine.
For example, if the dev machine has an assembly cache in their user profile containing newtsoft.json then
https://github.com/dotnet/cli/blob/c20e106d7570e8a1d24438765a7c9d9ce81de16c/test/dotnet-cache.Tests/GivenDotnetCachesAndPublishesProjects.cs#L104 is going to fail

Possible workarounds: 

On linux - We can just change the USER env to point to a different location
But on windows [SHSetKnownFolderPath ](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762249(v=vs.85).aspx) needs to be used to do this modification before and after the test

ramarag:
cc @gkhanna79 @piotrpMSFT

gkhanna79:
@ramarag There is only ever a single servicing location and it is looked up by hostfxr to ensure we can pickup the serviced hostpolicy. Later, ofcourse, hostpolicy does the same for other assemblies.

The already happens today.

That said, from the cache standpoint, I do agree.

piotrpMSFT:
I imagine this working similarly to `NUGET_PACKAGES`. We will be able to isolate ourselves within the context of a folder across the board. Dev isn't the only relevant experience here. I'm much more interested in ensuring isolation in CI!

ramarag:
@gkhanna79 , i filed this issue so that the test execution should not be modified due to the machine state.
So if there is a servicing done, the tests should not pick it up.
@piotrpMSFT : Generalized it to ensure each test do not interfere with others.

eerhardt:
Note - I removed the 'dead' tests because they were causing build warnings, which turn into build errors.

If this work item gets implemented, you revert https://github.com/dotnet/core-setup/commit/29c18275d557afcfa24f8007226556ffd5f2a5d5 to bring the tests back.

PR #2481, opened 2017-05-23
Roll fwd across hive

ramarag:
This change  disables user profile lookup of shared Fx
It also rolls forward across multiple probe locations

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -333,7 +333,7 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
     return false;
 }
 
-void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
+fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& version_list,
     const pal::string_t& fx_ver,
     const fx_ver_t& specified,
     const bool& patch_roll_fwd,
@@ -341,8 +341,7 @@ void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
 {
     trace::verbose(_X("Attempting FX roll forward starting from [%s]"), fx_ver.c_str());
 
-    std::vector<pal::string_t> list;
-    pal::readdir(fx_dir, &list);
+
     fx_ver_t most_compatible = specified;
     if (!specified.is_prerelease())
     {
@@ -350,10 +349,10 @@ void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
         {
             trace::verbose(_X("'Roll forward on no candidate fx' enabled. Looking for the least production greater than or equal to [%s]"), fx_ver.c_str());
             fx_ver_t next_lowest(-1, -1, -1);
-            for (const auto& version : list)
+            for (const auto& ver : version_list)
             {
-                fx_ver_t ver(-1, -1, -1);
-                if (fx_ver_t::parse(version, &ver, true) && (ver >= specified))
+                
+                if (!ver.is_prerelease() && ver >= specified)
                 {
                     next_lowest = (next_lowest == fx_ver_t(-1, -1, -1)) ? ver : std::min(next_lowest, ver);
                 }
@@ -371,12 +370,11 @@ void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
         if (patch_roll_fwd)
         {
             trace::verbose(_X("Applying patch roll forward from [%s]"), most_compatible.as_str().c_str());
-            for (const auto& version : list)
+            for (const auto& ver : version_list)
             {
-                trace::verbose(_X("Inspecting version... [%s]"), version.c_str());
-                fx_ver_t ver(-1, -1, -1);
+                trace::verbose(_X("Inspecting version... [%s]"), ver.as_str().c_str());
 
-                if (fx_ver_t::parse(version, &ver, true) && // true -- only prod. prevents roll forward to prerelease.
+                if (!ver.is_prerelease() && //  only prod. prevents roll forward to prerelease.
                     ver.get_major() == most_compatible.get_major() &&
                     ver.get_minor() == most_compatible.get_minor())
                 {
@@ -388,13 +386,12 @@ void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
     }
     else
     {
-        for (const auto& version : list)
+        for (const auto& ver : version_list)
         {
-            trace::verbose(_X("Inspecting version... [%s]"), version.c_str());
-            fx_ver_t ver(-1, -1, -1);
+            trace::verbose(_X("Inspecting version... [%s]"), ver.as_str().c_str());
 
-            if (fx_ver_t::parse(version, &ver, false) && // false -- implies both production and prerelease.
-                ver.is_prerelease() && // prevent roll forward to production.
+            //both production and prerelease.
+            if (ver.is_prerelease() && // prevent roll forward to production.
                 ver.get_major() == specified.get_major() &&
                 ver.get_minor() == specified.get_minor() &&
                 ver.get_patch() == specified.get_patch() &&
@@ -405,8 +402,8 @@ void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
             }
         }
     }
-    pal::string_t most_compatible_str = most_compatible.as_str();
-    append_path(&fx_dir, most_compatible_str.c_str());
+
+    return most_compatible;
 }
 
 pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode,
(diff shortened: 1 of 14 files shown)


ramarag:
This fixes https://github.com/dotnet/core-setup/issues/2398

@gkhanna79 @eerhardt @nguerrera @livarcocc PTAL

ramarag:
will make the necessary changes to doc once i get a signify on the code changes

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -144,125 +155,164 @@ public void SharedFxLookup_Must_Verify_Folders_in_the_Correct_Order()
...
+                .HaveStdErrContaining(Path.Combine(_exeSelectedMessage, "9999.0.3"));
+
+            DeleteAvailableSharedFxVersions(_exeSharedFxBaseDir, "9999.0.0", "9999.0.0-dummy0","9999.0.3");
+            DeleteAvailableSharedFxVersions(_globalSharedFxBaseDir, "9999.0.2", "9999.0.0-dummy1", "9999.0.0-dummy2");
         }
 
         [Fact]
         public void SharedFxLookup_Must_Not_Roll_Forward_If_Framework_Version_Is_Specified_Through_Argument()
         {
+            if (RuntimeEnvironment.OperatingSystemPlatform == Platform.Windows)
```
eerhardt: What is the plan here?  I don't think turning all these tests off on Windows is a great plan.

src/corehost/common/pal.unix.cpp
```diff
@@ -204,12 +204,18 @@ bool locate_dotnet_on_path(pal::string_t* dotnet_exe)
         append_path(&tok, _X("dotnet"));
         if (pal::realpath(&tok) && is_executable(tok))
         {
-            *dotnet_exe = tok;
+            dotnet_exes->push_back(tok);
```
eerhardt: When are you ever returning more than one here?  You are returning 'true' as soon as you push something on the list.

Do we really need to handle more than one?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -333,27 +333,26 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
     return false;
 }
 
-void fx_muxer_t::resolve_roll_forward(pal::string_t& fx_dir,
+fx_ver_t fx_muxer_t::resolve_most_compatible(const std::vector<fx_ver_t>& version_list,
```
gkhanna79: We should cotinue to call this resolve_roll_froward since that is what this function is doing.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -488,26 +482,62 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode,
...
 
+        pal::string_t resolved_ver_str = resolved_ver.as_str();
+        append_path(&fx_dir, resolved_ver_str.c_str());
+
         if (pal::directory_exists(fx_dir))
         {
             trace::verbose(_X("Chose FX version [%s]"), fx_dir.c_str());
-            return fx_dir;
+
+            if (!do_roll_forward)
```
gkhanna79: This check looks incorrect since for a given dir in the hiv_dir you enumerated above, resolve_most_compatible has already performed patch roll forward if it was applicable.

gkhanna79: Are you not trying to say "if (!multi-level-lookup-enabled)"?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -488,26 +482,62 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode,
...
+            {
+                return fx_dir;
+            }
+
+            //Continue to search for a better match if available
+            std::vector<fx_ver_t> version_list;
```
gkhanna79: I think this version_list should be at a higher scope where you add each "most compatibly resolved" version from each hive enumerated and will not need to add selected_ver.

Once the hive enumeration is complete, you invoke resolve_most_compatible for this list and will get the most applicable version that is patch roll forwarded if applicable.

gkhanna79: With the above said, I think this will work fine as well. Please add few comments here.

src/corehost/common/pal.h
```diff
@@ -211,7 +211,9 @@ namespace pal
     bool getenv(const char_t* name, string_t* recv);
     bool get_default_servicing_directory(string_t* recv);
     bool get_local_dotnet_dir(string_t* recv);
-    bool get_global_dotnet_dir(string_t* recv);
+    //As long as linux uses the path to find the global location of
```
gkhanna79: I would suggest we rephrase this to say that since, on Linux, we determine global location by enuemrating the location where dotnet is present on path, there could be multiple such locations. We should also say that on Windows, it will only be one.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
gkhanna79: Agreed. @ramarag why are we disabling this test?

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -366,6 +414,11 @@ public void Roll_Forward_On_No_Candidate_Fx_Fails_If_No_Higher_Version_Is_Availa
         [Fact]
         public void Roll_Forward_On_No_Candidate_Fx_Must_Look_Into_All_Lookup_Folders()
         {
+            if (RuntimeEnvironment.OperatingSystemPlatform == Platform.Windows)
```
gkhanna79: Same as above - this should endup doing the right thing by design. Is that not the case per your changes?

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -416,15 +469,21 @@ public void Roll_Forward_On_No_Candidate_Fx_Must_Look_Into_All_Lookup_Folders()
...
             // Remove dummy folders from user dir
-            DeleteAvailableSharedFxVersions(_userSharedFxBaseDir, "9999.1.1");
+            DeleteAvailableSharedFxVersions(_exeSharedFxBaseDir, "9999.1.1");
+            DeleteAvailableSharedFxVersions(_globalSharedFxBaseDir, "9999.1.0");
         }
 
         [Fact]
         public void Roll_Forward_On_No_Candidate_Fx_May_Be_Enabled_Through_Runtimeconfig_Json_File()
         {
+            if (RuntimeEnvironment.OperatingSystemPlatform == Platform.Windows)
```
gkhanna79: Same as above.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -461,12 +520,17 @@ public void Roll_Forward_On_No_Candidate_Fx_May_Be_Enabled_Through_Runtimeconfig
...
                 .Pass()
                 .And
-                .HaveStdErrContaining(_exeSelectedMessage);
+                .HaveStdErrContaining(_globalSelectedMessage);
         }
 
         [Fact]
         public void Roll_Forward_On_No_Candidate_Fx_May_Be_Enabled_Through_Argument()
         {
+            if (RuntimeEnvironment.OperatingSystemPlatform == Platform.Windows)
```
gkhanna79: As above.

src/corehost/common/pal.unix.cpp
ramarag: great catch ! will fix and add a test for it

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
ramarag: With the removal of userprofile the only other probe location is the global location on windows, and it is not easy to mock it unless we add knobs into the host itself.

I believe once the tracking item on containarizing the tests https://github.com/dotnet/core-setup/issues/1553 is done, then we can make the infra changes necessary for enabling these tests on windows

src/corehost/cli/fxr/fx_muxer.cpp
ramarag: no this is for roll_forward not enabled, will restructure the code for better readability

ramarag:
@gkhanna79 @eerhardt Any other concerns ?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -492,22 +486,60 @@ pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode,
...
+            }
+
+            fx_ver_t resolved_ver = resolve_framework_version(version_list, fx_ver, specified, patch_roll_fwd_enabled, roll_fwd_on_no_candidate_fx_val);
+
+            pal::string_t resolved_ver_str = resolved_ver.as_str();
+            append_path(&fx_dir, resolved_ver_str.c_str());
+
+            if (pal::directory_exists(fx_dir))
+            {
+                trace::verbose(_X("Chose FX version [%s]"), fx_dir.c_str());
```
gkhanna79: We should display this message only for the final version that is selected. Here, however, you can add logging to display the two versions being used to determine the resolved_ver.

src/corehost/common/pal.h
```diff
@@ -211,7 +211,9 @@ namespace pal
     bool getenv(const char_t* name, string_t* recv);
     bool get_default_servicing_directory(string_t* recv);
     bool get_local_dotnet_dir(string_t* recv);
-    bool get_global_dotnet_dir(string_t* recv);
+    //On Linux, we determine global location by enuemrating the location where dotnet is present on path, hence there could be multiple such locations
```
eerhardt: nit: type-o `enuemrating`

PR #2493, opened 2017-05-23
Fix Debian package build

rakeshsinghranchi:
- {shlibs:Depends} will pick up the correct packages for anything we link
  against so we don't need to include libssl and libcurl in the json config
  files.
- Add an explicit dependency on ICU (the runtime no longer links against it
  directly so dh_shlibdeps wasn't picking it up).

### DIFF
--- a/src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
+++ b/src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
@@ -30,8 +30,7 @@
 
     "debian_dependencies":{
         "%HOSTFXR_DEBIAN_PACKAGE_NAME%" : {},
-        "libssl1.0.0" : {},
-        "libcurl3" : {}
+        "%LIBICU_PACKAGE_NAME%": {} 
     },
 
     "debian_ignored_dependencies" : [
--- a/src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
+++ b/src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
@@ -30,6 +30,7 @@
 
     "debian_dependencies":{
         "%HOSTFXR_DEBIAN_PACKAGE_NAME%" : {},
+        "libssl1.0.0" : {},
         "%LIBICU_PACKAGE_NAME%": {} 
     },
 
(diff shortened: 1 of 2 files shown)


src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
```diff
@@ -30,8 +30,7 @@
 
     "debian_dependencies":{
         "%HOSTFXR_DEBIAN_PACKAGE_NAME%" : {},
-        "libssl1.0.0" : {},
```
ellismg: Can you confirm that the resulting package actually has a dependency on these two libraries (I'm worried that because of how we do the portable build, it will not).  If it doesn't can you please keep them in explicitly?

rakeshsinghranchi: Confirmed, the resulting package does not exhibit dependency to libssl1.0.0 . 

Depends: libc6 (>= 2.14), libcurl3 (>= 7.16.2), libgcc1 (>= 1:4.1.1), libgssapi-krb5-2 (>= 1.10+dfsg~), liblttng-ust0, libstdc++6 (>= 4.8), libunwind8, libuuid1 (>= 2.16), zlib1g (>= 1:1.1.4), dotnet-hostfxr-2.1.0-preview1-25323-9, libicu52

ellismg:
> @ellismg , can you review ? Added back libssl1.0.0 after verifying the resulting package.

I see that the libcurl3 dependency is removed.  Could you share the output of `dkpg -I` on the shared framework package that's generated with this PR, I just want to make sure the deps thing is doing the right thing.

rakeshsinghranchi:
Depends: libc6 (>= 2.14), libcurl3 (>= 7.16.2), libgcc1 (>= 1:4.1.1), libgssapi-krb5-2 (>= 1.10+dfsg~), liblttng-ust0, libstdc++6 (>= 4.8), libunwind8, libuuid1 (>= 2.16), zlib1g (>= 1:1.1.4), dotnet-hostfxr-2.1.0-preview1-25323-9, libicu52

ellismg:
> Depends: libc6 (>= 2.14), libcurl3 (>= 7.16.2), libgcc1 (>= 1:4.1.1), libgssapi-krb5-2 (>= 1.10+dfsg~), liblttng-ust0, libstdc++6 (>= 4.8), libunwind8, libuuid1 (>= 2.16), zlib1g (>= 1:1.1.4), dotnet-hostfxr-2.1.0-preview1-25323-9, libicu52

Awesome.  LGTM.

PR #2494, opened 2017-05-23
Ensure Duplicate items aren't present during publish

karajas:
skip ci please

@chcosta @eerhardt @rakeshsinghranchi @weshaggard @vivmishra 

This fix should fix the build break in master. The PortableBuild=true was determining the TargetRid (linux-x64) which decided the InstallerExtension and CombinedInstallerExtension (both became ''). Since these aren't set for a portable build this causes the wildcard to select everything. This is now conditioned and we also added a check for duplicates before we even try to publish.

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -30,7 +30,21 @@
              BuildInParallel="true" />
   </Target>
 
+  <Target Name="EnsureUniqueItemsToUpload">
+    <ItemGroup>
+      <DedupItems Include="@(ItemsToPublish->Distinct())" />
+    </ItemGroup>
+    <PropertyGroup>
+      <CountOld>@(ItemsToPublish->Count())</CountOld>
+      <CountNew>@(DedupItems->Count())</CountNew>
+      <_HasDups Condition="'$(CountOld)' != '$(CountNew)'">true</_HasDups>
+    </PropertyGroup>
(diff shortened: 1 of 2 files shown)


publish/publish.proj
```diff
@@ -30,7 +30,21 @@
...
+  <Target Name="EnsureUniqueItemsToUpload">
+    <ItemGroup>
+      <DedupItems Include="@(ItemsToPublish->Distinct())" />
+    </ItemGroup>
+    <PropertyGroup>
+      <CountOld>@(ItemsToPublish->Count())</CountOld>
+      <CountNew>@(DedupItems->Count())</CountNew>
+      <_HasDups Condition="'$(CountOld)' != '$(CountNew)'">true</_HasDups>
+    </PropertyGroup>
+    <Message Importance="High" Condition="'$(_HasDups)' != ''" Text="ItemsToPublish: '@(ItemsToPublish)'"/>
```
chcosta: nit:  merge this message and the error condition into one.

karajas: I wanted them separate to avoid having a giant error message but noted. I can add this to the refactor if we need to do that.

chcosta: I guess, if we're just displaying all of the items, and not just the duplicates, then it could be an obnoxious amount of red.  It's probably fine as-is.

PR #2495, opened 2017-05-23
Remove centos from PublishRids

karajas:
skip ci please

This is to remove centos from the publish rids which are used to evaluate the build legs. I am not removing any of the TargetsCentos condition based things as they also apply to RHEL.

/cc @vivmishra @weshaggard @rakeshsinghranchi @chcosta

### DIFF
--- a/publish/dir.props
+++ b/publish/dir.props
@@ -49,6 +49,5 @@
     <PublishRid Include="win-arm64" />
     <PublishRid Include="linux-arm" />
     <PublishRid Include="rhel.7-x64" />
-    <PublishRid Include="centos.7-x64" />
   </ItemGroup>
 </Project>


publish/dir.props
```diff
@@ -49,6 +49,5 @@
     <PublishRid Include="win-arm64" />
     <PublishRid Include="linux-arm" />
     <PublishRid Include="rhel.7-x64" />
```
weshaggard: @gkhanna79 @eerhardt @ericstj  Should we be removing all the non-portable RID's from this list?

eerhardt: We just added these when we made the RPM packages.

/cc @vivmishra 

Honestly, I think this whole concept can go away now that we have the "publish/finalize" leg in a separate build step.  All this is used for is to make sure that all the other legs finished.  PipeBuild already does that for us.

/cc @chcosta

karajas: I think the reason we have non portable rids is to express the portable based installers for those platforms. So it does sort of track what we expect the output to be.

gkhanna79: Are we porting this to 2.0.0?

vivmishra: Yes, the 2.0 PR was updated. It is still awaiting approval.

PR #2500, opened 2017-05-24
fix readme links

rakeshsinghranchi:
Fix the download links for master per the new package naming convention. 
skip ci please


rakeshsinghranchi:
cc @karajas

Issue #2216, opened 2017-04-27
Update Branding in Mac and Windows Runtime installers - Re-implement with new build system (master)


rakeshsinghranchi:
Issue :  https://github.com/dotnet/core-setup/issues/2155
Fix is in release/2.0.0 branch, similar fix is needed for master branch

PR #2504, opened 2017-05-24
Fixes to comply with RPM packaging guidelines - related #2216

vivmishra:
This PR brings in (first set of ) fixes for RPM packaging that were suggested for following packaging guidelines (https://fedoraproject.org/wiki/Packaging:Guidelines)

It is a follow-up to PR #2462


gkhanna79:
If you expect this packaging to happen in the CI, then please rebase to the tip and fix the CI as well since now, in master, it only targets portable builds.

PR #2514, opened 2017-05-24
Remove unecessary generateversionsourcefile and init-tools steps from offical build definitions

chcosta:
Fixes https://github.com/dotnet/core-setup/issues/2265

Validated that these steps are not necessary with private official build runs of individual legs

### DIFF
--- a/buildpipeline/Core-Setup-Linux-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Linux-Arm-BT.json
@@ -114,42 +114,6 @@
       "enabled": true,
       "continueOnError": false,
       "alwaysRun": false,
-      "displayName": "Init tools",
-      "timeoutInMinutes": 0,
-      "task": {
-        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
-        "versionSpec": "1.*",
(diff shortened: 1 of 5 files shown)


buildpipeline/Core-Setup-Linux-Arm-BT.json
```diff
@@ -242,6 +206,16 @@
     {
       "enabled": false,
```
eerhardt: What's this addition for?  Is it necessary?  Can it be removed?

chcosta: This is just what the VSTS download API returned, it could probably be removed, but it's getting added back whenever the definition is loaded into VSTS.  It would be nice if we could trim these down to the minimum data necessary, but for now we tend to just preserve whatever downloading the VSTS definition gives us.

eerhardt: Sounds good.

In general, it creates noise in reviews like this.  PR for "remove something" is adding new stuff to the file.  Seems odd.  But I'm not sure how to make it better.  Maybe separating out the "noise" from the real "changes" into separate commits?

PR #2515, opened 2017-05-24
Update NETCore.App dependencies for nca1.0 and nca.1.1

### LINKED ISSUE #2512: Microsoft.NETCore.App 2.0 doesn't list netcoreapp1.1 dependencies
## Steps to reproduce
Look at the dependencies listed in the NETCore.App 2.0 package:

https://www.nuget.org/packages/Microsoft.NETCore.App/2.0.0-preview1-002111-00
https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App/2.0.0-preview2-25324-02

## Expected  behavior
It should list the dependencies for netcoreapp1.0, netcoreapp1.1, and netcoreapp2.0

## Actual behavior
It only lists the dependencies for netcoreapp1.0 and netcoreapp2.0

@weshaggard @ericstj @gkhanna79

ericstj:
Dependencies were missing for netcoreapp1.1 due to missing import.  Fixes #2512

Update dependencies to latest stable versions to give folks serviced packages.  Related https://github.com/dotnet/standard/issues/354.

/cc @eerhardt @weshaggard @gkhanna79

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
+++ b/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
@@ -13,16 +13,22 @@
       <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
-      <Version>1.0.1</Version>
+      <Version>1.0.5</Version>
+    </NETCoreApp10Dependency>
+    <NETCoreApp10Dependency Include="Microsoft.NETCore.Platforms">
+      <Version>1.0.2</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.Runtime.CoreCLR">
-      <Version>1.0.4</Version>
+      <Version>1.0.7</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.VisualBasic">
       <Version>10.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="NETStandard.Library">
       <Version>1.6.0</Version>
+    </NETCoreApp10Dependency>
+    <NETCoreApp10Dependency Include="runtime.native.System.Security.Cryptography">
+      <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="System.Buffers">
       <Version>4.0.0</Version>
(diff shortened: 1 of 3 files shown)


src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
```diff
@@ -28,6 +28,12 @@
     <NETCoreApp11Dependency Include="NETStandard.Library">
```
weshaggard: Why didn't you update the runtime and host packages for 1.1? They should be up to 1.1.2

ericstj: Thanks, will do.

weshaggard: I would suggest diffing the nuspect for the 1.1.2 package with your local one to make sure the deps are updated.

ericstj: Sure.  I also had to lift some dependencies that were seemingly missed from 1.0/1.1 core.app packages.

src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
```diff
@@ -13,10 +13,13 @@
...
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
-      <Version>1.0.1</Version>
+      <Version>1.0.5</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.Runtime.CoreCLR">
-      <Version>1.0.4</Version>
+      <Version>1.0.7</Version>
+    </NETCoreApp10Dependency>
+    <NETCoreApp10Dependency Include="Microsoft.NETCore.Targets">
```
eerhardt: I don't see this dependency on [https://www.nuget.org/packages/Microsoft.NETCore.App/1.0.5](https://www.nuget.org/packages/Microsoft.NETCore.App/1.0.5)
But I DO see a dependency for Microsoft.NETCore.Platforms (>= 1.0.2), which I don't see here.

eerhardt: Also, 1.0.5 has a dependency on runtime.native.System.Security.Cryptography (>= 4.0.1) now.

---
In reply to: [118324972](https://github.com/dotnet/core-setup/pull/2515#discussion_r118324972) [](ancestors = 118324972)

ericstj: I'm doing a full diff to find gaps.

ericstj:
Ok, now the diff is clean between 1.1.2 package and this.  I'm doing some assets file inspection to make sure everything looks good there as well.

ericstj:
Assets file diff looks good.  One thing that's concerning that I mentioned above is that there are packages which have been serviced but those serviced versions aren't being consumed by the shared framework:
- NETCoreApp1.0
  - System.Diagnostics.StackTrace
    - Included: 4.0.1
    - Latest serviced: 4.0.2
  - System.IO.Compression
    - Included: 4.1.0
    - Latest serviced: 4.1.1

I'm not going to fix those here because that'd create a mismatch and make those packages go app-local.  I'll file an issue for servicing.

PR #2115, opened 2017-04-19
Coreclr UWP Packaging

rahku:
1. Build native UWP components
2. restore coreclr nuget package
3. Create appx package

PR #2520, opened 2017-05-24
Build CoreRuntime AppX Package for UWP

gkhanna79:
This is a rebase of original PR from @rahku at https://github.com/dotnet/core-setup/pull/2115 and fixing it to account for the new build system and does the following to support UWP scenarios:

1. Build native UWP components, sign them and publish to Azure blob.
2. Generate the Microsoft.NET.CoreRuntime.2.0 FW AppX package.

Once this is merged, @zamont will author the nuget package to carry required components for VS integration and also work with @ravimeda to consume the AppX package for CoreCLR AppCompat runs.

@zamont @weshaggard @ericstj PTAL


publish/dir.props
```diff
@@ -25,6 +25,9 @@
     <RuntimePackageFile Include="$(PackagesOutDir)**/runtime.*.nupkg" >
       <RelativeBlobPath>$(BinariesRelativePath)</RelativeBlobPath>
     </RuntimePackageFile>
+    <UWPArtifactsToUpload Include="$(UWPOutputDir)Microsoft.NET.CoreRuntime.2.0*.appx" >
```
weshaggard: Why the * in the path?

src/uwp/dir.props
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
```
weshaggard: Why do we need this? Why not just import ..\dir.props directly in build.proj?

src/uwp/build.cmd
```diff
@@ -0,0 +1,110 @@
...
+    ) 
+    goto :MissingVersion
+) 
+if "%VisualStudioVersion%"=="14.0" (
+    goto :VS2015
+) 
+
+:MissingVersion
+:: Can't find VS 2013+
+echo Error: Visual Studio 2015 required
```
weshaggard: Do we need VS2017 support?

publish/dir.props
zamont: Is the * for preview versions?

gkhanna79: No, it is a wildcard to pull in whatever version was built and named there.

zamont:
For the packaging and building of the appx package, LGTM. I'm not familiar with building the UWPHost or the cmake aspects of the changes so that is up to other reviewers.

src/uwp/dir.props
gkhanna79: just following the pattern.

src/uwp/build.cmd
gkhanna79: We do not support building across the stack with VS 2017 (it changed a lot of dependencies). We can look into that once we do intend to support building using it in (atleast in official builds).

PR #2524, opened 2017-05-24
Combine packaging props and publish props for deb packages

karajas:
The file rename caused issues with the debian package upload so refactoring the properties for reuse instead. PTAL

@rakeshsinghranchi @eerhardt @chcosta

### DIFF
--- a/publish/dir.targets
+++ b/publish/dir.targets
@@ -70,14 +70,14 @@
            Text=" Missing parameters for debian publish." />
            
     <ItemGroup>
-      <DebInstallerFile Include="$(SharedHostInstallerStart)$(ProductMoniker).deb" >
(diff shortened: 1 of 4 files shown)


publish/dir.targets
```diff
@@ -70,14 +70,14 @@
            Text=" Missing parameters for debian publish." />
            
     <ItemGroup>
-      <DebInstallerFile Include="$(SharedHostInstallerStart)$(ProductMoniker).deb" >
+      <DebInstallerFile Include="$(SharedHostInstallerFile)" >
         <PackageName>dotnet-host</PackageName>
```
karajas: Assuming this does not change.

eerhardt:
LGTM.

Issue #2533, opened 2017-05-25
Remove UserProfile lookup for Store


gkhanna79:
CC @eerhardt @morganbr @nguerrera @livarcocc @davidfowl

PR #2539, opened 2017-05-26
Remove Store user profile

ramarag:
fixes https://github.com/dotnet/core-setup/issues/2533

@eerhardt @gkhanna79 PTAL

### DIFF
--- a/Documentation/design-docs/DotNetCore-SharedPackageStore.md
+++ b/Documentation/design-docs/DotNetCore-SharedPackageStore.md
@@ -105,15 +105,18 @@ Note that this is different from current behavior of `dotnet run` for an applica
 The host will probe in the following order of precedence for `dotnet run` and application activations post `dotnet publish`:
 
 + `$CORE_SERVICING` on Unix or `%ProgramFiles(x86)%\coreservicing` on Windows.
++  Application Bin and Shared FX directory
 + `DOTNET_SHARED_STORE` in the chained order
-+ The dotnet.exe relative shared package store
-+ The global shared package store
-+ The Shared Framework directory
-+ ~~`DOTNET_HOSTING_OPTIMIZATION_CACHE` deprecated in favor of `DOTNET_SHARED_STORE`~~
-+ `--additionalprobingpaths` specified in the command line
-+ `runtimeOptions.additionalProbingPaths` (includes NuGet cache probe specified by the CLI for `dotnet run`)
-+ Application `bin` directory
++ Store locations
+    - The dotnet.exe relative shared package store
+    - The global shared package store
+    - ~~`DOTNET_HOSTING_OPTIMIZATION_CACHE` deprecated in favor of `DOTNET_SHARED_STORE`~~
++ Additional Probing Paths
+    - `--additionalprobingpaths` specified in the command line
+    - `runtimeOptions.additionalProbingPaths` (includes NuGet cache probe specified by the CLI for `dotnet run`)
 
+**NOTE:**  `--additionalprobingpaths` can be passed template paths like below and the host will interpret `|arch|/|tfm|` appropriately to look for assets: 
+           `%USERPROFILE%\user\cache\|arch|\|tfm|` or `$HOME/user/cache/|arch|/|tfm|`
 ## dotnet publish
 
 Publish will be enhanced to support a filter profile file specified as xml. This file explicitly lists all asset packages that need to be trimmed out of the publish output. The following are examples of how various application types can be published.
--- a/Documentation/design-docs/DotNetCore-SharedPackageStore.md
+++ b/Documentation/design-docs/DotNetCore-SharedPackageStore.md
@@ -5,7 +5,7 @@ To enable sharing of assemblies among all machine-wide .NET Core applications, a
 
 ### Packages Store
 
-The package store can be either a global system-wide folder or a user local folder:
+The package store can be either a global system-wide folder or a dotnet.exe relative folder:
 
 + **Global**:
     - The `dotnet` root location -- on Windows, the folder is located in `C:\Program Files (x86)\`. See layout below.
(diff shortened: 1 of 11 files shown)


Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -117,14 +106,16 @@ The host will probe in the following order of precedence for `dotnet run` and ap
 
 + `$CORE_SERVICING` on Unix or `%ProgramFiles(x86)%\coreservicing` on Windows.
```
gkhanna79: This ordering, in general, does not match the way things work. We should fix it. The host probes the locations in the following order:

1) Servicing
2) AppBin (including SharedFX)
3) Store locations
4) Additional Probing Paths

I think we should list the actual order so that readability and understanding of the concept remains simple. For specifics of each bullet (e.g. EXE or Global, one probe location over the other), we should have sub-bullets to explain those.

Documentation/design-docs/DotNetCore-SharedPackageStore.md
```diff
@@ -117,14 +106,16 @@ The host will probe in the following order of precedence for `dotnet run` and ap
...
++ The dotnet.exe relative shared package store
 + The global shared package store
 + The Shared Framework directory
 + ~~`DOTNET_HOSTING_OPTIMIZATION_CACHE` deprecated in favor of `DOTNET_SHARED_STORE`~~
 + `--additionalprobingpaths` specified in the command line
 + `runtimeOptions.additionalProbingPaths` (includes NuGet cache probe specified by the CLI for `dotnet run`)
 + Application `bin` directory
 
+**NOTE:**  `--additionalprobingpaths` can be passed template paths like below and the host will interpret `{arch}/{tfm}` appropriately to look for assets: 
+           `%USERPROFILE%\.dotnet\store\{arch}\{tfm}` or `$HOME/.dotnet/store/{arch}/{tfm}`
```
gkhanna79: I would suggest to give an example that does not tie in "store" so that it appears as generic as possible.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -831,12 +831,34 @@ int muxer_usage(bool is_sdk_present)
...
 
 // Convert "path" to realpath (merging working dir if needed) and append to "realpaths" out param.
-void append_realpath(const pal::string_t& path, std::vector<pal::string_t>* realpaths)
+void append_probe_realpath(const pal::string_t& path, std::vector<pal::string_t>* realpaths, const pal::string_t& tfm)
 {
-    pal::string_t real = path;
-    if (pal::realpath(&real))
+    pal::string_t probe_path = path;
+
+    if (pal::realpath(&probe_path) && pal::directory_exists(probe_path))
```
gkhanna79: Can I not create a folder that as "{arch}/{tfm}" in its name such that the call to directory exists succeeds? I think you need to search for the pattern first, transform it if applicable and then use the resultant value for the folder check.

src/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
```diff
@@ -255,5 +286,18 @@ private void MoveRuntimeConfigToSubdirectory(TestProjectFixture testProjectFixtu
 
             testProjectFixture.TestProject.RuntimeConfigJson = destRuntimeConfig;
         }
+
+         private string CreateAStrore(TestProjectFixture testProjectFixture)
```
gkhanna79: Store and not "Strore".

src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -235,6 +235,65 @@ private void ValidateRequiredDirectories(RepoDirectoriesProvider repoDirectories
...
+            DotNetCli dotnet = null,
+            string runtime = null,
+            string framework = "netcoreapp2.0",
+            string manifest = null,
+            string outputDirectory = null)
+        {
+            dotnet = dotnet ?? _sdkDotnet;
+            outputDirectory = outputDirectory ?? _testProject.OutputDirectory;
+
+            var publishArgs = new List<string>();
```
eerhardt: These aren't "publish" args, they are "store" args.

src/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
```diff
@@ -125,6 +125,37 @@ public void Muxer_Exec_activation_of_Build_Output_Portable_DLL_with_DepsJson_Loc
...
+                    "exec",
+                    "--additionalprobingpath", additionalProbingPath,
+                    appDll)
+                .CaptureStdErr()
+                .CaptureStdOut()
+                .Execute()
+                .Should()
+                .Pass()
+                .And
+                .HaveStdOutContaining("Hello World");
```
eerhardt: Should we verify that the Newtonsoft.Json assembly is loaded from the `store_path`?

src/corehost/cli/fxr/fx_muxer.cpp
eerhardt: I tested on my Windows and Linux machines yesterday, and I can create a directory named `{tfm}`.

There aren't many characters you can't use for a name in Linux.  Do we think we are safe with `{}`?  Should we double wrap the curly braces? `{{tfm}}`?

Another option is the `|` pipe character.  You can't use that on Windows, and you have to escape the `|` on Linux.  So I think maybe `|tfm|` and `|arch|` might be good choices.

---
In reply to: [118711887](https://github.com/dotnet/core-setup/pull/2539#discussion_r118711887) [](ancestors = 118711887)

ramarag:
@gkhanna79 @eerhardt i have address all the concerns above

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -831,12 +831,42 @@ int muxer_usage(bool is_sdk_present)
...
+    {
+        realpaths->push_back(probe_path);
+    }
+    else
     {
-        realpaths->push_back(real);
+        //Check if we can extrapolate {arch}<DIR_SEPARATOR>{tfm} for probing stores
```
eerhardt: out of date comment - `{arch}`

src/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
```diff
@@ -125,6 +125,40 @@ public void Muxer_Exec_activation_of_Build_Output_Portable_DLL_with_DepsJson_Loc
...
+                .EnvironmentVariable("COREHOST_TRACE", "1")
+                .CaptureStdErr()
+                .CaptureStdOut()
+                .Execute()
+                .Should()
+                .Pass()
+                .And
+                .HaveStdOutContaining("Hello World")
+                .And
+                .HaveStdErrContaining($"Adding tpa entry: {store_path}");
```
eerhardt: Can we check the full path to ensure the that `|arch|` and `|tfm|` were replaced correctly?

ramarag: done

PR #2561, opened 2017-05-30
Automated signing validation infrastructure

chcosta:
Adds a parallel (to publishing) build pipeline leg which does signing validation with the BinInspect.exe tool which VS has provided.  The BinInspect tool will validate installers, and also extract artifacts from installers / nupkgs / zips to validate signing those pieces.

BinInspect.exe currently lives on a share, so I copied the binaries into azure storage (dotnetcli container).  This infrastructure will download the BinInspect binaries, then download a Core-setup build and inspect those artifacts.

BinInspect does not have a baseline mechanism for binaries which we've determined are intentionally unsigned, so I added a task which uses the itemgroup defined in baseline.props (regex items), to ignore those items by comparing the baseline to the process output (an xml file).  

Most of this infrastructure could be moved into BuildTools, and utilized by other repo's in the future (and I think we should do that). 

I added this as a parallel (to publishing) build leg, because I was concerned about the time involved in downloading the tool, downloading artifacts, exploding those artifacts, and inspecting them.  My testing has shown that the entire task takes about 18 minutes for a core-setup drop, with 5 minutes of that being build leg overhead stuff.

/cc @eerhardt @weshaggard @gkhanna79


chcosta:
This work is also being tracked by https://github.com/dotnet/core-eng/issues/691

signingvalidation.proj
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
```
eerhardt: It's kind of weird that this folder is named `signingvalidation.proj`.  Can it just be renamed to `signingvalidation`?

Or maybe moving this new file and the `sign.proj` file into a `signing` folder?

buildpipeline/pipeline.json
```diff
@@ -147,6 +147,32 @@
...
+      "Name": "Signing validation",
+      "Parameters": {
+        "TreatWarningsAsErrors": "false"
+      },
+      "BuildParameters": {
+        "BuildConfiguration": "Release"
+      },
+      "Definitions": [
+        {
+          "Name": "Core-Setup-Signing-Validation",
```
eerhardt: Does this need to be named `Core-Setup`?  That seems redundant with the repo's name.

signingvalidation.proj
chcosta: This **file** is named `signingvalidation.proj`, there's also a `tools-local\signingvalidation` **folder**.  The reason for this was because I wanted an entry point, and then a separate file which the entry point can `MSBuild` after it did its "prerequisites" stuff.  There's no reason I couldn't move `sign.proj` and `signingvalidation.proj` into the same `signing` folder as `sign.proj` and `validate.proj` though..

buildpipeline/pipeline.json
chcosta: This is consistent with the other definitions in the branch, and it makes it easy to see what repo a build references or to search for it in `DEVDIV.visualstudio.com`.

signingvalidation.proj
```diff
@@ -0,0 +1,17 @@
...
+  <Target Name="Build"
+          DependsOnTargets="BuildPrerequisites;
+                            Validate" />
+
+  <Target Name="BuildPrerequisites"
+          DependsOnTargets="CreateVersionInfoFile;
+                            BuildCustomTasks" />
+
+  <Target Name="Validate">
+    <MSBuild Projects="$(MSBuildThisFileDirectory)tools-local\signingvalidation\dir.proj" />
```
eerhardt: Is there value in making this a separate project vs just making it a .targets file that gets imported?

Also, I really don't like this pattern of naming files `dir.proj`.  It makes it too confusing.  There's a bunch of "dir.proj" files in the repo, and when I have a few of them open in a text editor, it is impossible to tell which one is which.  It also makes it impossible to reference the file without giving the full path - which dir.proj file are you referring to?

chcosta:
@weshaggard suggested that BinInspect should be a nupkg instead of a tool stored in Azure so that it can follow the same restore mechanisms that other tools use. I'm following up with the tool owner to determine if there are any concerns with this.

signingvalidation.proj
chcosta: I want the signing validation to be a self-contained thing, so that in our build definition, we can just build one project and it does all the things necessary for it to run, including building its prerequisites (local MSBuild tasks, creating the buildversion.props file).  This is preferable to knowing that in order to do signing validation, I need to build the local tasks, create the version info file, and then build the validation project. My testing showed that I couldn't get MSBuild to re-evaluate its properties if I called MSBuild on the same project file.  So, I was forced to create a separate project file which I can MSBuild.

tools-local/signingvalidation/dir.proj
```diff
@@ -0,0 +1,115 @@
...
+    <PropertyGroup>
+      <BlobName>%(_BlobList.Identity)</BlobName>
+      <BlobExtension>%(_BlobList.Extension)</BlobExtension>
+      <BlobFilename>%(_BlobList.Filename)</BlobFilename>
+    </PropertyGroup>
+    <ItemGroup>
+      <DownloadArtifact Include="$(BlobName)" Condition="'$(BlobExtension)' == '.exe'" />
+      <DownloadArtifact Include="$(BlobName)" Condition="'$(BlobExtension)' == '.msi'" />
+      <DownloadArtifact Include="$(BlobName)" Condition="'$(BlobExtension)' == '.zip'" />
+      <DownloadArtifact Include="$(BlobName)" Condition="'$(BlobExtension)' == '.nupkg' and $(BlobFilename.StartsWith('runtime.win'))" />
```
eerhardt: What about the PlatformAbstractions and DependencyModel .nupkgs?

tools-local/tasks/ValidateBinInspectResults.cs
```diff
@@ -0,0 +1,86 @@
...
+        public ITaskItem [] BaselineFiles { get; set; }
+
+        [Output]
+        public ITaskItem[] ErrorResults { get; set; }
+
+        public override bool Execute()
+        {
+            XDocument xdoc = XDocument.Load(ResultsXml);
+
+            var rows = from data in xdoc.Descendants(s_RootElement)
```
eerhardt: Can't this be re-written as

```
var rows = xdoc.Descendants(s_RootElement).Descendants(s_RowElement);
```

?

chcosta:
This infrastructure is now using the internal feed to restore BinInspect via similar mechanisms to how IBCMerge is used.  I've removed the separate download from Azure storage and cleaned up the container I had added.

Before this merges, I'll need to ensure we are passing the OptionalTool variables from Pipebuild so that BinInspect package is restored.

signingvalidation.proj
weshaggard: Don't you want to move this under the signing directory now?

weshaggard: looks like you did but still have one in the root.

chcosta:
Thanks for the reviews.  I'll merge this tomorrow and monitor builds if there are no further concerns.

/cc @eerhardt @weshaggard

chcosta:
@dotnet-bot test Linux arm Release Build

Issue #2516, opened 2017-05-24
Update Runtime Installer [Windows]


leecow:
Strings and installer steps to be updated. Mac and Windows will receive similar updates and there will be separate issue covering Mac changes. Matching changes will be detailed in another issue for the SDK installer.

## Screen 1 (intro)

![runtime-1-new](https://cloud.githubusercontent.com/assets/2212879/26418024/9fcaa1c4-406f-11e7-966f-ddf809f7640f.png)

* [ ] Window title: `[productname] - [version] [arc] Installer`
* [ ] Remove "You just need a shell ..." text
* [ ] Update Introduction step content as detailed below.

### Intro step text elements
**.NET Core Runtime**

.NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform, and supported by Microsoft. We hope you enjoy it! 

**Learn more about .NET Core**
* [Core Documentation](https://aka.ms/dotnet-docs)
* [SDK Documentation](https://aka.ms/dotnet-cli-docs)
* [Privacy Statement](https://aka.ms/dev-privacy)
* [MIT License](https://aka.ms/dotnet-license)

## Screen 2 (progress)
* [ ] Window title: `[productname] - [version] [arc] Installer`

## Screen 3 (success)

![runtime-3-new](https://cloud.githubusercontent.com/assets/2212879/26418100/dd9bee72-406f-11e7-8975-a4a6e32d07c2.png)

* [ ] Window title: `[productname] - [version] [arc] Installer`
* [ ] Include installation location and list of all components installed
* [ ] Update Summary step content as detailed below.

### Summary step text elements

The following were installed at `C:\Program Files\dotnet`
* .NET Core Runtime 2.0.0
* ASP.NET Runtime Package Store

**Resources**
* [Core Documentation](https://aka.ms/dotnet-docs)
* [SDK Documentation](https://aka.ms/dotnet-cli-docs)
* [Release Notes](https://aka.ms/20-p2-rel-notes)
* [Tutorials](https://aka.ms/dotnet-tutorials)
* [.NET Core Telemetry](https://aka.ms/dotnet-cli-telemetry)

/cc @richlander @gkhanna79 @livarcocc

rakeshsinghranchi:
@leecow  , @gkhanna79 @Petermarcu @vivmishra 
1.  Windows title for Preview builds should have "Preview" branding too , can you please confirm?
[...]

PR #2574, opened 2017-06-01
Update Runtime Installer [Windows]

rakeshsinghranchi:
Fixes Issue : https://github.com/dotnet/core-setup/issues/2516 

Updated package branding
Added Welcome and Success page messages
Added dummyEula as required by Burn authoring . It doesn't show up in UI.


src/pkg/packaging/windows/sharedframework/bundle.wxs
```diff
@@ -27,7 +27,7 @@
 
     <swid:Tag Regid="microsoft.com" InstallPath="[DOTNETHOME]" />
 
-    <Variable Name="DOTNETHOME" Type="string" Value="[ProgramFiles6432Folder]dotnet" bal:Overridable="no" />
+    <Variable Name="DOTNETHOME" Type="string" Value="[$(var.Program_Files)]dotnet" bal:Overridable="no" />
```
gkhanna79: Why are we changing this to install in "Program Files" instead of the current "Program Files (x86)"?

rakeshsinghranchi: There is no change in the Install location for Bundles across Windows architecture.

As of today, X86 bundle on x64 Windows has DOTNETHOME value set to Program Files , However, during MSI installation, Windows Installer sets the right value to Program Files (x86) . 

MSI (s) (24:D8) [10:02:27:751]: WIN64DUALFOLDERS: 'C:\Program Files (x86)\' will substitute 17 characters in 'C:\Program Files\dotnet\' folder path. (mask argument = 0, the folder pair's iSwapAttrib member = 0).
MSI (s) (24:D8) [10:02:27:751]: PROPERTY CHANGE: Modifying DOTNETHOME property. Its current value is 'C:\Program Files\dotnet'. Its new value: 'C:\Program Files (x86)\dotnet\'.

We are ensuring that DOTNETHOME variable has correct value set to reflect the correct install location. This value is now used to show the install location of a bundle in Installer UI. 

Program_Files is a variable that is set correctly depending on the Package arch https://github.com/dotnet/core-setup/blob/master/src/pkg/packaging/windows/sharedframework/variables.wxi#L20

On Windows x64 :  
  - x86 runtime installer gets installed on Program Files (x86) 
  - x64 runtime installer gets installed on Program Files

On Windows x86 :
- x86 runtime installer gets installed on Program Files

rakeshsinghranchi: @leecow  has blessed the changes offline.

PR #2587, opened 2017-06-01
Trim leading zeros from AppX package version

gkhanna79:
AppX package version does not like leading zero that come from out build infrastructure. Trim them out to get the expected version.

Was able to repro locally and validate the fix.

@weshaggard PTAL.

### DIFF
--- a/src/pkg/projects/Microsoft.NET.CoreRuntime/Microsoft.NET.CoreRuntime.depproj
+++ b/src/pkg/projects/Microsoft.NET.CoreRuntime/Microsoft.NET.CoreRuntime.depproj
@@ -14,6 +14,9 @@
     <AppxIdentityName>Microsoft.NET.CoreRuntime.$(MajorVersion).$(MinorVersion)</AppxIdentityName>
     <AppxDisplayName>Microsoft.NET.CoreRuntime.$(MajorVersion).$(MinorVersion)</AppxDisplayName>
     
+    <BuildVersionAppX>$([System.Int32]::Parse($(BuildNumberMajor)))</BuildVersionAppX>
+    <RevisionVersionAppX>$([System.Int32]::Parse($(BuildNumberMinor)))</RevisionVersionAppX>
+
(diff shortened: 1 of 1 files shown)


gkhanna79:
@weshaggard Are you good with this? I would like to merge this to unblock official build of master.

weshaggard:
It seems like a hack but I don't fully understand the issue to provide an alternative. I wonder why it isn't an issue for our Assembly versions https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L24. At a minimum I would suggest pulling the version into a property to make it easier to read as that line is getting a little crazy.

gkhanna79:
It does not matter for Assembly versions since they are not processed by MakeAppX tool that follows a schema for version in AppXManifest.

gkhanna79:
@weshaggard Converted to properties.

src/pkg/projects/Microsoft.NET.CoreRuntime/Microsoft.NET.CoreRuntime.depproj
```diff
@@ -14,6 +14,9 @@
     <AppxIdentityName>Microsoft.NET.CoreRuntime.$(MajorVersion).$(MinorVersion)</AppxIdentityName>
     <AppxDisplayName>Microsoft.NET.CoreRuntime.$(MajorVersion).$(MinorVersion)</AppxDisplayName>
     
+    <BuildVersionAppX>$([System.Int32]::Parse($(BuildNumberMajor)))</BuildVersionAppX>
+    <RevisionVersionAppX>$([System.Int32]::Parse($(BuildNumberMinor)))</RevisionVersionAppX>
```
weshaggard: I think I would also move AppxVersion to be a property defined as `$(MajorVersion).$(MinorVersion).$(BuildVersionAppX).$(RevisionVersionAppX))` to keep the Replace statement a simple single property map.

PR #2608, opened 2017-06-03
Update BuildTools

ellismg:


ellismg:
test Windows_NT x64 Release Build please
test Windows_NT x64 Debug Build please

NuGet issues during restore.

ellismg:
test Windows_NT x64 Release Build please
test Windows_NT x64 Debug Build please

NuGet issues during restore.

ellismg:
test Windows_NT x86 Debug Build please

chcosta:
I'm completely fine pulling in your change to my (just submitted) PR, or resolving the conflict if you want to merge your PR.  https://github.com/dotnet/core-setup/pull/2625

weshaggard:
For my information why are we pulling in _.pdb locally as opposed to from the buildtools package?

chcosta:
There's no "_.pdb" in the buildtools package.  There's a "_._".  If "_.pdb" gets added (added back?) to buildtools, then I think the preference would be to use that.

ellismg:
What @chcosta said is correct.  If we want it to get added to buildtools, I'm happy to do that.

/cc @dagood.

weshaggard:
I think it would be nice for these placeholders to exist there but not necessary. From the change it looked like it was already there which is why I was asking the question.

Issue #2570, opened 2017-05-31
Improvements to error message about unfound SDK version


natemcmaster:
**Steps to reproduce**
Add a global.json file
```json
{
  "sdk": {
    "version": "99.99.99"
  }
}
```
`dotnet --version`

**Output**
```
The specified SDK version [] from global.json [/Users/namc/dev/aspnet/KestrelHttpServer/global.json] not found; install specified SDK version
Did you mean to run dotnet SDK commands? Please install dotnet SDK from: 
  http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
```

**Suggestions**
- include the version parsed from the global.json file in the error message
- list available sdk versions

**Environment data**
Using 2.0.0-preview2-006208

cc @johnbeisner

johnbeisner:
Closing: https://github.com/dotnet/core-setup/pull/2621

PR #2621, opened 2017-06-05
Defining Global json roll forward along patch versions behavior.

ghost:
If the version found in the Global json does not exist; roll forward along patch versions.

Addresses: 
https://github.com/dotnet/cli/issues/6688 - Global json roll forward
https://github.com/dotnet/core-setup/issues/2570 - minor BUG fix

1. If global.json does not contain a version, use latest version – consider release & pre-release
    1.1 If nothing can be found -- error out
2. If global.json does contain a version:
    2.1 If the version exists, use that
    2.2 If the version does not exist, use latest **patch** version – consider release & pre-release
    2.3 If nothing can be found -- error out

@dotnet/dotnet-cli

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -628,6 +628,50 @@ pal::string_t resolve_sdk_version(pal::string_t sdk_path, bool parse_only_produc
     return retval;
 }
 
+pal::string_t resolve_patch_sdk_version(pal::string_t global_cli_version, pal::string_t sdk_path, bool parse_only_production)
+{
+    fx_ver_t specified(-1, -1, -1);
+    if (!fx_ver_t::parse(global_cli_version, &specified, false))
+    {
+        trace::error(_X("The specified SDK version '%s' could not be parsed"), global_cli_version.c_str());
+        return pal::string_t();
+    }
+
+    trace::verbose(_X("--- Resolving SDK version from SDK dir [%s]"), sdk_path.c_str());
+
+    pal::string_t retval;
+    std::vector<pal::string_t> versions;
+
+    pal::readdir(sdk_path, &versions);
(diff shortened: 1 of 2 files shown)


MattGertz:
@livarcocc Is this for VS or the overlay?

Petermarcu:
@MattGertz, It's in the runtime host so it should be in the "overlay".

nguerrera:
@MattGertz @Petermarcu @livarcocc We also carry this code in VS with the resolver in order to make VS replicate the same behavior vs. global.json as dotnet on the command line.

Petermarcu:
@nguerrera does that code somehow get mirrored? VS doesn't distribute the host binary built from this repo does it? So there would be a different PR for that?

MattGertz:
Yeah, I wondered.  If it touches VS, then that'll need director approval.  @livarcocc Who owns this particular repo?  I'm embarrassed to say that I don't actually know that and am therefore unsure who owns the rollup to JoC.  If it's me, we'll need the right labels so that it will show up in the director queries properly (i.e. "Pending director approval", "Pending Shiproom approval" and "Approved to merge". @Petermarcu is also going to follow up with @nguerrera per the above.

MattGertz:
@Petermarcu clarified the ownership BTW -- it's both of us. :-)  I would roll it up in this case.

johnbeisner:
[RE Making dotnet SDK roll semantics smarter.txt](https://github.com/dotnet/core-setup/files/1053313/RE.Making.dotnet.SDK.roll.semantics.smarter.txt)

Petermarcu:
@MattGertz, Yup, for most cases it me but there is this very small bit of code for how to look up the SDK that is owned by the SDK team.

gkhanna79:
@Petermarcu is correct - the SDK parts of the host are owned by @livarcocc.

PR #2625, opened 2017-06-06
Remove local tasks and use buildtools tasks

chcosta:
These tasks have been moved into buildtools - https://github.com/dotnet/buildtools/commit/864caca896bae248dc1b5414bb14632d95201f35

Removing them from core-setup repo, in favor of the buildtools versions.

/cc @weshaggard @eerhardt


chcosta:
@dotnet-bot test Linux arm Release Build please

tools-local/tasks/core-setup.tasks.csproj
```diff
@@ -34,6 +34,9 @@
   </ItemGroup>
   <ItemGroup Condition="'$(TargetGroup)' != 'net45'">
     <Compile Include="netstandard/FileUtilities.netstandard.cs" />
+
+    <Reference Include="$(BuildToolsTaskCoreDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
```
eerhardt: Can you use `$(BuildToolsTaskDir)` instead?  That way you can put these two lines into a common spot.

src/pkg/projects/dir.targets
```diff
@@ -35,7 +35,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <File Include="$(ToolsDir)\_.pdb"
+      <File Include="$(MSBuildThisFileDirectory)../_.pdb"
```
eerhardt: Is this really where this file exists?  Who puts it there?

publish/publish.proj
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
-  <UsingTask TaskName="DownloadBlobsFromAzure" AssemblyFile="$(LocalBuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.Local.dll" />
+  <UsingTask TaskName="DownloadFromAzure" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
```
eerhardt: nit: `CloudTestTasks` seems like an unusual place for these tasks.  This has nothing to do with "Test".

publish/publish.proj
```diff
@@ -157,10 +157,11 @@
     </PropertyGroup>
     <MakeDir Directories="$(DownloadDirectory)"
              Condition="!Exists('$(DownloadDirectory)')" />
-    <DownloadBlobsFromAzure AccountName="$(AzureAccountName)"
+    <DownloadFromAzure AccountName="$(AzureAccountName)"
                            AccountKey="$(AzureAccessToken)"
                            ContainerName="$(ContainerName)"
                            BlobNames="@(_CoreHostPackages)"
+                           BlobNamePrefix="Runtime/$(SharedFrameworkNugetVersion)"
```
eerhardt: Why is this prefix necessary when we list the exact files we want to download in `BlobNames`?

build.proj
```diff
@@ -36,23 +36,21 @@
...
       <MSBuild Projects="tools-local/tasks/core-setup.tasks.builds" />
 
     <PropertyGroup>
       <DependencyModelTFM>netstandard1.3</DependencyModelTFM>
       <DependencyModelTFM Condition="'$(MSBuildRuntimeType)' != 'Core'">net451</DependencyModelTFM>
     </PropertyGroup>
     
-      <ItemGroup>
-        <CustomTaskDependencies Include="$(BuildToolsTaskDir)Newtonsoft.Json.dll" />
-        <CustomTaskDependencies Include="$(BuildToolsTaskDir)Microsoft.DotNet.PlatformAbstractions.dll" />
```
eerhardt: `DependencyModel` depends on this assembly.  If we need to copy DependencyModel to `$(LocalBuildToolsTaskDir)`, I would assume this is still necessary.  Unless it is getting copied there by some other mechanism.

publish/publish.proj
chcosta: In the DownloadFromAzure task, if you specify the BlobNamePrefix, it will trim that path from the directory it downloads to locally - https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.CloudTestTasks/DownloadFromAzure.cs#L99

So, by specifying the prefix we get local paths like `Bin\ForPublishing` instead of `Bin\ForPublishing\Runtime\blah`

publish/publish.proj
chcosta: I agree.  BuildTools needs to be partitioned better.  I originally thought to move this to something like `Microsoft.DotNet.BuildTools.Azure`, but Wes and I agreed that `CloudTestTasks` was fine for now since the other Azure tasks were there and he didn't care about the namespace so much.

build.proj
chcosta: I can add this back.  This was originally added when this repo was using this task, https://github.com/dotnet/core-setup/blob/master/tools-local/Microsoft.DotNet.Build.Tasks.Local/GetHostInformation.cs.  Now that we're using the task from BuildTools, it appeared safe to remove this, and local builds on Linux / Windows showed it was not necessary.  There are no other mechanisms copying this assembly there that I am aware of, so I'm not certain why the tasks in core-setup.tasks were not complaining about the missing dependency.

tools-local/tasks/core-setup.tasks.csproj
weshaggard: I think in this case he needs to use the independent properties so he can control if he gets the netfx or netstandard version of the task library.

chcosta: The problem with BuildToolsTaskDir is that it is dependent on what platform you're running on, not what framework you're targeting.  If we're building on Windows, we want to be able to reference the desktop assembly and the core assembly independently.

src/pkg/projects/dir.targets
chcosta: Thanks, I needed to force add this file.

Issue #2611, opened 2017-06-03
Remove global location lookup from path in non windows


ramarag:
On linux we infer the global location from the dotnet on the path, 
This needs to change. As a  first step  remove the PATH based current logic at 
https://github.com/dotnet/core-setup/blob/c3dd461c99bfdcb8d136bedf33e46156682fc348/src/corehost/common/pal.unix.cpp#L242
instead return an empty string.

The correct solution is to identify the global location for the various distros, i.e the location where our installers install dotnet:
 
1.	DEB packages:  /usr/share/dotnet
2.	RPM packages: /usr/share/dotnet
3.	PKG for OSX: /usr/local/share/dotnet

cc @eerhardt @gkhanna79 

@steveharter can you bring in the fix ?

gkhanna79:
cc @leecow @Petermarcu since this was discussed in shiproom last Thursday.

eerhardt:
I discussed this last week with a few people that had more Unix knowledge than I do.  The consensus was that if we felt it was necessary to define a "global" location on Unix, we should make it configurable/customizable.  The reason for this is that distros change file system layouts.  And also because end-users may have reasons for wanting to point to other "global" locations - they may not have read/write permissions to what we think is the "global" location, or that disk may be full, or on a network share, etc.

So with that in mind, if we have to define a "global" location on Unix, we should:

1. Not hard-code the default global locations into source code.  Instead provide them through a build-time variable.  This follows the makefile `./configure --prefix=/path/to/dotnet` pattern/usage.
1. Optionally, consider defining an environment variable that will override/augment the default global location. This allows an end-user to point to a custom location, if the default location doesn't work for them and they aren't building from source.
  - We should ensure using an environment variable isn't a security violation before we go through with this approach.  It could cause a privileged `dotnet` host to use a shared framework that is in a non-privileged area on disk, which may lead to an elevation of privilege. 

However, I am still not convinced that user scenario here is valuable enough to support this feature on Unix at all.  We could just completely remove the "global" location concept on Unix until we have a strong need for it.

[...]

PR #2633, opened 2017-06-06
Disable global location lookup for non Windows

steveharter:
Remove support for global location lookup based on PATH for Linux

For issue #2611

This PR may be extended if we decide there are additional semantics required.

### DIFF
--- a/src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
@@ -169,6 +169,11 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.MultilevelSharedFxLooku
             {
                 return;
             }
+            else
+            {
+                // Currently no support for global locations on Linux; remaining test code below is preserved for now.
+                return;
+            }
 
             var fixture = PreviouslyBuiltAndRestoredPortableTestProjectFixture
                 .Copy();
@@ -415,6 +420,11 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.MultilevelSharedFxLooku
             {
                 return;
             }
+            else
+            {
+                // Currently no support for global locations on Linux; remaining test code below is preserved for now.
+                return;
+            }
 
             var fixture = PreviouslyBuiltAndRestoredPortableTestProjectFixture
                 .Copy();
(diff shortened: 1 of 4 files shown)


ramarag:
can you also remove https://github.com/dotnet/core-setup/blob/e4197bc03d599f642b75983431d3ebad265c2588/src/test/TestUtils/Command.cs#L241 ?

src/test/TestUtils/Command.cs
```diff
@@ -238,41 +238,17 @@ public Command WithUserProfile(string userprofile)
...
-                        new_path.Append(sub_path);
-                        new_path.Append(Path.PathSeparator);
-                    }
-                }
-                _process.StartInfo.Environment["PATH"] = new_path.ToString();
-            }
-            return this;
-        }
+
         public Command WithGlobalLocation(string global)
```
eerhardt: This should be removed.  All it does it throw an exception.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -595,7 +615,12 @@ public void Ensure_On_NonWindows_Multiple_Global_Locations_Are_Probed()
             {
                 return;
             }
-            
+            else
+            {
+                // Currently no support for global locations on Linux; remaining test code below is preserved for now.
```
eerhardt: In general, I dislike this approach.  We have the test code preserved in source control history.  If we ever want to use it again, we can bring it back that way.

This just causes maintenance issues because as things get refactored, people will be forced to update this code to keep it building.  However, there is no way of knowing if the updates are valid or not because the test can't be executed.

IMO - we should just remove dead code, and use source code control for history purposes.

src/test/TestUtils/Command.cs
steveharter: The Windows side is eventually expected to use it, so I left it.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
steveharter: I agree on not having 'dead code' but there was a fairly recent change (issue #1553) that disabled the Windows side, and I wasn't sure on the priority of fixing that, although it seems pretty high to me since we still depend on \Program Files type functionality for the non-developer (production) case. So I was assuming we will fix the Windows side soon.

Issue #2513, opened 2017-05-24
Update Runtime Installer [Mac]


leecow:
Strings and installer steps to be updated. Mac and Windows will receive similar updates and there will be separate issue covering Windows changes. Matching changes will be detailed in another issue for the SDK installer.

## Introduction

<img width="682" alt="1 introduction - runtime" src="https://cloud.githubusercontent.com/assets/2212879/26412512/dca4f6d8-405d-11e7-8ad2-d22e4de53d3f.png">

* [ ] Window title: `Install [productname] - [version] arc`
* [ ] Remove "Welcome to ..." text
* [ ] Update Introduction step content as detailed below.

### Intro step text elements
**.NET Core Runtime**

.NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform, and supported by Microsoft. We hope you enjoy it! 

**Learn more about .NET Core**
* [Core Documentation](https://aka.ms/dotnet-docs)
* [SDK Documentation](https://aka.ms/dotnet-cli-docs)
* [Privacy Statement](https://aka.ms/dev-privacy)
* [MIT License](https://aka.ms/dotnet-license)

## License
* [ ] Remove License step from the installer. We've moved entirely to the MIT license so this is not longer needed. 

## Destination Select
* [ ] Window title: `Install [productname] - [version] arc`

## Installation Type
* [ ] Window title: `Install [productname] - [version] arc`

## Installation
* [ ] Window title: `Install [productname] - [version] arc`

## Summary

<img width="682" alt="6 success - runtime" src="https://cloud.githubusercontent.com/assets/2212879/26413167/f45748b0-405f-11e7-9169-b61521dd8aeb.png">

* [ ] Window title: `Install [productname] - [version] arc`
* [ ] Add 'success' graphic
* [ ] Include installation location and list of all components installed
* [ ] Update Summary step content as detailed below.

### Summary step text elements

The following were installed at `/usr/local/share/dotnet`
* .NET Core Runtime 2.0.0
* ASP.NET Runtime Package Store

**Resources**
* [Core Documentation](https://aka.ms/dotnet-docs)
* [SDK Documentation](https://aka.ms/dotnet-cli-docs)
* [Release Notes](https://aka.ms/20-p2-rel-notes)
* [Tutorials](https://aka.ms/dotnet-tutorials)
* [.NET Core Telemetry](https://aka.ms/dotnet-cli-telemetry)

[...]

PR #2635, opened 2017-06-07
Use MIT License for non Windows builds

ellismg:
The windows zips and installer still contain a few non MIT licensed
components, so we continue to use the .NET Library License for them.


ellismg:
@richlander This should do what we discussed.  The ZIPs will now have the MIT licenses in them, as well as the OSX PKGs.

All windows artifacts contain artifacts licensed with the MSFT license.

/cc @omajid

rakeshsinghranchi:
@ellismg , as per https://github.com/dotnet/core-setup/issues/2516 and https://github.com/dotnet/core-setup/issues/2513 , Windows and OSX installer do not show Eula screen. Instead, MIT License link is shown in the welcome page. With your change, do you plan to bring back the Eula in the UI ?

ellismg:
@rakeshsinghranchi I have no plans to make any changes to the UI in the OSX installers. @richlander is that something we should be doing now that things are MIT licensed?

src/pkg/packaging/windows/host/generatemsi.ps1
```diff
@@ -35,7 +35,7 @@ function RunCandle
         -out "$WixObjRoot\" `
         -ext WixDependencyExtension.dll `
         -dHostSrc="$SharedHostPublishRoot" `
-        -dMicrosoftEula="$PackagingRoot\osx\sharedhost\resources\en.lproj\eula.rtf" `
+        -dMicrosoftEula="$PackagingRoot\osx\license\msft\resources\en.lproj\eula.rtf" `
```
eerhardt: Does it still make sense for the Windows installer to get the license from `osx`, now that they aren't shared anymore?

Would it make more sense to put the msft license in the `packaging/windows` folder now?

ellismg: I can do that.

PR #2636, opened 2017-06-07
Fix VersionSuffix for the managed projects [release/2.0.0]

eerhardt:
VersionSuffix is getting set before $(BuildNumberMajor) and $(BuildNumberMinor) are being set.  When creating the DependencyModel nupkg, it is getting a bad version on its p2p reference to PlatformAbstractions.

The fix is to ensure VersionSuffix is defined correctly in the projects themselves - after the obj\BuildVersion.props file is created.

Workaround https://github.com/NuGet/Home/issues/4337

/cc @Petermarcu @mikeharder

### DIFF
--- a/BranchInfo.props
+++ b/BranchInfo.props
@@ -1,0 +1,13 @@
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MajorVersion>2</MajorVersion>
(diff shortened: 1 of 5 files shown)


chcosta:
For release/2.0.0, I think https://github.com/dotnet/core-setup/pull/2638 is a simpler fix.  For master, perhaps a bit of both.

eerhardt:
Closing in favor of #2638.  This refactoring is better served just going into master.

PR #2637, opened 2017-06-07
Fix VersionSuffix for the managed projects [master]

eerhardt:
VersionSuffix is getting set before $(BuildNumberMajor) and $(BuildNumberMinor) are being set.  When creating the DependencyModel nupkg, it is getting a bad version on its p2p reference to PlatformAbstractions.

The fix is to ensure VersionSuffix is defined correctly in the projects themselves - after the obj\BuildVersion.props file is created.

Workaround https://github.com/NuGet/Home/issues/4337


build.proj
```diff
@@ -56,19 +55,7 @@
 
   <Target Name="BatchRestorePackages" Condition="'$(RestoreDuringBuild)'=='true'">
     <Message Importance="High" Text="Restoring all packages..." />
-    <Exec Condition="'@(SdkRestoreProjects)' != ''" Command="$(DotnetRestoreCommand) &quot;%(SdkRestoreProjects.FullPath)&quot; %(SdkRestoreProjects.ExtraRestoreArgs)" StandardOutputImportance="Low" />
-  </Target>
-
-  <!-- Task from buildtools that uses lockfiles to validate that packages restored are exactly what were specified. -->
-  <UsingTask TaskName="ValidateExactRestore" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
-
-  <Target Name="ValidateExactRestore"
```
weshaggard: This looks somewhat unrelated, is this just extra clean-up?

src/managed/CommonManaged.props
```diff
@@ -17,6 +34,21 @@
     <PackageThirdPartyNoticesFile>$(RepoRoot)THIRD-PARTY-NOTICES.TXT</PackageThirdPartyNoticesFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(StabilizePackageVersion)' != 'true'">
+    <!-- Dev builds get a minor version, by default, of '0'.  Our tests restore packages with
```
weshaggard: Is this still true? I didn't think we are doing * versions any longer so we should just be able to do the defaults here.

src/managed/CommonManaged.props
```diff
@@ -5,7 +5,24 @@
 
   <PropertyGroup>
```
weshaggard: For my information. I assume this props file is essentially becoming our prop file for any SDK based projects.

build.proj
eerhardt: It's related because I had to move `BatchRestorePackages` lower in the list.  `ValidateExactRestore` should come after restore, so it would have been needed to be moved as well.  But it doesn't do anything since we don't have project.json files.  So removing it seemed best.

src/managed/CommonManaged.props
eerhardt: Yea, sort of.  Right now we only have the 2 projects, but if we added more SDK-based projects, they would want most of this stuff.

We do have 3 "test" projects that are SDK-based, but they don't import this props file.

Hopefully, we can get more "common" properties refactored into smaller .props files, so they can be shared and we can keep duplication to a minimum.

src/managed/CommonManaged.props
eerhardt: You are right - we aren't using `*` versions anymore. Looks like @mellinoe fixed it.

Here's what we are doing now:
https://github.com/dotnet/core-setup/search?utf8=%E2%9C%93&q=MNAVersion&type=

I'll remove this, and remove it from the dir.props I copied it from as well:
https://github.com/dotnet/core-setup/blob/674c3e3f09284c32321bd9eabd60852e02c0aefb/dir.props#L71-L75

eerhardt:
@weshaggard @chcosta - FYI I plan on merging this change today.  Please let me know if you have remaining feedback.

src/managed/CommonManaged.props
```diff
@@ -17,6 +19,26 @@
...
+  <!--
+  Once https://github.com/dotnet/buildtools/issues/1548 is fixed, the following version sections should be simply
+  <Import Condition="Exists('$(ToolsDir)BuildVersion.targets')" Project="$(ToolsDir)BuildVersion.targets" />
+  -->
+  <PropertyGroup>
+    <TodayTimeStamp>$([System.DateTime]::Now.ToString(yyyyMMdd))</TodayTimeStamp>
+    <BuildVersionFile Condition="'$(BuildVersionFile)'==''">$(ObjDir)BuildVersion-$(TodayTimeStamp).props</BuildVersionFile>
+  </PropertyGroup>
+
+  <Import Condition="Exists('$(BuildVersionFile)')" Project="$(BuildVersionFile)" />
```
chcosta: There should be a shim file in BuildTools that you import which determines the `BuildVersionFile` property and then imports the right file.  That would allow repo's to just import the single `BuildVersionFile` and not carry around this logic to determine the build version file name.  @eerhardt , I think we talked about this a while ago.  It's definitely unrelated to what you've merged here though and isn't blocking.

Thanks for getting this committed.

eerhardt: Right.  My hope once the above referenced issue (https://github.com/dotnet/buildtools/issues/1548) is fixed, that all of this logic just goes away, and this .props file has a simple import `<Import Condition="Exists('$(ToolsDir)BuildVersion.targets')" Project="$(ToolsDir)BuildVersion.targets" />` to get all this functionality.

PR #2650, opened 2017-06-07
Create branch dev/defaultinf

jcagme:
- We are enabling official build support for the dev/defaultintf and so we are adding this branch in the core-setup repo so we can consume packages from it.


chcosta:
What is dev/defaultinf?

ellismg:
> What is dev/defaultinf?

CoreCLR has a dev branch which is adding some new runtime support for default interface methods.  We need a corresponding Microsoft.NETCore.App which ingests this, so Roslyn can do language work on top of this,

dependencies.props
```diff
@@ -28,7 +28,7 @@
   <!-- Package dependency verification/auto-upgrade configuration. -->
   <PropertyGroup>
     <BaseDotNetBuildInfo>build-info/dotnet/</BaseDotNetBuildInfo>
-    <DependencyBranch>master</DependencyBranch>
+    <DependencyBranch>dev/defaultinf</DependencyBranch>
```
ellismg: FYI: This is no longer correct.  We should keep DependencyBranch as "master" and stop using it in the BuildInfoPath for CoreClr (and instead hard code `dev/defaultintf`).

PR #2652, opened 2017-06-08
Sign PlatformAbstractions and DependencyModel binaries

chcosta:
Microsoft.DotNet.PlatformAbstractions and Microsoft.Extensions.DependencyModel dll's were not getting signed because they were not output to a "signable" output folder, and they were not part of the "FilesToSign".

I'm not a huge fan of having to explicitly pass the `BaseOutputPath`  property to both the `dotnet build` and `dotnet pack` commands separately; but, I think this is the most targeted fix we can make for release/2.0.0.  In master, once `dir.props` is broken apart into smaller consumable chunks, we can just import the relevant chunk in `src\managed\CommonManaged.props`.

/cc @eerhardt @gkhanna79 @Petermarcu

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -267,6 +267,7 @@
     <IntermediateOutputRootPath Condition="'$(IntermediateOutputRootPath)' == ''">$(BaseIntermediateOutputPath)$(OSPlatformConfig)\</IntermediateOutputRootPath>
     <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == ''">$(IntermediateOutputRootPath)$(MSBuildProjectName)\</IntermediateOutputPath>
 
+    <ForPackagingDir>$(IntermediateOutputRootPath)sign/forPackaging/</ForPackagingDir>
     <CoreHostOutputDir>$(BaseOutputRootPath)corehost\</CoreHostOutputDir>
 
     <PackagesBasePath Condition="'$(PackagesBasePath)'==''">$(BinDir)$(OSPlatformConfig)/</PackagesBasePath>
--- a/dir.props
+++ b/dir.props
@@ -267,7 +267,6 @@
     <IntermediateOutputRootPath Condition="'$(IntermediateOutputRootPath)' == ''">$(BaseIntermediateOutputPath)$(OSPlatformConfig)\</IntermediateOutputRootPath>
     <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == ''">$(IntermediateOutputRootPath)$(MSBuildProjectName)\</IntermediateOutputPath>
 
-    <ForPackagingDir>$(IntermediateOutputRootPath)sign/forPackaging/</ForPackagingDir>
     <CoreHostOutputDir>$(BaseOutputRootPath)corehost\</CoreHostOutputDir>
 
     <PackagesBasePath Condition="'$(PackagesBasePath)'==''">$(BinDir)$(OSPlatformConfig)/</PackagesBasePath>
--- a/src/managed/dir.proj
+++ b/src/managed/dir.proj
@@ -16,7 +16,7 @@
       <BuildArgs>/p:Configuration=$(ConfigurationGroup)</BuildArgs>
       <BuildArgs Condition="'$(LatestCommit)' != ''">$(BuildArgs) /p:LatestCommit=$(LatestCommit)</BuildArgs>
       <BuildArgs Condition="'$(BuiltByString)' != ''">$(BuildArgs) /p:BuiltByString="$(BuiltByString)"</BuildArgs>
-      <BuildArgs>$(BuildArgs) /p:BaseOutputPath=$(ForPackagingDir)</BuildArgs>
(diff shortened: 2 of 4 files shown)


src/managed/dir.proj
```diff
@@ -16,6 +16,7 @@
       <BuildArgs>/p:Configuration=$(ConfigurationGroup)</BuildArgs>
       <BuildArgs Condition="'$(LatestCommit)' != ''">$(BuildArgs) /p:LatestCommit=$(LatestCommit)</BuildArgs>
       <BuildArgs Condition="'$(BuiltByString)' != ''">$(BuildArgs) /p:BuiltByString="$(BuiltByString)"</BuildArgs>
+      <BuildArgs>$(BuildArgs) /p:BaseOutputPath=$(BaseOutputRootPath)</BuildArgs>
```
gkhanna79: I do not follow - why do we need this to be passed? Who is using it?

chcosta: `BaseOutputPath` is used by the sdk to determine where to put binary output - https://github.com/dotnet/sdk/blob/c6542db9e04b376d8a2f38574bd6eb3ea36b78e5/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.DefaultOutputPaths.targets#L33-L43

Microbuild will only sign binaries which are under either the `OutDir` folder, or `IntermediateOutputPath` https://github.com/dotnet/sdk/blob/c6542db9e04b376d8a2f38574bd6eb3ea36b78e5/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.DefaultOutputPaths.targets#L33-L43

So, we need the PlatformAbstractions and DependencyModel binaries to be produced under `BaseOutputRootPath` so that they can be signed, and then from there, picked up for packaging.

chcosta: *fixed second link in previous comment

sign.proj
```diff
@@ -39,6 +39,15 @@
     </ItemGroup>
 
     <ItemGroup>
+      <FilesToSign Include="$(OutDir)$(ConfigurationGroup)/**/Microsoft.DotNet.PlatformAbstractions.dll">
```
eerhardt: The way this used to work is that we'd put the build result in the `$(IntermediateOutputPath)forPackaging` folder.  See below:

```
     <ItemGroup>
       <FilesToSign Include="$(IntermediateOutputPath)forPackaging/**/*.dll">
       </FilesToSign>
     </ItemGroup>
```

and here:  https://github.com/dotnet/core-setup/blob/v2.0.0-preview1/build_projects/dotnet-host-build/PackageTargets.cs#L216

I think we should set the BaseOutputPath to that folder and then we wouldn't have to change signing.proj at all.

chcosta: Sounds reasonable.  The `forPackaging` folder did stand out to me when I saw it in sign.proj because I didn't see it being used anywhere in the build.  I didn't want to do any cleanup in release/2.0.0 though, so I ignored it.  I'll change the `BaseOutputPath` as you suggested

src/managed/dir.proj
weshaggard: Can it not be imported in the common props file for these?

chcosta: Currently no, dir.props brings in too much stuff, and IntermediateOutputPath depends on a bunch of derived properties which happen in dir.props, so we just have to pass it explicitly.  As mentioned in the description, if dir.props gets refactored and we can pull in just the piece that we want, then we can add that import to the common props file.

chcosta:
Updated, per @eerhardt 's suggestion.  Validation showed that *something* is adding a `sign` folder to the `IntermediateOutputPath`.  doing a text search of `Tools` and Microbuild targets, didn't reveal where that folder suffix is originating from.

dir.props
```diff
@@ -267,6 +267,7 @@
     <IntermediateOutputRootPath Condition="'$(IntermediateOutputRootPath)' == ''">$(BaseIntermediateOutputPath)$(OSPlatformConfig)\</IntermediateOutputRootPath>
     <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == ''">$(IntermediateOutputRootPath)$(MSBuildProjectName)\</IntermediateOutputPath>
 
+    <ForPackagingDir>$(IntermediateOutputRootPath)sign/forPackaging/</ForPackagingDir>
```
eerhardt: How do you feel about just updated the existing `IntermediateOutputForPackaging` property?

[https://github.com/dotnet/core-setup/blob/8bda9c4328e1d59d884799b2375d8c04d8fcb36f/src/dir.props#L7](https://github.com/dotnet/core-setup/blob/8bda9c4328e1d59d884799b2375d8c04d8fcb36f/src/dir.props#L7)

It doesn't appear to be used anywhere.  At least then we don't have 2 "forPackaging" folders.

chcosta: I didn't see that one when I searched for "forPackaging", the important thing to me was to have a common definition somewhere.  I think that makes sense, I'll move it into root\dir.props

chcosta: hmm... actually, I probably don't need to move it, just update it

eerhardt:
test OSX x64 Release Build

chcosta:
test OSX x64 Release Build

PR #2654, opened 2017-06-08
Workaround for recent CMake bug w.r.t. CF guard

janvorli:
Recent versions of CMake generates the <LinkControlFlowGuard> node
under the <Link> node when it finds /guard:cf linker option in the
linker options. But that is not correct. It should generate it under
the <PropertyGroup> node so that the Microsoft.CppCommon.targets would
then generate <ControlFlowGuard> node under the <Link>.

To workaround the problem, change the casing of /guard:cf to uppercase.
CMake then doesn't detect the option as something known to it and passes
the option to the linker as additional option, which fixes the problem.


src/settings.cmake
```diff
@@ -155,8 +155,8 @@ if(WIN32)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /DEBUG /PDBCOMPRESS")
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:1572864")
 
-    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /guard:cf")
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /guard:cf")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /GUARD:CF")
```
gkhanna79: Do we not need to fix https://github.com/dotnet/core-setup/pull/2654/files#diff-5c2eb952c0365ec0c860557785f5d26fR139?

Likewise for the other repos too.

janvorli: We don't. The issue is only with the linker flag. And I have verified that all the target binaries contain the right cf guard stuff using dumpbin.

janvorli:

janvorli:
@dotnet-bot test Windows_NT x86 Debug Build please

sergiy-k:
@janvorli, is your PR ready?

janvorli:
@sergiy-k oh, I have thought that it was merged in a long time ago.

PR #2659, opened 2017-06-08
Update buildtools to 1708-03

karajas:


karajas:
This should unblock the core-setup maestro prs.

PR #2673, opened 2017-06-12
Move release/2.0.0 to Preview 3

rakeshsinghranchi:
Move release/2.0.0 to Preview 3. Also, remove branchinfo.txt file. As we have moved to new build tools & system, this file is not being used and causes confusion.


dir.props
```diff
@@ -15,9 +15,9 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
-    <PreReleaseLabel>preview2</PreReleaseLabel>
+    <PreReleaseLabel>preview3</PreReleaseLabel>
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
-    <ReleaseBrandSuffix>Preview 2</ReleaseBrandSuffix>
+    <ReleaseBrandSuffix>Preview 3</ReleaseBrandSuffix>
```
gkhanna79: Generally, looks good to me. Have you confirmed with a build that all produced packages have the correct pre-release label, including DependencyModel and PlatformAbstractions library?

gkhanna79: @rakeshsinghranchi Can you confirm the answer to the question above?

rakeshsinghranchi: I have looked at the packages produced by PR builds . All package artifacts have correct pre-release label, including DependencyModel and PlatformAbstractions.

PR #2691, opened 2017-06-14
Add symbol signing to core-setup packages.

crummel:
- Publish.proj changes to support signing.
  - Split `PublishCoreHostPackagesToFeed` into Download, Sign, Publish targets so we can insert the symbol signing in the correct place.
  - Call Embed-Index to index symbol files and signatures.
- Build definition changes
  - Install MicroBuild Signing plugin.
  - Clean up previous copies of the tooling.
  - Clone and download symbol signing tooling.
  - Additional variables for tooling location and certificate to use for signing.
  - Run on a machine with the Windows Dev Kit so we know we have `makecat`.
  - Use VS dev command prompt init script so we know were `makecat` is.
- Updated BuildTools for latest symbol signing fixes.

Unfortunately this cannot be tested in full without being in the official build - I tested it with publishing to MyGet, Azure, and dotnet/versions removed from publish.proj.  Example package output is at \\\\crummel-server\build_debug\crummel-core-setup-publish-testing-20170614-22.  Build definition used for testing is at https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?definitionId=6538&_a=completed.

### DIFF
--- a/buildpipeline/Core-Setup-Publish.json
+++ b/buildpipeline/Core-Setup-Publish.json
@@ -2,6 +2,24 @@
   "build": [
     {
       "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Install Signing Plugin",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "30666190-6959-11e5-9f96-f56098202fef",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "signType": "real",
+        "zipSources": "true",
+        "version": "",
+        "feedSource": "https://devdiv.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json"
+      }
+    },
+    {
+      "enabled": true,
       "continueOnError": true,
       "alwaysRun": false,
       "displayName": "Cleanup previous build if present",
@@ -17,6 +35,103 @@
         "arguments": "-path $(PB_SourcesDirectory)",
         "workingFolder": "",
         "inlineScript": "param($path)\n\nif (Test-Path $path){\n    # this will print out an error each time a file can't be deleted.\n    Remove-Item -Recurse -Force $path\n }\n\nif (Test-Path $path){\n    # in case vbcs is still alive\n     $p = Get-Process -Name \"VBCS\"\n     Stop-Process -InputObject $p\n }",
+        "failOnStandardError": "true"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": true,
+      "alwaysRun": false,
+      "displayName": "Cleanup previous tools source if present",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "inlineScript",
+        "scriptName": "",
+        "arguments": "-path $(PB_VsoToolsDir)",
+        "workingFolder": "",
+        "inlineScript": "param($path)\n\nif (Test-Path $path){\n    # this will print out an error each time a file can't be deleted.\n    Remove-Item -Recurse -Force $path\n }\n\nif (Test-Path $path){\n    # in case vbcs is still alive\n     $p = Get-Process -Name \"VBCS\"\n     Stop-Process -InputObject $p\n }",
+        "failOnStandardError": "true"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": true,
+      "alwaysRun": false,
+      "displayName": "Cleanup previous tools if present",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "inlineScript",
+        "scriptName": "",
+        "arguments": "-path $(ToolsRoot)",
+        "workingFolder": "",
+        "inlineScript": "param($path)\n\nif (Test-Path $path){\n    # this will print out an error each time a file can't be deleted.\n    Remove-Item -Recurse -Force $path\n }\n\nif (Test-Path $path){\n    # in case vbcs is still alive\n     $p = Get-Process -Name \"VBCS\"\n     Stop-Process -InputObject $p\n }",
+        "failOnStandardError": "true"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Run script $(VS140COMNTOOLS)\\VsDevCmd.bat",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "bfc8bf76-e7ac-4a8c-9a55-a944a9f632fd",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "$(VS140COMNTOOLS)\\VsDevCmd.bat",
+        "arguments": "",
+        "modifyEnvironment": "true",
+        "workingFolder": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Clone tools",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "git",
+        "arguments": "clone $(PB_VsoToolsRepo) $(PB_VsoToolsDir)",
+        "workingFolder": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Fetch symbol signing tooling",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "filePath",
+        "scriptName": "$(PB_VsoToolsDir)\\scripts\\DotNet-Trusted-Publish\\Fetch-Tools.ps1",
+        "arguments": "$(ToolsRoot)",
+        "workingFolder": "",
+        "inlineScript": "# You can write your powershell scripts inline here. \n# You can also pass predefined and custom variables to this scripts using arguments\n\n Write-Host \"Hello World\"",
         "failOnStandardError": "true"
       }
     },
@@ -302,7 +417,7 @@
     },
     "PB_VersionsRepo": {
       "value": "versions"
-    },    
+    },
     "PB_RepoName": {
       "value": "core-setup"
     },
@@ -314,10 +429,23 @@
     },
     "TeamName": {
       "value": "DotNetCore"
+    },
+    "PB_VsoToolsRepo": {
+      "value": "https://$(PB_VsoAccountName):$(PB_VsoPassword)@devdiv.visualstudio.com/DevDiv/_git/DotNet-BuildPipeline"
+    },
+    "PB_VsoToolsDir": {
+      "value": "$(Build.SourcesDirectory)\\toolsSource"
+    },
+    "ToolsRoot": {
+      "value": "$(Build.SourcesDirectory)\\tools"
+    },
+    "SymbolCatalogCertificateId": {
+      "value": "400"
     }
   },
   "demands": [
-    "Agent.OS -equals Windows_NT"
+    "Agent.OS -equals Windows_NT",
+    "WindowsKit",
   ],
   "retentionRules": [
     {
--- a/buildpipeline/Core-Setup-Publish.json
+++ b/buildpipeline/Core-Setup-Publish.json
@@ -52,7 +52,7 @@
       "inputs": {
         "scriptType": "inlineScript",
         "scriptName": "",
-        "arguments": "-path \"$(VsoToolsDir)\"",
+        "arguments": "-path \"$(PB_VsoToolsDir)\"",
         "workingFolder": "",
         "inlineScript": "param($path)\n\nif (Test-Path $path){\n    # this will print out an error each time a file can't be deleted.\n    Remove-Item -Recurse -Force $path\n }\n\nif (Test-Path $path){\n    # in case vbcs is still alive\n     $p = Get-Process -Name \"VBCS\"\n     Stop-Process -InputObject $p\n }",
         "failOnStandardError": "true"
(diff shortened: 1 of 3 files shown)


crummel:
CC/FYI @chcosta @karajas @gkhanna79 please let me know if you have any concerns.

gkhanna79:
I dont have any concerns. For validation, please chat with @ravimeda and @dagood on how to setup private official builds that could skip publishing but validate all other steps. Once that looks good, we can consider merging this and bringing it to 2.0.0.

CC @weshaggard

buildpipeline/Core-Setup-Publish.json
```diff
@@ -22,6 +40,103 @@
...
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "inlineScript",
+        "scriptName": "",
+        "arguments": "-path $(PB_VsoToolsDir)",
```
ravimeda: Two minor comments about - `"arguments": "-path $(PB_VsoToolsDir)",`
 

- Suggest wrapping path in quotes  `\"$(PB_VsoToolsDir)\"`
- Skip `-path`, since there is only one parameter. Positional is fine.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -22,6 +40,103 @@
...
+        "workingFolder": "",
+        "inlineScript": "param($path)\n\nif (Test-Path $path){\n    # this will print out an error each time a file can't be deleted.\n    Remove-Item -Recurse -Force $path\n }\n\nif (Test-Path $path){\n    # in case vbcs is still alive\n     $p = Get-Process -Name \"VBCS\"\n     Stop-Process -InputObject $p\n }",
+        "failOnStandardError": "true"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": true,
+      "alwaysRun": false,
+      "displayName": "Cleanup previous tools if present",
```
ravimeda: `"displayName": "Cleanup previous tools if present",`

Consider having a script or task in dotnet/buildtools that can accept path(s) to cleanup.  This avoids inline script that can be hard to read in json.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -22,6 +40,103 @@
...
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "filePath",
+        "scriptName": "$(PB_VsoToolsDir)\\scripts\\DotNet-Trusted-Publish\\Fetch-Tools.ps1",
+        "arguments": "$(ToolsRoot)",
+        "workingFolder": "",
+        "inlineScript": "# You can write your powershell scripts inline here. \n# You can also pass predefined and custom variables to this scripts using arguments\n\n Write-Host \"Hello World\"",
```
ravimeda: Empty the comment/text in inlineScript.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -314,10 +429,23 @@
     },
     "TeamName": {
       "value": "DotNetCore"
+    },
+    "PB_VsoToolsRepo": {
```
ravimeda: I think `PB_` prefix is suggested if the variable is consumed in PipeBuild. Same applies to `PB_VsoToolsDir` below.

dagood:
>  * Clone and download symbol signing tooling.

To clarify, the Fetch-Tools.ps1 script is for EmbedIndex, and the extract/catalog/sign/inject flow is handled in BuildTools.

buildpipeline/Core-Setup-Publish.json
crummel: Added quotes.  Left `path` - I think this makes it more clear what we're going to do with the argument.

buildpipeline/Core-Setup-Publish.json
crummel: I'll log an issue for this since this pattern is pretty widespread.

buildpipeline/Core-Setup-Publish.json
crummel: Fixed.

buildpipeline/Core-Setup-Publish.json
crummel: Fixed.

ravimeda:
LGTM

publish/publish.proj
```diff
@@ -164,14 +169,26 @@
...
                                  Condition="$([System.String]::new('%(_DownloadedPackages.Identity)').EndsWith('.symbols.nupkg'))" />
       <_DownloadedStandardPackages Include="@(_DownloadedPackages)"
                                   Exclude="@(_DownloadedSymbolsPackages)" />
     </ItemGroup>
+  </Target>
+
+  <Target Name="SignSymbolPackages" DependsOnTargets="InjectSignedSymbolCatalogIntoSymbolPackages">
+    <PropertyGroup>
+      <PowerShellExe Condition=" '$(PowerShellExe)'=='' ">%WINDIR%\System32\WindowsPowerShell\v1.0\powershell.exe</PowerShellExe>
+      <EmbedIndexScriptLocation Condition=" '$(EmbedIndexScriptLocation)'=='' ">$(PB_VsoToolsDir)\scripts\DotNet-Trusted-Publish\Embed-Index.ps1</EmbedIndexScriptLocation>
```
chcosta: It's weird to have `PB_` variables creep outside of build definitions and into the repo's.  The intent of `PB_` variables is that the prefix ensures there are no collisions with variables in the repo's.

publish/publish.proj
```diff
@@ -164,14 +169,26 @@
...
       <_DownloadedSymbolsPackages Include="%(_DownloadedPackages.Identity)" 
                                  Condition="$([System.String]::new('%(_DownloadedPackages.Identity)').EndsWith('.symbols.nupkg'))" />
       <_DownloadedStandardPackages Include="@(_DownloadedPackages)"
                                   Exclude="@(_DownloadedSymbolsPackages)" />
     </ItemGroup>
+  </Target>
+
+  <Target Name="SignSymbolPackages" DependsOnTargets="InjectSignedSymbolCatalogIntoSymbolPackages">
+    <PropertyGroup>
+      <PowerShellExe Condition=" '$(PowerShellExe)'=='' ">%WINDIR%\System32\WindowsPowerShell\v1.0\powershell.exe</PowerShellExe>
```
chcosta: I don't know the versioning scheme for powershell, but if you didn't fully qualify this path, wouldn't you be more likely to roll forward to whatever version is on the box?

publish/publish.proj
```diff
@@ -164,14 +169,26 @@
...
-      <_DownloadedPackages Include="@(_CoreHostPackages->'$(DownloadDirectory)%(Filename)%(Extension)')" />
+      <_DownloadedPackages Include="@(_CoreHostPackages->'$(PublishDirectory)%(Filename)%(Extension)')" />
       <_DownloadedSymbolsPackages Include="%(_DownloadedPackages.Identity)" 
                                  Condition="$([System.String]::new('%(_DownloadedPackages.Identity)').EndsWith('.symbols.nupkg'))" />
       <_DownloadedStandardPackages Include="@(_DownloadedPackages)"
                                   Exclude="@(_DownloadedSymbolsPackages)" />
     </ItemGroup>
+  </Target>
+
+  <Target Name="SignSymbolPackages" DependsOnTargets="InjectSignedSymbolCatalogIntoSymbolPackages">
```
chcosta: what happens when if you try to publish locally (non-official build)?  does this signing step skip? blow up?  something else?

buildpipeline/Core-Setup-Publish.json
```diff
@@ -314,10 +429,23 @@
     },
     "TeamName": {
       "value": "DotNetCore"
+    },
+    "VsoToolsRepo": {
```
chcosta: Shouldn't these be `PB_` variables?

buildpipeline/Core-Setup-Publish.json
```diff
@@ -14,14 +32,111 @@
...
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "filePath",
+        "scriptName": "$(VsoToolsDir)\\scripts\\DotNet-Trusted-Publish\\Fetch-Tools.ps1",
+        "arguments": "$(ToolsRoot)",
```
chcosta: We should explicitly pass variables where needed rather than depending on env variables

crummel:
Successful build including publishing to the test feed: https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=805932&_a=summary

publish/publish.proj
crummel: Nice catch, changed this in build definition JSON but not here.  Fixed.

buildpipeline/Core-Setup-Publish.json
crummel: I'm not sure.  All of these new variables are used in the core-setup definition, not used by pipebuild - that means they should be non-`PB_`, right?

publish/publish.proj
crummel: If the Windows Dev Kit or Driver Kit are installed (so makecat is present), everything should be fine - we stub out the SignFiles target that MicroBuild depends on.  We could do something wonky like check for `%ProgramFiles%\Microsoft SDKs\Windows` if blowing up without the WDK is unacceptable - this seems like a weird scenario though.  I'm not sure we should ever be publishing directly from a dev build because it would be putting unsigned bits on the official feeds.

buildpipeline/Core-Setup-Publish.json
crummel: I don't think there are any environment variables used here except `ToolPackageSource`, which is more of an option to override, I would expect it to be used only when working on changes like these.  Which ones are you looking at?

publish/publish.proj
crummel: Yes, that is better.  Fixed.

publish/publish.proj
chcosta: I agree that it's a slightly weird scenario.  For publishing validation though, I'll often change the azure account I'm publishing too and then exercise the publishing process locally.  If signing is a requirement it would mean that we didn't have a way to validate publishing outside of official builds.

buildpipeline/Core-Setup-Publish.json
chcosta: Anything that is defined as a variable in a VSTS build definition, should be prefixed with `PB_`.  We haven't fully made this true everywhere.  The idea is to prevent them from leaking into builds as environment variables which can make repro'ing failures difficult.

buildpipeline/Core-Setup-Publish.json
chcosta: None in particular.  Just highlighting that we shouldn't take a dependency on environment variables.

publish/publish.proj
crummel: Ok, would we prefer to skip or error out in this case?  For the purposes of testing publishing and for the official build it seems like erroring out would be better but it would make having makecat a requirement.  It wouldn't make signing a hard requirement though.

buildpipeline/Core-Setup-Publish.json
crummel: So for `VsoToolsDir` (or other things the repo build uses), should we have `PB_VsoToolsDir` in the definition JSON and then an additional `VsoToolsDir` that just references `$(PB_VsoToolsDir)`?  Or can client repos just use the `PB_` variables?

buildpipeline/Core-Setup-Publish.json
crummel: Agreed, it has made things difficult in other repos.  I'll double-check these changes.

dagood: I haven't been fully aware of the guidelines in these last two comment threads, so I may have been making things worse. :smile: Now I know, and they sound reasonable to me.

buildpipeline/Core-Setup-Publish.json
chcosta: In the build definition, you set `PB_VsoToolsDir`, and then any repo step which uses that variable should be explicitly passed `/p:VsoToolsdir=$(PB_VsoToolsDir)`

Side note:  Is it possible to make these into packages which are consumed in the build rather than having to clone a separate repo?

dagood:
> [chcosta] Side note: Is it possible to make these into packages which are consumed in the build rather than having to clone a separate repo?

The EmbedIndex tool package puts everything in a `tools\` dir, which I think makes things more difficult but I don't know how much more. If we change the symstore repo to produce a package with the typical `lib/...` convention then we could (easily) put it in BuildTools which would be great.

publish/publish.proj
crummel: Changed this to skip signing if the WDK is not present.

buildpipeline/Core-Setup-Publish.json
crummel: Changed variables to use PB_ correctly.  Putting all the pieces into a package is something we talked about but haven't done yet.  Logged https://github.com/dotnet/buildtools/issues/1598 for refactoring.

crummel:
@chcosta I've rebased and updated this PR for your comments when you have a chance to look again.  Thanks!

weshaggard:
@crummel is this ready to merge? Also does it need to get into release/2.0.0 branch?

crummel:
@weshaggard Yes, this is ready.  The release/2.0.0 PR is https://github.com/dotnet/core-setup/pull/2809.

publish/publish.proj
```diff
@@ -134,14 +134,17 @@
   </Target>
 
   <Target Name="PublishCoreHostPackagesToFeed"
```
chcosta: nit: I think that `PublishCoreHostPackagesToFeed` and `PushCoreHostPackagesToFeed` is (only slightly) confusing.  What about naming this `Publish` and `PublishCoreHostPackagesToFeed`?

publish/publish.proj
```diff
@@ -134,14 +134,17 @@
...
-          DependsOnTargets="CheckIfAllBuildsHavePublished"
+          DependsOnTargets="CheckIfAllBuildsHavePublished;DownloadCoreHostPackages;SignSymbolPackages;PushCoreHostPackagesToFeed"
           Condition="'@(_MissingBlobNames)' == ''">
     <Error Condition="'$(NuGetFeedUrl)' ==''" Text="Missing required property NuGetFeedUrl" />
     <Error Condition="'$(NuGetApiKey)' == ''" Text="Missing required property NuGetApiKey" />
     <Error Condition="'$(AzureAccessToken)' == ''" Text="Missing required property 'AzureAccessToken'" />
     <Error Condition="'$(AzureAccountName)' == ''" Text="Missing required property 'AzureAccountName'" />
     <Error Condition="'$(ContainerName)' == ''" Text="Missing required property 'ContainerName'" />
+    <Message Condition="'$(WindowsSdkDir)' == ''" Text="Windows SDK not found.  Symbols packages will not be signed." />
```
chcosta: Is there any check that this script is running on Windows, or is it just assumption by convention?

publish/publish.proj
crummel: There's several different "Publish" things going on in this file - publishing packages to Azure, publishing packages to NuGet, publishing .deb packages to their repo, publish debian tools to NuGet.  I think we should continue to be specific here or it could get confusing.  We could rename `PushCoreHostPackagesToFeed` to `PushCoreHostPackagesToNuGetFeed` maybe?

publish/publish.proj
crummel: The pipeline always runs the publish job on a Windows machine.  For dev builds, this could cause issues if the machine is not Windows but has `WindowsSdkDir` set but that seems like a really outside case on something that's already an outside case (publishing from a dev build) to me.

publish/publish.proj
chcosta: That's fine, or make it a verb like `DoPublishCoreHostPackagesToFeed`.

crummel: Fixed.

publish/publish.proj
chcosta: Actually, looking at this again.  I think the behavior you have here is fine.  If a dev is going to publish locally, then signing wouldn't be happening and they could still publish from whatever OS.

chcosta: Also, to be clear, I'm not saying that supporting publishing from Unix, or even from a dev box really, is a functional requirement, it's just nice to have.

Issue #2658, opened 2017-06-08
Language pack story for .NET Core apps


nguerrera:
If my .NET Core app is localized and I have satellite assemblies with the translations, I will often want to distribute the satellite assemblies separately from my main implementation assemblies. This saves disk space for users that don't need the other languages.

On .NET Core, this is problematic because my satellite assemblies need to be listed in my .deps.json. If they are on disk, but not listed, then they don't get used. If they are listed, but not on disk, my program fails. This means that unlike .NET Framework apps, a language pack installer cannot just lay down satellite assemblies in the right place, it also needs to update the deps.json somehow (and roll it back on language pack uninstall).

@gkhanna79 @Petermarcu @eerhardt @livarcocc I'm raising this in response to concerns about size-on-disk impact of satellite assemblies in .NET CLI. We likely don't have the runway to build language packs anyhow for 2.0, but with this situation, I'm not even sure how to make them work reliably.

gkhanna79:
CC @steveharter as well.

nguerrera:
Possible solution:

Add locale-specific deps to separate locale-specifc json:

```
MyApp.dll
MyApp.deps.json
   fr\
      MyApp.resources.dll
      MyApp.resources.deps.json
```

gkhanna79:
I think that layout can be made to work if that is how the language installer will deploy and lay them out.

gkhanna79:
Can there be a resources subfolder under which resources exist under IETF code subfolders? In the example above, "fr" would become "resources\fr".

Is that feasible?

nguerrera:
That could work, but it goes against the .NET Framework tradition. I guess the problem is you wouldn't know which subfolders to scan?

gkhanna79:
> I guess the problem is you wouldn't know which subfolders to scan?
Yes - unless we have a list somewhere that is always kept upto date.

gkhanna79:
Can the language pack installer update the app's "runtimeconfig.json" by adding a property (e.g. resourceFolders) that is a delimited list of IETF codes? If so, the host can read that to augment the resource lookup.

nguerrera:
The LP installer should not modify any files, just drop new ones. If you have to replace/edit/rollback files, you have a much more complicated problem on your hands. We should achieve the same experience as .NET Framework where locale dirs can be added and removed with no other steps required.

gkhanna79:
Can the LP installer drop a "runtimeconfig.resources.json" with that information?

[...]

PR #2717, opened 2017-06-21
Allow missing satellite assemblies

steveharter:
Currently when a satellite assembly is missing, that is treated as an error and the application does not start. This PR changes the behavior to ignore the missing assembly, allow the application to run, and logs a trace informational message about the missing assembly.

This enables scenarios where the core application does not need to provide all satellite assemblies with the main install (to save on disk size, for example), and they can be provided later. However, note that the .deps.json file must still contain entries for all expected languages \ satellite assemblies in order for future satellite assemblies to be dropped in and work (this behavior did not change in the PR).

Fixes https://github.com/dotnet/core-setup/issues/2658

This is expected to be ported to release/2.0.0

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -13,6 +13,16 @@
 #include "fx_ver.h"
 #include "libhost.h"
 
+const pal::string_t MissingAssemblyMessage = _X(
+    "%s:\n"
+    "  An assembly specified in the application dependencies manifest (%s) was not found:\n"
+    "    package: '%s', version: '%s'\n"
+    "    path: '%s'");
+
+const pal::string_t ManifestListMessage = _X(
+    "  This assembly was expected to be in the local runtime store as the application was published using the following target manifest files:\n"
+    "    %s");
+
 namespace
 {
 // -----------------------------------------------------------------------------
@@ -303,25 +313,47 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
     return false;
 }
 
-bool report_missing_assembly_in_manifest(const deps_entry_t& entry)
+bool report_missing_assembly_in_manifest(const deps_entry_t& entry, bool isNotError = false)
 {
-    trace::error(_X(
-        "Error:\n"
-        "  An assembly specified in the application dependencies manifest (%s) was not found:\n"
-        "    package: '%s', version: '%s'\n"
-        "    path: '%s'"),
-        entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+    bool showManifestListMessage = !entry.runtime_store_manifest_list.empty();
 
-    if (!entry.runtime_store_manifest_list.empty())
+    if (entry.asset_type == deps_entry_t::asset_types::resources)
     {
-        trace::error(_X(
-            "  This assembly was expected to be in the local runtime store as the application was published using the following target manifest files:\n"
-            "    %s"),
-            entry.runtime_store_manifest_list.c_str());
+        // Treat missing resource assemblies as informational.
+        isNotError = true;
+
+        trace::info(MissingAssemblyMessage.c_str(), _X("Info"),
+            entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+
+        if (showManifestListMessage)
+        {
+            trace::info(ManifestListMessage.c_str(), entry.runtime_store_manifest_list.c_str());
+        }
+    }
+    else if (isNotError)
+    {
+        trace::warning(MissingAssemblyMessage.c_str(), _X("Warning"),
+            entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+
+        if (showManifestListMessage)
+        {
+            trace::warning(ManifestListMessage.c_str(), entry.runtime_store_manifest_list.c_str());
+        }
+    }
+    else
+    {
+        trace::error(MissingAssemblyMessage.c_str(), _X("Error"),
+            entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+
+        if (showManifestListMessage)
+        {
+            trace::error(ManifestListMessage.c_str(), entry.runtime_store_manifest_list.c_str());
+        }
     }
 
-    return false;
+    return isNotError;
 }
+
 /**
  *  Resovle the TPA assembly locations
  */
@@ -583,9 +615,7 @@ bool deps_resolver_t::resolve_probe_dirs(
             // because of rid-fallback could happen (ex: CentOS falling back to RHEL)
             if ((entry.asset_name == _X("apphost")) && ends_with(entry.library_name, _X(".Microsoft.NETCore.DotNetAppHost"), false))
             {
-                trace::warning(_X("Warning: assembly specified in the dependencies manifest was not found -- package: '%s', version: '%s', path: '%s'"), 
-                    entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
-                return true;
+                return report_missing_assembly_in_manifest(entry, true);
             }
 
             return report_missing_assembly_in_manifest(entry);
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -313,14 +313,14 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
     return false;
 }
 
-bool report_missing_assembly_in_manifest(const deps_entry_t& entry, bool isNotError = false)
+bool report_missing_assembly_in_manifest(const deps_entry_t& entry, bool continueResolving = false)
 {
     bool showManifestListMessage = !entry.runtime_store_manifest_list.empty();
 
     if (entry.asset_type == deps_entry_t::asset_types::resources)
     {
         // Treat missing resource assemblies as informational.
-        isNotError = true;
+        continueResolving = true;
 
         trace::info(MissingAssemblyMessage.c_str(), _X("Info"),
             entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
@@ -330,7 +330,7 @@ bool report_missing_assembly_in_manifest(const deps_entry_t& entry, bool isNotEr
             trace::info(ManifestListMessage.c_str(), entry.runtime_store_manifest_list.c_str());
         }
     }
-    else if (isNotError)
+    else if (continueResolving)
     {
         trace::warning(MissingAssemblyMessage.c_str(), _X("Warning"),
             entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
@@ -351,7 +351,7 @@ bool report_missing_assembly_in_manifest(const deps_entry_t& entry, bool isNotEr
         }
     }
 
-    return isNotError;
+    return continueResolving;
 }
 
 /**
--- a/dir.props
+++ b/dir.props
@@ -84,7 +84,7 @@
 
     <SharedFrameworkNugetVersion>$(ProductVersion)</SharedFrameworkNugetVersion>
 
-    <UWPCoreRuntimeSdkVersion>5.4.0</UWPCoreRuntimeSdkVersion>
+    <UWPCoreRuntimeSdkVersion>6.0.0</UWPCoreRuntimeSdkVersion>
     <UWPCoreRuntimeSdkFullVersion>$(UWPCoreRuntimeSdkVersion)$(ProductVersionSuffix)</UWPCoreRuntimeSdkFullVersion>
 
     <NuGetVersion>$(SharedFrameworkNugetVersion)</NuGetVersion>
--- a/src/pkg/packaging/deb/dotnet-hostfxr-debian_config.json
+++ b/src/pkg/packaging/deb/dotnet-hostfxr-debian_config.json
@@ -7,7 +7,7 @@
 
     "short_description": "%HOSTFXR_BRAND_NAME% %HOSTFXR_NUGET_VERSION%",
     "long_description": ".NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform and is supported by Microsoft. We hope you enjoy using it! If you do, please consider joining the active community of developers that are contributing to the project on GitHub (https://github.com/dotnet/core). We happily accept issues and PRs.",
-    "homepage": "https://dotnet.github.io/core",
+    "homepage": "https://dotnet.github.io",
 
     "release":{
         "package_version":"1.0.0.0",
--- a/src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
+++ b/src/pkg/packaging/deb/dotnet-sharedframework-debian_config.json
@@ -7,7 +7,7 @@
 
     "short_description": "%SHARED_FRAMEWORK_BRAND_NAME% %SHARED_FRAMEWORK_NUGET_NAME% %SHARED_FRAMEWORK_NUGET_VERSION%",
     "long_description": ".NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform and is supported by Microsoft. We hope you enjoy using it! If you do, please consider joining the active community of developers that are contributing to the project on GitHub (https://github.com/dotnet/core). We happily accept issues and PRs.",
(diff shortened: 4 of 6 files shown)


src/corehost/cli/deps_resolver.cpp
```diff
@@ -303,25 +313,40 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
...
+    {
+        // Treat missing resource assemblies as a warning, not an error.
+        isWarning = true;
+    }
+
+    if (isWarning)
+    {
+        trace::warning(MissingAssemblyMessage.c_str(), _X("Warning"),
+            entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+    }
```
nguerrera: We don't want a warning either. We are going to be relying on this behavior in production when the language pack is not installed.

steveharter: AFAIK there are no implications of a warning other than a log message if logging is enabled. Are you suggesting we write an informational message or nothing? I think having some sort of message would be useful for scenarios where all declared satellites are expected.

nguerrera: Informational works for me. I don't think anything by-design should be logged as a warning.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -303,25 +313,47 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
...
-    trace::error(_X(
-        "Error:\n"
-        "  An assembly specified in the application dependencies manifest (%s) was not found:\n"
-        "    package: '%s', version: '%s'\n"
-        "    path: '%s'"),
-        entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+    bool showManifestListMessage = !entry.runtime_store_manifest_list.empty();
 
-    if (!entry.runtime_store_manifest_list.empty())
+    if (entry.asset_type == deps_entry_t::asset_types::resources)
```
gkhanna79: If you make this "if" block only to set isNotError state for resources, then you can avoid duplication of the remaining contents of the "if" block" with the immediately following "else if".

src/corehost/cli/deps_resolver.cpp
```diff
@@ -583,9 +615,7 @@ bool deps_resolver_t::resolve_probe_dirs(
             // because of rid-fallback could happen (ex: CentOS falling back to RHEL)
             if ((entry.asset_name == _X("apphost")) && ends_with(entry.library_name, _X(".Microsoft.NETCore.DotNetAppHost"), false))
             {
-                trace::warning(_X("Warning: assembly specified in the dependencies manifest was not found -- package: '%s', version: '%s', path: '%s'"), 
-                    entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
-                return true;
+                return report_missing_assembly_in_manifest(entry, true);
```
gkhanna79: Also, I believe @Petermarcu wanted this to become "false" by default.

steveharter: Do you want to log an issue for this, or change it now? If we return false, this condition will cause the app not to load. A case of misconfigured deps data? Looking at history, it's been this way (returning true) since the code was added in May 2016.

src/corehost/cli/deps_resolver.cpp
steveharter: The next elseif logs as `warning`, not `info`. So there are three cases: 1) log as `info` for resources, 2) log as `warning` for "apphost" case also discussed below, 3) log as `error` for all other cases. False (which will cause app start to fail) is returned only for the error case.

gkhanna79: You can use a variable to conditionalize the value. It is your call.

src/corehost/cli/deps_resolver.cpp
gkhanna79: I was referring to pass "false" for the second argument (isNotError) by default.

steveharter: Hmm the current code (with the second commit) does have "false" being the default value of that argument, in which case the method will return "false" as well unless the entry is a resource\satellite assembly.

steveharter: I talked to Peter; I understand your feedback now. The intent was for all types of deps entries to not fail; will have another commit soon for that.

steveharter:
Per previous comments, the code is now updated to ignore all missing assemblies (resource, native and managed\tpa) instead of just resources. A trace message is logged with the information about what was not found.

@jkotas do you see any issue with this, especially with not passing all assemblies in the tpa list? This means at runtime, when there is an attempt to load the assembly, there will be a FileNotFoundException (or similar exception), or perhaps it will succeed and find the assembly anyway although possibly not the intended version, etc.

Also on the original issue with satellite assemblies - since only the root path of the resource directories are passed to coreclr, as long as a additional drop-in satellite assembly shares a root path with a valid resource assembly (that exists on disk and is in the .deps.json file) then they will "light up" and be used at runtime provided the locale matches the directory. For example, if `mydir\de\MyApp.Resources.dll` exists in the .deps.json file and on disk, then `mydir` is passed to the coreclr and thus `mydir\fr\MyApp.Resources.dll` will be found when `fr` locale is used even if it was not in the .deps.json file).

cc @Petermarcu

steveharter:
@eerhardt also per our discussion this won't fix https://github.com/dotnet/sdk/issues/1362 however you will now no longer get the error message `An assembly specified in the application dependencies manifest...` when the app is run. Instead, the missing assembly will be ignored and a FileNotFoundException or similar will likely result. The intent of https://github.com/dotnet/sdk/issues/1362 however is to catch this error condition at publish time, not run time.

Petermarcu:
@gkhanna79 , do you mind also taking a look and weighing in on whether there is anything we need to think about WRT to the TPA list?

steveharter:
I will hold on this until I get confirmation from @gkhanna79 and\or @jkotas. Then I will start the port to release/2.0.0 asap.

eerhardt:
@steveharter just reminded me that we did work for the runtime store specifically in this area to give a good error message when the assembly wasn't able to be found.

It sounds like we are regressing that behavior now. Instead of getting a nice error message that says "the assembly can't be loaded, might be missing a runtime store for manifest 'foo'", now the message will be hidden in the dotnet host trace (if enabled). And instead, we're back to the generic "FileNotFound" error message without any indication.

Maybe we still want to error out with a good error if the assembly is associated with a runtime store?

nguerrera:
Maybe just plumb the error through to the eventual FileNotFoundException if any. (He said with zero understanding of the technical challenge that poses.)

nguerrera:
Does the store work with resource assemblies? I am thinking we're getting in to feature creep for v2.0 by scoping this beyond resource assemblies. Maybe tackle the rest for vNext? We are only blocked on the resource assembly case.

jkotas:
I agree that it would be nice to make this more fine grained and still do the hard validation in most cases. We are losing diagnosability and potential making our support cost higher with the proposed change.

> We are only blocked on the resource assembly case.

What about:
- Do not require the localized resource assemblies to be listed in .deps.json
- Leave it up to the runtime to look for them by convention

nguerrera:
@jkotas That would be ideal and also fix the scenario (that we do not have for CLI 2.0) where you don't know all of the languages up-front and you reserve the option to just compile and ship a new language pack without changing the main installer. My understanding from talking to @gkhanna79 was that simply not failing up-front if satellite in deps is not found was the safer alternative for the tight 2.0 schedule. However, if what you propose is not risky and easy enough, then I think we should do that.

livarcocc:
The satellite assembly not listed in the deps.json comes with other issues, like, where should the host look to find them? What about the deps.json itself? We would have to go change the deps.json to not list these assemblies, which would spread the work for this as well.

I am fine if we decide to do something different here, but if that's the case, I would like some of us to get together and to design the new approach, rather than figure it out in the PR.

Though, my vote for 2.0 is to either do what we are doing here (don't error out, but keep the store scenario sane), or just scope this to satellite assemblies.

steveharter:
With the current pr it does support drop-in satellite assemblies provided they follow folder naming convention along with shared base directory as I explained earlier, or if they are nested under the current standalone app directory.

Peter and others have scenarios where the error check for non-resource assemblies specified in the .deps.json got in the way, thus the latest changes to this pr... So whether we can wait post 2.0 to address this depends on the magnitude of those other scenarios. @Petermarcu and others with such scenarios, can you elaborate on these please. Thanks.

P.s. A 'strict' option was also discussed but considered too risky at this point.

Petermarcu:
Lets sync up tomorrow on this then. For the case of the runtime store, it does make sense to have a hard expectation that it is found in the runtime store.

There have been many cases where users have wanted to do lightup/reflection, plugin models, or IoC scenarios where the assembly they want to load shows up later than when the app was built. I want to make sure apps have the freedom to take a dependency on an assembly and own the responsibility of ensuring that it is shipped with the app and not have to hack the deps file to do it.

jkotas:
> There have been many cases where users have wanted to do lightup/reflection, plugin models, or IoC scenarios where the assembly they want to load shows up later than when the app was built. I want to make sure apps have the freedom to take a dependency on an assembly and own the responsibility of ensuring that it is shipped with the app and not have to hack the deps file to do it.

@Petermarcu This is orthogonal to the problem that this PR is trying to address: There is `Microsoft.NETCore.DotNetHostPolicy.SetAppPaths` property that you can set in your `runtimeconfig.json` that will cause the runtime to pick up everything in the app directory. It is off by default because of it is the more secure and serviceable option. (You can also achieve the same effect of probing the app directory by a couple of lines of code for assembly resolve handler.)

Petermarcu:
@jkotas , and the assemblies will be loaded without being rejected because they weren't in the deps file?

Petermarcu:
Will it load automatically if you set `Microsoft.NETCore.DotNetHostPolicy.SetAppPaths` or to people have to do an explicit load?

jkotas:
It will load automatically. It behaves as if `<your app directory>\*.dll` was added to your deps file, without the package metadata - so extra .dlls you pick up this way are not serviceable, cannot be replaced by copies from store, etc.

Petermarcu:
Ok, that should solve my scenario.

steveharter:
This is my understanding:
1) Ignore missing satellite assemblies (the original issue).
2) Keep existing behavior of failing on other missing native and managed assemblies that are in the deps file. If someone wants a late-bound plug-in model, they will have to use an alternative approach of either an explicit load or SetAppPaths.
3) Live with the drop-in satellite assemblies quirk where you need at least one reference to a satellite assembly. The `PLATFORM_RESOURCE_ROOTS` property must be set, which is passed to coreclr to specify where to look for resources. In order for that to occur, there must be at least one resource \ satellite assembly specified already - e.g. `myApp/fr/MyApp.Resources.dll`. So in a typical scenario, the `en` resources are going to be embedded in `MyApp.dll` with no satellite assemblies specified in the deps because the assumption is a language pack will come later. However, just dropping in the satellite assemblies from the language pack later will *not* work because `PLATFORM_RESOURCE_ROOTS` will still be empty - there needed to be at least one satellite assembly to begin with that is in the deps file. This means tooling around this scenario have at least one resource assembly specified in the deps file, even though the physical satellite assembly does not need to exist until the language pack drops. It also means other\additional languages can be added later and still work even if they were not in the deps file.

Petermarcu:
@nguerrera @livarcocc are you good with 3? @steveharter is there not a good option for setting a default location?

nguerrera:
With no tooling changes, we will have *all* satellites listed in deps, and they may or may not be on disk depending if language pack is deployed or not. So above will work for us, but the general semantics are murkier than the conclusion in our last meeting. I think we should meet again to talk through consequences and evaluate alternatives.

steveharter:
I just sent out a meeting invite for later today to continue discussion.

Petermarcu:
I won't be able to do the meeting but I do like the idea of having a default. The common practice is to have your default language in your main assembly. In that case it would be nice if dropping in the resource assemblies for other languages just lit up if they are put in the right place.

eerhardt:
Does this design handle resources for assemblies that aren't in your app's directory?  For example - 'drop-in' resource assemblies in the runtime store, or 'drop-in' resource assemblies in the shared framework?

Basically, if I have an assembly foo.dll that was loaded from `/path/to/runtimestore/foo.dll`, and I drop some resource assembly in `/path/to/runtimestore/de/foo.resources.dll`, will the `de` resource get picked up?  Or will it only look in my app's directory for resources?

By default, ASP.NET assemblies won't be in the app's directory - they are loaded from the runtime store.  So if we ever wanted to localize the ASP.NET assemblies with drop-in targeting packs, would this design work?

Petermarcu:
Its a good question. One of the drivers for this is Roslyn's resources I believe. @livarcocc if Roslyn is in the runtime store, do you need this to work for the store to see the benefits?

nguerrera:
Yes, we do need Roslyn to be translated. I'm assuming we're protected if we have all of its satellites listed in csc and vbc's deps files.  I think the drop-in scenario with no listing in deps file is compelling, but again, it is not blocking us.  It feels like what's described in (3) or (3) + app path default is not the right, deliberate design. If we want to support removing resources from the deps altogether, I think every loaded assembly should get the chance to find its satellites next to it by the long-standing convention, not just assemblies that have at least one satellite of any language listed, or just assemblies loaded from app path.

gkhanna79:
Even with making resource assemblies opyi9nally present, the fundamental premise does not chabfe... that path to resource assemblies comes from the deos.json file.

Based upon discussion last werk, that premise is not changing with langpacks.

Are we trying to design a more general resource lightup?



Thanks,
Gaurav


nguerrera:
@gkhanna79 We're not asking for a more general light-up feature, though I think it would be useful. What's concerning to me is that it sounds that when we just remove the up-front check, then a half-baked lightup feature emerges as described in (3) of @steveharter's list.

nguerrera:
@steveharter and I spoke on the phone. Unfortunately, nobody else was available. After talking it through, my concern about a "half-baked light-up feature emerging" from removing the up-front check was wrong. This behavior already exists: iff you specify at least one resource, then other resources (even other languages) can be found even if they are not in the deps. 

So I think @steveharter's proposal is fine. 

Any general light-up scenarios should be considered separately and I think that we should not tie them to the v2.0 release or at least not to this PR.

steveharter:
The current code now implements 1 above (ignoring satellite assemblies) and does nothing for 2 and 3 as those are existing semantics. I plan on committing this and then creating the port to release/2.0.0

steveharter:

```NuGet.targets(97,5): error : Unable to obtain lock file access on 'C:\Users\dotnet-bot\AppData\Local\Temp\NuGetScratch\lock\daa510a16fa60d5c9e9ed25962062ce28c4b99e3' for operations on 'D:\j\workspace\debug_windows---b57b94ba\packages\runtime.ubuntu.16.10-x64.runtime.native.system.security.cryptography.openssl\4.3.0\runtime.ubuntu.16.10-x64.runtime.native.system.security.cryptography.openssl.4.3.0.nupkg'. This may mean that a different user or administator is holding this lock and that this process does not have permission to access it. If no other process is currently performing an operation on this file it may mean that an earlier NuGet process crashed and left an inaccessible lock file, in this case removing the file 'C:\Users\dotnet-bot\AppData\Local\Temp\NuGetScratch\lock\daa510a16fa60d5c9e9ed25962062ce28c4b99e3' will allow NuGet to continue. [D:\j\workspace\debug_windows---b57b94ba\src\pkg\projects\Microsoft.NETCore.UniversalWindowsPlatform\src\Microsoft.NETCore.UniversalWindowsPlatform.depproj] [D:\j\workspace\debug_windows---b57b94ba\src\pkg\projects\Microsoft.NETCore.UniversalWindowsPlatform\src\Microsoft.NETCore.UniversalWindowsPlatform.depproj]```

File: src/corehost/cli/deps_resolver.cpp [hunk 4]
```diff
@@ -303,25 +313,47 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
...
+            entry.deps_file.c_str(), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+
+        if (showManifestListMessage)
+        {
+            trace::info(ManifestListMessage.c_str(), entry.runtime_store_manifest_list.c_str());
+        }
     }
+    else if (continueResolving)
+    {
+        trace::warning(MissingAssemblyMessage.c_str(), _X("Warning"),
```
nguerrera: Isn't this new warning case more than (1).

steveharter: No since the previous code logged a warning before here https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/deps_resolver.cpp#L586. Now it just calls this same helper method to ensure to ensure it logs an info for missing satellite assemblies, and a warning for other assemblies. As a benefit, however, it also outputs with the improved error message format.

PR #2724, opened 2017-06-22
Update BuildTools to 2.0.0-prerelease-01721-01.

crummel:
This should fix the issue with "An item with the specified name already exists" in the core-setup Maestro definitions.


crummel:
Yes, I don't think all builds clean up after themselves correctly.  There's a few issues for CleanupVSTSAgent but I'm not immediately seeing one for making sure it's on all builds, I'll dig a little deeper and possibly create one.

PR #2726, opened 2017-06-23
Update uap10.1 moniker in packages to uap10.0.15138 to support NS2.0 projects in VS

zamont:
VS is setting the NuGet TFM to the build number of the Windows SDK TargetPlatformVersion. For example, when you build in VS and need to restore your NuGet packages, you may see the following in your build output:

```
Restoring packages for UAP,Version=v10.0.10586...
```

Thus, we will set the TFM to a value that is higher than the latest released Windows SDK: `10.0.15063.0`. This will allow us to restore newer libraries for NS2.0.

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
+++ b/src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
@@ -31,7 +31,7 @@
     </Dependency>
 
     <ProjectReference Include="..\Microsoft.Net.UWPCoreRuntimeSdk\Microsoft.Net.UWPCoreRuntimeSdk.pkgproj">
-      <TargetFramework>uap10.1</TargetFramework>
+      <TargetFramework>uap10.0.15138</TargetFramework>
     </ProjectReference>
 
     <!-- Add a placeholder to make sure NuGet understands we still support older versions,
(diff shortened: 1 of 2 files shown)


ericstj:
We need to make a holistic change to plumb this through the system and centralize its definition.  I don't want to see this number appear in more than one place in each repo.   It also needs to be in CoreFx where we build other packages which need to install in UAP.

zamont:
@joperezr @ericstj Does this change look good? As we discussed, this will need to be better implemented later but right now it's going to unblock us. I ran a local build without issues building the various packaging projects.

zamont:
I will need to double check that VS will restore this package as I expect it to. Depending on the results of that, I'd like to check in what I have right now.

zamont:
I'm merging to unblock us. If you're concerned about this PR merging, please let me know and we can fix it afterwards.

joperezr:
LGTM

src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.pkgproj
```diff
@@ -31,7 +31,7 @@
     </Dependency>
 
     <ProjectReference Include="..\Microsoft.Net.UWPCoreRuntimeSdk\Microsoft.Net.UWPCoreRuntimeSdk.pkgproj">
-      <TargetFramework>uap10.1</TargetFramework>
+      <TargetFramework>uap10.0.15138</TargetFramework>
```
weshaggard: Does nuget already have mapping for this value? What about the VS templates are they now using this value? What about all the corefx packages, they are still using 10.1 so none of them will work.

joperezr: > Does nuget already have mapping for this value?

No

> What about the VS templates are they now using this value? 

Zach should be able to answer this but I believe so.

> What about all the corefx packages, they are still using 10.1 so none of them will work.

For now we just wanted to do the Hello World scenario which only requires the UWP metapackage. Once we have the real moniker (as this one is just temporary for now) then we will change all of our corefx OOB packages and the metapackage to use that moniker instead.

zamont: VS is using the moniker "10.0.xxxxx" where xxxxx is the Windows build number. To complete these changes down the road, we will set this value to the exact value of the RS3 build number.

PR #2752, opened 2017-06-28
Fix publish error introduced by recent UwpCoreRuntimeSdk work

zamont:
Modify publish/dir.props to wildcard select the CoreRuntime appx package to upload so it's not uploaded on distros where it has not been built.

### DIFF
--- a/publish/dir.props
+++ b/publish/dir.props
@@ -22,7 +22,7 @@
     <RuntimePackageFile Include="$(PackagesOutDir)**/runtime.*.nupkg" >
       <RelativeBlobPath>$(BinariesRelativePath)</RelativeBlobPath>
     </RuntimePackageFile>
-    <UWPArtifactsToUpload Include="$(MicrosoftNetCoreRuntimeAppxOutputPath)" >
+    <UWPArtifactsToUpload Include="$(UWPOutputDir)Microsoft.NET.CoreRuntime.*.appx" >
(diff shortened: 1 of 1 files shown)


zamont:
@ravimeda this is the proposed fix. The previous commit hard-coded a path regardless if it was present. This commit makes it a wildcard expansion so it will only get expanded if the file is there.

publish/dir.props
```diff
@@ -22,7 +22,7 @@
     <RuntimePackageFile Include="$(PackagesOutDir)**/runtime.*.nupkg" >
       <RelativeBlobPath>$(BinariesRelativePath)</RelativeBlobPath>
     </RuntimePackageFile>
-    <UWPArtifactsToUpload Include="$(MicrosoftNetCoreRuntimeAppxOutputPath)" >
+    <UWPArtifactsToUpload Include="$(UWPOutputDir)Microsoft.NET.CoreRuntime.*.appx" >
```
weshaggard: I would not duplicate this but instead do condition exists check. `<UWPArtifactsToUpload Include="$(MicrosoftNetCoreRuntimeAppxOutputPath)" Condition="Exists('$(MicrosoftNetCoreRuntimeAppxOutputPath)')">`

zamont: Okay, yeah I can do that instead.

PR #2755, opened 2017-06-29
Update prerelease label to preview2

ellismg:
release/uwp6.0 is going to be using preview1


branchinfo.txt
```diff
@@ -6,6 +6,6 @@ MINOR_VERSION=1
 PATCH_VERSION=0
 STABILIZE_PACKAGE_VERSION=false
 LOCK_HOST_VERSION=false
-RELEASE_SUFFIX=preview1
+RELEASE_SUFFIX=preview2
```
weshaggard: I'm not sure this file is used anymore but probably still worth updating until it gets deleted.

eerhardt: Can we just delete it now?  No one uses it, and it has been deleted in release/2.0.0

PR #2770, opened 2017-06-30
Sign UWP artifacts

chcosta:
These are mostly @nattress 's changes, with the exception that I moved this section...

```
      <FilesToSign Include="$(UWPOutputDir)*.appx"> 
      </FilesToSign> 
```

The appx files are produced during the "Build NuGet Packages" step, so they wouldn't be available for signing during the binaries phase.

FYI @joshfree @zamont @nattress

### DIFF
--- a/tools-local/tasks/ValidateBinInspectResults.cs
+++ b/tools-local/tasks/ValidateBinInspectResults.cs
@@ -52,7 +52,7 @@ namespace Microsoft.DotNet.Build.Tasks
             {
                 foreach(var baselineFile in BaselineFiles)
                 {
-                    IEnumerable<XElement> baselineExcluded = errorRows.Where(f => Regex.IsMatch(f.Descendants(s_NameElement).First().Value), baselineFile.ItemSpec, RegExOptions.IgnoreCase);
+                    IEnumerable<XElement> baselineExcluded = errorRows.Where(f => Regex.IsMatch(f.Descendants(s_NameElement).First().Value), baselineFile.ItemSpec, RegexOptions.IgnoreCase);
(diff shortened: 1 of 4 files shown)


tools-local/tasks/ValidateBinInspectResults.cs
```diff
@@ -52,8 +52,7 @@ public override bool Execute()
             {
                 foreach(var baselineFile in BaselineFiles)
                 {
-                    Regex ignoreRegEx = new Regex(baselineFile.ItemSpec);
-                    IEnumerable<XElement> baselineExcluded = errorRows.Where(f => ignoreRegEx.IsMatch(f.Descendants(s_NameElement).First().Value));
+                    IEnumerable<XElement> baselineExcluded = errorRows.Where(f => Regex.IsMatch(f.Descendants(s_NameElement).First().Value), baselineFile.ItemSpec, RegexOptions.IgnoreCase);
```
zamont: The Regex.IsMatch call seems to be causing the build failures

zamont: Based on this StackOverflow [link](https://stackoverflow.com/questions/2439965/case-insensitive-regex-without-using-regexoptions-enumeration), you might be able to keep the old code, except make the first line:

```
Regex ignoreRegEx = new Regex($"(?i){baselineFile.ItemSpec}");
```

chcosta: Thanks, this is what I get for trying to write code from my phone when I'm half asleep.  I'll take a look.

chcosta:
Test windows_nt arm Debug build

PR #2779, opened 2017-07-03
Release build failure: Increase timeout to allow core-setup downloads to complete

ChadNedzlek:
These builds download a huge amount of data by opening literally hundreds of simultaneous connections.  In the official build logs, some of these requests end up at the end of the queue and timeout while waiting their turn.  Most of those eventually get a turn, but some of them don't.  I'm raising the timeout 6 fold (the default timeout is 100 seconds), and hoping that's enough.


ChadNedzlek:
Yeah, throttling would be a good idea, but we'd have to switch the build around a bit, since it's just invoking hundreds of parallel tasks. We'd have to switch to a single task that does throttling internally (probably allowing 4 or 5 parallel downloads at a time). But that's a more substantial change that I didn't want to try to make in this release branch.

ChadNedzlek:

PR #2793, opened 2017-07-05
Filter out version = -1.-1.-1 & tests: SDK Muxer

ghost:
version = -1.-1.-1 is a placeholder version and should not be considered
add tests for the same...

Other:
-test directories are isolated; we don't need to delete SDK test versions from them anymore
-better failure comment for the global.json lookup
-combining methods: 'resolve_sdk_version' and 'resolve_patch_sdk_version' 

Addresses: https://github.com/dotnet/core-setup/issues/1760

@dotnet/dotnet-cli

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -595,48 +595,20 @@ pal::string_t fx_muxer_t::resolve_cli_version(const pal::string_t& global_json)
     return retval;
 }
 
-pal::string_t resolve_sdk_version(pal::string_t sdk_path, bool parse_only_production)
+pal::string_t resolve_sdk_version(pal::string_t sdk_path, bool parse_only_production, pal::string_t global_cli_version)
 {
-    trace::verbose(_X("--- Resolving SDK version from SDK dir [%s]"), sdk_path.c_str());
+    fx_ver_t specified(-1, -1, -1);
 
-    pal::string_t retval;
-    std::vector<pal::string_t> versions;
-
-    pal::readdir(sdk_path, &versions);
-    fx_ver_t max_ver(-1, -1, -1);
-    for (const auto& version : versions)
+    //   Validate the global cli version if specified
+    if (!global_cli_version.empty())
     {
-        trace::verbose(_X("Considering version... [%s]"), version.c_str());
-
-        fx_ver_t ver(-1, -1, -1);
-        if (fx_ver_t::parse(version, &ver, parse_only_production))
+        if (!fx_ver_t::parse(global_cli_version, &specified, false))
         {
-            max_ver = std::max(ver, max_ver);
+            trace::error(_X("The specified SDK version '%s' could not be parsed"), global_cli_version.c_str());
+            return pal::string_t();
         }
     }
 
-    pal::string_t max_ver_str = max_ver.as_str();
-    append_path(&sdk_path, max_ver_str.c_str());
-
-    trace::verbose(_X("Checking if resolved SDK dir [%s] exists"), sdk_path.c_str());
-    if (pal::directory_exists(sdk_path))
-    {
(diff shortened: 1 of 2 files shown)


johnbeisner:
@steveharter 
@wtgodbe 
Can someone take a look at this PR please.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -796,7 +796,8 @@ bool fx_muxer_t::resolve_sdk_dotnet_path(const pal::string_t& own_dir, const pal
         }
     }
 
-    if (!cli_version.empty())
+    pal::string_t neg_version = _X("|-1.-1.-1|");
```
steveharter: Where does the pipe `|` come from? Special character?

src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
```diff
@@ -190,6 +190,55 @@ public void SdkLookup_Global_Json_Patch_Rollup()
...
+                .Environment(s_DefaultEnvironment)
+                .EnvironmentVariable("DOTNET_MULTILEVEL_LOOKUP", "0")
+                .CaptureStdOut()
+                .CaptureStdErr()
+                .Execute()
+                .Should()
+                .Pass()
+                .And
+                .HaveStdErrContaining(Path.Combine(_exeSelectedMessage, "9999.0.4", _dotnetSdkDllMessageTerminator));
+        }
```
steveharter: No DeleteAvailableSdkVersions?

Note I suspect tests can collide with each other, we should either make sure they have unique versions (across tests) or run the tests in series. I was planning on investigating this.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
```diff
@@ -279,9 +327,6 @@ public void SdkLookup_Must_Pick_The_Highest_Semantic_Version()
                 .Pass()
                 .And
                 .HaveStdErrContaining(Path.Combine(_exeSelectedMessage, "10000.0.0", _dotnetSdkDllMessageTerminator));
-
-            // Remove dummy folders from user dir
-            DeleteAvailableSdkVersions(_userSdkBaseDir, "9999.0.2");
```
steveharter: Causing a random issue? I notice there are other sdk versions added for this test as well that are not removed (see above on comment on possible colliding)

src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
johnbeisner: Test directories are isolated. Each test creates a numbered directory [0,1,2 in this particular case corresponding to the 3 'facts' in ''GivenThatICareAboutMultilevelSDKLookup'] with "cmd", "exe", and "user" as subdirectories - no collisions.

johnbeisner: So 'DeleteAvailableSdkVersions' might be useful if we wanted to test the de-installation of an SDK version.

steveharter: OK I randomly get exceptions regarding unable to delete directory...

steveharter: Tests running simultaneously could get the same ID (numbered directory) in a race condition. Not sure if that is the reason for the random failures I get.

steveharter (approved):

src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
johnbeisner: "Tests running simultaneously could get the same ID" - if this were the case, the tests are invalid as they would mix the versions from the tests running simultaneously.

PR #2798, opened 2017-07-06
Fix appx ordering so that the signed appx is included in the nupkg

chcosta:
Fixes #1110

Validated with local builds, private "official builds" producing signed appx package, and @nattress 's package scrutiny.

/cc @nattress @zamont @joshfree


chcosta:
Looks like I caused some of the logic to leak across OS's.  I'll take a look at fixing this.

PR #2800, opened 2017-07-07
Fix framework assembly deployment for CoreCLR

nattress:
Before NS2.0, the framework assemblies were drawn from a variety of
individual Nuget packages. Filtering was performed to not copy a Nuget
package's contents if it was the CoreCLR runtime package, since that is
provided by an AppX package dependency. The test for "is this a CoreCLR
runtime package" was the presence of mscorlib.[ni.]dll. With NS2.0,
mscorlib is just a type forwarder in the main framework assembly payload
so the filter broke. Instead, use System.Private.CoreLib.[ni.]dll which is
now present in the CoreCLR runtime package. CoreCLR NS2.0 runtime Nuget
packages will similarly contain System.Private.CoreLib.dll.


nattress:
@sergiy-k @chcosta PTAL

PR #2801, opened 2017-07-07
Using new version of WCF in the UWP metapackage

joperezr:
@dagood @zhenlan @ericstj 

Adding new uap implementation of WCF to the metapackage.


dagood:
While you're in this file, could you comment out the WCF `RemoteDependencyBuildInfo` too, so that other auto-upgrades start working again? Right now the auto-upgrade build tries to create a WCF build-info and fails because the commit hash is commented out.

joperezr:
done.

PR #2802, opened 2017-07-07
Build / sign appx separately so that you can include other signed binaries

chcosta:
Appx consumes uwphost and other binaries, we need to ensure those are built / signed first, then build the appx, sign it, then build packages.  This updates Windows and Windows (arm) definitions to build appx separately.

Validated that uwphost and appx are signed in private official Windows (x64 and arm) builds.

### DIFF
--- a/buildpipeline/Core-Setup-Windows-BT.json
+++ b/buildpipeline/Core-Setup-Windows-BT.json
@@ -15,8 +15,8 @@
         "scriptType": "inlineScript",
         "scriptName": "",
         "arguments": "-path $(PB_SourcesDirectory) -rootPath $(Build.SourcesDirectory)",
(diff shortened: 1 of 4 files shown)


buildpipeline/Core-Setup-Windows-BT.json
```diff
@@ -635,8 +689,7 @@
       "value": "dn-bot"
     },
     "PB_VsoPassword": {
-      "value": null,
-      "isSecret": true
+      "value": "cwzvo2apfo6wdij6q3vyruzfmbbcc2wkyus6widyltfkev7vumfa"
```
nattress: Is this leaking a password?

chcosta: checking

dagood: Thanks @nattress, revoked and cycled.

PR #2814, opened 2017-07-11
RTM branding

rakeshsinghranchi:


weshaggard:
What about removing the version number? Does this change handle that?

rakeshsinghranchi:
@weshaggard 

This change just takes care of RTM branding in setup UI . 
Package name has version and prereleaselabel that is tied with StabilizePackageVersion value .
https://github.com/dotnet/core-setup/blob/release/2.0.0/dir.props#L76

weshaggard:
OK if this is the only change let me roll it into my stabilization PR I will be submitting. So I'm going to close this PR.

PR #2815, opened 2017-07-12
Stabilize packages for .NET Core 2.0 release

weshaggard:
FYI @Petermarcu, @rakeshsinghranchi 

This PR will not pass CI until the other repo's are stable and we have the auto-update PR into this repo. I tested the build locally with some private built stable packages.


weshaggard:
Ignoring tizen legs as they don't have stable packages yet.

PR #2824, opened 2017-07-12
Adding multi-level SDK look-up tests: SDK Muxer

ghost:
When the 'current' and 'user' directories were removed from consideration for the SDK Muxer; just the 'exe' directory was left as a viable directory to search for an SDK - therefore, SDK multi-level lookups could not be tested. Adding a faux 'global' directory so we can properly test multi-level SDK look-ups.

@dotnet/dotnet-cli

### DIFF
--- a/src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using Microsoft.DotNet.InternalAbstractions;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
```diff
@@ -23,12 +23,15 @@ public class GivenThatICareAboutMultilevelSDKLookup
         private string _currentWorkingDir;
```
wli3: Off topic discussion, the name of the file is hilarious. "Given" part should be an abbreviation of the test set up. In this case, Given mockDirsOfDifferentLevelAndTestProject

src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
```diff
@@ -95,6 +101,86 @@ public GivenThatICareAboutMultilevelSDKLookup()
...
+
+            // Add some dummy versions
+            AddAvailableSdkVersions(_globalSdkBaseDir, "9999.0.1");
+
+            // Specified CLI version: none
+            // CWD: empty
+            // User: empty
+            // Exe: 9999.0.0, 9999.0.0-dummy
+            // global: 9999.0.0, 9999.0.1
+            // Expected: 9999.0.1 from global dir
```
steveharter: Won't it look in Exe location before global?

johnbeisner:
The faux 'global' directory is not working as I expected; closing the PR.

PR #2826, opened 2017-07-13
[release/1.1] Add AppleCrypto shim as a direct NETCore.App dependency for servicing

bartonjs:
runtime.native.System.Security.Cryptography.Apple is an indirect
dependency, so building new versions of it out of CoreFX doesn't result
in the tooling pushing it to a higher minimum version when we service it.

This change adds the oldest 4.3.1-servicing package as a direct dependency,
and the version will roll forward via the normal upgrade-to-newer-builds
process.


pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -42,6 +42,7 @@
     "System.Threading.ThreadPool": "4.3.0",
     "Microsoft.DiaSymReader.Native": "1.4.1",
     "Libuv": "1.9.1",
+    "runtime.native.System.Security.Cryptography.Apple": "4.3.1-servicing-25511-01",
```
bartonjs: I notice that System.Net.Http and System.Diagnostics.DiagnosticSource are both frozen at an older build number.  Microsoft.NETCore.Runtime.CoreCLR was the last version bump (several days ago), but this package didn't exist at 25505-01.

Do I need to wait until we re-baseline 1.1 on 25511-01 or higher?

weshaggard: Those look like the slipped in after our last release. They should be marked as stable. Can you please update them to remove the pre-release identifier for those. Essentially revert https://github.com/dotnet/core-setup/commit/225035b171bfa82816be22d9d5f7364e7e34d375.

weshaggard:
test Ubuntu16.04 x64 Release Build

PR #2827, opened 2017-07-13
Only symbol-sign and index symbol packages

dagood:
Change the glob pattern to only match `*.symbols.nupkg`, and run EmbedIndex once per symbols.nupkg to avoid its directory-listing behavior.


publish/publish.proj
```diff
@@ -162,7 +162,7 @@
 	       copying them to the indexed directory -->
       <PublishDirectory Condition="'$(WindowsSdkDir)' == ''">$(DownloadDirectory)</PublishDirectory>
       <!-- list of packages that we want to embed symbol signatures in -->
-      <SymbolPackagesToPublishGlob Condition="'$(SymbolPackagesToPublishGlob)' == ''">$(DownloadDirectory)/**/*.nupkg</SymbolPackagesToPublishGlob>
+      <SymbolPackagesToPublishGlob Condition="'$(SymbolPackagesToPublishGlob)' == ''">$(DownloadDirectory)/**/*.symbols.nupkg</SymbolPackagesToPublishGlob>
```
dagood: This is a property, not an item like I assumed. Fixing.

weshaggard: Does anything else use this property that might be affected?

dagood: In BuildTools, this is used to decide which zips to make signed catalogs for. Nowhere else, according to searches.

weshaggard: Why not just change it to an item group directly in the target below then?

publish/publish.proj
```diff
@@ -182,12 +182,20 @@
   </Target>
 
   <Target Name="SignSymbolPackages" DependsOnTargets="InjectSignedSymbolCatalogIntoSymbolPackages" Condition="'$(WindowsSdkDir)' != ''">
-    <PropertyGroup>
-      <PowerShellExe Condition=" '$(PowerShellExe)'=='' ">powershell.exe</PowerShellExe>
-      <EmbedIndexScriptLocation Condition=" '$(EmbedIndexScriptLocation)'=='' ">$(VsoToolsDir)\scripts\DotNet-Trusted-Publish\Embed-Index.ps1</EmbedIndexScriptLocation>
-    </PropertyGroup>
+    <ItemGroup>
+      <SymbolPackagesToEmbedIndex Include="$(SymbolPackagesToPublishGlob)" />
```
weshaggard: I would move the glob directly here, esspecially now that you are doing some filtering that depends on the same paths below with NormalPackagesToNotEmbedIndex.

dagood: I'll look at moving them into the same target, but I think they'll both need to be here because the symbol package glob is used by the BuildTools targets.

dagood:
PTAL, I think it's in a reasonable state now. I've only been doing local testing of `SignSymbolPackages` and its dependencies, however--I would appreciate help with broader-scope tests.

publish/publish.proj
```diff
@@ -181,13 +179,32 @@
...
+
+  <Target Name="SignSymbolPackages"
+          DependsOnTargets="FindDownloadedPackagesForSigning;InjectSignedSymbolCatalogIntoSymbolPackages"
+          Condition="'$(WindowsSdkDir)' != ''">
+    <ItemGroup>
+      <EmbedIndexArgPairs Include="@(SymbolPackagesToEmbedIndex -> '%(Identity) $(PublishDirectory)%(Filename)%(Extension)')" />
+    </ItemGroup>
+
+    <!-- Inline the EmbedIndex.ps1 script used in CoreFX and CoreCLR: directly call EmbedIndex. -->
+    <Exec Command="$(DotNetToolDir)\dotnet $(EmbedIndexToolDir)\tools\EmbedIndex.dll %(EmbedIndexArgPairs.Identity)" />
```
crummel: I think you'll need changes in the Publish pipeline json to pass `DotNetToolDir` and `EmbedIndexToolDir`.  You should be able to reuse `DotnetCliPath` from BuildTools instead of `DotNetToolDir`.

crummel: Currently we pass in `VsoToolsDir` so you can probably replace that with `EmbedIndexToolDir`.

dagood: Hmm, my logic is that `EmbedIndex.ps1` takes `EmbedIndexToolDir` and `DotNetToolDir` from env, so msbuild should have it in its env, too.

I'm not comfortable making the change to use Core-Setup's CLI rather than the downloaded one due to potential version differences. (In the future, when EmbedIndex becomes a real part of buildtools, this won't be an issue.)

dagood: I can go ahead and pass them explicitly--no real harm in that.

crummel: Ah, that makes sense on CLI.  Sounds good on passing them in - I think we've been preferring that to depending on the environment more lately anyway.  It's generally easier to see what's going on too.

publish/publish.proj
dagood: Sorry, missed this comment. Currently, it has to be a property because BuildTools expects it to be to run some of its targets. I don't think I properly anticipated how Core-Setup was going to use this functionality.

dagood:
@weshaggard I couldn't implement your suggestion, but I think this is somewhat better until we get buildtools improvements. Does this look ok to merge?

publish/publish.proj
weshaggard: We should definitely avoid depending on env variables but aren't these setup from the root dir.props? Why wouldn't we have them already defined?

weshaggard:
>@weshaggard I couldn't implement your suggestion, but I think this is somewhat better until we get buildtools improvements. Does this look ok to merge?

Yeah that's fine.

publish/publish.proj
dagood: They are set as VSTS vars using VSTS logging commands in `Fetch-Tool.ps1` (from the BuildPipeline repo), then they're available in build definition steps and injected into the environment for later steps. This same kind of thing happens in CoreFX, except that the `Fetch-Tool.ps1` output vars are used in other build definition steps rather than inside msbuild. Adding the `/p:Foo=$(Foo)` doesn't change much, but at least it's an explicit link.

This will all get better and less confusing when EmbedIndex is shipped with BuildTools rather than downloaded by some external script.

dagood:
Thanks, merging, and I'll keep an eye on it.

publish/publish.proj
weshaggard: I think I'm still missing something. Why wouldn't the publish.proj not just use the CLI that we initialize via init-tools? Or is it that the publish step doesn't do the normal setup, which I could believe.

dagood: It could use the init-toolsed CLI, and that would probably work just fine. I just don't want to needlessly rock the boat with this change: `Fetch-Tool.ps1` already gets its own CLI that historically we've always run EmbedIndex on. (And it will get this extra CLI whether we use it or not.) I want to just leave this alone for now until we can trim out all this stuff when shipping EmbedIndex via BuildTools.

dagood:
Looks like this worked: a build I kicked off was successful, and https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.win-x64.Microsoft.NETCore.DotNetHostPolicy/2.1.0-preview2-25513-03 is showing good package contents. The logs also showed that EmbedIndex was only running on symbol packages.

PR #2854, opened 2017-07-18
Adding a mutex when selecting a directory index: SDK Muxer tests

ghost:
Steve Harter had stated that it is possible for a multi-threaded test-run to pick the same directory ID for more than one test instance. Therefore, we should lock the code that chooses the ID down to one thread.

Refer to line nos: 5, 13, & 446-463

@dotnet/dotnet-cli


johnbeisner:
@wtgodbe 
@steveharter

PR #2856, opened 2017-07-18
Fix local build for UWPHost 

zamont:
Fix this by adding header files from CoreCLR repo. This is where mscoree.h originated from, so additional header files should be fine.


zamont:
PTAL @chcosta @nattress @eerhardt

PR #2877, opened 2017-07-20
Update to readme to have 2.0.0 links

weshaggard:
@Petermarcu I'm updating the table to just have 2.0.0 links but I've not been able to verify all the links yet until our stable build is out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,33 +53,33 @@ Daily Builds
 [win-x64-badge-2.0.X]: https://dotnetcli.blob.core.windows.net/dotnet/Runtime/release/2.0.0/sharedfx_win-x64_Release_version_badge.svg
 [win-x64-installer-2.0.X]: https://dotnetcli.blob.core.windows.net/dotnet/Runtime/release/2.0.0/dotnet-runtime-latest-win-x64.exe
 [win-x64-zip-2.0.X]: https://dotnetcli.blob.core.windows.net/dotnet/Runtime/release/2.0.0/dotnet-runtime-latest-win-x64.zip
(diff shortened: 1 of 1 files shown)


weshaggard:
@chcosta while updating the links in the readme.txt I noticed we no longer have zip/tar.gz files that contain the symbols for the target. Is that expected?

@leecow @dagood do we do anything with those symbol zips when we do a release?

dagood:
For releases, I gather up all the `*.symbols.nupkg`s and put them in a new Core-Setup zip for DLC. I don't know anywhere the build output symbol zips/tar.gz's are/were used.

weshaggard:
>For releases, I gather up all the *.symbols.nupkgs and put them in a new Core-Setup zip for DLC. I don't know anywhere the build output symbol zips/tar.gz's are/were used.

I updated the readme links to point the symbols.nupkg packages.

weshaggard:
Actually @chcosta just pointed the symbol zip out to me in our blob storage. I just overlooked them. I'm still not entirely sure what they are used for however but they are a little easier to download if you just want the symbols so I will redirect the links in the readme to those.

leecow:
@weshaggard - we do post them to the download center and provide links in release docs.

dagood:
> [leecow] we do post them to the download center and provide links in release docs.

Just to make sure we're on the same page, are the ones here a good example to look at: https://github.com/dotnet/core/blob/master/release-notes/download-archives/2.0.0-preview2-download.md? Those are assembled from the CoreCLR (Runtime) and CoreFX (Shared Framework) symbol packages. There's also a zip assembled from Core-Setup symbol packages, but it's not linked there.

Those are all made by zipping up a bunch of symbol packages, but it's not related to the symbol zip. The symbol packages are based on the NuGet format, and the symbol zip has all the symbol files of a Core-Setup build in a flat dir (and none of the runnable binaries).

PR #2884, opened 2017-07-22
Make sure to use stable package when restoring

weshaggard:
This fixes the invalid package dependency to the prerelease package
version instead of the stable package version

Fixes the error:
error NU1603: Microsoft.Extensions.DependencyModel 2.0.0 depends on Microsoft.DotNet.PlatformAbstractions (>= 2.0.0-servicing-25521-01) but Microsoft.DotNet.PlatformAbstractions 2.0.0-servicing-25521-01 was not found. An approximate best match of Microsoft.DotNet.PlatformAbstractions 2.0.0 was resolved.

Without this Microsoft.Extensions.DependencyModel ends up with a prerelease package depdency on Microsoft.Extensions.PlatformAbstractions.

cc @ericstj @Petermarcu @eerhardt

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -140,7 +140,7 @@
     <SdkRestoreProjects Include="$(MSBuildThisFileDirectory)src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
     <SdkRestoreProjects Include="$(MSBuildThisFileDirectory)src\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj">
       <!-- Workaround https://github.com/NuGet/Home/issues/4337 -->
-      <ExtraRestoreArgs>/p:VersionSuffix=$(VersionSuffix)</ExtraRestoreArgs>
+      <ExtraRestoreArgs Condition="'$(StabilizePackageVersion)' != 'true'">/p:VersionSuffix=$(VersionSuffix)</ExtraRestoreArgs>
     </SdkRestoreProjects>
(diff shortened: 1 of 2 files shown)


dir.props
```diff
@@ -140,7 +140,7 @@
     <SdkRestoreProjects Include="$(MSBuildThisFileDirectory)src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
     <SdkRestoreProjects Include="$(MSBuildThisFileDirectory)src\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj">
       <!-- Workaround https://github.com/NuGet/Home/issues/4337 -->
-      <ExtraRestoreArgs>/p:VersionSuffix=$(VersionSuffix)</ExtraRestoreArgs>
+      <ExtraRestoreArgs Condition="'$(StabilizePackageVersion)' != 'true'">/p:VersionSuffix=$(VersionSuffix)</ExtraRestoreArgs>
```
eerhardt: Kind of a silly question, but shouldn't VersionSuffix be empty when StabilizePackageVersion is set to true?

weshaggard: We could set it to empty but I guess I've been working under the assumption it is always empty unless we set it explicitly. Although I get your point as something could end up setting it somewhere else.

weshaggard:
> I would have thought VersionSuffix would be empty, thus the RestoreArgs should set VersionSuffix to empty as well.

I get what you were saying now. I updated to do exactly that just in case there are other places that use VersionSuffix.

Issue #2876, opened 2017-07-20
Core-setup requires Windows SDK to be on the machine


maririos:
@karajas commented on [Tue Jul 18 2017](https://github.com/dotnet/core-eng/issues/1249)

Core-setup now builds the uwpshim and requires onecore.lib which is a part of the Windows SDK.

Our build machines have this dependency installed on them currently. 
/cc @shawnro 

---

@karajas commented on [Tue Jul 18 2017](https://github.com/dotnet/core-eng/issues/1249#issuecomment-316220646)

According to @weshaggard, we should look into removing this dependency as it may no longer be required.

steveharter:
@nattress PTAL

nattress:
This can be closed. We don't build uwpshim in this branch anymore.

PR #2913, opened 2017-07-27
Build UWPShim as a Store App

nattress:
Alter build of UWPShim so it is a Windows Store exe instead of a desktop command line application. This prevents a command window popping up when running CoreCLR apps in VS.
This requires splitting the uwp cmake build into two pieces: a desktop piece for CopyWin32Resources and UWPHost, and a Store app piece for UWPShim
Statically link vcruntime.lib into CopyWin32Resources and UWPHost to avoid a runtime requirement that vcruntime140.dll is present.


nattress:
@sergiy-k PTAL

nattress:
This fixes https://github.com/dotnet/core-setup/issues/2876

PR #2942, opened 2017-07-31
Enable RHEL6 and CentOS 6 RID detection in build

janvorli:
This change adds RHEL6 and CentOS 6 RID detection to src/corehost/build.sh.
These distros don't have the /etc/os-release file and so we need to use another
source - the /etc/redhat-release file.
It is a clone of similar change merged in for CoreCLR.

### DIFF
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -14,35 +14,38 @@ init_rid_plat()
         else
             if [ -e $ROOTFS_DIR/etc/os-release ]; then
                 source $ROOTFS_DIR/etc/os-release
-                export __rid_plat="$ID.$VERSION_ID"
+                __rid_plat="$ID.$VERSION_ID"
             fi
             echo "__rid_plat is $__rid_plat"
         fi
     else
+        __rid_plat=""
         if [ -e /etc/os-release ]; then
             source /etc/os-release
-
             if [[ "$ID" == "rhel" && $VERSION_ID = 7* ]]; then
-                export __rid_plat="rhel.7"
+                __rid_plat="rhel.7"
             elif [[ "$ID" == "centos" && "$VERSION_ID" = "7" ]]; then
-                export __rid_plat="rhel.7"
+                __rid_plat="rhel.7"
             else
-                export __rid_plat="$ID.$VERSION_ID"
+                __rid_plat="$ID.$VERSION_ID"
             fi
-        else
-            export __rid_plat=
+        elif [ -e /etc/redhat-release ]; then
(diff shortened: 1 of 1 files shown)


src/corehost/build.sh
```diff
@@ -19,18 +19,21 @@ init_rid_plat()
...
                 export __rid_plat="rhel.7"
             else
                 export __rid_plat="$ID.$VERSION_ID"
             fi
-        else
-            export __rid_plat=
+        elif [ -e /etc/redhat-release ]; then
+            local redhatRelease=$(</etc/redhat-release)
+            if [[ $redhatRelease == "CentOS release 6."* || $redhatRelease == "Red Hat Enterprise Linux Server release 6."* ]]; then
+               __rid_plat="rhel.6"
```
weshaggard: Don't you still need to export the __rid_plat variable?

janvorli: There is no need to export it, the export is needed only if we need to pass it to a child process, which we don't do. I have somehow missed the fact that the rest of the function uses the export, so I'll unify it.
.

src/corehost/build.sh
```diff
@@ -14,35 +14,38 @@ init_rid_plat()
...
             elif [[ "$ID" == "centos" && "$VERSION_ID" = "7" ]]; then
-                export __rid_plat="rhel.7"
+                __rid_plat="rhel.7"
             else
-                export __rid_plat="$ID.$VERSION_ID"
+                __rid_plat="$ID.$VERSION_ID"
+            fi
+        elif [ -e /etc/redhat-release ]; then
+            local redhatRelease=$(</etc/redhat-release)
+            if [[ $redhatRelease == "CentOS release 6."* || $redhatRelease == "Red Hat Enterprise Linux Server release 6."* ]]; then
```
weshaggard: This seems very fragile but I guess we will have to deal with it if we don't have any other way.

src/corehost/build.sh
weshaggard: sounds good.

src/corehost/build.sh
janvorli: I know it doesn't look nice, but it is described in the RedHat doc as a way to detect RedHat version:
https://access.redhat.com/articles/3078

weshaggard: Given we are only detecting redhat here I wonder if we should use this for both 6 and 7.

janvorli: I prefer it this way as a fallback for the now standard /etc/os-release.

PR #2961, opened 2017-08-02
Enable building debian9 packages

rakeshsinghranchi:
This work enables building Debian 9 packages. This is the first step towards building this new distro specific package. This PR uses Debian 8.2 docker build image with static set of pre-requisite package dependency required for Debian 9.

### DIFF
--- a/buildpipeline/Core-Setup-Linux-BT.json
+++ b/buildpipeline/Core-Setup-Linux-BT.json
@@ -526,6 +526,96 @@
         "failOnStandardError": "false"
       }
     },
+	{
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Initialize docker - Debian 9",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "$(PB_DockerHost_ToolsDirectory)/scripts/docker/init-docker.sh",
+        "arguments": "$(DockerImageName_Debian9)",
+        "workingFolder": "$(PB_DockerHost_Sandbox)",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Init tools - Debian 9 container",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "docker",
+        "arguments": "run --rm $(DockerCommonRunArgs_Debian9) /bin/bash -c \"HOME=$(PB_GitDirectory); git clean -X -d -f; $(PB_GitDirectory)/init-tools.sh\"",
+        "workingFolder": "",
+        "failOnStandardError": "false"
+      }
+    },
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Build traversal build dependencies - Debian 9",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "d9bafed4-0b18-4f58-968d-86655b4d2ce9",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "filename": "docker",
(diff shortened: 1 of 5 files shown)


rakeshsinghranchi:
cc @Petermarcu  , @karajas . Can you take a look at the PR ? Thanks.

Petermarcu:
@weshaggard

src/pkg/packaging/deb/dotnet-sharedframework-debian9_config.json
```diff
@@ -0,0 +1,41 @@
...
+
+    "package_name": "%SHARED_FRAMEWORK_DEBIAN_PACKAGE_NAME%",
+    "install_root": "/usr/share/dotnet",
+
+    "short_description": "%SHARED_FRAMEWORK_BRAND_NAME% %SHARED_FRAMEWORK_NUGET_NAME% %SHARED_FRAMEWORK_NUGET_VERSION%",
+    "long_description": ".NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform and is supported by Microsoft. We hope you enjoy using it! If you do, please consider joining the active community of developers that are contributing to the project on GitHub (https://github.com/dotnet/core). We happily accept issues and PRs.",
+    "homepage": "https://dotnet.github.io",
+
+    "release":{
+        "package_version":"1.0.0.0",
```
weshaggard: Is this the correct version? Should it align with our product version?

buildpipeline/Core-Setup-Linux-BT.json
```diff
@@ -526,6 +526,96 @@
         "failOnStandardError": "false"
       }
     },
+	{
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Initialize docker - Debian 9",
```
weshaggard: It looks like you are doing another full clean build for Debian 9. Is that expected? I thought we were only doing the package build but not rebuilding everything. If we plan to rebuild everything why not have a separate build leg so we can do it in parallel?

karajas:
I believe another issue mentioned is that we don't have a cli to perform init-tools on debian 9?

buildpipeline/Core-Setup-Linux-BT.json
rakeshsinghranchi: No,we are not rebuilding everything . Like other native installers, we consume prebuilt Linux-x64 portable binaries by volume mapping the prebuilt binaries and build native installers for each distro using corresponding docker container.

weshaggard:
```
"arguments": "run --rm $(DockerCommonRunArgs_Debian9) /bin/bash -c \"HOME=$(PB_GitDirectory); git clean -X -d -f; $(PB_GitDirectory)/init-tools.sh\"",
```

The first step you run is `git clean` which to me appears to be cleaning out the entire repo and then rebuilding. Perhaps I've missed where you are creating another clone but it certainly looks like you are using the same clone.

src/pkg/packaging/deb/dotnet-sharedframework-debian9_config.json
rakeshsinghranchi: This is the default value that always gets set to the correct product version during the package build step. see https://github.com/dotnet/core-setup/blob/master/src/pkg/packaging/deb/package.targets#L178 .

buildpipeline/Core-Setup-Linux-BT.json
rakeshsinghranchi: Look at DockerCommonRunArgs_Debian9 build variable. It does the volume mapping to prebuilt sharedframeworkpublish folder from binary build. Also, distro specific build steps do not contain binary build step. for e.g look at https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=875149

weshaggard: OK thanks for the information.

src/pkg/packaging/deb/dotnet-sharedframework-debian9_config.json
weshaggard: cool.

buildpipeline/Core-Setup-Linux-BT.json
```diff
@@ -916,6 +1010,18 @@
     "DockerCommonRunArgs_Debian8": {
       "value": "--name $(PB_DockerContainerName)$(DockerTag_Debian8) -v \"$(PB_SourcesDirectory):$(PB_GitDirectory)\" -v $(Build.StagingDirectory)/sharedFrameworkPublish/:/root/sharedFrameworkPublish/ -w=\"$(PB_GitDirectory)\" $(DockerImageName_Debian8)"
     },
+    "DockerTag_Debian9": {
+      "value": "debian-8.2-debpkg-9f87c3c-20173003023006"
```
eerhardt: @rakeshsinghranchi - why are we using a `debian-8.2` docker image to build the `debian9` deb package?

Looking at https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/tags/, I don't see any debian 9 images.  Is this expected?  I think we should be using debian 9 machines here.

/cc @MichaelSimons

PR #2977, opened 2017-08-04
[RHEL6] Enable build from latest corefx release/2.0.0 on RHEL 6

janvorli:
This is a change to make sure that when building  for RHEL 6 and using current release/2.0.0 corefx branch, the build picks the right package from corefx.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -17,7 +17,7 @@
 
   <PropertyGroup>
     <PlatformPackageVersion>2.0.0</PlatformPackageVersion>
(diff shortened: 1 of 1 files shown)


weshaggard:
@janvorli this change appears to be missing the update to the OfficialBuildRid list in https://github.com/dotnet/core-setup/blob/release/2.0.0/src/pkg/projects/netcoreappRIDs.props. Once we get a corefx and coreclr build working we likely need to add that here to get core-setup to produce the correct assets.

janvorli:
@weshaggard I've missed that core-setup also contains the official rids, I am creating new PRs for adding them.

weshaggard:
This PR is no longer needed so closing.

PR #2993, opened 2017-08-07
[release/2.0.0] Only publish when destination deliberately specified 

dagood:
See https://github.com/dotnet/core-setup/pull/1869 (earlier similar work for internal builds of release/1.1.0).

Allows configuration of which publish steps are performed by changing Pipebuild global parameters. This will let us set up a new internal build that only publishes to (different) blob storage--official builds will continue unchanged.

As of submitting, I haven't tested this yet. We should wait to merge until we start making 2.0.0 servicing builds, but I wanted early review.

/cc @weshaggard @dleeapho

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -11,9 +11,13 @@
   
   <Target Name="PublishFinalOutput"
           Condition="'$(Finalize)' == 'true'"
-          DependsOnTargets="PublishCoreHostPackagesToFeed;FinalizeBuildInAzure;UpdatePublishedVersions" />
-          
-  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+          DependsOnTargets="PublishCoreHostPackagesToFeed;FinalizeBuildInAzure;UpdateVersionsRepo" />
+
+  <Target Name="UpdateVersionsRepo"
+          Condition="'$(GitHubAuthToken)' != ''"
+          DependsOnTargets="UpdatePublishedVersions" />
+
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdateVersionsRepo" >
     <ItemGroup>
       <PackagesToShip Include="$(BinDir)/ForPublishing/*.nupkg" Exclude="$(BinDir)/ForPublishing/*.symbols.nupkg" />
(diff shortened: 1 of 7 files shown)


publish/publish.proj
```diff
@@ -11,9 +11,13 @@
   
   <Target Name="PublishFinalOutput"
           Condition="'$(Finalize)' == 'true'"
-          DependsOnTargets="PublishCoreHostPackagesToFeed;FinalizeBuildInAzure;UpdatePublishedVersions" />
-          
-  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
+          DependsOnTargets="PublishCoreHostPackagesToFeed;FinalizeBuildInAzure;UpdateVersionsRepo" />
+
+  <Target Name="UpdateVersionsRepo"
+          Condition="'$(GitHubAuthToken)' != ''"
```
crummel: Do we want to standardize on either triggering this skip on one particular missing variable (`GitHubAuthToken`) or skipping on any missing variable (`DebRepoId` etc below)?  `GitHubUser` and `GitHubEmail` are also required for this, correct?

publish/publish.proj
```diff
@@ -185,8 +193,7 @@
   </Target>
   
   <Target Name="PublishDebToolPackageToFeed"
-          Condition="'$(PublishDebToolToFeed)' == 'true'">
-    <Error Condition="'$(CliNuGetFeedUrl)' ==''" Text="Missing required property CliNuGetFeedUrl" />
+          Condition="'$(PublishDebToolToFeed)' == 'true' AND '$(CliNuGetFeedUrl)' != ''">
```
crummel: Should we promote `PublishDebToolToFeed` and make a `PB_PublishDebToolToFeed` that gets passed in since this already has a mechanism to skip?  Not opposed to doing it this way either.

publish/publish.proj
dagood: Right now, I just copied what I did for 1.0.0/1.1.0.

I think it makes sense to skip based on auth key/token only and error on other blanks, because in some cases (e.g. dotnet/versions repo name) there are defaults--and api keys never have defaults. Erroring out rather than skipping when the user missed some non-obvious config sounds easier to figure out to me.

publish/publish.proj
dagood: I strongly dislike having config plus a boolean that turns it on/off. I think it encourages people to leave sensitive info in the build but set up variables that are intended to disable using them, which leads to bad dev builds. With removing the info being the switch, if something bad happens (e.g. building a commit that doesn't support disabling the publish step), at worst you get a failed build.

Issue #2820, opened 2017-07-12
Microsoft.Build.Net.CoreRuntimeTask.dll needs strong-name signing


nattress:
When multiple projects targeting different versions of the UWP tooling are present (either in the same solution in VS, or compiled in two different VS instances in short succession so they hit the same MSBuild worker node), only one copy of Microsoft.Build.Net.CoreRuntimeTask.dll will be loaded into MSBuild's process. They have different paths and different versions.  This is caused by https://github.com/Microsoft/msbuild/issues/663

This can cause a build failure if the task name / parameters have changed, or silent incorrect behavior.

To ensure the correct copy gets loaded, all copies of Microsoft.Build.Net.CoreRuntimeTask.dll need strong name signing.

PR #2994, opened 2017-08-07
Strongname sign Microsoft.Build.Net.CoreRuntimeTask

nattress:
Fixes https://github.com/dotnet/core-setup/issues/2820


nattress:
@chcosta PTAL - I can't really test this outside of the VSTS build and am unsure if it's sufficient to get the binary strongname signed. Is MSFT key appropriate here? I wasn't sure whether to use that or Open.

PR #2996, opened 2017-08-07
Incrememnt package version to 2.0.1

wtgodbe:
CC @weshaggard does anything else need to be done here? Looks like the PreRelaseLabel was already set to "servicing"

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -13,8 +13,8 @@
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>0</PatchVersion>
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
+    <PatchVersion>1</PatchVersion>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
     <PreReleaseLabel>servicing</PreReleaseLabel>
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
     <ReleaseBrandSuffix></ReleaseBrandSuffix>
--- a/src/pkg/packaging/dir.proj
+++ b/src/pkg/packaging/dir.proj
@@ -164,8 +164,12 @@
 
   <Target Name="GenerateCombinedInstallers" DependsOnTargets="InitPackage;GenerateBundles" />
 
+  <PropertyGroup>
+    <BuildAllPackages>false</BuildAllPackages>
+  </PropertyGroup>
+
   <Target Name="GenerateNugetPackages" DependsOnTargets="InitPackage" Condition="'$(UsePrebuiltPortableBinariesForInstallers)' == 'false'">
-    <ItemGroup>
+    <ItemGroup Condition="'$(BuildAllPackages)' == 'true'">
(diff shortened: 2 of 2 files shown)


dir.props
```diff
@@ -13,8 +13,8 @@
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>0</PatchVersion>
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
+    <PatchVersion>1</PatchVersion>
```
weshaggard: This is the first release we have core-setup on BuildTools so I'm not sure if this is the only change needed or not. Can you please do a full build and look at the versions of the assets to make sure they all match?

wtgodbe: Looks like DependencyModel & PlatformAbstractions are still 2.0.0-servicing, but everything else is 2.0.1-servicing. I'll see if I can figure out where they have their package version defined.

weshaggard: Actually for those 2 packages I would like to do something similar to corefx and disable the packaging of them instead of bumping them. I would prefer we not ship them with every servicing release.

src/managed/dir.proj
```diff
@@ -19,6 +19,10 @@
       <BuildArgs>$(BuildArgs) /p:BaseOutputPath=$(IntermediateOutputForPackaging)</BuildArgs>
     </PropertyGroup>
 
-    <Exec Command="$(DotnetToolCommand) build $(BuildArgs) %(PackageProjects.Identity)" />
+    <PropertyGroup>
+      <PackageManagedComponents>false</PackageManagedComponents>
+    </PropertyGroup>
+
+    <Exec Command="$(DotnetToolCommand) build $(BuildArgs) %(PackageProjects.Identity)" Condition="'$(PackageManagedComponents)' == 'true'"/>
```
wtgodbe: @weshaggard I believe this should stop them from being generated, running a build now to confirm

src/pkg/packaging/dir.proj
```diff
@@ -164,7 +164,11 @@
 
   <Target Name="GenerateCombinedInstallers" DependsOnTargets="InitPackage;GenerateBundles" />
 
-  <Target Name="GenerateNugetPackages" DependsOnTargets="InitPackage" Condition="'$(UsePrebuiltPortableBinariesForInstallers)' == 'false'">
+  <PropertyGroup>
+    <PackageManagedComponents>false</PackageManagedComponents>
+  </PropertyGroup>
+
+  <Target Name="GenerateNugetPackages" DependsOnTargets="InitPackage" Condition="'$(UsePrebuiltPortableBinariesForInstallers)' == 'false' and '$(PackageManagedComponents)' == 'true'">
```
wtgodbe: @weshaggard my last change was wrong, but this one seems to have done the trick. DependencyModel/PlatformAbstractions packages are no longer being generated, but the .DLL's are still present in Bin.

weshaggard: Instead of going this route can you instead put a condition on the ItemGroup for the PackageProjects below similar to what we are going to do in corefx and only include them if the BuildAllPackages property is true.

weshaggard: Yes it is fine to have them build to the bin folder we just don't want to package and ship the packages unless we need to fix something in the library.

PR #3003, opened 2017-08-08
Move sign legs to PROD

jcagme:
Current sign build legs use the test pool. Moving them to use PROD's

### DIFF
--- a/buildpipeline/Core-Setup-Signing-Windows-x64.json
+++ b/buildpipeline/Core-Setup-Signing-Windows-x64.json
@@ -516,10 +516,10 @@
   "queue": {
     "pool": {
       "id": 83,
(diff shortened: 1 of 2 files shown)


buildpipeline/Core-Setup-Signing-Windows-x64.json
```diff
@@ -516,10 +516,10 @@
   "queue": {
     "pool": {
       "id": 83,
-      "name": "DotNetCore-Test"
+      "name": "DotNet-Build"
     },
     "id": 159,
```
ravimeda: Please check if the `Id` should be updated too.

jcagme: Updated

PR #3005, opened 2017-08-08
Fixing the package name on dependencies.props so that corefx gets updated in core-setup

joperezr:
Fixing the package name on dependencies.props so that corefx gets updated in core-setup

cc: @weshaggard @joshfree @shmao @nattress @zhenlan

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -16,7 +16,8 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <CoreFxVersion>4.5.0-preview1-25608-01</CoreFxVersion>
+    <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.5.0-preview1-25531-01</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
+    <MicrosoftPrivateCoreFxUAPPackageVersion>4.5.0-preview1-25608-01</MicrosoftPrivateCoreFxUAPPackageVersion>
     <PlatformPackageVersion>2.1.0-preview1-25608-01</PlatformPackageVersion>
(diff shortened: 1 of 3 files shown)


joperezr:
cc: @dagood

dependencies.props
```diff
@@ -59,7 +59,7 @@
     <XmlUpdateStep Include="CoreFx">
       <Path>$(MSBuildThisFileFullPath)</Path>
       <ElementName>CoreFxVersion</ElementName>
-      <PackageId>Microsoft.Private.CoreFx.NETCoreApp</PackageId>
+      <PackageId>Microsoft.Private.CoreFx.UAP</PackageId>
```
dagood: I don't understand why this would change anything... aren't these packages always built with the same version?
https://github.com/dotnet/versions/blob/8d7763e82daa7cf6afd4abcb75b1f48be5e2e83d/build-info/dotnet/corefx/master/Latest_Packages.txt#L5-L6

I think `CoreFxVersion` should be renamed to something more specific if they aren't.

joperezr: Nope, in release/uwp6.0 branch we turned off all of the packages that we won't be shipping in uwp release, which includes the netcoreapp package. That is what caused the corefx package version to stop getting updated. The file you are looking at in corefx is the one from master, if you check the one from release/uwp6.0 you'll see what I mean: https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/uwp6.0/Last_Build_Packages.txt

weshaggard: Can you follow my lead that I did in the release/2.0.0 branch and rename this? https://github.com/dotnet/core-setup/blob/release/2.0.0/dependencies.props#L21. We should probably also put those changes into master.

dagood: Ah, forgot which build-info this pointed to. Makes sense, thanks.

dagood: Looks good to me!

PR #3016, opened 2017-08-10
Compile Microsoft.Build.Net.CoreRuntimeTask with MSFT public key

nattress:
Strongname signing was disabled for this assembly. Opt it into signing
with the MSFT key.

My previous attempt (https://github.com/dotnet/core-setup/commit/e46a14f5c924bfcf52536436f5386b33b347eb04) was insufficient since signing was disabled in that assembly's project file.


nattress:
@chcosta @dagood PTAL

dagood:
I don't really know this area. (I think @chcosta does though.)

PR #3017, opened 2017-08-10
Add Redhat6 to official build of core-setup

smile21prc:
Add Redhat6 to official build of core-setup. As it's not a portable build, it's parameter is different from rhel.7.2-x64.

Seems core-setup doesn't have test run specs in its json file, which is different from corefx and coreclr. If it actually has, please let me know and I'll enable its test run targeting on Redhat6 as well.

This is the coreclr effort to resolve dotnet/core-eng#1427.

For the reference, the corefx part of this change was merged into corefx release/2.0.0 as
dotnet/corefx@204cdd7

and the coreclr part is:
https://github.com/dotnet/coreclr/pull/13301

### DIFF
--- a/buildpipeline/pipeline.json
+++ b/buildpipeline/pipeline.json
@@ -31,6 +31,21 @@
           }
         },
         {
+          "Name": "Core-Setup-Linux-BT",
+          "Parameters": {
+            "PB_DistroRid": "rhel.6.9-x64",
+            "PB_DockerTag": "centos-6-783abde-20171304101322",
+            "PB_AdditionalBuildArguments":"-PortableBuild=false -strip-symbols",
+            "PB_PortableBuild": "false"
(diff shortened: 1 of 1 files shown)


buildpipeline/pipeline.json
```diff
@@ -31,6 +31,21 @@
           }
         },
         {
+          "Name": "Core-Setup-Linux-BT",
+          "Parameters": {
+            "PB_DistroRid": "rhel.6.9-x64",
```
weshaggard: I think the RID should probably be "rhel.6-x64" I don't think we have a RID for 6.9.

smile21prc: Thanks for pointing that out @weshaggard , I've renamed the PB_DistroRid value as suggested.

PR #3064, opened 2017-08-16
Fix homebrew link

richlander:
* The homebrew link in the conclusion.html is broken.
* Removes "www."
* Switched to https.
* Also noticed a commit-specific GitHub link; switched to master branch.


karelz:

Issue #2981, opened 2017-08-04
Cannot resolve assembly paths for reference compilation libraries


tuespetre:
There is no implementation of `ICompilationAssemblyResolver` that will resolve paths for a `CompilationLibrary` with a `Type` of `"reference"`, so when trying to use libraries like `Microsoft.AspNetCore.Mvc` that call `ResolveReferencePaths`, an exception is thrown:

```
System.InvalidOperationException: Can not find compilation library location for package 'MyLibrary'
```

This is a regression with 2.x, as the following used to work:

```
<Reference Include="MyLibrary">
    <HintPath>../../path/to/MyLibrary.dll</HintPath>
</Reference>
```

I have been reading everything I can find surrounding this issue and came across the following comment:

https://github.com/dotnet/sdk/issues/1213#issuecomment-302734715

However, this is not an applicable solution for projects that do in fact require `PreserveCompilationContext` to be `true`.

Here is what I came up with as a workaround, that lets us develop locally with MVC and also use the precompilation tool for deploys:

```
    public class MvcConfiguration : IDesignTimeMvcBuilderConfiguration
    {
        private class DirectReferenceAssemblyResolver : ICompilationAssemblyResolver
        {
            public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> assemblies)
            {
                if (!string.Equals(library.Type, "reference", StringComparison.OrdinalIgnoreCase))
                {
                    return false;
                }

                var paths = new List<string>();

                foreach (var assembly in library.Assemblies)
                {
                    var path = Path.Combine(ApplicationEnvironment.ApplicationBasePath, assembly);

                    if (!File.Exists(path))
                    {
                        return false;
                    }

                    paths.Add(path);
                }

                assemblies.AddRange(paths);

                return true;
            }
        }

        public void ConfigureMvc(IMvcBuilder builder)
        {
            // .NET Core SDK v1 does not pick up reference assemblies so
            // they have to be added for Razor manually. Resolved for
            // SDK v2 by https://github.com/dotnet/sdk/pull/876 OR SO WE THOUGHT
            /*builder.AddRazorOptions(razor =>
            {
                razor.AdditionalCompilationReferences.Add(
                    MetadataReference.CreateFromFile(
                        typeof(PdfHttpHandler).Assembly.Location));
            });*/

            // .NET Core SDK v2 does not resolve reference assemblies' paths
            // at all, so we have to hack around with reflection
            typeof(CompilationLibrary)
[...]

PR #3065, opened 2017-08-16
Allow AppBaseCompilation assembly resolver to resolve 'reference'

mattvperry:
Just allow the AppBaseCompilation resolver to move forward if the library type is 'reference.' This will allow it to find direct DLL references in the build output.

This fixes this issue which I just filed: dotnet/sdk#1500

This issue seems to be specifically related to VS 15.3. I wasn't able to figure out why exactly but this issue is reproducible with ASP.Net Core 1.1 and ASP.Net Core 2.0 in VS 2017 15.3 but seems to work fine in the earlier version of VS 2017.

Edit: This has nothing to do with Visual Studio. I just made the assumption based on the fact that I had recently upgraded.

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/Resolution/AppBaseCompilationAssemblyResolver.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/Resolution/AppBaseCompilationAssemblyResolver.cs
@@ -45,9 +45,10 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
                 string.Equals(library.Type, "msbuildproject", StringComparison.OrdinalIgnoreCase);
 
             var isPackage = string.Equals(library.Type, "package", StringComparison.OrdinalIgnoreCase);
+            var isReferenceAssembly = string.Equals(library.Type, "referenceassembly", StringComparison.OrdinalIgnoreCase);
             if (!isProject &&
                 !isPackage &&
-                !string.Equals(library.Type, "referenceassembly", StringComparison.OrdinalIgnoreCase) &&
+                !isReferenceAssembly &&
                 !string.Equals(library.Type, "reference", StringComparison.OrdinalIgnoreCase))
             {
                 return false;
@@ -57,7 +58,7 @@ namespace Microsoft.Extensions.DependencyModel.Resolution
             var isPublished = _fileSystem.Directory.Exists(refsPath);
 
             // Resolving reference assebmlies requires refs folder to exist
-            if (!isProject && !isPackage && !isPublished)
+            if (isReferenceAssembly && !isPublished)
             {
                 return false;
             }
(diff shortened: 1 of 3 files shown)


mattvperry:
After further investigation, the root cause of this seems to be this change here: https://github.com/dotnet/sdk/commit/e64e26cbd927148a140b8aa94920e051d8f8898e

The generated deps files now list direct DLL references with the type "reference." The `DependencyModel` project's `AppBaseCompilationAssemblyResolver` couldn't handle deps with type "reference." My PR should fix that.

nguerrera:
@eerhardt

mattvperry:
This should also address dotnet/core-setup#2981

steveharter:
@mattvperry there have been a couple people working on this besides yourself, although if you want to take it you can at this point although we would like to get it in ASAP into master so we can verify fix before likely upcoming patch release.

Please evaluate @tuespetre which has an implementation at #2981 that uses another class (DirectReferenceAssemblyResolver) which is nice because it does not pollute the existing reference-based resolver (AppBaseCompilationAssemblyResolver). I also have a private branch based on that.

Also note code review comment if deciding to keep current implementation.

src/managed/Microsoft.Extensions.DependencyModel/Resolution/AppBaseCompilationAssemblyResolver.cs
```diff
@@ -47,7 +47,8 @@ public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> ass
             var isPackage = string.Equals(library.Type, "package", StringComparison.OrdinalIgnoreCase);
             if (!isProject &&
                 !isPackage &&
-                !string.Equals(library.Type, "referenceassembly", StringComparison.OrdinalIgnoreCase))
+                !string.Equals(library.Type, "referenceassembly", StringComparison.OrdinalIgnoreCase) &&
+                !string.Equals(library.Type, "reference", StringComparison.OrdinalIgnoreCase))
```
steveharter: If you decide against a separate class then the code below this
```if (!isProject && !isPackage && !isPublished)```
will need to change to
```if (library.Type equals “referenceassembly” && !isPublished)```

**Comment by mattvperry** [comment]
@steveharter , I hadn't noticed that @tuespetre had an open suggestion which is why I tracked this down myself. 

I decided to keep my current implementation, let me explain. I agree that `AppBaseCompilationAssemblyResolver` might be getting a bit polluted but from the current layout of code it seems like these different AssemblyResolvers are describing _where_ to look for references not _what_ references to look for. This is backed up by the fact that libraries of type "referenceassembly" are actually looked for with two resolvers. Since we are actually looking for direct references in the App's Base directory it made sense to me to just add it here. If there are strong feelings otherwise, I can add the extra resolver however I feel their might be a good bit of code duplication.

**Comment by tuespetre** [comment]
I agree with @mattvperry, the only reason my workaround was separate was because it was just a workaround and I didn't want to copy-paste the whole source of `AppBaseCompilationAssemblyResolver` and then add modifications to it.

File: src/managed/Microsoft.Extensions.DependencyModel/Resolution/AppBaseCompilationAssemblyResolver.cs [hunk 2]
```diff
@@ -57,7 +58,7 @@ public bool TryResolveAssemblyPaths(CompilationLibrary library, List<string> ass
             var isPublished = _fileSystem.Directory.Exists(refsPath);
 
             // Resolving reference assebmlies requires refs folder to exist
```
steveharter: nit: can you correct the 'assebmlies' type please

PR #3076, opened 2017-08-21
[1.1.0] Remove finalize tracking for removed build legs

dagood:
See https://github.com/dotnet/core-setup/pull/3075.

> Some platforms are being checked as a finalization prereq even though the legs aren't being run. See https://github.com/dotnet/core-eng/issues/1540. Removing the checks unblocks build finalization.

### DIFF
--- a/build_projects/dotnet-host-build/PublishTargets.cs
+++ b/build_projects/dotnet-host-build/PublishTargets.cs
@@ -124,14 +124,11 @@ namespace Microsoft.DotNet.Host.Build
                         "win.arm64.version",
                         "ubuntu.x64.version",
                         "ubuntu.16.04.x64.version",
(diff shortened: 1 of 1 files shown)


build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -121,17 +121,13 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
                     {
                         "win.x86.version",
                         "win.x64.version",
-                        "win.arm64.version",
```
weshaggard: Are we not building arm64?

dagood: My mistake, arm64 just had an error in the build I looked at--the leg is still running. Reverting this line.

PR #3114, opened 2017-08-29
Use ResourceHandlingTask to extract embedded resources for portable class libraries

zamont:
Add property <NetCoreGeneratePrisForPortableLibraries> to CoreRuntime.targets and re-write ResourceHandlingTask to extract ResW files from embedded resources in non-Framework assemblies as well, since current ResGen breaks on NS2.0 C# UWP libraries. Behavior for non-Framework resources are written to match ResGen behavior.

# Main Changes
- Separate code paths for Framework vs non-Framework assemblies. Framework assemblies processed in the same manner as before.
- Instead of assuming one ResW files per assembly, assume 1+ ResW files per assembly.
- Change persisted state so each assembly is associated with multiple ResW files. Each ResW file has all the information it needs: Path, LastWriteTime, and MSBuild metadata.
- Simulate ResGen behavior by processing main assemblies before satellite assemblies, correctly processing the NeutralResourcesLanguageAttribute in main assemblies, and correctly processing satellite assemblies.

# Testing
- Build private metapackage and installed in VS2017 on RS3 box. Behavior in RS3 project (where this was consumed) matches pre-RS3 project.

### DIFF
--- a/src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
+++ b/src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
+using System.Diagnostics;
 using System.IO;
 using System.Xml.Linq;
 using System.Collections;
@@ -160,7 +161,6 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
             UnprocessedAssemblyList = null;
 
             List<ITaskItem> unprocessedAssemblyList = new List<ITaskItem>();
-            List<ITaskItem> frameworkReswList = new List<ITaskItem>();
             List<ITaskItem> reswList = new List<ITaskItem>();
 
             _state = ReadStateFile(StateFile);
@@ -226,13 +226,13 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
                                     newTaskItem.SetMetadata("NeutralResourceLanguage", reswInfo.NeutralResourceLanguage);
                                 }
 
-                                if (containsFrameworkResources)
-                                {
-                                    frameworkReswList.Add(newTaskItem);
-                                }
-                                else
+                                if (!containsFrameworkResources)
                                 {
                                     newTaskItem.SetMetadata("OriginalItemSpec", reswInfo.ResWPath); // Original GenerateResource behavior creates this additional metadata item on processed non-framework assemblies
+                                    reswList.Add(newTaskItem);
+                                }
+                                else if (!SkipFrameworkResources)
+                                {
                                     reswList.Add(newTaskItem);
                                 }
                             }
@@ -241,16 +241,7 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
                     }
 
                     UnprocessedAssemblyList = unprocessedAssemblyList.ToArray(); // For now this list will always be empty
-                    
-                    if (SkipFrameworkResources)
-                    {
-                        ReswFileList = reswList.ToArray();
-                    }
-                    else
-                    {
-                        ReswFileList = reswList.Concat(frameworkReswList).ToArray();
-                    }
-
+                    ReswFileList = reswList.ToArray();
                     WriteStateFile(StateFile, _state);
                 } 
                 catch (Exception e)
@@ -335,7 +326,8 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
                 resourceIndexName = resourceFileName;
                 neutralResourceLanguage = "en-US";
             }
-            else {
+            else
+            {
                 string culturePath = "";
                 string culture = assembly.Culture;
                 string assemblyName = assembly.Name.Value;
@@ -348,10 +340,11 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
                 {
                     culturePath = neutralResourceLanguage + Path.DirectorySeparatorChar;
                 }
+                // Do not handle the case where culture is Invariant and no NeutralResourcesLanguageAttribute is declared
+                // This should already be taken care of in method ExtractAssemblyResWList
                 else
                 {
-                    // Do not handle the case where culture is Invariant and no NeutralResourcesLanguageAttribute is declared
-                    // This should already be taken care of in method ExtractAssemblyResWList
+                    Debug.Assert(false, "Assembly with the Invariant culture and no NeutralResourcesLanguageAttribute is being extracted for embedded resources. This should have been caught by earlier checks.");
                 }
 
                 if (resourceFileName.EndsWith("." + culture, StringComparison.OrdinalIgnoreCase))
@@ -434,7 +427,7 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
             if (String.IsNullOrEmpty(assembly.Culture) && !TryGetNeutralResourcesLanguageAttribute(assembly, out neutralResourceLanguage))
             {
                 // Must have NeutralResourcesLanguageAttribute
-                // warning MSB3817: The assembly "C:\Users\zamont\source\repos\08_22_ResGenBeforeAttempt_CSharpApp\ClassLibrary1\bin\x86\Debug\ClassLibrary1.dll" does not have a NeutralResourcesLanguageAttribute on it. To be used in an app package, portable libraries must define a NeutralResourcesLanguageAttribute on their main assembly (ie, the one containing code, not a satellite assembly).
+                // warning MSB3817: The assembly "<FullPath>\ClassLibrary1.dll" does not have a NeutralResourcesLanguageAttribute on it. To be used in an app package, portable libraries must define a NeutralResourcesLanguageAttribute on their main assembly (ie, the one containing code, not a satellite assembly).
                 return null;
             }
 
--- a/src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
+++ b/src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
@@ -20,11 +20,19 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
     public sealed class ResourceHandlingTask : Task
     {
         [Serializable()]
+        public sealed class ResWInfo
+        {
+            public DateTime ResWTimeUtc;
+            public string   ResWPath;
+            public string   ResourceIndexName;
+            public string   NeutralResourceLanguage;
+        }
+        [Serializable()]
         public sealed class PortableLibraryResourceStateInfo
         {
             public DateTime PLibTimeUtc;
-            public DateTime ResWTimeUtc;
-            public string   ResWPath;
+            public bool ContainsFrameworkResources;
+            public List<ResWInfo> ResWInfoList;
         }
 
         [Serializable()]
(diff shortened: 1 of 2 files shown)


zamont:
@nattress @DnlHarvey @morganbr PTAL
cc: @joshfree

nattress:
@sergiy-k Please give this a close look too

joshfree:
@tarekgh

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -66,21 +76,37 @@ public bool IsUpToDate(string assemblyPath, out string reswFilePath)
                         _logger.LogMessage(MessageImportance.Low, Resources.Message_CachedReswNotUpToDateAssemblyNewer, assemblyPath); 
                         return false;
                     }
-                    if (info.ResWPath == null || !File.Exists(info.ResWPath))
+                    if (info.ResWInfoList == null)
                     {
-                        _logger.LogMessage(MessageImportance.Low, Resources.Message_CachedReswNotExists, assemblyPath, info.ResWPath);
```
tarekgh: >_logger.LogMessage(M [](start = 24, length = 20)

why removed the logging here? we can just update it.

zamont: This logging is kept in the foreach loop below.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -144,55 +174,81 @@ public override bool Execute()
...
+                                    newTaskItem.SetMetadata("OriginalItemSpec", reswInfo.ResWPath); // Original GenerateResource behavior creates this additional metadata item on processed non-framework assemblies
+                                    reswList.Add(newTaskItem);
+                                }
+                            }
                         }
 
                     }
 
-                    UnprocessedAssemblyList = unprocessedAssemblyList.ToArray();
+                    UnprocessedAssemblyList = unprocessedAssemblyList.ToArray(); // For now this list will always be empty
```
tarekgh: >unprocessedAssemblyList [](start = 46, length = 23)

why we didn't keep the list of assemblies that not having resources here? would it be better to have IsUpToDate return this info so you don't have to process the assemblies that not having resources at all?

zamont: We've always discarded assemblies that do not contain resources. Also, Inserting into this list would defeat the purpose of the change because items that are inserted into here will be passed to ResGen after this, which may cause ResGen to crash. For example, an RS3 C# UWP Library that has no resources

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -144,55 +174,81 @@ public override bool Execute()
...
                     
-                    if (!SkipFrameworkResources)
+                    if (SkipFrameworkResources)
```
tarekgh: >SkipFrameworkResources [](start = 24, length = 22)

would it be better passing SkipFrameworkResources instead of containsFrameworkResources through the code and avoid using 2 lists for the framework and non framework? maybe this can simplify the code?

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
tarekgh: why we cannot get rid of UnprocessedAssemblyList then?

---
In reply to: [135929570](https://github.com/dotnet/core-setup/pull/3114#discussion_r135929570) [](ancestors = 135929570)

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -267,18 +323,82 @@ private void WriteStateFile(string stateFile, ResourceHandlingState state)
...
+
+            if (containsFrameworkResources)
             {
-                if (stream == null)
-                    return null;
+                reswFilePath = OutResWPath + Path.AltDirectorySeparatorChar + resourceFileName + ".resw";
+                resourceIndexName = resourceFileName;
+                neutralResourceLanguage = "en-US";
+            }
+            else {
```
tarekgh: >else { [](start = 12, length = 6)

nit: please move the parenthesis to new line

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -295,61 +415,70 @@ private void WriteResW(Stream stream, string reswFilePath)
...
 
-            assemblyName = assembly.Name.Value;
-            string resourcesName = "FxResources." + assemblyName + ".SR.resources";
-            int resourceCount = 0;
+            string neutralResourceLanguage;
+            if (String.IsNullOrEmpty(assembly.Culture) && !TryGetNeutralResourcesLanguageAttribute(assembly, out neutralResourceLanguage))
+            {
+                // Must have NeutralResourcesLanguageAttribute
+                // warning MSB3817: The assembly "C:\Users\zamont\source\repos\08_22_ResGenBeforeAttempt_CSharpApp\ClassLibrary1\bin\x86\Debug\ClassLibrary1.dll" does not have a NeutralResourcesLanguageAttribute on it. To be used in an app package, portable libraries must define a NeutralResourcesLanguageAttribute on their main assembly (ie, the one containing code, not a satellite assembly).
+                return null;
```
tarekgh: should we produce the same warning as VS used to do?

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: You're right, I can probably get rid of it. I wanted to make the least possible diff so as to reduce risk in this last-minute change, so setting this list to be empty was my attempt to do that. I'd like to keep this if you don't object to it.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: Ok.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -267,18 +323,82 @@ private void WriteStateFile(string stateFile, ResourceHandlingState state)
...
+
+                if (!String.IsNullOrEmpty(culture))
+                {
+                    culturePath = culture + Path.DirectorySeparatorChar;
+                }
+                else if (TryGetNeutralResourcesLanguageAttribute(assembly, out neutralResourceLanguage))
+                {
+                    culturePath = neutralResourceLanguage + Path.DirectorySeparatorChar;
+                }
+                else
```
tarekgh: > else [](start = 15, length = 5)

please comment this else

tarekgh: Actually should we never hit this case here? I mean maybe we can keep the else and assert inside it?

---
In reply to: [135931755](https://github.com/dotnet/core-setup/pull/3114#discussion_r135931755) [](ancestors = 135931755)

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: We will always need two different Booleans. `SkipFrameworkResources` is an input and `containsFrameworkResources` will tell us how to process an input Framework assembly.

I've made a code change to address this.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -267,18 +323,82 @@ private void WriteStateFile(string stateFile, ResourceHandlingState state)
...
+            foreach (ICustomAttribute attribute in assembly.AssemblyAttributes)
+            {
+                if (TypeHelper.GetTypeName(attribute.Type, NameFormattingOptions.None).Equals("System.Resources.NeutralResourcesLanguageAttribute"))
+                {
+                    if (attribute.Arguments.Count() > 0)
+                    {
+                        IMetadataConstant metadataConstant = attribute.Arguments.ElementAt(0) as IMetadataConstant;
+                        if (metadataConstant == null)
+                        {
+                            return false; // Unable to parse
```
tarekgh: > return false; // Unable to parse [](start = 27, length = 33)

should we continue the loop instead of returning false? just in case we have multiple neutral attribute definition? I am not sure if we can have such case though.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: We should but I don't think I have time to do this completely. We need to check this in asap. I've edited the code to remove my PII.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: The VS UI will only set one attribute, so unless people hand-write the NeutralResourcesAttribute then I think we'll be fine. I think it makes sense to return.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: Adding an assert.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
tarekgh: we are just inconsistent as if we have 
if (attribute.Arguments.Count() == 0) we continue the loop. 

anyway, this is minor point. i was hoping to be consistent.

---
In reply to: [135934366](https://github.com/dotnet/core-setup/pull/3114#discussion_r135934366) [](ancestors = 135934366)

zamont:
@tarekgh I'm going to upload a new commit shortly to address your feedback minus the MSBuild warning. Please let me know if that is satisfactory. I would like to remind you, we are shooting to check this in ASAP so if you have further feedback that are small nit's, I would like to address those later.

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -295,61 +408,70 @@ private void WriteResW(Stream stream, string reswFilePath)
...
+                        }
+                        ms.Write(buffer, 0, index);
+                        ms.Seek(0, SeekOrigin.Begin);
+                        
+                        string resourceFileName = resourceReference.Name.Value.Remove(resourceReference.Name.Value.Length - 10);
+                        if (resourceReference.Name.Value.Equals(frameworkResourcesName, StringComparison.OrdinalIgnoreCase))
+                        {
+                            containsFrameworkResources = true;
+                            reswInfoList.Add(ExtractResourcesFromStream(ms, assembly, resourceFileName, true));
+                            return reswInfoList;
```
tarekgh: >return reswInfoList; [](start = 28, length = 20)

why returning here? should we continue the loop. just in case the assembly has FX resources and some other resources?

tarekgh:
I am going to sign off. I hope you already tested this well. 

---
In reply to: [325830508](https://github.com/dotnet/core-setup/pull/3114#issuecomment-325830508) [](ancestors = 325830508)

src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
zamont: In this case I went with the previous logic that in Framework assemblies we only expect one embedded resource and nothing else.

tarekgh: This will not work well with the test assemblies. but I am fine with what you have. small comment will be good. thanks for your effort getting this ready quickly.

PR #3115, opened 2017-08-29
Clean up core-setup build

eerhardt:
Clean up core-setup build.

Set DisableImplicitNuGetFallbackFolder=true to ensure we never use the NuGet fallback folder that comes with the SDK.  

Enable SDK-based projects to import the root dir.props file.  This spawned a few one-off changes since things like the BaseIntermediateOutputPath is now under RepoRoot\Bin and TreatWarningsAsErrors=true.

Note: I removed a bunch of dead test code and logged it in the issue tracking those tests - https://github.com/dotnet/core-setup/issues/1553.


dir.props
```diff
@@ -2,6 +2,11 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
 
+  <PropertyGroup>
+    <!-- Directory.Build.props imports this .props file, so ensure it isn't imported again -->
+    <ImportDirectoryBuildProps>false</ImportDirectoryBuildProps>
```
chcosta: Consider flipping this property and putting it in `Directory.Build.props` instead.  Like an `<ImportedDirectoryBuildProps>true</ImportedDirectoryBuildProps>`

dir.props
```diff
@@ -388,7 +393,7 @@
     <MicrosoftNetCoreRuntimeAppxOutputPath>$(UWPOutputDir)Microsoft.NET.CoreRuntime.$(MajorVersion).$(MinorVersion).appx</MicrosoftNetCoreRuntimeAppxOutputPath>
   </PropertyGroup>
 
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(UsingNETSdkCompiler)' != 'true'">
```
chcosta: Merge these two property groups under the same condition?

src/test/TestUtils/RepoDirectoriesProvider.cs
```diff
@@ -71,5 +71,27 @@ public class RepoDirectoriesProvider
...
                 _builtDotnet = builtDotnet ?? Path.Combine(baseArtifactsFolder, "obj", _buildRID+".Release", "sharedFrameworkPublish");
         }
+
+        private static string GetRepoRootDirectory()
+        {
+            string currentDirectory = Directory.GetCurrentDirectory();
+
+            while (currentDirectory != null)
+            {
+                var gitDirOrFile = Path.Combine(currentDirectory, ".git");
```
chcosta: nit: `var -> string`

src/managed/CommonManaged.props
```diff
@@ -19,17 +15,6 @@
     <PackageThirdPartyNoticesFile>$(RepoRoot)THIRD-PARTY-NOTICES.TXT</PackageThirdPartyNoticesFile>
   </PropertyGroup>
 
-  <!--
-  Once https://github.com/dotnet/buildtools/issues/1548 is fixed, the following version sections should be simply
-  <Import Condition="Exists('$(ToolsDir)BuildVersion.targets')" Project="$(ToolsDir)BuildVersion.targets" />
-  -->
-  <PropertyGroup>
```
chcosta: Is this now handled via dir.props?

Directory.Build.props
```diff
@@ -0,0 +1,28 @@
...
+
+  <Import Project="dir.props" />
+
+  <PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true'">
+    <!--
+    The Microsoft.NET.Sdk changed the definition of what it assumes $(BaseIntermediateOutputPath) is.
+    In the "old" .csproj files, BaseIntermediateOutputPath could be shared with many projects.
+    In the SDK-based .csproj files, BaseIntermediateOutputPath is assumed to be project-specific.
+    -->
+    <BaseIntermediateOutputPath>$(IntermediateOutputPath)</BaseIntermediateOutputPath>
```
weshaggard: Can we not just set this in dir.props?

dir.props
eerhardt: Done.

src/test/TestUtils/RepoDirectoriesProvider.cs
eerhardt: Done.

src/managed/CommonManaged.props
eerhardt: Yes!  With this refactoring, all the projects are importing the root `dir.props`, like they should be.

dir.props
eerhardt: It doesn't work like that.  `ImportDirectoryBuildProps` is a property [declared in MSBuild's `Microsoft.Common.props`](https://github.com/Microsoft/msbuild/blob/38fe952a2ec5885b922e8581d33aa09e0904cf8b/src/Tasks/Microsoft.Common.props#L19).

`Directory.Build.props` imports `dir.props` so it can reuse all the properties in it.  If a non-SDK project imports `dir.props`, and then later imports `Microsoft.Common.props`, it will import `Directory.Build.props`.  We actually don't want `Directory.Build.props` imported at all in the current build for non-SDK projects.

When we take a new CLI, we will be able to detect if we are in an SDK project or not.  Then some more refactoring can happen here.  My thoughts are that we move ALL the properties in `dir.props` into `Directory.Build.props`, and switch the imports around - a blank root `dir.props` that just imports `Directory.Build.props`.

Directory.Build.props
eerhardt: Not really.  That would cause ripple changes to all the places that currently use `$(BaseIntermediateOutputPath)`.  Looking, there appears to be a few places in BuildTools that use it.  For example, `BuildVersion.targets` uses it for the `$(BuildVersionFilePath)` property.

For non-SDK projects (like our build.proj and dir.proj files), `$(BaseIntermediateOutputPath)` == `$(RepoRoot)\Bin\obj`.
This causes a problem for SDK-based projects (or really any MSBuild v15 projects), since the definition of `$(BaseIntermediateOutputPath)` changed to allow for package managers to lay down `.props/.targets` files that are automatically imported.  See the comments [here](https://github.com/Microsoft/msbuild/blob/38fe952a2ec5885b922e8581d33aa09e0904cf8b/src/Tasks/Microsoft.Common.props#L45-L61).

For SDK-based projects, `$(BaseIntermediateOutputPath)` MUST be unique per project and if you want to change it from `$ProjectDirectory)\obj`, you MUST set it before Microsoft.Common.props is imported.

eerhardt:
*sigh* finally green.  Let me know if there is anymore feedback.  I'll merge tomorrow morning if none.

Directory.Build.props
weshaggard: Sigh. Ok.

PR #3125, opened 2017-08-30
[release/2.0.0] Create ShippedNuGetPackage items before versions repo update

dagood:
~~Make `BeforeTargets` reference the exact `UpdatePublishedVersions` target that needs the package items, rather than the placeholder `UpdateVersionsRepo`.~~

Change to use `DependsOnTargets` to order. The `BeforeTargets` pointed at the wrong target, causing the package items to be created after they were needed.

Without the `ShippedNuGetPackage` items, release/2.0.0 builds were making empty versions repo commits.

This problem was added by https://github.com/dotnet/core-setup/pull/2993 to release/2.0.0 only.

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -13,11 +13,17 @@
           Condition="'$(Finalize)' == 'true'"
           DependsOnTargets="PublishCoreHostPackagesToFeed;FinalizeBuildInAzure;UpdateVersionsRepo" />
 
+  <!--
+    Target wrapping UpdatePublishedVersions: ensures that ShippedNuGetPackage items are created and
+    disables versions repo update if no auth token is defined. Otherwise, not specifying an auth
+    token would cause an error.
+  -->
   <Target Name="UpdateVersionsRepo"
           Condition="'$(GitHubAuthToken)' != ''"
-          DependsOnTargets="UpdatePublishedVersions" />
+          DependsOnTargets="ExcludeSymbolsPackagesFromPublishedVersions;
+                            UpdatePublishedVersions" />
 
-  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdateVersionsRepo" >
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions">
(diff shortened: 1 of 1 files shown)


publish/publish.proj
```diff
@@ -17,7 +17,7 @@
           Condition="'$(GitHubAuthToken)' != ''"
           DependsOnTargets="UpdatePublishedVersions" />
 
-  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdateVersionsRepo" >
+  <Target Name="ExcludeSymbolsPackagesFromPublishedVersions" BeforeTargets="UpdatePublishedVersions" >
```
weshaggard: Isn't this still going to

weshaggard: Actually it seems like this is making the UpdateVersionsRepo target completely useless now.

dagood: I think we need to keep `UpdateVersionsRepo` because `UpdatePublishedVersions` comes from BuildTools, and the condition doesn't make sense there. (If someone depends on `UpdatePublishedVersions` but doesn't have an auth token, an error seems like the most useful signal.)

This target will always be before `UpdateVersionsRepo` simple because `UpdateVersionsRepo` depends on `UpdatePublishedVersions`, but even if not, `UpdateVersionsRepo` doesn't use the items.

I think another way to do this would be to change `UpdateVersionsRepo` to `DependsOnTargets="ExcludeSymbolsPackagesFromPublishedVersions;UpdatePublishedVersions"` and remove the `BeforeTargets`. Maybe that would be clearer?

weshaggard: yeah that would probably be a little clearer.

dagood: Moved to DependsOnTargets and added a comment about what this stub target's doing.

dagood:
@weshaggard @crummel is release/2.0.0 in an ok place to merge this now?

weshaggard:
>@weshaggard @crummel is release/2.0.0 in an ok place to merge this now?

Yes I want this for our stable builds.

PR #3134, opened 2017-09-01
Allow AppBaseCompilation assembly resolver to resolve 'reference' (#3…

steveharter:
Shiproom mail to be sent for approval.

Direct port of https://github.com/dotnet/core-setup/pull/3065

For issue https://github.com/dotnet/core-setup/issues/2981

Fixes InvalidOperationException "Cannot find compilation library location for package" issue with having direct assembly references.

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
+++ b/src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
@@ -8,7 +8,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
+    <!-- Since Microsoft.DotNet.PlatformAbstractions isn't being serviced, take a dependency on the shipped package. -->
+    <!--<ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />-->
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
   </ItemGroup>
 
--- a/src/managed/CommonManaged.props
+++ b/src/managed/CommonManaged.props
@@ -5,7 +5,7 @@
 
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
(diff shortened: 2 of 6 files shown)


eerhardt:
Ok, I've made the necessary changes for shipping the DependencyModel package.

@weshaggard can you take a look?

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -8,7 +8,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
+    <!-- Since Microsoft.DotNet.PlatformAbstractions isn't being serviced, take a dependency on the shipped package. -->
+    <!--<ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />-->
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.0.0" />
```
weshaggard: I wonder if we can come up with a better approach here. Something that determines whether or not we want to reference the project or package based on whether or not we are building the package. 

To be clear I'm fine with what you have as I believe it is correct I'm just think about the future.

eerhardt: Agreed.  I was thinking about basing the Project or Package Reference off `$(BuildAllPackages)`.  But the issue is that these properties aren't flown very well.  The outer/uber build gets MSBuild properties set, but they aren't flown into these SDK-based projects when built with `dotnet build`, `dotnet pack`, etc.

I've been making advancements in `master`, but it still isn't as clean as I'd like.  Depending on the level of changes we need to make going forward in the `release/2.0.0` branch, we might consider pulling some of those build refactorings from `master` here to help with things like this.

Another thought:  if we would use the MSBuild that comes with the dotnetcli for the outer/uber build, we could change the we build these SDK-based projects.  Instead of doing `<Exec Command="dotnet build"` to build the SDK projects, we could just do `<MSBuild Target="Build|Pack|etc"`, which would make the flowing of these MSBuild properties easier.

weshaggard: While keying off the BuildAllPackages might work in some scenarios I think the real question should be are we building a new package or not. If we are building a new package it should remain a projectreference but if not it should become a package refernece. 

Another option would be to move the version number into these projects directly in servicing mode and so we would always keep the ProjectReference which would still keep the correct version.

weshaggard: The later is basically what we do in corefx.

PR #3136, opened 2017-09-02
Support all resource types in ResourceHandlingTask

nattress:
Remove unnecessary cast. Casting to a string in
`ResXResourceWriter.AddResource` prevents adding object / byte[]
resources. Copying any such resources will cause an
`InvalidCastException`. This technically allows binary data into the generated .resw and eventual PRI file, but it matches MSBuild's behavior. See https://github.com/Microsoft/msbuild/blob/4ceeb0936bd51eccc5fdf27105e1fb7e7de93249/src/Tasks/GenerateResource.cs#L3534. The final PRI file thus contains a string-ized representation of the base64 serialized binary data.


nattress:
@zamont @tarekgh PTAL

nattress:
@sergiy-k @morganbr FYI

tarekgh:
> The final PRI file thus contains a string-ized representation of the base64 serialized binary data.

Was this officially supported on the desktop? or this is something new are supporting?

PR #3137, opened 2017-09-02
Improve performance of ResourceHandlingTask

zamont:
This is done by removing the eager check to see if the assembly is a properly-built MainAssembly: InvariantCulture with a NeutralResourcesLanguage attribute. This used to run on all assemblies, including Framework assemblies and satellite assemblies, which should never require the check.

This had dramatic wins in our TFS enlistment, likely because of Microsoft.Cci inefficiencies, but the change only yields small wins when built here with a different Microsoft.Cci to compile with. It would still be good to have so we don't regress C# Rebuild scenario in VS. I have collected results for two apps, let me know if I should investigate more.

Also, the change results in absolutely no differences in resw files produced before/after.

### Performance Results

- Blank App
  - 130 CPU sample win
  - Here the only portablie libraries were Framework assemblies
- SyncFusion sample
  - 30 CPU sample win
  - Lots of portable libraries, but only one had embedded resources, so we only hit one additional assembly

### DIFF
--- a/src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
+++ b/src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
@@ -340,11 +340,13 @@ namespace Microsoft.Build.Net.CoreRuntimeTask
                 {
                     culturePath = neutralResourceLanguage + Path.DirectorySeparatorChar;
                 }
-                // Do not handle the case where culture is Invariant and no NeutralResourcesLanguageAttribute is declared
-                // This should already be taken care of in method ExtractAssemblyResWList
+                // Must have NeutralResourcesLanguageAttribute
+                // warning MSB3817: The assembly "<FullPath>\ClassLibrary1.dll" does not have a NeutralResourcesLanguageAttribute on it.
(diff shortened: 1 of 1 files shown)


src/uwp/Microsoft.Build.Net.CoreRuntimeTask/ResourceHandlingTask.cs
```diff
@@ -457,15 +451,18 @@ private List<ResWInfo> ExtractAssemblyResWList(string assemblyFilePath, out bool
                         ms.Seek(0, SeekOrigin.Begin);
                         
                         string resourceFileName = resourceReference.Name.Value.Remove(resourceReference.Name.Value.Length - 10);
-                        if (resourceReference.Name.Value.Equals(frameworkResourcesName, StringComparison.OrdinalIgnoreCase))
+                        bool isFrameworkResource = resourceReference.Name.Value.Equals(frameworkResourcesName, StringComparison.OrdinalIgnoreCase);
```
tarekgh: >isFrameworkResource [](start = 29, length = 19)

nit: you can use containsFrameworkResources instead of defining a new variable here and that will safe restoring the value again.

tarekgh:
:+1:

PR #3158, opened 2017-09-08
Updating coreclr and corefx package dependencies

joperezr:
Updating one last time corefx and coreclr before release.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -9,17 +9,17 @@
        These ref versions are pulled from https://github.com/dotnet/versions.
   -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


joperezr:
FYI: @nattress @joshfree @zamont

joperezr:
there was a maestro PR open already but it was updating way more than we needed, so this PR will only update corefx and coreclr which is what we need before shipping.

joperezr:
@dotnet/dnceng looks like there is a problem with Jenkins, this PR and #3112 have been waiting for almost 4 hours to get the Windows_NT builds but they are still the ~350th on the queue. I checked the actual jenkins build definition and looks like the problem is that the actual machine pool is offline. I have buddy built these changes on my local machine for all three archs in windows and made sure that build succeeds so I'll go ahead and merge this now.

PR #3162, opened 2017-09-08
Make non-stable and update to 1.0.7

safern:
cc: @weshaggard


safern:
I suspect this needs the corefx and coreclr build to produce the servicing packages.

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -62,7 +62,7 @@ public override string ToString()
         //
 
         // Full versions and package information.
-        public static bool EnsureStableVersion => true;
+        public static bool EnsureStableVersion => false;
         public string LatestHostPrerelease => "servicing";
         public string LatestHostBuildMajor => CommitCountString;
         public string LatestHostBuildMinor => "00";
```
weshaggard: You need to bump the 6 to 7 below to fix your CI issues.

safern: Duh, missed that one. Thanks for taking a look.

PR #3163, opened 2017-09-09
Move to non-stable 1.1.4

safern:
cc: @weshaggard


weshaggard:
ubuntu 16.10 CI legs are broken and should actually be removed. @safern can you please work on a PR to remove that leg?

safern:
> @safern can you please work on a PR to remove that leg?

Yes, will do that.

PR #3171, opened 2017-09-11
Moving shipping packages to stable

nattress:


nattress:
@dotnet-bot Please test

nattress:
@dotnet-bot test

joperezr:
The failures in OSX and Linux are expected. The actual build succeeds in all three legs (Linux arm, Linux x64 and OSX x64) but when the build is done, The Jenkins build definition tries to find the .tar.gz file of the shared framework in order to archive it and will fail if nothing was produced. This is the line where we set that: https://github.com/dotnet/core-setup/blob/master/netci.groovy#L108

Given that is not an actual problem with the build itself, I would use this PR to comment out that line as well (So subsequent PRs against this branch <if any> won't fail) and just merge this PR with those three builds "failing". cc: @weshaggard @mmitche

nattress:
I thought the same so I fixed it in https://github.com/dotnet/core-setup/commit/e10cf56bff9a823671039a571d7c31e9c3aa4a63 but this still fails artifact publishing. I will try merging this and spinning the VSO build.

PR #3180, opened 2017-09-11
Fixing official build definitions

joperezr:
cc: @chcosta @weshaggard @joshfree @nattress 

Fixing the build definitions so that we can produce a final build of release/uwp6.0 branch

- Removing Linux and OSX builds since they are no-ops in this branch.
- Removing steps in Windows builds that will build the shared framework and installers and sign them, since that is .NET Core specific and not required for UWP.


joperezr:
I also added one more commit that will stop building non-Windows builds in CI so subsequent builds won't fail.

joperezr:
@dotnet-bot test ci please

PR #3187, opened 2017-09-11
Add RHEL 6 and CentOS 6 distro detection

janvorli:
This change adds RHEL 6 and CentOS 6 distro detection. These distros
don't have the /etc/os-release file and so we use an alternative source
of the truth that is the /etc/redhat-release file.
We use the same detection way in the build.sh scripts in all repos.


janvorli:
@eerhardt yes, I will create a porting PR today.

janvorli:
Fixes #3189

PR #3194, opened 2017-09-12
Removing NetStandard.Library reference from the UWP metapackage

joperezr:
cc: @joshfree @weshaggard @nattress @Petermarcu 

In case we decide to remove netsandard dependency from the UWP metapackage, this is the PR that would take care of that. I'm adding a no-merge label until we decide in our meeting what we will want to do, but I wanted to get the PR ready and CI running in case we decide to take it.


joperezr:
Closing this since we decided to go with a different approach and actually fix the netstandard library package dependency.

PR #3205, opened 2017-09-13
Fix PlatformManifest generation

eerhardt:
When we generate the PlatformManifest.txt, and an assembly doesn't have a valid FileVersion, we are never updating the first entry that gets put into the fileVersions collection.  This is because the check for AssemblyVersion only updates the entry if both current and existing AssemblyVersions are non-null.

Fixing it check for current != null and existing == null, to update the entry.

Fix https://github.com/dotnet/corefx/issues/23988

/cc @rynowak 

## Note

I compared the PlatformManifest.txt from https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App/2.1.0-preview2-25708-01 and my locally built file, and only 2 differences show up:

image

image

### DIFF
--- a/tools-local/tasks/GenerateFileVersionProps.cs
+++ b/tools-local/tasks/GenerateFileVersionProps.cs
@@ -58,15 +58,21 @@ namespace Microsoft.DotNet.Build.Tasks
 
                 if (fileVersions.TryGetValue(fileName, out existing))
                 {
-                    if (current.AssemblyVersion != null &&
-                        existing.AssemblyVersion != null &&
-                        current.AssemblyVersion != existing.AssemblyVersion)
+                    if (current.AssemblyVersion != null)
                     {
-                        if (current.AssemblyVersion > existing.AssemblyVersion)
+                        if (existing.AssemblyVersion == null)
                         {
                             fileVersions[fileName] = current;
+                            continue;
                         }
-                        continue;
+                        else if (current.AssemblyVersion != existing.AssemblyVersion)
+                        {
+                            if (current.AssemblyVersion > existing.AssemblyVersion)
+                            {
+                                fileVersions[fileName] = current;
+                            }
+                            continue;
+                        }
                     }
 
                     if (current.FileVersion != null && 


tools-local/tasks/GenerateFileVersionProps.cs
```diff
@@ -58,15 +58,21 @@ public override bool Execute()
 
                 if (fileVersions.TryGetValue(fileName, out existing))
                 {
-                    if (current.AssemblyVersion != null &&
-                        existing.AssemblyVersion != null &&
-                        current.AssemblyVersion != existing.AssemblyVersion)
+                    if (current.AssemblyVersion != null)
                     {
-                        if (current.AssemblyVersion > existing.AssemblyVersion)
+                        if (existing.AssemblyVersion == null)
```
ericstj: Curious how we had any assembly version that was null once and not null the second time.  Was the first time a cross-gen image for a different platform?  If so, maybe we should switch our assembly version reading to always use SRM (see https://github.com/dotnet/sdk/issues/1502)

ericstj:
I think I had an assumption when I wrote this code that if we had ourselves a file and that file was present it'd always either be an assembly (thus have a version) or not, but this shows otherwise.  I had a suggestion above for a potential root cause where my assumption holds.  This is a good fix regardless.

tools-local/tasks/GenerateFileVersionProps.cs
eerhardt: When I debugged it on my machine, it is because the first file encountered doesn't exist.

`F:\git\core-setup\Bin\win-x64.Debug\crossgen\Microsoft.NETCore.App\linux-x64\runtimes\linux-x64\lib\netcoreapp2.1\System.Runtime.CompilerServices.Unsafe.dll`

Notice that it is `crossgen\Microsoft.NETCore.App\linux-x64` and I'm on a Windows machine.

https://github.com/dotnet/core-setup/blob/a0b82c4ec4a5186d1ecf611503bd075a308be7c6/tools-local/tasks/GenerateFileVersionProps.cs#L123-L137

Notice that if the file doesn't exist, it tries to get the AssemblyVersion from the metadata, which also doesn't exist.

ericstj: I see, that makes sense.  I bet we broke this when we started cross-gening.  Prior to that we'd actually restore all the RIDs when building out this manifest just to read the file info out of them.

PR #3212, opened 2017-09-14
Enable end to end Alpine core-setup build

janvorli:
This change enables full end to end build of core-setup including
managed code on Alpine Linux.

### DIFF
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -22,13 +22,14 @@ init_rid_plat()
         __rid_plat=""
         if [ -e /etc/os-release ]; then
             source /etc/os-release
-            if [[ "$ID" == "rhel" && $VERSION_ID = 7* ]]; then
-                __rid_plat="rhel.7"
-            elif [[ "$ID" == "centos" && "$VERSION_ID" = "7" ]]; then
-                __rid_plat="rhel.7"
-            else
-                __rid_plat="$ID.$VERSION_ID"
+            if [[ "$ID" == "centos" ]]; then
+                ID="rhel"
             fi
+            if [[ "$ID" == "rhel" || "$ID" == "alpine" ]]; then
+                # remove the last version number
+                VERSION_ID=${VERSION_ID%.*}
+            fi
+            __rid_plat="$ID.$VERSION_ID"
         elif [ -e /etc/redhat-release ]; then
             local redhatRelease=$(</etc/redhat-release)
             if [[ $redhatRelease == "CentOS release 6."* || $redhatRelease == "Red Hat Enterprise Linux Server release 6."* ]]; then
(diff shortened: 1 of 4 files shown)


src/corehost/common/pal.unix.cpp
```diff
@@ -257,16 +257,27 @@ static
 pal::string_t normalize_linux_rid(pal::string_t rid)
 {
     pal::string_t rhelPrefix(_X("rhel."));
+    pal::string_t alpinePrefix(_X("alpine."));
+    size_t minorVersionSeparatorIndex = std::string::npos;
```
eerhardt: calling this `minorVersionSeparatorIndex` is misleading now, since we use it for the patch version separator on alpine.

The same applies to the corresponding C# code below.

src/pkg/projects/netcoreappRIDs.props
```diff
@@ -18,6 +18,7 @@
   <ItemGroup>
     <OfficialBuildRID Include="linux-x64" />
     <OfficialBuildRID Include="rhel.6-x64" />
+    <OfficialBuildRID Include="alpine.3.6-x64" />
```
eerhardt: I think we will need CoreCLR and CoreFX packages for `alpine.3.6-x64` before we can check this in.

src/corehost/build.sh
```diff
@@ -22,13 +22,14 @@ init_rid_plat()
...
-                __rid_plat="rhel.7"
-            elif [[ "$ID" == "centos" && "$VERSION_ID" = "7" ]]; then
-                __rid_plat="rhel.7"
-            else
-                __rid_plat="$ID.$VERSION_ID"
+            if [[ "$ID" == "centos" ]]; then
+                ID="rhel"
+            fi
+            if [[ "$ID" == "rhel" || "$ID" == "alpine" ]]; then
+                # remove the minor version digit
```
eerhardt: For alpine, it isn't the "minor" version digit.  So maybe this comment should say "remove the last version digit"

src/corehost/common/pal.unix.cpp
janvorli: I understand, I had the "minor" as the "last" version component in my head, but I can see that it is confusing.

src/pkg/projects/netcoreappRIDs.props
janvorli: Ah, I didn't know having that in would cause troubles, I'll remove it and put that in after we have the official ones. Is it the same case for corefx / coreclr? I've already merged this in into coreclr.

eerhardt: I think you should be fine putting this in for corefx and coreclr.

The difference here is that core-setup needs to restore packages from coreclr and corefx for all those RIDs in order to build its "platform manifest" file.

PR #3215, opened 2017-09-14
Adding FilesToSign error check to all ItemGroups

jcagme:
This error check was added to only one item group. Adding it to all


weshaggard:
Lets wait until after the current servicing is done before merging.

jcagme:
Changes will be merged until we have stable builds in place. We don't want extra builds being triggered and using the only two agent machines we have in the pool we use for signing just for this change.

eerhardt:
Can this PR be merged now?

weshaggard:
Yes :)

PR #3219, opened 2017-09-14
Add error checks

jcagme:
Adding error checks to ensure we fail the build when there are no files to sign


eerhardt:
We aren't planning on shipping a release/2.0.0 build in the near future.  We can take this change.

PR #3224, opened 2017-09-15
Don't check for UWP files while signing in arm64

jcagme:
UWP is not supported in arm64, hence do not check for the existence of files to sign if the RID is win-arm64

### DIFF
--- a/signing/sign.proj
+++ b/signing/sign.proj
@@ -76,7 +76,7 @@
         <Authenticode>Appx</Authenticode>
       </FilesToSign>
     </ItemGroup>
-    <Error Condition="'@(FilesToSign)' == ''" Text="There are no files to sign. FilesToSign group is empty."/>
+    <Error Condition="'@(FilesToSign)' == '' AND '$(TargetRid)' != 'win-arm64'" Text="There are no files to sign. FilesToSign group is empty."/>
   </Target>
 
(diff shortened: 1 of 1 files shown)


signing/sign.proj
```diff
@@ -76,7 +76,7 @@
         <Authenticode>Appx</Authenticode>
       </FilesToSign>
     </ItemGroup>
-    <Error Condition="'@(FilesToSign)' == ''" Text="There are no files to sign. FilesToSign group is empty."/>
+    <Error Condition="'@(FilesToSign)' == '' AND '$(TargetRid)' != 'win-arm64'" Text="There are no files to sign. FilesToSign group is empty."/>
```
eerhardt: This may be to nitpicky - but can we use the same condition that is in the build?

`'$(Platform)' != 'arm64'`

https://github.com/dotnet/core-setup/blob/f0fed403e0f455ebaf37626a9561b7de5fdbb160/src/pkg/projects/Microsoft.NETCore.UniversalWindowsPlatform/Microsoft.NETCore.UniversalWindowsPlatform.builds#L6

eerhardt:
FYI - @zamont @nattress

signing/sign.proj
jcagme: Changed

eerhardt:
Merging to unblock the build.  sign.proj isn't used in CI anyway.

PR #3228, opened 2017-09-19
add basic FreeBSD support

wfurt:
This is part for https://github.com/dotnet/corefx/issues/1626

This change adds recognition for FreeBSD and implements function to get OS version. 
There should be no impact for other platforms. 
 
./dotnet --info
Runtime Environment:
 OS Name:     FreeBSD
 OS Version:  11.0-RELEASE-p1
 OS Platform: FreeBSD
 RID:         linux-x64
 Base Path:   /mnt/dotnetcli4/dotnetcli.OK/sdk/2.0.0-preview1-005977/

(RID fixup is not part of this change)

### DIFF
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -41,14 +41,21 @@ init_rid_plat()
     if [ "$(uname -s)" == "Darwin" ]; then
         __rid_plat=osx.10.12
     fi
+    if [ "$(uname -s)" == "FreeBSD" ]; then
+        major_ver=`uname -U | cut -b1-2`
+        __rid_plat=freebsd.$major_ver
+    fi
 
     if [ $__linkPortable == 1 ]; then
         if [ "$(uname -s)" == "Darwin" ]; then
             __rid_plat="osx"
+        elif [ "$(uname -s)" == "FreeBSD" ]; then
+            __rid_plat="freebsd"
         else
             __rid_plat="linux"
         fi
     fi
+    echo __rid_plat=$__rid_plat
 }
 
 usage()
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -55,7 +55,6 @@ init_rid_plat()
             __rid_plat="linux"
         fi
     fi
-    echo __rid_plat=$__rid_plat
 }
 
 usage()
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -257,7 +257,8 @@ pal::string_t pal::get_current_os_rid_platform()
     if (ret == 0)
     {
         char *pos = strchr(str,'.');
-        if (pos) {
+        if (pos)
+        {
             *pos = '\0';
         }
         ridOS.append(_X("freebsd."));
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -208,7 +208,7 @@ pal::string_t trim_quotes(pal::string_t stringToCleanup)
 pal::string_t pal::get_current_os_rid_platform()
 {
     pal::string_t ridOS;
-    
+
     char str[256];
 
     // There is no good way to get the visible version of OSX (i.e. something like 10.x.y) as
@@ -240,6 +240,28 @@ pal::string_t pal::get_current_os_rid_platform()
             ridOS.append(_X("osx.10."));
             ridOS.append(pal::to_string(minorVersion));
         }
+    }
+
+    return ridOS;
+}
+#elif defined(__FreeBSD__)
+// On FreeBSD get major verion. Minors should be compatible
+pal::string_t pal::get_current_os_rid_platform()
+{
+    pal::string_t ridOS;
+
+    char str[256];
+
+    size_t size = sizeof(str);
+    int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
+    if (ret == 0)
+    {
+        char *pos = strchr(str,'.');
+        if (pos) {
+            *pos = '\0';
+        }
+        ridOS.append(_X("freebsd."));
+        ridOS.append(str);
     }
 
     return ridOS;
@@ -436,17 +458,17 @@ bool pal::get_own_executable_path(pal::string_t* recv)
     if (error_code == ENOMEM)
     {
         size_t len = sysctl(mib, 4, NULL, NULL, NULL, 0);
-        std::unique_ptr<char[]> buffer = new (std::nothrow) char[len];
+        std::unique_ptr<char[]> buffer (new (std::nothrow) char[len]);
 
         if (buffer == NULL)
         {
             return false;
         }
 
-        error_code = sysctl(mib, 4, buffer, &len, NULL, 0);
+        error_code = sysctl(mib, 4, buffer.get(), &len, NULL, 0);
         if (error_code == 0)
         {
-            recv->assign(buffer);
+            recv->assign(buffer.get());
             return true;
         }
     }
--- a/src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
+++ b/src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
@@ -74,10 +74,14 @@ namespace Microsoft.DotNet.PlatformAbstractions.Native
 
         private static string GetFreeBSDVersion()
         {
-            String verson = RuntimeInformation.OSDescription;
+            // This is same as sysctl kern.version
+            // FreeBSD 11.0-RELEASE-p1 FreeBSD 11.0-RELEASE-p1 #0 r306420: Thu Sep 29 01:43:23 UTC 2016     root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC
+            // What we want is major release as minor releases should be compatible.
+            String version = RuntimeInformation.OSDescription;
             try
             {
-                return RuntimeInformation.OSDescription.Split()[1];
+                // second token up to first dot
+                return RuntimeInformation.OSDescription.Split()[1].Split('.')[0];
             }
             catch
             {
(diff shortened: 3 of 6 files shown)


eerhardt:
@wfurt - the RID calculation code exists in 2 places - once in C# and once in C++.  Since you are changing the C# RID calculation code, you should also change the C++ RID calculation:

https://github.com/dotnet/core-setup/blob/master/src/corehost/common/pal.unix.cpp#L207-L400

wfurt:
Thanks for the pointer @eerhardt I'll get that as separate PR.
I'm still not quite sure how one or the other is used. 
This change is needed so cli can properly determine file extensions 
as fas as RID I did not find direct use yet.

eerhardt:
> I'll get that as separate PR.

Please don't separate this.  These two pieces of logic (managed and native RID calculation) need to remain in sync, it doesn't make sense for one to be updated, without the other.

> I'm still not quite sure how one or the other is used.

The managed RID calculation is pretty self-explanatory - anywhere managed code needs to get the current machine's RID, it calls into `RuntimeEnvironment.GetRuntimeIdentifier()`.

The native RID calculation is used by the `dotnet` host in order to pick the correct native assets at runtime.

It would be ideal if these two places could be shared, but it currently isn't easily possible.  The managed code is used outside of .NET Core, so it can't depend on an API that is only available in .NET Core.  The native piece is invoked before the CLR is even loaded, so it cannot invoke managed code.  This is why we have the calculation in two places.

src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
```diff
@@ -101,6 +103,8 @@ private static string GetRIDOS()
                     return OperatingSystem.ToLowerInvariant();
                 case Platform.Darwin:
                     return "osx";
+                case Platform.FreeBSD:
+                    return "FreeBSD";
```
eerhardt: Is this officially what we want the RID to be?  It looks inconsistent with our existing RIDs.  The other platforms don't have capital letters - `win`, `linux`, `osx`, `ubuntu`, etc.

I'm not sure how we make decisions on what string we use for RIDs.  @ericstj - do you know?

src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
```diff
@@ -68,6 +72,19 @@ private static string GetDarwinVersion()
             }
         }
 
+        private static string GetFreeBSDVersion()
+        {
+            String verson = RuntimeInformation.OSDescription;
+            try
+            {
+                return RuntimeInformation.OSDescription.Split()[1];
```
eerhardt: I'm not familiar with the versioning story of FreeBSD, but is there any normalization that should happen here?

For example, with RedHat, RedHat guarantees that minor versions are backwards compatible.  This allows us to use 1 RID for RedHat 7.1, 7.2, 7.3, etc.

See https://github.com/dotnet/core-setup/blob/b5c917c0b24171ae27c7cfcfd2617b1088d41c11/src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs#L143-L150

If FreeBSD makes similar guarantees, we can simplify the RIDs used on that platform.  Ex. we could just have `FreeBSD.11` instead of both `FreeBSD.11.0` and `FreeBSD.11.1`.

wfurt: I don't know if it does, I can ask the FreeBSD community. I think as far as RID it should be be freebsd.11, The "display" Name should be FreeBSD. I would expect the OS Version presents true OS version.  
Would hat make sense?

wfurt:
set on rid freebsd.11-x64

src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
```diff
@@ -214,6 +231,11 @@ private static Platform DetermineOSPlatform()
             {
                 return Platform.Darwin;
             }
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Create("FREEBSD")))
```
janvorli: Why is this different from the other platforms? I mean, why not OSPlatform.FREEBSD instead of OSPlatform.Create("FREEBSD")?

src/corehost/build.sh
```diff
@@ -41,14 +41,21 @@ init_rid_plat()
...
     if [ $__linkPortable == 1 ]; then
         if [ "$(uname -s)" == "Darwin" ]; then
             __rid_plat="osx"
+        elif [ "$(uname -s)" == "FreeBSD" ]; then
+            __rid_plat="freebsd"
         else
             __rid_plat="linux"
         fi
     fi
+    echo __rid_plat=$__rid_plat
```
eerhardt: Is this just for your testing?  or is it intended to check in?

wfurt: yes, it was for testing. removed.

src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
```diff
@@ -60,6 +60,8 @@ private static string GetRIDVersion()
                     return $".{OperatingSystemVersion}";
                 case Platform.Darwin:
                     return $".{OperatingSystemVersion}";
+                case Platform.FreeBSD:
+                    return $".{OperatingSystemVersion.Split('.')[0]}";
```
eerhardt: This is a little inconsistent with our other platforms.  For example, if you are on a RedHat 7.3 machine, `RuntimeEnvironment.OperatingSystemVersion` will return `7` and the RID is `rhel.7-x64`.  But on FreeBSD, `RuntimeEnvironment.OperatingSystemVersion` will return `11.1` and the RID is freebsd.11-x64.

Basically - the RID and the `OperatingSystemVersion` match on all our other platforms.  But here they aren't matching.

janvorli: I agree with @eerhardt, the OperatingSystemVersion should return 11 only here.

wfurt: I can fix that. FreeBSD has also text in version file like RELEASE or STABLE 
In my case the version is 11.0-RELEASE-p1.
Maybe for core-setup/cli we can keep the version consistent as you describe and have Platform functions to return true version as recognized by OS.

Would that be reasonable?

src/corehost/common/pal.unix.cpp
```diff
@@ -244,6 +244,28 @@ pal::string_t pal::get_current_os_rid_platform()
...
+    pal::string_t ridOS;
+
+    char str[256];
+
+    size_t size = sizeof(str);
+    int ret = sysctlbyname("kern.osrelease", str, &size, NULL, 0);
+    if (ret == 0)
+    {
+        char *pos = strchr(str,'.');
+        if (pos) {
```
janvorli: A nit - please move the opening brace to the next line.

src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
wfurt: That was rejected by  API review board.  https://github.com/dotnet/corefx/issues/23989 and https://github.com/dotnet/corefx/pull/23986
For now, they did not want to add FreeBSD member.

src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
janvorli: Actually, everywhere (corefx, coreclr, core-setup), we use the normalized version only. So for RHEL, we have versions 6 and 7, although they are reported as 7.x and 6.x by the OS. Similar for Alpine, their version is 3.6.2, but we report 3.6. Basically, if the normalized version changes, it means a major OS change happened and not just minor update that fixes bugs, but keeps stuff compatible.

src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
janvorli: Ah, ok, that makes sense.

src/managed/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs
```diff
@@ -68,6 +72,23 @@ private static string GetDarwinVersion()
...
+        private static string GetFreeBSDVersion()
+        {
+            // This is same as sysctl kern.version
+            // FreeBSD 11.0-RELEASE-p1 FreeBSD 11.0-RELEASE-p1 #0 r306420: Thu Sep 29 01:43:23 UTC 2016     root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC
+            // What we want is major release as minor releases should be compatible.
+            String version = RuntimeInformation.OSDescription;
+            try
+            {
+                // second token up to first dot
+                return RuntimeInformation.OSDescription.Split()[1].Split('.')[0];
```
janvorli: This looks fragile. Is the version string guaranteed to have the format you assume? It seems it would be nice to use the same method to extract the version in the C++ pal::get_current_os_rid_platform and here. That would also make the version extraction less fragile. You'd need to add a new method to NativeMethods.Unix.cs to pinvoke the `sysctlbyname` and call it from here with the "kern.osrelease"

wfurt: https://www.freebsd.org/releases/
There always should be number and dot. 
I was thinking about kern.osrelease but the OSDescription was conveniently works and also works in net45 branch (after adding proper reference)

janvorli: Ok, let's keep it this way for now.

bartonjs:
Based on how we do Ubuntu (every numbered release) this should probably include the minor.  So on a latest-and-greatest build we'd want it to be `freebsd.11.2-x64` (to allow is to insert new assets aligned to a minor release, if needed)

PR #3230, opened 2017-09-19
When loading coreclr under Windows add PATH locations

steveharter:
Change the LoadLibrary options so that the PATH environment variable is also used when loading coreclr. This allows any static library references to be loaded that exist in %PATH% but not in the other 'standard' locations. This does not affect the behavior of other dlls loaded once the host->coreclr handoff occurs including loading of dlls specified in the deps file such as managed assemblies and pinvoke\native assemblies.

Fixes https://github.com/dotnet/core-setup/issues/3213. Note: if we don't fix this, then any dlls that need to be loaded must first be moved into one of the 'standard' locations.

The primary way we want to load the coreclr is to ensure the coreclr directory is searched first (the library we're loading), and then the 'standard' locations. This is usually done by one of two ways:
`LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` (current implementation)
or
`LOAD_WITH_ALTERED_SEARCH_PATH` (proposed change)

The `LOAD_WITH_ALTERED_SEARCH_PATH` will include the PATH while `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` will not.

Some links (note we do specify an absolute path in the _lpFileName_ argument)
https://msdn.microsoft.com/en-us/library/windows/desktop/ms682586(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/ms684179(v=vs.85).aspx

Also, the intent is to affect the loading of coreclr, but the helper function is also used to load hostfxr.dll.

### DIFF
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -104,7 +104,8 @@ bool pal::load_library(const string_t* in_path, dll_t* dll)
     //Adding the assert to ensure relative paths which are not just filenames are not used for LoadLibrary Calls
     assert(!LongFile::IsPathNotFullyQualified(path) || !LongFile::ContainsDirectorySeparator(path));
 
-    *dll = ::LoadLibraryExW(path.c_str(), NULL, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
+    // LOAD_WITH_ALTERED_SEARCH_PATH looks in the directory specified by the first argument, and then the default locations.
+    *dll = ::LoadLibraryExW(path.c_str(), NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
     if (*dll == nullptr)
     {
         trace::error(_X("Failed to load the dll from [%s], HRESULT: 0x%X"), path, HRESULT_FROM_WIN32(GetLastError()));
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -88,10 +88,6 @@ bool pal::load_library(const string_t* in_path, dll_t* dll)
 {
     string_t path = *in_path;
 
-    // LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR:
-    //   In portable apps, coreclr would come from another directory than the host,
-    //   so make sure coreclr dependencies can be resolved from coreclr.dll load dir.
-
     if (LongFile::IsPathNotFullyQualified(path))
     {
         if (!pal::realpath(&path))
@@ -101,10 +97,15 @@ bool pal::load_library(const string_t* in_path, dll_t* dll)
         }
     }
     
-    //Adding the assert to ensure relative paths which are not just filenames are not used for LoadLibrary Calls
+    // Adding the assert to ensure relative paths which are not just filenames are not used for LoadLibrary Calls
     assert(!LongFile::IsPathNotFullyQualified(path) || !LongFile::ContainsDirectorySeparator(path));
 
+    // Prevent LoadLibrary from looking in the current directory.
+    ::SetDllDirectoryW(_X(""));
+
     // LOAD_WITH_ALTERED_SEARCH_PATH looks in the directory specified by the first argument, and then the default locations.
+    // This is necessary when using the shared runtime because coreclr is in a different directory than the host,
+    // so make sure coreclr dependencies can be resolved in the directory coreclr.dll exists in.
     *dll = ::LoadLibraryExW(path.c_str(), NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
     if (*dll == nullptr)
     {


janvorli:
@steveharter the change has possible security consequences. Before your change, the library search path would not include the current directory. With your change though, the current directory would be included in the lookup before the PATH is consulted. Can you please add `SetDllDirectory("")` call somewhere to the startup path to make sure the current directory is always excluded?
@morganbr, do you agree?

steveharter:
@janvorli I added the call to prevent the current directory from being used. Thanks

morganbr:
Where are the places coreclr is actually expected to be? Both the old and new versions look fairly broad.

steveharter:
@morganbr coreclr will normally be in a shared location or a standalone location. E.g.
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.0.0
C:\git\standaloneapp\bin\Debug\netcoreapp2.0\win10-x64\publish
C:\Program Files\standaloneapp

However the PR doesn't change the location of where we look for coreclr. The PR is about helping coreclr.dll load successfully by finding coreclr.dll's static dependencies that happen to exist in %PATH%

morganbr:
I guess I have the same question for the dependencies then. In other words, do we really need to look all over the path or do we have a pretty good idea of where to find dependencies? 

The general security issues in dll loading are:
1. Loading from the current working directory is very dangerous (but you've handled that).
2. Loading from the application directory is fine if it's in Program Files, but concerning if it might not be (e.g. if it's in Downloads, there could be a malicious dll that's been downloaded separately and waiting to run)
3. Loading from the path depends on what's on the path. If it includes unsecured locations, then there could be security issues. There can also be normal "dll hell" issues if a different version of the same dll shows up earlier on the path.

steveharter:
For clarity here's the locations Windows will search for dependencies when loading coreclr, taken from a link above with the current directory scratched out.

_If SafeDllSearchMode is enabled, the alternate search order is as follows:
The directory specified by lpFileName.
The system directory. Use the GetSystemDirectory function to get the path of this directory. 
The 16-bit system directory. There is no function that obtains the path of this directory, but it is searched. 
The Windows directory. Use the GetWindowsDirectory function to get the path of this directory.
~~The current directory.~~
The directories that are listed in the PATH environment variable. Note that this does not include the per-application path specified by the App Paths registry key. The App Paths key is not used when computing the DLL search path._

steveharter:
@dpodder can you talk a bit about the priority; I know you have a hack to copy the file(s) to where they will be found. Thanks.

dpodder:
@steveharter the HACK is in dotnet/coreclr, and is specifically being tracked by dotnet/coreclr#13970. The "problem" is that VC++ doesn't install PGI runtime dependencies into any system paths (but this is by design). They're instead made available on PATH when running in a "developer command prompt." The documented and supported way to run PGI assemblies is to run through the developer command prompt--except, in the case of coreclr when loaded through dotnet.exe, that still fails since PATH is being ignored.

I can't speak to priorities anymore, and can only speak as a community member--I actually left the dotnet team last Friday. @dotnet/rap-team should have the right context to weigh in though.

brianrob:
Thanks @dpodder!

@steveharter, from what I'm reading here, it sounds like the issue that we're likely to be broken by toolset updates in the future if we have to keep the hack that @dpodder describes above, since the internals of how PGO works can change.  So, we are looking for a workaround for a fix to allow us to remove the hack and use the PATH by running inside of a VS command prompt.  Hopefully that helps to answer your question - if not, just let me know.

morganbr:
If this is just for our development, is there a way to limit this change to non-production builds?

brianrob:
I think it's going to be difficult to decide what builds this should and shouldn't be in.  We use this functionality in essentially every build out of master, since we run automation to recompute the PGO counts.  When it comes to release builds, there is a policy decision as to which builds this will occur in, but it will be at least one build out of release right around escrow.  So, I'm not sure how we'd be able to pick and choose when this functionality works and when it doesn't.  Are we concerned that in order to use the PGO functionality as required by the compiler, that we're exposing a security bug?

steveharter:
>> Are we concerned that in order to use the PGO functionality as required by the compiler, that we're exposing a security bug?

I wouldn't say it is a security "bug", but @morganbr did express a concern that since the directories in %PATH% are not necessarily secured, it could be exploited in some way. Perhaps it could also cause some versioning issues.

Two mitigating factors are that the %PATH% is the last option in the search logic, and that this is only for 'new' static library references (since existing static references are not in %PATH%) which basically means a custom-built coreclr.

I am deferring merging this until Morgan and\or Jan give approval, as the semantic change really is in the coreclr, not core-setup.

steveharter:
@morganbr @janvorli any final comments on risk and sign-off vs no sign-off? Unless I get sign off on this, I'm going to close this PR and the associated issue. Thanks.

jorive:
I just want to clarify that this is not related to the VS version. This is generally broken because the PGI runtime is made available on PATH through the "Developer Command Prompt for VS".

steveharter:
ping @morganbr @janvorli 

Should I set up a meeting to discuss?

brianrob:
@steveharter, I think a meeting would be good so hopefully we can close on this.  Thanks.

steveharter:
Meeting request sent for today; if that doesn't work let's try next week. Thanks

janvorli:
@steveharter I've actually spent some time this morning thinking about possible solutions that would not lower the security and fix the problem. And it seems that we can do that with the help of delay loading. We would enable delay loading for the pgo dll when building coreclr.dll with PGO enabled. We would then override the delay loading hook function in coreclr.dll and let it do the LoadLibrary for the pgo.dll. There we can specify the option we need to enable LoadLibrary to look at the PATH too.
Benefits:
* It would happen only for the coreclr.dll built with PGO instrumentation enabled
* The PATH would be consulted only for loading the pgo.dll (this is probably not a big deal since the bullet before already ensures that it is used in the PGO instrumented runs and in such case, the security seems to not to be a problem)
* No changes to the dotnet tool would be necessary
I will try to do a proof of concept experiment over the weekend since I have not used the delay loading with hooks ever before and all of the idea came from reading the MSDN doc on that.

janvorli:
@steveharter proof of concept implemented and verified to be working as expected.

steveharter:
Closing this PR; assigning issue to @janvorli

PR #3235, opened 2017-09-26
Port build changes from master to dev/release/2.0.0

eerhardt:
In support of removing the https://github.com/dotnet/source-build patches for core-setup, I am porting the build changes that were made in master after release/2.0.0 was forked.  This will allow the source-build work to progress.

The first 9 commits are straight cherry-picks from master.  The last 2 commits I had to do some manual fix ups.


src/managed/CommonManaged.props
```diff
@@ -15,15 +15,6 @@
     <PackageThirdPartyNoticesFile>$(RepoRoot)THIRD-PARTY-NOTICES.TXT</PackageThirdPartyNoticesFile>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(StabilizePackageVersion)' != 'true'">
```
weshaggard: Why did you remove the stable version suffix workaround?

eerhardt: Because this is handled by the root dir.props now.

https://github.com/dotnet/core-setup/blob/bb6dac8404513d5ac5a1a216e362ec2ff5ea8ef5/dir.props#L74-L76

Technically, this change should be made in `master` as well.

weshaggard: I see. Interestingly it doesn't show up when I do "view all" diff it only shows up on the individual commit which is a little odd.

eerhardt: That's because this addition came in a cherry-picked commit, and then I removed it as part of "fixing up changes from master".  Originally this was needed because the "managed" projects didn't import the root dir.props.  So this change was needed.  But then later I changed it so these projects do import the root dir.props.  That makes this code unnecessary now.

I'll port this change to master as well, since it is unnecessary there.

eerhardt:
@weshaggard I'm going to merge this PR into dev/release/2.0.0.  The CI run in release/2.0.0 (#3246) is green.

PR #3249, opened 2017-09-28
Mark as packages non-stable for 1.1.5

safern:
cc: @weshaggard

### DIFF
--- a/src/Microsoft.DotNet.PlatformAbstractions/project.json
+++ b/src/Microsoft.DotNet.PlatformAbstractions/project.json
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.1.4-servicing-*",
+  "version": "1.1.5-servicing-*",
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.PlatformAbstractions/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.1.4-servicing-*",
```
weshaggard: We haven't been shipping PlatformAbstractions and DependencyModel so lets not bump there versions until we think we need to ship them (we may even want to consider downgrading them to one higher then the last one we shipped).

cc @eerhardt

safern: Ok. Will fix that then. So 1.1.4-servicing-* is good? Or should we downgrade to 1.1.3?

weshaggard: Just revert to 1.1.4 for now.

safern: Ok

eerhardt: This sounds like a good plan to me.

Note, for release/2.0.0, when we merge in [the build changes from master](https://github.com/dotnet/core-setup/pull/3235), these packages will automatically have the version number that the whole repo uses.  We won't have these versions spread all over the repo.

Of course, we will be able to override the version number for each package, but by default these numbers won't be duplicated.

safern:
test Ubuntu x64 Release Build

safern:
Oops. Closed it unintentionally

Issue #3243, opened 2017-09-27
Unable to compile with VS2017


steveharter:
## Steps to reproduce
Install VS 2017 without VS 2015

## Expected  behavior
Run the build command (master branch), no errors.

## Actual behavior
Run the build, the first error will contain text like:
```
 Calling "C:\git\coresetup\src\corehost\Windows\gen-buildsys-win.bat C:\Users\sharter\source\ vs2017 x64 eb60f0bf20c52b49032953763519d904b134c473 2.1.0-preview1-25719-0 2.1.0-preview1-25719-0.1.0-preview1-25719-0 2.1.0-preview1-25719 -0"
  CMake Error in CMakeLists.txt:
    No CMAKE_CXX_COMPILER could be found.
```

This can be worked around by changing the hard-coded version of VS in src\corehost\Windows\gen-buildsys-win.bat to `_VSString=15 2017` (yes we need a better way to set this for `14 2015` as well)

After this, there are issues with generating a vcxproj. Using the latest CMake 3.9.3 (some earlier versions have different errors):
```
   Project "C:\git\coresetup2\Bin\obj\win-x64.Debug\corehost\CMakeFiles\3.9.3\VCTargetsPath.vcxproj" on node 1 (default targets).
C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets(354,5): error MSB8013: This project doesn't contain the Configuration and Platform
combination of Debug|Win32. [C:\git\coresetup2\Bin\obj\win-x64.Debug\corehost\CMakeFiles\3.9.3\VCTargetsPath.vcxproj] [C:\git\coresetup2\src\corehost\build.proj]
      Done Building Project "C:\git\coresetup2\Bin\obj\win-x64.Debug\corehost\CMakeFiles\3.9.3\VCTargetsPath.vcxproj" (default targets) -- FAILED.
```

Also, previously building with VS2015 and VS2017 side-by-side resulted in [UWP errors](https://github.com/dotnet/core-setup/issues/3073). I'm not sure if this is addressed yet, as I've been building with just VS2015 installed.

steveharter:
@weshaggard PTAL. Thanks

weshaggard:
@steveharter you need to upgrade your version of CMake (not sure the min version) as the later versions have templates for VS2017.

weshaggard:
[...]

PR #3251, opened 2017-09-28
Only change platform to AnyCPU for csproj files

weshaggard:
Fixes https://github.com/dotnet/core-setup/issues/3243

cc @eerhardt 
@steveharter this should fix the CMake issue, although I cannot build the UWP projects still locally. I think I just need some new install to support that I'm looking at that next. 

After @eerhardt added the Directory.Build.targets file it started setting Platform to AnyCPU which messed up the native builds so to fix this we should only apply these settings for csproj.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,7 @@
     We are currently on too old of an SDK to detect - it doesn't have https://github.com/dotnet/sdk/pull/1242.
     But once we move to a new version, we can remove this property setting, and everything else will work correctly.
     -->
-    <UsingMicrosoftNETSdk>true</UsingMicrosoftNETSdk>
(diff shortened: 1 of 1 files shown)


weshaggard:
Note that this didn't break CI because the CI machines still use VS2015 which doesn't have the default imports of Directory.Build.targets.

steveharter:
@weshaggard thanks I can compile the host now, but I also can't compile the UWP host.

However if I change UWP's gen-buildsys-win.bat to use -G with VS 2017 then I can compile fine: change
`set __VSString=14 2015`
to
`set __VSString=15 2017`

If there a reason we want to keep the existing hard-coded __VSString, or should we change it to use VS2017 if it exists?

weshaggard:
@steveharter if updating the generator fixes it then yes please update it.

steveharter:
@weshaggard OK I will. Thanks

weshaggard:
@steveharter I already put a PR out https://github.com/dotnet/core-setup/pull/3254

PR #3252, opened 2017-09-29
Change to current directory after calling VSDevCmd.bat

weshaggard:
VSDevCmd.bat switches to a user directory so we need to ensure
we are switched back to the root of the repo in order for the
commands that take relative paths to projects to work. This will
allow you to build the repo from a clean cmd prompt again (still
requires VS installed)

cc @eerhardt @steveharter

### DIFF
--- a/run.cmd
+++ b/run.cmd
@@ -19,5 +19,6 @@ call "%_VSCOMNTOOLS%\VsDevCmd.bat"
 :Run
 :: We do not want to run the first-time experience.
 set DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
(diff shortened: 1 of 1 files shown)


eerhardt:
I'm a bit confused at why we have so many different versions of this script.

https://github.com/dotnet/coreclr/blob/master/run.cmd is different from
https://github.com/dotnet/corefx/blob/master/run.cmd is different from this

Would it make sense to get alignment across our repos?

weshaggard:
Yes it would make sense to align. This is a prime example of a script that I would like the auto-update scripting feature for.

cc @dagood @dleeapho

PR #3266, opened 2017-10-03
Handle RIDs better during the build.

eerhardt:
Currently, when building non-portable on a win10 machine, core-setup restores and produces `win7` RID assets.  This causes problems in source-build since corefx produces `win10` assets on the same machine. Thus, core-setup can't use the corefx assets because win10 is not compatible with win7.

In fixing this, I refactored the way RIDs are handled during the build.  We have 3 main RID usages:
* What is the RID of the current machine?  $(HostMachineRid)
* What is the RID of the toolset we are using to build?  core-setup doesn't have this scenario yet, but corefx calls this $(ToolRuntimeRID)
* What is the RID that the current build is producing outputs for?  $(OutputRid)

## Notes

I expect the ARM CIs to fail since I needed to modify the netci.groovy file.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -16,6 +16,7 @@
     <TraversalBuildDependencies>
       CreateOrUpdateCurrentVersionFile;
       CreateVersionInfoFile;
+      CreateHostMachineInfoFile;
       BatchRestorePackages;
       BuildCustomTasks;
     </TraversalBuildDependencies>
@@ -33,6 +34,26 @@
 
   <Target Name="BuildTraversalBuildDependencies"
           DependsOnTargets="$(TraversalBuildDependencies)" />
+
+  <Target Name="CreateHostMachineInfoFile">
+    <GetTargetMachineInfo>
+      <Output PropertyName="HostMachineRid" TaskParameter="RuntimeIdentifier" />
+    </GetTargetMachineInfo>
+
+    <PropertyGroup>
+      <HostMachineInfoPropsContent>
+&lt;Project&gt;
+  &lt;PropertyGroup&gt;
+    &lt;HostMachineRid&gt;$(HostMachineRid)&lt;/HostMachineRid&gt;
+  &lt;/PropertyGroup&gt;
+&lt;/Project&gt;
+      </HostMachineInfoPropsContent>
+    </PropertyGroup>
+
+    <WriteLinesToFile File="$(HostMachineInfoProps)"
+                      Lines="$(HostMachineInfoPropsContent)"
+                      Overwrite="True" />
+  </Target>
 
   <Target Name="BuildCustomTasks">
       <MSBuild Projects="tools-local/Microsoft.DotNet.Build.Tasks.Local/Microsoft.DotNet.Build.Tasks.Local.builds" />
--- a/buildpipeline/Core-Setup-Linux-BT.json
+++ b/buildpipeline/Core-Setup-Linux-BT.json
@@ -141,7 +141,7 @@
       },
       "inputs": {
         "filename": "docker",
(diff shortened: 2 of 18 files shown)


build.proj
```diff
@@ -34,6 +35,26 @@
   <Target Name="BuildTraversalBuildDependencies"
           DependsOnTargets="$(TraversalBuildDependencies)" />
 
+  <Target Name="CreateHostMachineInfoFile">
+    <GetTargetMachineInfo>
+      <Output PropertyName="HostMachineRid" TaskParameter="RuntimeIdentifier" />
```
weshaggard: Why not capture TargetOS as OsGroup here instead of parsing the HostMachineRid in dir.props?

buildpipeline/Core-Setup-Linux-BT.json
```diff
@@ -889,9 +883,6 @@
     "DockerTag_Ubuntu1604": {
       "value": "ubuntu-16.04-debpkg-e5cf912-20174703024721"
     },
-    "DistroRid_Ubuntu1604": {
-      "value": "ubuntu.16.04-$(PB_TargetArchitecture)"
```
weshaggard: Maybe I've missed it but where are these distro specific values getting set now?

eerhardt: They are being inferred from the current machine.

`RuntimeEnvironment.GetRuntimeIdentifier()` will return `ubutnu.16.04-x64` and that gets put into `$(HostMachineRid)`.  and by default `$(OutputRid)` is set to `$(HostMachineRid)-$(TargetArchitecture)`.

eerhardt: Note that all these steps do the same loop:

* Set up docker for a specific distro
* Build the installer for that specific distro
* Publish the installer for that specific distro

Since the build and publish are happening on `ubuntu.16.04` machines, `$(HostMachineRid)` just works.

build.proj
eerhardt: Because none of those properties are "good enough".

https://github.com/dotnet/buildtools/blob/9714ddaaef88b530a9054ca6f99687fc090fd057/src/Microsoft.DotNet.Build.Tasks/GetTargetMachineInfo.cs#L41-L50

TargetOS only gets set to `Windows_NT`, not `win10` or `win7`.  It also just says `Linux`, not `ubuntu.14.04` or `centos.7.1`.

eerhardt: Note: corefx uses a completely different BuildTools Task for this:

https://github.com/dotnet/buildtools/blob/03988127f489bcd1cf58d71131ac628620d53de9/src/Microsoft.DotNet.Build.Tasks/GenerateConfigurationProps.cs#L135-L155

Which does the same parsing and sets the `RuntimeOS` property.

eerhardt:
test ci please

buildpipeline/Core-Setup-Linux-BT.json
weshaggard: great.

rakeshsinghranchi: can we call $(MachineRid) instead of $(HostMachineRid) ? It might be confusing in case of host machine running docker container for building installer. 
For e.g. It can be a Ubuntu14.04 host machine that runs Ubuntu16.04 container to build the installers. In this case, $(HostMachineRid) will be ubuntu.16.04-x64 which is confusing as it may infer Ubuntu.14.04-x64 per the naming convention.

eerhardt: The intention of the property is the RID of the current machine that is running the build.  Just calling it `MachineRid` can also be confused with "this is RID of the machine that we are building for".

For what it's worth, the dotnet/cli calls this property [`HostRid`](https://github.com/dotnet/cli/blob/master/build/HostInfo.targets#L15).  One intention here is to try to start using the same name for the same thing across repos.

Issue #3269, opened 2017-10-04
Re-enable including source files in Microsoft.NETCore.App package during source-build


eerhardt:
We are disabling including source files in Microsoft.NETCore.App package during source-build because of long path issues.

See https://github.com/dotnet/source-build/pull/206#issuecomment-334028708

If we start building with a build system that supports long paths, we can enable this again.  Or if the product requirements no longer need source files included in the symbols packages, we can just remove this all together.

/cc @ericstj @weshaggard @dagood

dagood:
Closing: source files are no longer needed, and the transport packages from CoreCLR and CoreFX already don't include them. (We rely on SourceLink instead.)

PR #3270, opened 2017-10-04
Disable including source files in nupkgs during source-build to work around the long path limitation.

eerhardt:
/cc @weshaggard @ericstj @dagood 

I've logged https://github.com/dotnet/core-setup/issues/3269 to track turning this back on.

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -163,7 +163,7 @@
       </FilesToPackage>
     </ItemGroup>
 
-    <ItemGroup>
+    <ItemGroup Condition="'$(IncludeSourceFilesInPackage)' == 'true'">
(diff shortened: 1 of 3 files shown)


ericstj:
What changed that started breaking this?  I've had to jump through hoops in the past to keep it working: https://github.com/dotnet/core-setup/commit/38565036f3d3c67e873184336528c63c93fce1e8#diff-5e85be3c1f3e808f5c649b7357c3a686

eerhardt:
> What changed that started breaking this?

See https://github.com/dotnet/source-build/pull/206#issuecomment-334028708

Basically we are building this repo nested inside of the `source-build` repo, and the root repo path to core-setup is now: `D:\j\workspace\Windows_NT_Re---a3f4c8f6\src\core-setup\`

eerhardt:
> Back when I hit this the last time folks insisted I keep the source packages working 

Who were the folks?  I'd like to get their thoughts on where we are heading here.

ericstj:
> Who were the folks?

I've forwarded you an email thread with some more context.

PR #3272, opened 2017-10-04
Build custom tasks for net45 on Windows, even when building in source-build.

eerhardt:
Our custom tasks need to be build for net45 when running on Desktop MSBuild.

@weshaggard

### DIFF
--- a/tools-local/dir.props
+++ b/tools-local/dir.props
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\dir.props" />
+
+  <PropertyGroup>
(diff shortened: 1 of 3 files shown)


tools-local/dir.props
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\dir.props" />
+
+  <PropertyGroup>
+    <BuildCustomTasksForDesktop Condition="'$(DotNetBuildFromSource)' != 'true' OR '$(MSBuildRuntimeType)' != 'Core'">true</BuildCustomTasksForDesktop>
```
weshaggard: Why do we even care about the DotNetBuildFromSource flag? I would just base it on MsBuildRuntimeType.

eerhardt: Good point.  changing.

PR #3188, opened 2017-09-12
Port to 2.0.0: Add RHEL 6 and CentOS 6 distro detection

janvorli:
This change adds RHEL 6 and CentOS 6 distro detection. These distros
don't have the /etc/os-release file and so we use an alternative source
of the truth that is the /etc/redhat-release file.

PR #3273, opened 2017-10-04
Bump versions to 2.0.3

safern:
cc: @weshaggard

### DIFF
--- a/src/managed/CommonManaged.props
+++ b/src/managed/CommonManaged.props
@@ -5,7 +5,7 @@
 
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
-    <VersionPrefix>2.0.2</VersionPrefix>
+    <VersionPrefix>2.0.3</VersionPrefix>
     <AssemblyFileVersion>$(VersionPrefix)</AssemblyFileVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
--- a/dir.props
+++ b/dir.props
@@ -13,7 +13,7 @@
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>0</MinorVersion>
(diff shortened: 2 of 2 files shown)


src/managed/CommonManaged.props
```diff
@@ -5,7 +5,7 @@
 
   <PropertyGroup>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../..'))/</RepoRoot>
-    <VersionPrefix>2.0.2</VersionPrefix>
+    <VersionPrefix>2.0.3</VersionPrefix>
```
weshaggard: I don't think we need to bump these unless we decide to ship one of these libraries.

safern: I added them because currently they are under the servicing list of packages. @eerhardt added them in this commit: https://github.com/dotnet/core-setup/commit/69f6efc7395c83721eab65fd30184a1bfb253ad3

safern: It was because of this change:

https://github.com/dotnet/core-setup/pull/3188

eerhardt: Yes, we had to change PlatformAbstractions in order to support RHEL 6.

FYI - when dev/release/2.0.0 comes in, these version numbers will all roll up to the product-wide version number in the root.  See https://github.com/dotnet/core-setup/commit/df6ab9c2a48b47181bedfb2704888b074e8a269a#diff-f9d387bebca621498ce66c5220efb38dR25

eerhardt:
Did we ever ship a `2.0.2`?  I'm confused why we are bumping past `2.0.2`, is there some background context?

src/managed/CommonManaged.props
safern: Makes sense. I will keep an eye on. Thanks for the heads up.

weshaggard: OK keep these bumped given we need to ship them.

weshaggard:
>Did we ever ship a 2.0.2? I'm confused why we are bumping past 2.0.2, is there some background context?

We didn't we are bumping to 2.0.3 to align with the SDK version for this release as they did ship a 2.0.2 for a different context.

PR #3293, opened 2017-10-09
Updating CLI to 2.0 RTM

joperezr:
cc: @eerhardt @weshaggard 

Updating CLI to 2.0 RTM on dev/release/2.0 branch


init-tools.cmd
```diff
@@ -71,6 +71,8 @@ if NOT exist "%BUILD_TOOLS_PATH%init-tools.cmd" (
 :afterbuildtoolsrestore
 
 echo Initializing BuildTools...
+:: Setting RunnintInitTools to true since the restore of project.csproj will throw some warnings for package downgrades and we want to disable TreatWarningsAsErrors.
+set RunningInitTools=true
```
eerhardt: I'm not crazy about this....

joperezr: Agreed, but the problem is that TreatWarningsAsErrors is set to true for all projects on core-setup by dir.props, and the new CLI will throw warnings for package downgrades closure of the tool-runtime. The other way to do this would be to have a change in buildtools, but I think this would be more scoped.

eerhardt: Which packages are causing the downgrades?  Is it only `Cci`?

/cc @weshaggard

weshaggard: I'm not crazy about this either and I'm also interested in the downgrade warnings

joperezr: Yup looks like it's mostly the CCI ones plus one saying that one package used Desktop TFM instead of netcoreapp.
```
  Restoring packages for /Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj...
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Collections.NonGeneric (>= 4.0.1-rc4-24217-00) but System.Collections.NonGeneric 4.0.1-rc4-24217-00 was not found. An approximate best match of System.Collections.NonGeneric 4.0.1-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.IO.FileSystem.Primitives (>= 4.0.1-rc4-24217-00) but System.IO.FileSystem.Primitives 4.0.1-rc4-24217-00 was not found. An approximate best match of System.IO.FileSystem.Primitives 4.0.1-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Collections (>= 4.0.11-rc4-24217-00) but System.Collections 4.0.11-rc4-24217-00 was not found. An approximate best match of System.Collections 4.0.11-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Diagnostics.Debug (>= 4.0.11-rc4-24217-00) but System.Diagnostics.Debug 4.0.11-rc4-24217-00 was not found. An approximate best match of System.Diagnostics.Debug 4.0.11-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Globalization (>= 4.0.11-rc4-24217-00) but System.Globalization 4.0.11-rc4-24217-00 was not found. An approximate best match of System.Globalization 4.0.11-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.IO (>= 4.1.0-rc4-24217-00) but System.IO 4.1.0-rc4-24217-00 was not found. An approximate best match of System.IO 4.1.0-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.IO.FileSystem (>= 4.0.1-rc4-24217-00) but System.IO.FileSystem 4.0.1-rc4-24217-00 was not found. An approximate best match of System.IO.FileSystem 4.0.1-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Reflection (>= 4.1.0-rc4-24217-00) but System.Reflection 4.1.0-rc4-24217-00 was not found. An approximate best match of System.Reflection 4.1.0-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Resources.ResourceManager (>= 4.0.1-rc4-24217-00) but System.Resources.ResourceManager 4.0.1-rc4-24217-00 was not found. An approximate best match of System.Resources.ResourceManager 4.0.1-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Runtime (>= 4.1.0-rc4-24217-00) but System.Runtime 4.1.0-rc4-24217-00 was not found. An approximate best match of System.Runtime 4.1.0-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Runtime.Extensions (>= 4.1.0-rc4-24217-00) but System.Runtime.Extensions 4.1.0-rc4-24217-00 was not found. An approximate best match of System.Runtime.Extensions 4.1.0-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Runtime.InteropServices (>= 4.1.0-rc4-24217-00) but System.Runtime.InteropServices 4.1.0-rc4-24217-00 was not found. An approximate best match of System.Runtime.InteropServices 4.1.0-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Text.Encoding (>= 4.0.11-rc4-24217-00) but System.Text.Encoding 4.0.11-rc4-24217-00 was not found. An approximate best match of System.Text.Encoding 4.0.11-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1603: Microsoft.Cci 4.0.0-rc4-24217-00 depends on System.Threading (>= 4.0.11-rc4-24217-00) but System.Threading 4.0.11-rc4-24217-00 was not found. An approximate best match of System.Threading 4.0.11-rc4-24217-01 was resolved.
/Users/joperezr/git/core-setup/packages/microsoft.dotnet.buildtools/2.0.0-servicing-02104-04/lib/tool-runtime/project.csproj : error NU1701: Package 'Microsoft.Net.Compilers.Targets.NetCore 0.1.5-dev' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.
```

joperezr: I'm fine if you guys rather have me fix the package dependencies instead if that is what we want.

eerhardt: I think the cleanest thing to do here would be to fix the dependencies.  But I don't think that is feasible at this moment.

The next step down would be to suppress these warnings in project.csproj in buildtools.  Outside repos shouldn't have to do workarounds for buildtools' problems.  buildtools should fix them itself.

joperezr: sounds good, I'll fix that from buildtools side then

joperezr: PR for buildtools changes are up: https://github.com/dotnet/buildtools/pull/1731  

Once that is merged and buildtools is available I'll update this PR.

weshaggard: While I'm fine with disabling this in the project directly. For the CCI package warnings we could also fix it by uploading those packages to nuget.org. Ultimately we want to update the CCI package but we could fix it by uploading those dependent packages.

BuildToolsVersion.txt
```diff
@@ -1 +1 @@
-2.0.0-servicing-02029-02
\ No newline at end of file
+1.0.0-servicing-02110-02
```
eerhardt: `1.0.0` ??

joperezr: Good catch! fixing...

init-tools.sh
```diff
@@ -66,15 +66,15 @@ if [ ! -e $__DOTNET_PATH ]; then
             *)
                 echo "Unsupported OS '$OSName' detected. Downloading linux-x64 tools."
                 OS=Linux
-                __DOTNET_PKG=dotnet-dev-linux-x64
+                __DOTNET_PKG=dotnet-sdk-${__DOTNET_TOOLS_VERSION}-linux-x64
```
weshaggard: For consistency we should do the same thing we did in the other init-tools.sh scripts and factor this out a little differently.

weshaggard: Although if you didn't do it in the release branches yet then we can do that separately.

eerhardt: What's our plan for keeping these scripts in sync?

init-tools.sh
```diff
@@ -46,18 +46,18 @@ if [ ! -e $__DOTNET_PATH ]; then
...
             Linux)
-                __DOTNET_PKG=dotnet-dev-linux-x64
+                __DOTNET_PKG=dotnet-sdk-${__DOTNET_TOOLS_VERSION}-linux-x64
                 OS=Linux
 
                 if [ -e /etc/redhat-release ]; then
                     redhatRelease=$(</etc/redhat-release)
                     if [[ $redhatRelease == "CentOS release 6."* || $redhatRelease == "Red Hat Enterprise Linux Server release 6."* ]]; then
-                        __DOTNET_PKG=dotnet-dev-rhel.6-x64
+                        __DOTNET_PKG=dotnet-sdk-${__DOTNET_TOOLS_VERSION}-rhel.6-x64
```
eerhardt: @janvorli - do we have a `2.0.0` rhel.6 SDK?  I assume 'no'.

janvorli: We do. I needed to create that as a new bootstrap cli when corefx was migrated to 2.0.0

eerhardt: Awesome!  Then this works.

init-tools.sh
joperezr: Yeah I only did it so far on master branches of corefx and coreclr, release branches haven't been changed yet. If we want that change in, I can do it in a separate PR.

>What's our plan for keeping these scripts in sync?

AFAIK there is no plan yet. We have tried to make them as short and consice as possible by moving all of the shared logic to the init-tools scripts on buildtools, and have these just as a bootstrap step. We could do what we do in buildtools repo, where we initialize most of the tools using the CLI bootstrap scripts, but even then we would still have to keep some small scripts that will call into those.

joperezr:
@weshaggard @eerhardt is this good to go? this is the only one missing from all of release branches.

init-tools.sh
eerhardt: One way of making these shorter would be to dogfood our `dotnet-install` scripts that we tell our customers to use.  https://github.com/dotnet/cli/blob/master/scripts/obtain/

Then when a new platform/RID comes along, we add it to the one place (the `dotnet-install` scripts) and this wouldn't even have to change.

weshaggard: >What's our plan for keeping these scripts in sync?

We don't have an implemented plan yet. The general idea is to have auto-updated scripts where we can submit PR's to update these common scripts.

>One way of making these shorter would be to dogfood our dotnet-install scripts that we tell our customers to use. https://github.com/dotnet/cli/blob/master/scripts/obtain/

We experimented with this approach in BuildTools and we found that it causes some very interesting failure modes which completely block folks if we cannot download the script (I know same could be said about the CLI itself), and there is also security concerns pulling a script to bootstrap. We can definitely think about it some more but we don't have a clear answer for these yet.

janvorli: One quick idea on a possible approach: we might use a git submodule with all the stuff that we want to share between our repos. Besides the init-tools.sh, it could also contain platform detection code that our build.sh scripts have.

weshaggard: Yes that is an idea but taking a dependency on git submodules in all our core repo's has its own set of challenges as well. Now folks  need to know and remember to run git submodule update when syncing.

PR #3304, opened 2017-10-13
Mark as stable for 1.0.8

safern:
cc: @weshaggard


pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -42,7 +42,7 @@
     "Libuv": "1.9.2",
     "Microsoft.NETCore.Platforms": "1.0.2",
     "System.Net.Http": "4.1.3",
-    "runtime.native.System.Security.Cryptography": "4.0.2"
+    "runtime.native.System.Security.Cryptography": "4.0.3"
   },
```
weshaggard: Please be sure to check any of the new package versions we are shipping to ensure they are explicitly listed here (assuming they are in Microsoft.NETCore.App).

safern: Just validated this and looks like expected. Will go ahead and merge

PR #3305, opened 2017-10-13
Mark packages as stable for 1.1.5

safern:
cc: @weshaggard


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -70,7 +70,7 @@
       <Version>1.0.7</Version>
```
weshaggard: shouldn't this become 1.0.8?

pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -43,7 +43,7 @@
     "Microsoft.DiaSymReader.Native": "1.4.1",
     "Libuv": "1.9.2",
     "runtime.native.System.IO.Compression": "4.3.1",
-    "runtime.native.System.Security.Cryptography": "4.3.2",
+    "runtime.native.System.Security.Cryptography": "4.3.3",
     "runtime.native.System.Security.Cryptography.Apple": "4.3.1",
     "runtime.native.System.Security.Cryptography.OpenSsl": "4.3.2"
```
weshaggard: Please be sure to check any of the new package versions we are shipping to ensure they are explicitly listed here (assuming they are in Microsoft.NETCore.App).

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
safern: Yup you're right. Will update that.

safern:
@weshaggard updated the version that needs to be fixed, and added OpenSsl in project.json and in Microsoft.NETCore.App.pkgproj for backward compatibility.

PR #3309, opened 2017-10-13
Add missing dependency for servicing 1.0.8

safern:
cc: @weshaggard

### DIFF
--- a/pkg/projects/Microsoft.NETCore.App/project.json
+++ b/pkg/projects/Microsoft.NETCore.App/project.json
@@ -34,6 +34,7 @@
     "System.Security.Cryptography.Encoding": "4.0.0",
     "System.Security.Cryptography.Primitives": "4.0.0",
     "System.Security.Cryptography.X509Certificates": "4.1.1",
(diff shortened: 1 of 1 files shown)


pkg/projects/Microsoft.NETCore.App/project.json
```diff
@@ -34,6 +34,7 @@
     "System.Security.Cryptography.Encoding": "4.0.0",
     "System.Security.Cryptography.Primitives": "4.0.0",
     "System.Security.Cryptography.X509Certificates": "4.1.1",
+    "System.Security.Cryptography.OpenSsl": "4.0.1",
```
weshaggard: nit: we generally try to keep these in alphabetical order to we can more easily find packages in the log list.

safern: Ok. I will submit a PR to update this json files to fix the sorting.

weshaggard: don't worry about it. You can get it next time :)

Issue #3280, opened 2017-10-06
Deps only packages


Petermarcu:
We need to provide packages that install just our dependencies on each distro to facilitate people deploying self contained applications.

PR #3317, opened 2017-10-16
Enable building debian installer for Dotnet Runtime Dependencies package

rakeshsinghranchi:
Enable building meta deb package for all supported debian distro and versions. This meta package installs the set of system packages that runtime exhibits dependency on. It will be very useful for self-contained application where system dependencies can be installed via this newly built packages.

Fixes https://github.com/dotnet/core-setup/issues/3280


Petermarcu:
LGTM but someone with more context on how this works should also give a +1

wli3:
@rakeshsinghranchi  that is package you were talking about that only has native dependency but no content?  So all the other packages like dotnet-runtime don't have to be dirstro specific

Should package like dotnet-runtime depend on this package?

wli3:
like that?
image

rakeshsinghranchi:
@wli3 , the purpose of this meta package is to support self-contained deployment scenario. The existing native installers will remain as is . Adding another installer ( meta package ) in the dependency layer will add an overhead to the end users consuming the packages manually. 

Regarding the packaging name, I would let @Petermarcu and @leecow share their thoughts here .

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
```diff
@@ -0,0 +1,49 @@
...
+    "license": {
+        "type": "MIT",
+        "full_text": "Copyright (c) 2017 Microsoft\nPermission is hereby granted, free of charge, to any person obtaining a copy\nof this software and associated documentation files (the \"Software\"), to deal\nin the Software without restriction, including without limitation the rights\nto use, copy, modify, merge, publish, distribute, sublicense, and/or sell\ncopies of the Software, and to permit persons to whom the Software is\nfurnished to do so, subject to the following conditions:\nThe above copyright notice and this permission notice shall be included in all\ncopies or substantial portions of the Software.\nTHE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\nIMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\nFITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE\nAUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER\nLIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,\nOUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE\nSOFTWARE."
+    },
+
+    "debian_dependencies":{
```
leecow: We've seen some situations (eg the Debian 9 Docker image) where gnupg is a required dependency.

leecow:
I would like to see the expected customer usage of this package written out.

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
rakeshsinghranchi: @Petermarcu , who can help provide context on why gnupg is a required dependency for runtime in Debian 9 ?

rakeshsinghranchi:
cc @Petermarcu 
@leecow , yes, I am assuming there will be usage added to self-contained deployment page and in core-setup readme page .

Petermarcu:
@rakeshsinghranchi to @wli3 's comment about this being a dependency of the runtime. Would there be a benefit of those packages being more reusable in any way?

rakeshsinghranchi:
@Petermarcu ,
As discussed with @wli3 couple of months back,  the benefit will lead to building single distro agnostic runtime debs and upstream installers ( assuming SDK and ASP.NET do not exhibit any system package dependencies on their own) . If this approach is implemented fully and correctly, new distro support addition will require adding new meta package in core-setup branch.

However, it will still require building meta package for each distro versions.Also,Repo feed update to each distro and version remains the same.

Petermarcu:
The metapackages are these deps packages right? It sure would be nice to just have a single deb and or rpm instead of having to build them for each distro. That would also make it so if we can just get these metapackages into the distros, then we have much simpler instructions to get them.

src/pkg/packaging/deb/package.targets
```diff
@@ -195,6 +195,7 @@
   </Target>
 
   <Target Name="GenerateSharedFrameworkDeb">
+
```
eerhardt: unnecessary change

src/pkg/packaging/deb/package.targets
```diff
@@ -283,6 +284,88 @@
...
+    <RemoveDir Condition="Exists('$(debIntermediatesDir)')" Directories="$(debIntermediatesDir)" />
+    <MakeDir Directories="$(debIntermediatesDir)" />
+
+    <!-- Compute ICU Version --> 
+    <Exec Command="apt-cache search --names-only '^libicu[0-9]+$' | cut -d' ' -f1" 
+          ConsoleToMSBuild="true"> 
+       <Output TaskParameter="ConsoleOutput" PropertyName="LibIcuPackageName" /> 
+    </Exec> 
+	
+	<!-- Compute LIBSSL Version -->
```
eerhardt: (nit) you are using tabs here.  we should stick with spaces.

src/pkg/packaging/deb/package.targets
```diff
@@ -283,6 +284,88 @@
...
+      <debLayoutPackageRoot>$(debLayoutDirectory)package_root</debLayoutPackageRoot>
+      <debLayoutSamples>$(debLayoutDirectory)samples</debLayoutSamples>
+      <debLayoutDocs>$(debLayoutDirectory)docs</debLayoutDocs>
+    </PropertyGroup>
+
+    <RemoveDir Condition="Exists('$(debIntermediatesDir)')" Directories="$(debIntermediatesDir)" />
+    <MakeDir Directories="$(debIntermediatesDir)" />
+
+    <!-- Compute ICU Version --> 
+    <Exec Command="apt-cache search --names-only '^libicu[0-9]+$' | cut -d' ' -f1"
```
eerhardt: This appears to be duplicated below.  Can you refactor it to eliminate the duplication?

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
```diff
@@ -0,0 +1,49 @@
+{
+    "maintainer_name":"Microsoft",
+    "maintainer_email": "dotnetcore@microsoft.com",
+
+    "package_name": "%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME%",
+    "install_root": "/usr/share/dotnet",
+
+    "short_description": "%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME% %RUNTIME_DEPS_VERSION%",
+    "long_description": ".NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform and is supported by Microsoft. We hope you enjoy using it! If you do, please consider joining the active community of developers that are contributing to the project on GitHub (https://github.com/dotnet/core). We happily accept issues and PRs.",
```
eerhardt: This package isn't really the .NET Core development platform.  Can we update this?

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
```diff
@@ -0,0 +1,49 @@
+{
+    "maintainer_name":"Microsoft",
+    "maintainer_email": "dotnetcore@microsoft.com",
+
+    "package_name": "%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME%",
+    "install_root": "/usr/share/dotnet",
```
eerhardt: Is this the best place for this?  Do we technically install anything here?

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
```diff
@@ -0,0 +1,49 @@
...
+        "type": "MIT",
+        "full_text": "Copyright (c) 2017 Microsoft\nPermission is hereby granted, free of charge, to any person obtaining a copy\nof this software and associated documentation files (the \"Software\"), to deal\nin the Software without restriction, including without limitation the rights\nto use, copy, modify, merge, publish, distribute, sublicense, and/or sell\ncopies of the Software, and to permit persons to whom the Software is\nfurnished to do so, subject to the following conditions:\nThe above copyright notice and this permission notice shall be included in all\ncopies or substantial portions of the Software.\nTHE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\nIMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\nFITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE\nAUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER\nLIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,\nOUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE\nSOFTWARE."
+    },
+
+    "debian_dependencies":{
+    "libc6":{},
+	"libcurl3":{},
```
eerhardt: (nit) tabs here vs spaces in other spots.

src/pkg/packaging/deb/package.targets
```diff
@@ -283,6 +284,88 @@
           UseHardlinksIfPossible="False" />
 
   </Target>
+  
+  <Target Name="GenerateRuntimeMetaDeb">
```
eerhardt: This may just be a personal preference, but I don't really like the name "MetaDeb".  It doesn't really tell me what this package is. Can this be renamed to "RuntimeDependenciesDeb"?  That gives the reader a better indication of what this thing is.

dir.props
```diff
@@ -380,6 +383,8 @@
     <HostFxrDebPkgName>$(HostFxrDebPkgName.ToLower())</HostFxrDebPkgName>
     <SharedFxDebPkgName>$(DotnetRuntimeString)$(SharedFrameworkNugetVersion)</SharedFxDebPkgName>
     <SharedFxDebPkgName>$(SharedFxDebPkgName.ToLower())</SharedFxDebPkgName>
+    <RuntimeMetaDebPkgName>$(DotnetRuntimeMetaPackageString)$(HostResolverVersion)</RuntimeMetaDebPkgName>
```
eerhardt: It is incorrect to use `$(HostResolverVersion)` version here.  It should be `$(ProductVersion)`

dir.props
```diff
@@ -263,10 +263,12 @@
     <DotnetHostString>dotnet-host-</DotnetHostString>
     <DotnetHostFxrString>dotnet-hostfxr-</DotnetHostFxrString>
     <DotnetRuntimeString>dotnet-runtime-</DotnetRuntimeString>
+    <DotnetRuntimeMetaPackageString>dotnet-runtime-dep-</DotnetRuntimeMetaPackageString>
```
eerhardt: These names should align.  It doesn't make sense for the variable to be named `RuntimeMetaPackage` but the value to be `runtime-dep`.

eerhardt: The Docker container uses the term `runtime-deps`.  We should align here and add a `s`.  /cc @michaelsimons

eerhardt:
To make this "complete", it feels like our other debian packages should depend on this new package.  That way the dependencies are always aligned, and in a single place.  This is how the Docker images are structured, as far as I know.  The "deps" image has all the dependencies, and then the image that contains the shared fx depends on the "deps" image.  /cc @MichaelSimons

dir.props
MichaelSimons: I agree - `runtime-deps` - see https://github.com/dotnet/dotnet-docker/blob/master/README.md#L11

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
rakeshsinghranchi: install_root is a required field for package schema. No, no payload is installed in this location.

src/pkg/packaging/deb/package.targets
rakeshsinghranchi: Are you referring newline ? Other targets too have a newline, following the same pattern here.

eerhardt: 4 out of the 5 existing targets in this file don't start with a newline.

rakeshsinghranchi: You are right. I noticed just the one with the extra newline. Will fix this one and TestDebuild target.

rakeshsinghranchi:
I have opened a separate issue https://github.com/dotnet/core-setup/issues/3325 to track the work that involves exhibiting dependency on this runtime-deps package . Goal here is to get this PR complete and keep #3325 as the follow up WI to complete the scenario for building distro agnostic runtime and upstream packages.

src/pkg/packaging/deb/package.targets
rakeshsinghranchi: will take care of this as part of #3325

src/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json
```diff
@@ -0,0 +1,49 @@
+{
+    "maintainer_name":"Microsoft",
+    "maintainer_email": "dotnetcore@microsoft.com",
+
+    "package_name": "%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME%",
+    "install_root": "/usr/share/dotnet",
+
+    "short_description": "%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME% %RUNTIME_DEPS_VERSION%",
+    "long_description": ".NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. This package installs all the system dependencies for Dotnet Core Runtime.",
```
eerhardt: I think we should explicitly say that this doesn't install the .NET Core development platform itself, but just its dependencies.

src/pkg/packaging/deb/package.targets
eerhardt: It feels wrong to add duplication now, and plan on refactoring it later as part of a proposal that we aren't sure we are going to do.  I think this duplication shouldn't be that hard to fix now.

PR #3320, opened 2017-10-17
Add auto dependency flow repo API

eerhardt:
I first needed to fix the build for the new 2.0.0 RTM version of the .NET Core SDK.

The 2nd commit implements the auto-dependency flow.


dependencies.props
```diff
@@ -16,14 +16,18 @@
...
-    <PlatformPackageVersion>2.0.0</PlatformPackageVersion>
+    <MicrosoftNETCorePlatformsPackageVersion>2.0.0</MicrosoftNETCorePlatformsPackageVersion>
     <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.4.2-servicing-25708-01</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
     <MicrosoftPrivateCoreFxUAPPackageVersion>4.4.0-preview3-25526-01</MicrosoftPrivateCoreFxUAPPackageVersion>
     <MicrosoftNETCoreRuntimeCoreCLRPackageVersion>2.0.2-servicing-25708-01</MicrosoftNETCoreRuntimeCoreCLRPackageVersion>
-    <MicrosoftNETCoreRuntimeJitPackageVersion>$(MicrosoftNETCoreRuntimeCoreCLRPackageVersion)</MicrosoftNETCoreRuntimeJitPackageVersion>
-    <NETStandardVersion>2.0.0</NETStandardVersion>
-    <DiaSymReaderNativeVersion>1.4.1</DiaSymReaderNativeVersion>
-    <WcfVersion>4.4.0-preview3-25503-01</WcfVersion>
+    <NETStandardLibraryVersion>2.0.0</NETStandardLibraryVersion>
```
dagood: `NETStandardLibraryVersion` -> `NETStandardLibraryPackageVersion` ?

eerhardt: doh.  Nice catch.

eerhardt:
> Of the changes for CLI 2.0.0, I don't understand the DisableImplicitAssetTargetFallback and NoWarn,

This is a bit complicated, but I'll try to explain.

Back in the `project.json` world, we had an alternate `dotnet test` design.  Instead of using the `dotnet test` execution stuff like we do today from https://github.com/microsoft/vstest, we had a different way of running tests.  Basically, we had a convention approach where we would look for assemblies with a name formatted in a certain way, and `dotnet test` would execute that assembly as the "Main" assembly.

There are tests for the `dotnet.exe` host that ensure the way that `dotnet test` was invoking the host for these test harnesses continues to work.  The tests are using a bunch of pre-release, old packages/assemblies that are no longer being serviced and supported.  

Moving to the 2.0.0 SDK, NuGet no longer can restore these old packages by default.  So we need to add these properties to these projects in order to keep the old restore behavior.

I didn't want to nuke these tests in this change.  I've logged https://github.com/dotnet/core-setup/issues/3323 for us to look into whether deleting these tests is the right thing to do or not.

PR #3322, opened 2017-10-18
Use new publish mechanism in core-setup

wtgodbe:
This should allow us to publish to the dotnetfeed transport feed in core-setup. Will also require adding:

`PB_TransportFeedAccessToken=[AzureKeyVault=EngKeyVault,SecretName=dotnetfeed-storage-access-key-1]`

Into the pipebuild definition. Will queue off a private test build after getting some feedback.

@karajas @markwilkie @chcosta @jcagme PTAL

### DIFF
--- a/init-tools.msbuild
+++ b/init-tools.msbuild
@@ -3,9 +3,10 @@
     <TargetFramework>netcoreapp1.0</TargetFramework>
     <EnableDefaultItems>false</EnableDefaultItems>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
-    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)Tools/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath>$(ToolsDir)/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.DotNet.BuildTools" Version="$(BuildToolsPackageVersion)" />
+    <PackageReference Include="$(FeedTasksPackage)" Version="$(FeedTasksPackageVersion)" />
   </ItemGroup>
 </Project>
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -7,6 +7,8 @@
   <UsingTask TaskName="ListAzureBlobs" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
 
+  <Import Project="$(PackagesDir)/$(FeedTasksPackage)/$(FeedTasksPackageVersion)/build/$(FeedTasksPackage).targets" />
+
   <Target Name="Build"
           DependsOnTargets="PublishToAzure;PublishDebFilesToDebianRepo;PublishDebToolPackageToFeed;PublishFinalOutput" />
   
@@ -149,8 +151,8 @@
                    ForcePublish="true" />
   </Target>
 
-  <Target Name="PublishCoreHostPackagesToFeed"
-          DependsOnTargets="CheckIfAllBuildsHavePublished;DownloadCoreHostPackages;SignSymbolPackages;DoPushCoreHostPackagesToFeed"
+  <Target Name="PublishCoreHostPackages"
+          DependsOnTargets="CheckIfAllBuildsHavePublished;DownloadCoreHostPackages;SignSymbolPackages;DoPushCoreHostPackagesToFeed;DoPushCoreHostPackagesToAzure"
           Condition="'@(_MissingBlobNames)' == '' AND '$(NuGetFeedUrl)' != ''">
     <Error Condition="'$(NuGetFeedUrl)' ==''" Text="Missing required property NuGetFeedUrl" />
     <Error Condition="'$(NuGetApiKey)' == ''" Text="Missing required property NuGetApiKey" />
@@ -239,6 +241,48 @@
     <ExecWithRetriesForNuGetPush Condition="'@(_DownloadedSymbolsPackages)' != ''"
                                  Command="$(NuGetPushSymbolsCommand) %(_DownloadedSymbolsPackages.Identity)"
                                  IgnoredErrorMessagesWithConditional="@(IgnorableErrorMessages)" />
+  </Target>
+
+  <Target Name="DoPushCoreHostPackagesToAzure"
+          DependsOnTargets="PublishToAzureBlobFeed;PublishSymbolsToAzureBlobFeed"
+          Condition="'$(OfficialPublish)' == 'true'"/>
+
+  <Target Name="PublishToAzureBlobFeed">
+    <Error Condition="'$(PackagesUrl)'==''" Text="Missing property PackagesUrl" />
+    <Error Condition="'$(TransportFeedAccessToken)'==''" Text="Missing property TransportFeedAccessToken" />
+    <PropertyGroup>
+      <ExpectedFeedUrl>$(PackagesUrl)</ExpectedFeedUrl>
+      <PublishFlatContainer Condition="'$(PublishFlatContainer)' == ''">false</PublishFlatContainer>
+    </PropertyGroup>
+    <ItemGroup>
+      <ItemsToPush Remove="*.nupkg" />
+      <ItemsToPush Include="@(_DownloadedStandardPackages)"/>
+    </ItemGroup>
+    <PushToBlobFeed ExpectedFeedUrl="$(ExpectedFeedUrl)"
+                    AccountKey="$(TransportFeedAccessToken)"
+                    ItemsToPush="@(ItemsToPush)"
+                    PublishFlatContainer="$(PublishFlatContainer)"
+                    Overwrite="$(OverwriteOnPublish)" 
+                    IndexDirectory="$(ObjDir)"/>
+  </Target>
+
+  <Target Name="PublishSymbolsToAzureBlobFeed">
+    <Error Condition="'$(SymbolPackagesUrl)'==''" Text="Missing property SymbolPackagesUrl" />
+    <Error Condition="'$(TransportFeedAccessToken)'==''" Text="Missing property TransportFeedAccessToken" />
+    <PropertyGroup>
+      <ExpectedFeedUrl>$(SymbolPackagesUrl)</ExpectedFeedUrl>
+      <PublishFlatContainer Condition="'$(PublishFlatContainer)' == ''">true</PublishFlatContainer>
+    </PropertyGroup>
+    <ItemGroup>
+      <ItemsToPush Remove="*.nupkg" />
+      <ItemsToPush Include="@(_DownloadedSymbolsPackages)"/>
+    </ItemGroup>
+    <PushToBlobFeed ExpectedFeedUrl="$(ExpectedFeedUrl)"
+                    AccountKey="$(TransportFeedAccessToken)"
+                    ItemsToPush="@(ItemsToPush)"
+                    PublishFlatContainer="$(PublishFlatContainer)"
+                    Overwrite="$(OverwriteOnPublish)" 
+                    IndexDirectory="$(ObjDir)"/>
   </Target>
 
   <Target Name="PublishDebToolPackageToFeed"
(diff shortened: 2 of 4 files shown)


init-tools.msbuild
```diff
@@ -3,9 +3,23 @@
     <TargetFramework>netcoreapp1.0</TargetFramework>
     <EnableDefaultItems>false</EnableDefaultItems>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
-    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)Tools/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath>$(ToolsDir)/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
+    <FeedTasksVersion>1.0.0-prerelease-02111-01</FeedTasksVersion>
```
eerhardt: What's the strategy for keeping this version updated?

publish/publish.proj
```diff
@@ -7,12 +7,19 @@
   <UsingTask TaskName="ListAzureBlobs" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
 
+  <PropertyGroup>
+    <FeedsDir Condition="'$(RunningOnCore)' != 'true'">$(ToolsDir)Microsoft.Dotnet.Build.Tasks.Feed/net46/</FeedsDir>
+    <FeedsDir Condition="'$(RunningOnCore)' == 'true'">$(ToolsDir)microsoft.dotnet.build.tasks.feed/netstandard1.5/</FeedsDir>
```
eerhardt: I wish we would start doing this differently.

Instead of following this old pattern that we use in BuildTools, where we do a bunch of restoring, copying, probing, etc manually, we would take advantage of the MSBuild features built into NuGet.  This isn't how we tell customers to provide custom build tasks.

Instead, it would be better if the `microsoft.dotnet.build.tasks.feed` package had a `build` folder that injected Tasks/Targets/etc into projects that `<PackageReference>` to it.

This is the normal way our customers inject custom MSBuild constructs into projects.

/cc @weshaggard

eerhardt: See https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package for more info on how this works.

Basically, if you have .targets and .props files in the `build` folder of your NuGet package, they are automatically included in the project.

init-tools.msbuild
karajas: Currently it's just checked in. Is it better to make it similar to how BuildToolsVersion.txt works?

eerhardt: My experience is that having hard-coded version numbers that have no strategy for getting updated regularly get out of date/sync really quick.  And then when someone does need to update them, there are a bunch of breaking changes they need to work through.

I think it would be better if we had an automated approach to updating this version, just like we do with other dependencies.

publish/publish.proj
```diff
@@ -226,6 +233,48 @@
...
 
+  <Target Name="DoPushCoreHostPackagesToAzure"
+          DependsOnTargets="DoPushStandardPackagesToAzure;DoPushSymbolPackagesToAzure"
+          Condition="'$(OfficialPublish)' == 'true'"/>
+
+  <Target Name="DoPushStandardPackagesToAzure">
+    <PropertyGroup>
+      <RelativePath>$(PackagesRelativePath)</RelativePath>
+      <RelativePath Condition="'$(RelativePath)' == ''">$(ConfigurationGroup)/pkg</RelativePath>
+      <ExpectedFeedUrl>https://$(TransportFeedAccountName).blob.core.windows.net/$(TransportFeedContainerName)/$(RelativePath)/index.json</ExpectedFeedUrl>
```
chcosta: I'm surprised deriving the feed url is not part of the feed api

publish/publish.proj
```diff
@@ -226,6 +233,48 @@
...
+    </ItemGroup>
+    <PushToBlobFeed ExpectedFeedUrl="$(ExpectedFeedUrl)"
+                    AccountKey="$(TransportFeedAccessToken)"
+                    ItemsToPush="@(ItemsToPush)"
+                    PublishFlatContainer="$(PublishFlatContainer)"
+                    Overwrite="$(OverwriteOnPublish)" 
+                    IndexDirectory="$(ObjDir)"/>
+  </Target>
+
+  <Target Name="DoPushSymbolPackagesToAzure">
```
chcosta: You may want to consider changing `DoPushSymbolPackagesToAzure` and `DoPushStandardPackagesToAzure` to `GatherSymbolPackages`, `GatherStandardPackages`, and a `PushPackagesToAzure` target that just batches on the `ExpectedFeedUrl` which would be metadata of the `ItemsToPush` itemgroup.

publish/publish.proj
wtgodbe: @karajas is this something you can implement in the buildtools package?

publish/publish.proj
wtgodbe: I believe the api just expects a URL, @karajas is that correct?

karajas: The URL needs to be specified currently.

publish/publish.proj
karajas: The thing is this will break how we've told CLI and CoreCLR how to update. Do we want this change only for core-setup?

karajas: Do we have any examples of props being imported via a PackageReference in core-setup? I agree it's the right thing to do but we'd really like to unblock this today if possible

init-tools.msbuild
chcosta: If there was a small buildtools package with a props file which defined the packages / versions that buildtools produced, then init-tools.cmd/sh could restore that project directly and init-tools.msbuild could use the props file to determine the set of packages to restore.

weshaggard: We can definitely add an updater to dependencies.props to get it auto-updated which is what I would expect.

publish/publish.proj
weshaggard: I would like to restructure the package such that we can get a common layout and support targets for this task library but from init-tools.msbuild project not necessary in the publish.proj. This is not a normal project and so we aren't going to call restore on it in order to use a PackageReference directly in this project.

publish/publish.proj
weshaggard: It shouldn't be hardcoded in the task library I actually expect this to be passed in from the build definition though not set here.

publish/publish.proj
```diff
@@ -7,12 +7,19 @@
   <UsingTask TaskName="ListAzureBlobs" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
   <UsingTask TaskName="UploadToAzure" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
 
+  <PropertyGroup>
```
weshaggard: Why isn't this using BuildToolsTaskDir property instead of directly doing this here?

wtgodbe: Where is that property defined? I'm not familiar with it

weshaggard: It is in build.common.props from BuildTools.

publish/publish.proj
weshaggard: It allows the task dll loading to be handled by a targets file that comes in the package.

wtgodbe:
Any final comments? I believe this should be ready to merge

init-tools.cmd
```diff
@@ -61,8 +61,8 @@ if NOT exist "%DOTNET_LOCAL_PATH%" (
 
 if exist "%BUILD_TOOLS_PATH%" goto :afterbuildtoolsrestore
 echo Restoring BuildTools version %BUILDTOOLS_VERSION%...
-echo Running: "%DOTNET_CMD%" restore "%INIT_TOOLS_RESTORE_PROJECT%" --no-cache --packages %PACKAGES_DIR% --source "%BUILDTOOLS_SOURCE%" /p:BuildToolsPackageVersion=%BUILDTOOLS_VERSION% >> "%INIT_TOOLS_LOG%"
-call "%DOTNET_CMD%" restore "%INIT_TOOLS_RESTORE_PROJECT%" --no-cache --packages %PACKAGES_DIR% --source "%BUILDTOOLS_SOURCE%" /p:BuildToolsPackageVersion=%BUILDTOOLS_VERSION% >> "%INIT_TOOLS_LOG%"
+echo Running: "%DOTNET_CMD%" restore "%INIT_TOOLS_RESTORE_PROJECT%" --no-cache --packages %PACKAGES_DIR% --source "%BUILDTOOLS_SOURCE%" /p:BuildToolsPackageVersion=%BUILDTOOLS_VERSION% /p:ToolsDir=%TOOLRUNTIME_DIR% >> "%INIT_TOOLS_LOG%"
```
karajas: This change is no longer needed.

init-tools.sh
```diff
@@ -98,8 +98,8 @@ fi
 
 if [ ! -e $__BUILD_TOOLS_PATH ]; then
     echo "Restoring BuildTools version $__BUILD_TOOLS_PACKAGE_VERSION..."
-    echo "Running: $__DOTNET_CMD restore \"$__INIT_TOOLS_RESTORE_PROJECT\" --no-cache --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE /p:BuildToolsPackageVersion=$__BUILD_TOOLS_PACKAGE_VERSION" >> $__init_tools_log
-    $__DOTNET_CMD restore "$__INIT_TOOLS_RESTORE_PROJECT" --no-cache --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE /p:BuildToolsPackageVersion=$__BUILD_TOOLS_PACKAGE_VERSION >> $__init_tools_log
+    echo "Running: $__DOTNET_CMD restore \"$__INIT_TOOLS_RESTORE_PROJECT\" --no-cache --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE /p:BuildToolsPackageVersion=$__BUILD_TOOLS_PACKAGE_VERSION /p:ToolsDir=$__TOOLRUNTIME_DIR" >> $__init_tools_log
```
karajas: This one also.

init-tools.msbuild
karajas: This has now been updated.

publish/publish.proj
karajas: This has now been updated.

init-tools.msbuild
```diff
@@ -3,9 +3,11 @@
     <TargetFramework>netcoreapp1.0</TargetFramework>
     <EnableDefaultItems>false</EnableDefaultItems>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
-    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)Tools/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
+    <BaseIntermediateOutputPath>$(ToolsDir)/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
+    <PackageTargetFallback>$(PackageTargetFallback);netframework</PackageTargetFallback>
```
weshaggard: This fallback shouldn't be needed and should be removed.

publish/publish.proj
weshaggard: Yes can we please be consistent with the target naming, in particular the entry point targets. Lets call it PublishToAzureBlobFeed as in @karajas PR https://github.com/dotnet/corefx/pull/24784/files

publish/publish.proj
```diff
@@ -241,6 +243,48 @@
...
+                    PublishFlatContainer="$(PublishFlatContainer)"
+                    Overwrite="$(OverwriteOnPublish)" 
+                    IndexDirectory="$(ObjDir)"/>
+  </Target>
+
+  <Target Name="DoPushSymbolPackagesToAzure">
+    <PropertyGroup>
+      <RelativePath>$(SymbolsRelativePath)</RelativePath>
+      <RelativePath Condition="'$(RelativePath)' == ''">$(ConfigurationGroup)/symbolpkg</RelativePath>
+      <ExpectedFeedUrl>https://$(TransportFeedAccountName).blob.core.windows.net/$(TransportFeedContainerName)/$(RelativePath)/index.json</ExpectedFeedUrl>
```
weshaggard: Did we figure out why this needs to be defaulted here? Isn't the build definition going to be passing it in?

weshaggard: Otherwise we will not have the extensibility point we need to override the feed end-point

chcosta:
@wtgodbe, can you address the feedback in this PR and get this merged today?  If there's any additional feedback, it can be handled separately.

wtgodbe:
Updated to address feedback, going to merge & queue off a build.

PR #3330, opened 2017-10-19
Fix output directories.

eerhardt:
1. Use PackageOutputPath variable exclusively everywhere instead of our own name for the same path.
2. Split nuget packages and installer assets into separate paths - packages and assets.

This should fix the issues in https://github.com/dotnet/source-build/pull/249#issuecomment-337791094

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -215,6 +215,11 @@
     <TestTargetRid Condition="'$(TestTargetRid)' == ''">$(OutputRid)</TestTargetRid>
   </PropertyGroup>
 
+  <!-- Produce assets into the specified blob feed. -->
+  <PropertyGroup Condition="'$(DotNetOutputBlobFeedDir)' != ''">
+    <AssetOutputPath>$(DotNetOutputBlobFeedDir)assets/</AssetOutputPath>
+  </PropertyGroup>
+
   <!-- Set up the default output and intermediate paths -->
   <PropertyGroup>
     <OSPlatformConfig>$(OutputRid).$(ConfigurationGroup)</OSPlatformConfig>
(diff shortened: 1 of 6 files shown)


src/pkg/packaging/deb/package.targets
```diff
@@ -17,14 +17,7 @@
     </PropertyGroup>
 
     <Exec Command="$(DotnetRestoreCommand) $(dotnetDebToolSource)"/>
-    <Exec Command="$(DotnetToolCommand) pack $(dotnetDebToolSource) --output $(PackagesOutDir)intermediate/ $(VersionSuffixArg)"/>
-
-    <ItemGroup>
-      <pkFiles Include="$(PackagesOutDir)intermediate/dotnet-deb-tool.*.nupkg"
-               Exclude="$(PackagesOutDir)intermediate/*.symbols.nupkg" />
-    </ItemGroup>
-
```
weshaggard: Is this copy not needed any longer?

dir.props
```diff
@@ -230,8 +235,9 @@
     <CoreHostOutputDir>$(BaseOutputRootPath)corehost\</CoreHostOutputDir>
 
     <PackagesBasePath Condition="'$(PackagesBasePath)'==''">$(BinDir)$(OSPlatformConfig)/</PackagesBasePath>
-    <PackagesOutDir Condition="'$(PackagesOutDir)'==''">$(PackagesBasePath)packages/</PackagesOutDir>
-    <SymbolPackagesOutDir Condition="'$(SymbolPackagesOutDir)'==''">$(PackagesOutDir)</SymbolPackagesOutDir>
+    <PackageOutputPath Condition="'$(PackageOutputPath)'==''">$(PackagesBasePath)packages/</PackageOutputPath>
+    <AssetOutputPath Condition="'$(AssetOutputPath)'==''">$(PackageOutputPath)</AssetOutputPath>
+    <SymbolPackageOutputPath Condition="'$(SymbolPackageOutputPath)'==''">$(PackageOutputPath)</SymbolPackageOutputPath>
```
weshaggard: Do we want to default these to a assets and symbols sub-dir by default?

dagood: These being in the same folder as the packages actually means that there are two packages with same ident+version but different contents in the packages dir "feed". Symbol packages are currently spec'd to go in `assets/`. So far, I've just put the symbol packages in the assets dir root. We should figure out where symbol packages should all go because eventually we'll need to gather them up to be able to release a build properly.

dagood: (Deleted old comment--I severely misread this.)

The isolated build has always had symbol packages in the same dir as lib packages, but it would be more like the other Core repos to have a separate symbol package dir. Leaving it the same as it was seems like a safer bet, though.

src/pkg/packaging/deb/package.targets
eerhardt: No, it wasn’t necessary to begin with. What this was doing before was packing the nupkg to an intermediate folder and then copying the nupkg to the output folder. I just made the pack output straight to the final destination. As for the symbol package, this doesn’t get produced by default from my testing. And if it did, we can just shut it off, instead of making it and not copying it.

dir.props
eerhardt: Right - that was my thinking - leave the defaults alone.  I didn't want to change it here.

PR #3335, opened 2017-10-24
Move Version attribute to child element to allow MSBuild 14 to continue building core-setup.tasks.csproj.

eerhardt:
Trying to fix the official build again.


eerhardt:
merged to unblock the build.

PR #3340, opened 2017-10-24
Fix Debian 9 official build 

eerhardt:
by explicitly specifying the OutputRid since we are using a Debian 8.2 image for the Debian 9 leg.


rakeshsinghranchi:
Looks good to me. This is a temporary fix until BuildTools in core-setup is updated to 2.0.0 RTM and Debian.9 deps image is available for building the installers. https://github.com/dotnet/buildtools/issues/1633

PR #3346, opened 2017-10-25
Fix PublishCoreHostPackages target name

karajas:
skip ci please

This should fix the build break.

/cc @wtgodbe @eerhardt @chcosta

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -14,7 +14,7 @@
   
   <Target Name="PublishFinalOutput"
           Condition="'$(Finalize)' == 'true'"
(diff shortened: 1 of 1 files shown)


eerhardt:
In general, we shouldn't be "skip ci" when we are editing MSBuild files.  More or less, you should almost never use "skip ci" unless you are editing a markdown/text file.  Something you are guaranteeing that will not affect the build.

karajas:
I agree 100%, we've had this discussion before :) I'm not sure publish.proj runs in CI.

wtgodbe:
Don't know how I missed this, thought I had ctrl-f'd all of the target name changes. My mistake.

PR #3350, opened 2017-10-25
Merge dev/release/2.0.0 in to master

weshaggard:
PTAL @dagood @eerhardt 

cc @rakeshsinghranchi


eerhardt:
Looks like BuildTools is failing.

Directory.Build.props
```diff
@@ -27,4 +27,9 @@
     <BaseIntermediateOutputPath>$(IntermediateOutputPath)</BaseIntermediateOutputPath>
   </PropertyGroup>
 
+  <!-- Workaround nuget restore warnings turning into errors in the BuildTools tool-runtime/project.csproj -->
```
eerhardt: Why can’t BuildTools do this?  This feels wrong

dependencies.props
```diff
@@ -17,15 +17,18 @@
...
     <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.5.0-preview1-25824-02</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
     <MicrosoftPrivateCoreFxUAPPackageVersion>4.6.0-preview1-25824-02</MicrosoftPrivateCoreFxUAPPackageVersion>
-    <PlatformPackageVersion>2.1.0-preview1-25824-02</PlatformPackageVersion>
     <MicrosoftNETCoreRuntimeCoreCLRPackageVersion>2.1.0-preview1-25824-51</MicrosoftNETCoreRuntimeCoreCLRPackageVersion>
-    <MicrosoftNETCoreRuntimeJitPackageVersion>$(MicrosoftNETCoreRuntimeCoreCLRPackageVersion)</MicrosoftNETCoreRuntimeJitPackageVersion>
-    <MicrosoftNetNativeCompilerPackageVersion>2.0.0-preview-25524-00</MicrosoftNetNativeCompilerPackageVersion>
-    <NETStandardVersion>2.1.0-preview1-25825-02</NETStandardVersion>
-    <DiaSymReaderNativeVersion>1.4.1</DiaSymReaderNativeVersion>
-    <WcfVersion>4.5.0-preview2-25825-02</WcfVersion>
+    <NETStandardVersion>2.1.0-preview1-25824-08</NETStandardVersion>
```
eerhardt: This needs to be `NETStandardLibraryPackageVersion`

eerhardt: Also - you are downgrading the version.

Directory.Build.props
dagood: After looking up `Directory.Build.props` (had no idea this was auto-imported) it sounds like we can create the file in the `tool-runtime` dir with this property and we will be ok. It actually seems like we *should* do this to prevent the repo from modifying the BuildTools init behavior. (And prevent the environment from playing a part, if the repo has no `Directory.Build.props`.)

dependencies.props
```diff
@@ -17,15 +17,18 @@
...
+    <MicrosoftDiaSymReaderNativePackageVersion>1.4.1</MicrosoftDiaSymReaderNativePackageVersion>
+    <WcfVersion>4.5.0-preview2-25824-01</WcfVersion>
```
dagood: This `WcfVersion` property should be removed: the updater below changed to `SystemServiceModelDuplexPackageVersion`. (Probably auto-updated on `master` and caused a merge conflict.)

Directory.Build.props
eerhardt: IMO - the root cause here is the warnings, which should be fixed.  We need to get BuildTools to a "clean" state, which it isn't in right now.  

These package mis-alignments have caused problems up and down repos and we keep pushing fixing them off.  This is just another case of them causing problems.

weshaggard: I don't plan on checking in this workaround I want to make the change in buildtools, minimally to disable the warnings for this project but ideally fixing the packages. I just submitted this as a temp workaround before the EOD yesterday to try and flush out other issues in this merge.

dependencies.props
weshaggard: I had to keep rebasing to deal with conflicts I must have messed this merge up the last time I did it. I will fix it.

Issue #2983, opened 2017-08-05
Don't assume dotnet.exe is the only host


davidfowl:
It seems that the logic in hostfxr.dll assumes dotnet.exe is the process that is activating it. I'm trying to launch the .NET Core host from a native IIS module and I can't exactly replicate the lookup behavior because it assumes the initial search path is the executing process instead of the hive where hostfxr.dll is loaded from:

```
--- Invoked hostfxr [commit hash: 40c565230930ead58a50719c0ec799df77bddee9] main
Own dll path 'C:\Program Files\IIS Express\iisexpress.dll'
Checking if CoreCLR path exists=[C:\Program Files\IIS Express\coreclr.dll]
--- Executing in muxer mode...
Detected a non-standalone application, expecting app.dll to execute.
Treating application 'C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.dll' as a managed executable.
App runtimeconfig.json from [C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.dll]
Runtime config is cfg=C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.runtimeconfig.json dev=C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.runtimeconfig.dev.json
Attempting to read runtime config: C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.runtimeconfig.json
Attempting to read dev runtime config: C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.runtimeconfig.dev.json
Runtime config [C:\Users\dfowler\Documents\GitHub\AspNetCoreModule\samples\SampleServer\bin\Debug\netcoreapp2.0\SampleServer.runtimeconfig.json] is valid=[1]
Ignoring host interpreted additional probing path C:\Users\dfowler\.dotnet\store\x64\netcoreapp2.0 as it does not exist.
--- Resolving FX directory, specified ''
Searching FX directory in [C:\Program Files\IIS Express\]
[...]

PR #3356, opened 2017-10-27
Support shared hosts other than dotnet.exe

steveharter:
The current code in hostfxr sets own_dir to be the location of the currently running executable, which in turn is used to determine the location to search for the runtime. This prevents other hosts such as IIS (w3wp.exe) from using CoreClr when it does its own implementation of dotnet.exe and calls hostfxr.dll itself.

The fix is to use argv[0] as the path to the dotnet.exe which from which own_dir will be obtained. Normally argv[0] will be the current path to the executable. Also similar changes were made to hostpolicy to use argv[0] since it uses own_dir to set up global store paths.

Fixes https://github.com/dotnet/core-setup/issues/2983

Tested by creating a simple Windows console test compiled for x64, and renaming of the `C:\program files\dotnet` directory to `C:\program files\dotnet2` so that the global lookup doesn't occur and cause a false positive on the test. Code:
```C
#include "stdafx.h"
#include "windows.h"

typedef int(*hostfxr_main_fn) (const int argc, const WCHAR* argv[]);

int main()
{
    HMODULE fxrDll = ::LoadLibraryExW(L"C:\\Program Files\\dotnet2\\host\\fxr\\2.0.0\\hostfxr.dll", NULL, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
    if (!fxrDll)
    {
        printf("LoadLibraryExW failed with error %d.\n", GetLastError());
        return 0;
    }

    puts("Loaded hostfxr...");

    hostfxr_main_fn main_fn = (hostfxr_main_fn)GetProcAddress(fxrDll, "hostfxr_main");
    if (!main_fn)
    {
        printf("GetProcAddress failed with error %d.\n", GetLastError());
        return 0;
    }

    WCHAR **new_argv = new WCHAR*[3];
    new_argv[0] = L"C:\\Program Files\\dotnet2\\dotnet.exe";
    // relative paths also work since they are converted to absolute:
    // new_argv[0] = L"..\\..\\..\\..\\..\\..\\Program Files\\dotnet2\\dotnet.exe";
    new_argv[1] = L"exec";
    new_argv[2] = L"C:\\git\\repros\\console\\bin\\Debug\\netcoreapp2.0\\console.dll";

    puts("Invoking...");
    int rc = main_fn(3, (const WCHAR**)new_argv);

    ::FreeLibrary(fxrDll);

    puts("Success!");

    return rc;
}
```

### DIFF
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -66,7 +66,7 @@ bool parse_arguments(
     if (argc >= 1)
     {
         args.own_path = argv[0];
-        if (!pal::realpath(&args.own_path) || !pal::file_exists(args.own_path))
+        if (!args.own_path.empty() && (!pal::realpath(&args.own_path) || !pal::file_exists(args.own_path)))
         {
             trace::warning(_X("Failed to resolve argv[0] as path [%s]. Using location of current executable instead."), args.own_path.c_str());
             args.own_path.clear();
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -66,8 +66,9 @@ bool parse_arguments(
     if (argc >= 1)
     {
         args.own_path = argv[0];
-        if (!pal::realpath(&args.own_path))
+        if (!pal::realpath(&args.own_path) || !pal::file_exists(args.own_path))
         {
+            trace::warning(_X("Failed to resolve argv[0] as path [%s]. Using location of current executable instead."), args.own_path.c_str());
             args.own_path.clear();
         }
     }
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -61,8 +61,19 @@ bool parse_arguments(
     const int argc, const pal::char_t* argv[], arguments_t* arg_out)
 {
     arguments_t& args = *arg_out;
+
+    // Try to use argv[0] as own_path to allow for hosts located elsewhere
+    if (argc >= 1)
+    {
+        args.own_path = argv[0];
+        if (!pal::realpath(&args.own_path))
+        {
+            args.own_path.clear();
+        }
+    }
+
     // Get the full name of the application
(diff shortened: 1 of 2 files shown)


src/corehost/cli/args.cpp
```diff
@@ -61,8 +61,19 @@ bool parse_arguments(
...
 {
     arguments_t& args = *arg_out;
+
+    // Try to use argv[0] as own_path to allow for hosts located elsewhere
+    if (argc >= 1)
+    {
+        args.own_path = argv[0];
+        if (!pal::realpath(&args.own_path))
+        {
+            args.own_path.clear();
```
eerhardt: Do we want to trace this situation, so we know it happened?

src/corehost/cli/args.cpp
```diff
@@ -61,8 +61,19 @@ bool parse_arguments(
...
+        }
+    }
+
     // Get the full name of the application
-    if (!pal::get_own_executable_path(&args.own_path) || !pal::realpath(&args.own_path))
+    if (args.own_path.empty() && (!pal::get_own_executable_path(&args.own_path) || !pal::realpath(&args.own_path)))
     {
         trace::error(_X("Failed to resolve full path of the current executable [%s]"), args.own_path.c_str());
```
eerhardt: When `own_path` is empty, this message isn't going to be the greatest.  But thinking more about it, I guess we will be able to tell it was empty - :)

src/corehost/cli/args.cpp
steveharter: I added a trace. Also checked to see if the actual file exists to prevent any backwards compatibility issues (or forward compatibility issues if the file ends up being used for something).

src/corehost/cli/args.cpp
steveharter: This will be covered by the new check for exists()

PR #3359, opened 2017-10-28
Enable build pipeline for Alpine

janvorli:
Now that the coreclr and corefx packages are produced for Alpine 3.6, we can finally enable producing core-setup ones.

### DIFF
--- a/buildpipeline/pipeline.json
+++ b/buildpipeline/pipeline.json
@@ -30,7 +30,7 @@
           }
         },
         {
-          "Name": "Core-Setup-Linux-Arm-BT",
+          "Name": "Core-Setup-Linux-BT",
           "Parameters": {
             "PB_DockerTag": "centos-6-376e1a3-20174311014331",
             "PB_TargetArchitecture": "x64",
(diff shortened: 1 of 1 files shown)


buildpipeline/pipeline.json
```diff
@@ -45,6 +45,21 @@
           }
         },
         {
+          "Name": "Core-Setup-Linux-BT",
+          "Parameters": {
+            "PB_DistroRid": "alpine.3.6-x64",
+            "PB_DockerTag": "alpine-3.6-3148f11-20171119021156",
+            "PB_TargetArchitecture": "x64",
+            "PB_AdditionalBuildArguments":"-TargetArchitecture=x64 -DistroRid=alpine.3.6-x64 -PortableBuild=false -strip-symbols",
```
eerhardt: `-DistroRid=alpine.3.6-x64`

We no longer use `DistroRid` on core-setup.  Most distros will calculate the Rid if we have the PlatformAbstractions updated correctly.  So it should just calculate the correct RID based on the machine.

For the cases where we can't calculate the RID correctly, use `/p:OutputRid=<rid>`.  We use that for the `tizen` RID.

buildpipeline/pipeline.json
```diff
@@ -30,7 +30,7 @@
           }
         },
         {
-          "Name": "Core-Setup-Linux-Arm-BT",
+          "Name": "Core-Setup-Linux-BT",
```
eerhardt: I think this broke the official build for RHEL6.  The error I see is:

```
2017-10-31T16:08:02.3207280Z /root/coresetup/packages/transport.runtime.linux-x64.microsoft.netcore.runtime.coreclr/2.1.0-preview1-25831-04/tools/crossgen: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /root/coresetup/packages/transport.runtime.linux-x64.microsoft.netcore.runtime.coreclr/2.1.0-preview1-25831-04/tools/crossgen) 
```

eerhardt: I don't think we were crossgen'ing before.

Issue #3377, opened 2017-11-01
RHEL 6 official builds don't work with crossgen enabled


janvorli:
For some reason, when using Core-Setup-Linux-BT in the official build instead of the Core-Setup-Linux-Arm-BT as the related "Name" element, the build fails when trying crossgen some files. The issue is that it tries to use crossgen from transport.runtime.linux-x64.microsoft.netcore.runtime.coreclr instead of the one from the rhel.6-x64 package.
We need to investigate and fix that.

steveharter:
Is this causing the current build failures in master?
```
/root/coresetup/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj(262,5): error MSB3073: The command "/root/coresetup/packages/transport.runtime.linux-x64.microsoft.netcore.runtime.coreclr/2.1.0-preview1-25901-01/tools/crossgen @/root/coresetup/Bin/obj/linux-x64.Release/Microsoft.NETCore.App/linux-x64/linux-x64/crossgen/Microsoft.CSharp.rsp" exited with code 1.
```

steveharter:
@janvorli is this still an active issue?

PR #3378, opened 2017-11-01
Fix Alpine and RHEL official builds

janvorli:
I have broken the RHEL build by my recent change that was supposed to
enable Alpine official builds. The Alpine build didn't work either since
there was a recent change that modified the way we specify the distro
RID and I have missed that.

The RHEL 6 should work with the Core-Setup-Linux-BT name, but for some 
reason it tries to use crossgen from a linux-x64 coreclr package instead of 
the rhel.6-x64 one and obviously fails. I've created issue #3377 to investigate 
and fix that.
The Core-Setup-Linux-Arm-BT that was there before disables the crossgen, so 
that's why it works.

skip ci please

### DIFF
--- a/buildpipeline/pipeline.json
+++ b/buildpipeline/pipeline.json
@@ -30,7 +30,7 @@
           }
         },
         {
-          "Name": "Core-Setup-Linux-BT",
+          "Name": "Core-Setup-Linux-Arm-BT",
           "Parameters": {
             "PB_DockerTag": "centos-6-376e1a3-20174311014331",
             "PB_TargetArchitecture": "x64",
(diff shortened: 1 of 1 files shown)


buildpipeline/pipeline.json
```diff
@@ -45,12 +45,13 @@
           }
         },
         {
-          "Name": "Core-Setup-Linux-BT",
+          "Name": "Core-Setup-Linux-Arm-BT",
           "Parameters": {
             "PB_DistroRid": "alpine.3.6-x64",
```
eerhardt: this can be removed.

buildpipeline/pipeline.json
```diff
@@ -45,12 +45,13 @@
...
             "PB_DockerTag": "alpine-3.6-3148f11-20171119021156",
             "PB_TargetArchitecture": "x64",
-            "PB_AdditionalBuildArguments":"-TargetArchitecture=x64 -DistroRid=alpine.3.6-x64 -PortableBuild=false -strip-symbols",
+            "PB_AdditionalBuildArguments":"-TargetArchitecture=x64 -PortableBuild=false -strip-symbols",
+            "PB_AdditionalMSBuildArguments":"/p:OutputRid=alpine.3.6-x64",
```
eerhardt: This shouldn't be necessary once we move up to the latest .NET Core SDK. However, since we are using 2.0.0 now to build, this is necessary.

Issue #3193, opened 2017-09-12
Host fails cryptically when app is missing runtimeconfig.json


ericstj:
1. dotnet new console
2. dotnet build
3. del bin\debug\netcoreapp2.0\*runtimeconfig.json
4. dotnet bin\debug\netcoreapp2.0\<app>.dll

Expect:
Error message that indicates runtimeconfig.json is missing (as one of the options).

Actual: 
Host assumes application is selfcontained and logs:
```
A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in 'F:\scratch\bin\Debug\netcoreapp2.0\'.
```

I've seen this come up a number of times when folks are deploying their applications and they don't realize what to do in order to fix it.

PR #3382, opened 2017-11-01
Improve message when runtimeconfig.json is missing or has no framework

steveharter:
Add an error message to help users identify the cause of not being able to find hostpolicy.dll when the runtimeconfig.json file is missing or has no framework specified.

When there is no runtimeconfig.json file or no framework specified in that file, the current code treats the app as a standalone app, so the new messages explain that.

Fixes https://github.com/dotnet/core-setup/issues/3193

The two messages are prefixed with this existing message:
```
A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in 'C:\git\repros\console\bin\Debug\netcoreapp2.0\'.
```

The additional message when there is no runtimeconfig.json:
```
Failed to run as a standalone app because there is no 'C:\git\repros\console\bin\Debug\netcoreapp2.0\console.runtimeconfig.json' file. If this should be a portable app instead, add that file specifying the appropriate framework.
```

The additional message when there is no runtime section:
```
Failed to run as a standalone app because there is no framework specified in 'C:\git\repros\console\bin\Debug\netcoreapp2.0\console.runtimeconfig.json'. If this should be a portable app instead, specify the appropriate framework in that file.
```

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -482,6 +482,17 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
 
     // If it still couldn't be found, somebody upstack messed up. Flag an error for the "expected" location.
     trace::error(_X("A fatal error was encountered. The library '%s' required to execute the application was not found in '%s'."), LIBHOSTPOLICY_NAME, expected.c_str());
+    if (mode == host_mode_t::muxer && !config.get_portable())
+    {
+        if (!pal::file_exists(config.get_path()))
+        {
+            trace::error(_X("Running as a standalone app because there is no '%s' file. If this should be a portable app instead, add that file specifying the appropriate framework."), config.get_path().c_str());
+        }
+        else if (config.get_fx_name().empty())
+        {
+            trace::error(_X("Running as a standalone app because there is no framework specified in '%s'. If this should be a portable app instead, specify the appropriate framework in that file."), config.get_path().c_str());
+        }
+    }
     return false;
 }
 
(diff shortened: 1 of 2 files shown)


steveharter:
@eerhardt PTAL thanks.

eerhardt:
Would it be worthwhile to add a test?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -482,6 +482,17 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
 
     // If it still couldn't be found, somebody upstack messed up. Flag an error for the "expected" location.
     trace::error(_X("A fatal error was encountered. The library '%s' required to execute the application was not found in '%s'."), LIBHOSTPOLICY_NAME, expected.c_str());
+    if (mode == host_mode_t::muxer && !config.get_portable())
+    {
+        if (!pal::file_exists(config.get_path()))
+        {
+            trace::error(_X("Running as a standalone app because there is no '%s' file. If this should be a portable app instead, add that file specifying the appropriate framework."), config.get_path().c_str());
```
eerhardt: (minor) Instead of saying `Running as a standalone app`, would it be better to say `Attempted to run as a standalone app`?  Since we didn't really run anything.

steveharter: Hmm. I changed it to "Failed to run as a standalone app"

eerhardt: Sounds good to me.  Thanks.

teyc:
It would be super if the error message took me to a documentation link with a sample to get me quickly started. The `runtimepolicy.config` is a new concept even to experienced .NET people, and I've no idea what it needs to look like. The top results on Google are all github issues, and the top MS docs article for .NET native, and only provides the schema.

steveharter:
@kathleendollard what do you think of adding a link to a sample for runtimeconfig.json?

Issue #3192, opened 2017-09-12
Host inconsistently validates shared framework deps.json


ericstj:
See https://github.com/dotnet/buildtools/pull/1671

There was a bug where buildtools was writing the wrong extension to the deps file it created.
EG: xunit.console.netcore.dll

This file did not exist in the shared framework folder, however xunit.console.netcore.exe did.

Now, if you ran any app that also had xunit.console.netcore.exe or xunit.console.netcore.dll the host would not complain.  But if the application was missing xunit.console.netcore.* then it would result in the error:
```
An assembly specified in the application dependencies manifest (Microsoft.NETCore.App.deps.json) was not found:
 package: 'xunit.console.netcore', version: '99.99.99'
 path: 'xunit.console.netcore.dll'
```

The host needs to be consistent about this error message and also be consistent about wether its considering extension or not.

PR #3383, opened 2017-11-01
Remove workaround for deps not containing entry for app

steveharter:
Verify that there are no duplicate entries with different file extensions (normally .exe and .dll) within the deps file(s) and including the managed app name specified to execute.

The root cause of the exe\dll issue is that there has been a workaround existing for a long time that just adds the managed app without verifying it is in the deps file. If the deps has a different entry (`exe` instead of `dll` for example) that was previously ignored; the TODO in the code says remove this workaround however it must remain.

So the fix is to look at each entry in the deps and determine if it was processed before with a different file extension. This catches the original exe\dll issue as well as any duplicates within just the deps file. The same message is shown for both cases:
```
Error:
  An assembly specified in the application dependencies manifest (PortableApp.deps.json) has already been found but with a different file extension:
    package: 'PortableApp', version: '1.0.0'
    path: 'PortableApp.dll'
    previously found assembly: 'D:\git\coresetup1\Bin\tests\win-x64.Debug\288\PortableApp\bin\PortableApp.exe'
```

Fixes https://github.com/dotnet/core-setup/issues/3192

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -26,32 +26,6 @@ const pal::string_t ManifestListMessage = _X(
 namespace
 {
 // -----------------------------------------------------------------------------
-// A uniqifying append helper that doesn't let two entries with the same
-// "asset_name" be part of the "output" paths.
-//
-void add_tpa_asset(
-    const pal::string_t& asset_name,
-    const pal::string_t& asset_path,
-    std::unordered_set<pal::string_t>* items,
-    pal::string_t* output)
-{
-    if (items->count(asset_name))
-    {
-        return;
-    }
-
-    trace::verbose(_X("Adding tpa entry: %s"), asset_path.c_str());
-
-    // Workaround for CoreFX not being able to resolve sym links.
-    pal::string_t real_asset_path = asset_path;
-    pal::realpath(&real_asset_path);
-    output->append(real_asset_path);
-
-    output->push_back(PATH_SEPARATOR);
-    items->insert(asset_name);
-}
-
-// -----------------------------------------------------------------------------
 // A uniqifying append helper that doesn't let two "paths" to be identical in
 // the "output" string.
 //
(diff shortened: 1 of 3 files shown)


src/corehost/cli/deps_resolver.cpp
```diff
@@ -397,11 +397,6 @@ bool deps_resolver_t::resolve_tpa_list(
     };
 
     // First add managed assembly to the TPA.
-    // TODO: Remove: the deps should contain the managed DLL.
-    // Workaround for: csc.deps.json doesn't have the csc.dll
```
eerhardt: Looking at the latest SDK, it looks like csc.deps.json still doesn't have the csc.dll/.exe.

Are we sure we want to break this?

steveharter: No I was just asserting until I found a reason not to. I will keep this in. Thanks

eerhardt: FYI, it looks like in the absolute latest SDK that the Roslyn pieces have changed, and now csc.deps.json does have a csc.dll listing.  But I think this new host will still need to be able to launch already released SDKs (since users can pin an SDK version with a global.json file).

steveharter: I've updated the code to include explicit checks for duplicates, and left the original work-around in.

src/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
```diff
@@ -60,6 +60,29 @@ public void Muxer_activation_of_Build_Output_Portable_DLL_with_DepsJson_and_Runt
...
+        public void Muxer_activation_of_Build_Output_Portable_DLL_with_DepsJson_having_Assembly_with_Different_File_Extension_Fails()
+        {
+            var fixture = PreviouslyBuiltAndRestoredPortableTestProjectFixture
+                .Copy();
+
+            var dotnet = fixture.BuiltDotnet;
+
+            // Change *.dll to *.exe
+            var appDll = fixture.TestProject.AppDll;
+            var appDll2 = appDll.Replace(".dll", ".exe");
```
eerhardt: can `appDll2` be renamed to `appExe`?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -89,8 +63,41 @@ void add_unique_path(
...
+    }
+
+    return path.substr(name_pos + 1);
+}
+
 } // end of anonymous namespace
 
+  // -----------------------------------------------------------------------------
+  // A uniqifying append helper that doesn't let two entries with the same
+  // "asset_name" be part of the "output" paths.
```
eerhardt: What is the `output` paths here?  The parameter that was `output` removed.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -89,8 +63,41 @@ void add_unique_path(
     existing->insert(real);
 }
 
+pal::string_t get_deps_filename(const pal::string_t& path)
```
eerhardt: Why is this named "get_deps_filename"?  Isn't it a utility to find any file name?

Don't we already have this utility somewhere?  [https://github.com/dotnet/core-setup/blob/88da334986d5f560c98248c2cfff6e5816cdde7a/src/corehost/common/utils.h#L26](https://github.com/dotnet/core-setup/blob/88da334986d5f560c98248c2cfff6e5816cdde7a/src/corehost/common/utils.h#L26)

steveharter: Note the hard-coded backslash for deps files. get_filename will use the platform-specific backslash\forwardslash

src/corehost/cli/deps_resolver.cpp
steveharter: Update comment. Thanks

src/test/HostActivationTests/GivenThatICareAboutPortableAppActivation.cs
steveharter: Done. Thanks

src/corehost/cli/deps_resolver.cpp
eerhardt: Can you rename the method and/or add a comment to make that more obvious?

steveharter: I added a comment

PR #3390, opened 2017-11-03
Update build definitions

weshaggard:
cc @dagood @eerhardt

### DIFF
--- a/buildpipeline/Core-Setup-Windows-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Windows-Arm-BT.json
@@ -77,7 +77,6 @@
     },
     {
       "enabled": true,
-
       "continueOnError": false,
       "alwaysRun": false,
       "displayName": "Build traversal build dependencies",
--- a/buildpipeline/Core-Setup-Windows-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Windows-Arm-BT.json
@@ -505,9 +505,9 @@
       "allowOverride": true
     },
     "PB_CommonMSBuildArgs": {
-      "value": "/p:ConfigurationGroup=$(BuildConfiguration) /p:TargetArchitecture=$(PB_TargetArchitecture) /p:PortableBuild=$(PB_PortableBuild) /p:DisableCrossgen=true $(PB_AdditionalBuildArguments)"
+      "value": "/p:ConfigurationGroup=$(BuildConfiguration) /p:TargetArchitecture=$(PB_TargetArchitecture) /p:PortableBuild=$(PB_PortableBuild) /p:DisableCrossgen=true $(PB_AdditionalMSBuildArguments)"
     },
-    "PB_AdditionalBuildArguments": {
+    "PB_AdditionalMSBuildArguments": {
       "value": ""
     },
     "OfficialBuildId": {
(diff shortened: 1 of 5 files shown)


buildpipeline/pipeline.json
```diff
@@ -121,6 +124,7 @@
           "Parameters": {
             "PB_TargetArchitecture": "x64",
             "PB_PortableBuild": "true",
+            "PB_AdditionalMSBuildArguments": "/p:SkipTests=$(PB_SkipTests) /p:DotNetRestoreSources=$(PB_RestoreSource)",
```
eerhardt: Why do we sometimes `-SkipTests` and other times `/p:SkipTests`?

buildpipeline/Core-Setup-Windows-Arm-BT.json
```diff
@@ -76,6 +77,7 @@
     },
     {
       "enabled": true,
+
```
eerhardt: unnecessary new line??

eerhardt:
Why aren't you changing any of the Linux build defs?

weshaggard:
>Why aren't you changing any of the Linux build defs? I'd assume we'd need to change for publishing.

The linux definitions already supported configuring the tests so I didn't need to modify them.

buildpipeline/pipeline.json
weshaggard: I did't rework all the definitions so if they were parameters to build.sh/cmd then it is just `-SkipTests` but if they were msbuild args then just `/p:SkipTests`. While I don't like the inconsistency either I didn't want to rework the entire build defintion at this point in time.

eerhardt:
> The linux definitions already supported configuring the tests so I didn't need to modify them.

How about publish?

buildpipeline/pipeline.json
eerhardt: fair enough.  it would be nice to be consistent.

weshaggard:
>How about publish?

For this PR all I did was disable the final publish but I'm trying to decouple the publish logic now to make it more consistent with the other repos and ensure we only do "public" publishing from the publish leg which we can configure. That change is coming next week.

buildpipeline/Core-Setup-Windows-Arm-BT.json
```diff
@@ -62,6 +62,7 @@
       "alwaysRun": false,
       "displayName": "Install Signing Plugin",
       "timeoutInMinutes": 0,
+      "condition": "and(succeeded(), in(variables.PB_SignType, 'real', 'test'))",
```
dagood: Later in this file the type is hard-coded to `real`, is this ok or should it use `$(PB_SignType)`?

https://github.com/weshaggard/core-setup/blob/75bed989ccb49da15048108cf1fe8ac379c3e461/buildpipeline/Core-Setup-Windows-Arm-BT.json#L65-L72

weshaggard: That is OK and is coming from the microbuild task. We ultimately override it by passing in SignType to our build command so it is never really used.

eerhardt:
> For this PR all I did was disable the final publish but I'm trying to decouple the publish logic now to make it more consistent with the other repos and ensure we only do "public" publishing from the publish leg which we can configure. 

When you do this, make sure all the assets continue going to the exact same places.  We have other pieces that are dependent on the exact location of each asset.

For example, for a single build, the assets go to 2 places.

1. To a folder with the build number `2.1.0-preview1-1234`.
1. To the "channel" folder - i.e for master the folder named "master", or for "release/2.0.0", etc.

We have links and tools that require these assets to be in both places.

PR #3392, opened 2017-11-07
Merge dev/release/2.0.0 into release/2.0.0

weshaggard:
@dagood @eerhardt


Directory.Build.props
```diff
@@ -0,0 +1,30 @@
+<Project>
+
+  <PropertyGroup>
+    <!--
+    NOTE: This .props file assumes it is only imported by projects that use the SDK="Microsoft.NET.Sdk" in their project.
+    We are currently on too old of an SDK to detect - it doesn't have https://github.com/dotnet/sdk/pull/1242.
```
eerhardt: We should be on a valid version of the SDK by now, since we are on 2.0.0.  So we should be able to remove this.  (doesn't have to be done on the merge commit, but it should be done in both master and release/2.0.0).

PR #3395, opened 2017-11-07
Enable Alpine official RID

janvorli:
This is a leftover from the Alpine enabling changes that I've forgotten
to make after corefx and coreclr official builds were enabled.

### DIFF
--- a/src/pkg/projects/netcoreappRIDs.props
+++ b/src/pkg/projects/netcoreappRIDs.props
@@ -8,6 +8,7 @@
   <ItemGroup>
     <OfficialBuildRID Include="linux-x64" />
     <OfficialBuildRID Include="rhel.6-x64" />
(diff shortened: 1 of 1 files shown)


janvorli:
cc: @MichaelSimons

janvorli:
@eerhardt I have removed this from my original change that enabled Alpine support in core-setup based on your feedback, but forgotten to add it after the official builds of coreclr / corefx were out.

eerhardt:
@janvorli - are we enabling Jenkins legs for alpine and rhel6?

janvorli:
@eerhardt I've enabled RHEL 6 in corefx only so far (optionally). I've discussed it with @mmitche some time ago and the recommendation was to wait with coreclr / core-setup until the ci moves to the new way of using pipelines for CI build.

PR #3402, opened 2017-11-09
Update NuGet.ProjectModel reference to respect RepoAPI package version.

dseefeld:
Updated reference to NuGet.ProjectModel to use version from RepoAPI and moved default value to dependencies.props.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -28,6 +28,8 @@
     <SystemServiceModelNetTcpPackageVersion>$(SystemServiceModelDuplexPackageVersion)</SystemServiceModelNetTcpPackageVersion>
     <SystemServiceModelPrimitivesPackageVersion>$(SystemServiceModelDuplexPackageVersion)</SystemServiceModelPrimitivesPackageVersion>
     <SystemServiceModelSecurityPackageVersion>$(SystemServiceModelDuplexPackageVersion)</SystemServiceModelSecurityPackageVersion>
+
+    <NuGetProjectModelPackageVersion>4.3.0-preview2-4095</NuGetProjectModelPackageVersion>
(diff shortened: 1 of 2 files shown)


dependencies.props
```diff
@@ -28,6 +28,8 @@
     <SystemServiceModelNetTcpPackageVersion>$(SystemServiceModelDuplexPackageVersion)</SystemServiceModelNetTcpPackageVersion>
     <SystemServiceModelPrimitivesPackageVersion>$(SystemServiceModelDuplexPackageVersion)</SystemServiceModelPrimitivesPackageVersion>
     <SystemServiceModelSecurityPackageVersion>$(SystemServiceModelDuplexPackageVersion)</SystemServiceModelSecurityPackageVersion>
+
+    <NuGetProjectModelPackageVersion>4.3.0-preview2-4095</NuGetProjectModelPackageVersion>
```
dagood: This should go in the `infrastructure and test only dependencies` property group down below, just for categorization.

eerhardt:
I'm a bit confused by this.  The NuGet version that our local core-setup.tasks.csproj needs is actually tied to the version of NuGet that BuildTools has.  It shouldn't use the version that is built during source-build.

In the long run (or maybe even sooner), we should move core-setup.tasks.csproj to BuildTools itself, so we aren't building MSBuild tasks as part of the core-setup build.

dagood:
Interesting, not sure how to make that work with the offline build, where I don't think we can have that package. Dependency on the DLL in the Tools dir? If that doesn't work, might just want to move them to buildtools now.

dseefeld:
Closing to investigate other strategies based on @eerhardt 's comment above.

PR #3408, opened 2017-11-13
Replace deprecated VSTS task for copy/publish artifacts for Linux template

maririos:
Updating the Linux template too as I missed it in PR https://github.com/dotnet/core-setup/commit/58873e3c28538790ecc3e75cbf0477d9c8a9deeb .

cc: @chcosta

### DIFF
--- a/buildpipeline/Core-Setup-Linux-BT.json
+++ b/buildpipeline/Core-Setup-Linux-BT.json
@@ -949,20 +949,42 @@
       "enabled": true,
       "continueOnError": true,
       "alwaysRun": false,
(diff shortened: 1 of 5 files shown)


buildpipeline/Core-Setup-Linux-BT.json
```diff
@@ -909,20 +909,42 @@
       "enabled": true,
       "continueOnError": true,
       "alwaysRun": false,
```
chcosta: I'd expect alwaysrun to be true for these two steps

maririos: Good call. I changed all the other files too

PR #3420, opened 2017-11-15
Add retry logic for setup of dotnet cli

dseefeld:
Copy init-tools.sh retry logic from corefx.  Also adding display_error_message() function.

Fixes dotnet/core-eng#1966


weshaggard:
Same feedback as with corefx. Lets please make sure they exactly match unless there is a reason they cannot.

dseefeld:
Updated to match master.

dseefeld:
@weshaggard @dagood Can you take a look at this?  Thanks

PR #3425, opened 2017-11-16
Update corefx current ref and dependencies

safern:
cc: @weshaggard @eerhardt


safern:
Now that we published the latest Platforms package we need to update this manually (already did in the versions repo).

safern:
> LGTM. This should have happened automatically with the next corefx release/2.0.0 build, but this works just as well.

Yes, we didn't want to wait for the next corefx release/2.0.0 build as we already queued that one before updating the version's repo. Thanks for reviewing

PR #3433, opened 2017-11-17
Use BuildTools PackageVersionPropsUrl tooling

dagood:
https://github.com/dotnet/core-eng/issues/2037
BuildTools PR this is using: https://github.com/dotnet/buildtools/pull/1787

These changes are the same as https://github.com/dotnet/corefx/pull/25319

---

All definitions need PB_PackageVersionPropsUrl to default to empty string, not undefined. If it were undefined, VSTS could leave in $(PB_PackageVersionPropsUrl) which would make the build fail when it fails to download it.

For Docker, map the environment variable on each run command.

For OSX/Windows, have VSTS set the environment variable by creating a definition variable.

### DIFF
--- a/buildpipeline/Core-Setup-Linux-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Linux-Arm-BT.json
@@ -356,7 +356,7 @@
       "allowOverride": true
     },
     "DockerCommonRunArgs": {
-      "value": "--name $(PB_DockerContainerName) -v \"$(PB_SourcesDirectory):$(PB_GitDirectory)\" -w=\"$(PB_GitDirectory)\" $(PB_DockerImageName)"
+      "value": "--name $(PB_DockerContainerName) -v \"$(PB_SourcesDirectory):$(PB_GitDirectory)\" -w=\"$(PB_GitDirectory)\" -e \"PACKAGEVERSIONPROPSURL=$(PB_PackageVersionPropsUrl)\" $(PB_DockerImageName)"
     },
     "PB_VsoRepoUrl": {
       "value": "--branch $(PB_Branch) https://$(PB_VsoAccountName):$(PB_VsoPassword)@devdiv.visualstudio.com/DevDiv/_git/DotNet-Core-Setup-Trusted"
@@ -414,6 +414,9 @@
     "PB_ChecksumAzureAccessToken": {
       "value": null,
       "isSecret": true
+    },
+    "PB_PackageVersionPropsUrl": {
+      "value": ""
     }
   },
   "demands": [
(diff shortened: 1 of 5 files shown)


buildpipeline/Core-Setup-Linux-Arm-BT.json
```diff
@@ -356,7 +356,7 @@
       "allowOverride": true
     },
     "DockerCommonRunArgs": {
-      "value": "--name $(PB_DockerContainerName) -v \"$(PB_SourcesDirectory):$(PB_GitDirectory)\" -w=\"$(PB_GitDirectory)\" $(PB_DockerImageName)"
+      "value": "--name $(PB_DockerContainerName) -v \"$(PB_SourcesDirectory):$(PB_GitDirectory)\" -w=\"$(PB_GitDirectory)\" -e \"PACKAGEVERSIONPROPSURL=$(PB_PackageVersionPropsUrl)\" $(PB_DockerImageName)"
```
eerhardt: Dumb question - why is this using an environment variable and not passing the value in to the MSBuild properties?

dagood: Short answer: the package version props download happens during init-tools (https://github.com/dotnet/buildtools/pull/1787).

MSBuild reentrancy would have been neat, but there were enough complications that I decided not to do it for this initial implementation. I think I wrote them down somewhere, but I can't find it now. Top things that stick out are that there doesn't seem to be a way to "soft cancel" the requested target (to download, instead), and I couldn't find any reasonable way to pass down *all* properties into the inner layer.

eerhardt: We already have places that generate .props files during the build.

https://github.com/dotnet/buildtools/blob/78260239e4db68ad5c2df774a27d5f9a54fd92e5/src/Microsoft.DotNet.Build.Tasks/PackageFiles/BuildVersion.targets#L6
https://github.com/dotnet/buildtools/blob/03988127f489bcd1cf58d71131ac628620d53de9/src/Microsoft.DotNet.Build.Tasks/GenerateConfigurationProps.cs#L15

I'm not sure I understand why this couldn't have been done similarly.  Downloading a props file from the internet, and generating it on the fly, work the same way.  The values aren't available in the root `build.proj` project.  But for all the projects that get invoked from it, the values are available.

dagood: Maybe that would work, I don't know if we need the values in the root build.proj. Filed https://github.com/dotnet/buildtools/issues/1813.

weshaggard:
@mmitche this should enable auto-dependency flow for core-setup

weshaggard:
@dagood do you have a similar PR for corefx yet?

dagood:
Yep, linked in the description: https://github.com/dotnet/corefx/pull/25319

weshaggard:
Sorry I missed it :)

PR #3450, opened 2017-11-22
Invoke VsDevCmd in scripts without changing working directory

### LINKED ISSUE #3429: Couldn't build from sources on windows
Couldn't build from sources on windows:
## Steps to reproduce
Clone project into disk D:\ root and run build.cmd
## Expected  behaviour
Project built

## Actual behaviour
D:\core-setup>build.cmd
**********************************************************************
** Visual Studio 2017 Developer Command Prompt v15.4.4
** Copyright (c) 2017 Microsoft Corporation
**********************************************************************
Installing dotnet cli...
Restoring BuildTools version 2.0.0-prerelease-02215-01...
Initializing BuildTools...
Done initializing tools.
Running: D:\core-setup\Tools\msbuild.cmd /nologo /verbosity:minimal /clp:Summary /maxcpucount /nodeReuse:false /p:OSGroup=Windows_NT /p:TargetArchitecture=x64  /flp:v=normal  build.proj
MSBUILD : error MSB1009: Project file does not exist.
Switch: build.proj
Command execution failed with exit code 1.

## Environment data
Visual Studio 15.4.4
Windows 10

4creators:
Fixes #3429

### DIFF
--- a/clean.cmd
+++ b/clean.cmd
@@ -1,8 +1,20 @@
 @if not defined _echo @echo off
 setlocal EnableDelayedExpansion
 
-echo Stop VBCSCompiler.exe execution.
-for /f "tokens=2 delims=," %%F in ('tasklist /nh /fi "imagename eq VBCSCompiler.exe" /fo csv') do taskkill /f /PID %%~F
+set NO_DASHES_ARG=%1
+if not defined NO_DASHES_ARG goto no_help
+if /I [%NO_DASHES_ARG:-=%] == [?] goto Usage
+if /I [%NO_DASHES_ARG:-=%] == [h] goto Usage
+
+:no_help
+
+:: Check if VBCSCompiler.exe is running
+tasklist /fi "imagename eq VBCSCompiler.exe" |find ":" > nul
+:: Compiler is running if errorlevel == 1
+if errorlevel 1 (
+	echo Stop VBCSCompiler.exe execution.
+	for /f "tokens=2 delims=," %%F in ('tasklist /nh /fi "imagename eq VBCSCompiler.exe" /fo csv') do taskkill /f /PID %%~F
+)
 
 :: Strip all dashes off the argument and use invariant
 :: compare to match as many versions of "all" that we can
--- a/run.cmd
+++ b/run.cmd
@@ -9,11 +9,12 @@ if exist %_VSWHERE% (
 )
 if not exist "%_VSCOMNTOOLS%" set _VSCOMNTOOLS=%VS140COMNTOOLS%
 if not exist "%_VSCOMNTOOLS%" (
-  echo Error: Visual Studio 2015 or 2017 required.
-  echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
-  exit /b 1
+    echo Error: Visual Studio 2015 or 2017 required.
+    echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
+    exit /b 1
 )
 
+set VSCMD_START_DIR="%~dp0"
 call "%_VSCOMNTOOLS%\VsDevCmd.bat"
 
 :Run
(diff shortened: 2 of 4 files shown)


run.cmd
```diff
@@ -1,19 +1,24 @@
 @if "%_echo%" neq "on" echo off
 setlocal
 
-if defined VisualStudioVersion goto :Run
+if defined VisualStudioVersion (
+    if not defined __VSVersion echo %__MsgPrefix%Detected Visual Studio %VisualStudioVersion% developer command ^prompt environment
```
weshaggard: I don't think this applies to this repo.

run.cmd
```diff
@@ -1,19 +1,24 @@
...
 if not exist "%_VSCOMNTOOLS%" (
-  echo Error: Visual Studio 2015 or 2017 required.
-  echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
-  exit /b 1
+    echo %__MsgPrefix%Error: Visual Studio 2015 or 2017 required.
+    echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
+    exit /b 1
 )
 
+set VSCMD_START_DIR="%~dp0"
```
weshaggard: This is the key. Thanks for fixing.

run.cmd
4creators: will remove it

run.cmd
4creators: should I fix it in all other scripts as well ?

weshaggard: Yeah I think setting in any script before we call VsDevCmd is the correct way to fix this issue throughout our repos.

run.cmd
```diff
@@ -1,19 +1,24 @@
 @if "%_echo%" neq "on" echo off
 setlocal
 
-if defined VisualStudioVersion goto :Run
+if defined VisualStudioVersion (
+    if not defined __VSVersion echo %__MsgPrefix%Detected Visual Studio %VisualStudioVersion% developer command ^prompt environment
+    goto :Run
+)
 
+echo %__MsgPrefix%Searching ^for Visual Studio 2017 or 2015 installation
```
4creators: @weshaggard  Should I remove this comment as well?

weshaggard: Yeah I don't think it adds much value. What would actually add more value is a message that says which version of VsDevCmd it is running. I've wanted to know that on a number of occasions.

4creators: You mean 4 number version of VS?

weshaggard: I think just the full path to VSDevCmd would be enough to point at which VS was installed but if there is also a version that would be good as well. I know the VS2017 prompt prints out a version already which is useful as well but that only works in VS2017.

4creators: I will try to get something reasonable printed

weshaggard:
@4creators do have some time to update this PR?

4creators:
@weshaggard Yes I will finish it. Sorry for the delay it just slipped off my radar

clean.cmd
```diff
@@ -1,8 +1,20 @@
...
+
+:no_help
+
+:: Check if VBCSCompiler.exe is running
+tasklist /fi "imagename eq VBCSCompiler.exe" |find ":" > nul
+:: Compiler is running if errorlevel == 1
+if errorlevel 1 (
+	echo Stop VBCSCompiler.exe execution.
+	for /f "tokens=2 delims=," %%F in ('tasklist /nh /fi "imagename eq VBCSCompiler.exe" /fo csv') do taskkill /f /PID %%~F
+)
```
4creators: Porting fixes for VBCSCompiler.exe cleanup and adding logic to enable help message display (was not working till now)

4creators:
@weshaggard PTAL sorry for delay

1. Introduced safeguards for all VsDevCmd.bat invocations
2. Cleaned up clean.cmd code

run.cmd
```diff
@@ -9,11 +9,12 @@ if exist %_VSWHERE% (
 )
 if not exist "%_VSCOMNTOOLS%" set _VSCOMNTOOLS=%VS140COMNTOOLS%
 if not exist "%_VSCOMNTOOLS%" (
-  echo Error: Visual Studio 2015 or 2017 required.
-  echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
-  exit /b 1
+    echo %__MsgPrefix%Error: Visual Studio 2015 or 2017 required.
```
weshaggard: We don't set __MsgPrefix anywhere in this repo that I could find from a quick search so can undo this part of the change.

weshaggard:
Thanks @4creators!

Issue #3457, opened 2017-11-27
Add support for multiple frameworks \ platforms


steveharter:
This is the main feature issue to support layered frameworks in core-setup which affects hostfxr.dll, hostpolicy.dll and apphost.exe.

Currently a single framework is supported, identified by name and base version in the runtimeconfig.json section:
```
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "2.1.0"
    }
```

In order for ASP.NET (and others tbd) to obtain the same benefit of roll-forward semantics and a self-contained directory structure, the following design has been prototyped and approved:

1) Multiple framework through "layering" (a single inheritance hierarchy)
2) The existing runtimeconfig.json and deps.json schema remains
2a) Each new framework has its own runtimeconfig.json which will reference the next (lower-level) framework. E.g. ASP.NET's runtimeconfig.json will reference Microsoft.NETCore.App
2b) Settings in the runtimeconfig.json will carry forward to the next lower level, although the exact semantics differ depending on the setting - some override, while others are additive

Additional design information is located at https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/multilevel-sharedfx-lookup.md, and additional information regarding runtimeconfig.json settings will be added soon.

Known constraints:
A) `--fx-framework <version>` only applies to the first framework
B) The RID graph must be located in the lowest level framework's deps.json (Microsoft.NETCore.App)
C) hostpolicy.dll must be located in the lowest level framework's directory (Microsoft.NETCore.App)
D) The store functionality remains as-is for now
E) The FX_DEPS_FILE property variable passed to the CLR only refers to the root (last) framework's deps file
F) The tfm is defined by the apps' runtimeconfig.json (it can't be changed in fx runtimeconfig.json)

jkotalik:
cc/ @JunTaoLuo

muratg:
Thanks @steveharter!

PR #3460, opened 2017-11-27
Add support for multiple shared platforms

steveharter:
Initial PR for the multiple shared framework feature.

Fixes https://github.com/dotnet/core-setup/issues/3457

Design points:
1) Added fx_definition_t type which encapsulates a framework including name, requested version, found version, deps file, runtimeconfig file.
2) Refactor code to use an array of fx_definition_t instead of the raw single parameter app or framework values (framework name, version, deps file, etc) in order to force the code to handle multiple frameworks which helps avoid issues\bugs etc.
3) fx_definition_t[0] is the application, while fx_definition_t[size-1] is Microsoft.NetCore.App. In between [0] and [size-1] are the additional frameworks (ASP.NET). In a standalone case there is only the application, so size=1.
4) The interface from hostfxr to hostpolicy is not breaking; a new hostfxr.dll still passes the correct information to older hostpolicy.dlls.

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -221,10 +221,12 @@ void deps_resolver_t::setup_probe_config(
         m_probes.push_back(probe_config_t::svc(ext_pkgs));
     }
 
-    if (pal::directory_exists(m_fx_dir))
+    for (int i = 1; i < init.fx_definitions.size(); ++i)
     {
-        // FX probe
-        m_probes.push_back(probe_config_t::fx(m_fx_dir, m_fx_deps.get()));
+        if (pal::directory_exists(init.fx_definitions[i]->get_dir()))
+        {
+            m_probes.push_back(probe_config_t::fx(init.fx_definitions[i]->get_dir(), &init.fx_definitions[i]->get_deps()));
+        }
     }
 
     // The published deps directory to be probed: either app or FX directory.
@@ -372,7 +374,7 @@ bool deps_resolver_t::resolve_tpa_list(
     const std::vector<deps_entry_t> empty(0);
     dir_assemblies_t items;
 
-    auto process_entry = [&](const pal::string_t& deps_dir, deps_json_t* deps, const deps_entry_t& entry) -> bool
+    auto process_entry = [&](const pal::string_t& deps_dir, const deps_entry_t& entry) -> bool
     {
         if (entry.is_serviceable)
         {
@@ -428,10 +430,11 @@ bool deps_resolver_t::resolve_tpa_list(
     // Workaround for: csc.deps.json doesn't have the csc.dll
     pal::string_t managed_app_asset = get_filename_without_ext(m_managed_app);
     add_tpa_asset(managed_app_asset, m_managed_app, &items);
-    const auto& deps_entries = m_deps->get_entries(deps_entry_t::asset_types::runtime);
+
+    const auto& deps_entries = get_deps().get_entries(deps_entry_t::asset_types::runtime);
     for (const auto& entry : deps_entries)
     {
-        if (!process_entry(m_app_dir, m_deps.get(), entry))
+        if (!process_entry(m_app_dir, entry))
         {
             return false;
         }
@@ -439,7 +442,7 @@ bool deps_resolver_t::resolve_tpa_list(
 
     // If the deps file wasn't present or has missing entries, then
     // add the app local assemblies to the TPA.
-    if (!m_deps->exists())
+    if (!get_deps().exists())
     {
         dir_assemblies_t local_assemblies;
 
@@ -458,7 +461,7 @@ bool deps_resolver_t::resolve_tpa_list(
         auto additional_deps_entries = additional_deps->get_entries(deps_entry_t::asset_types::runtime);
         for (auto entry : additional_deps_entries)
         {
-            if (!process_entry(m_app_dir, additional_deps.get(), entry))
+            if (!process_entry(m_app_dir, entry))
             {
                 return false;
             }
@@ -466,12 +469,15 @@ bool deps_resolver_t::resolve_tpa_list(
     }
 
     // Probe FX deps entries after app assemblies are added.
-    const auto& fx_entries = m_portable ? m_fx_deps->get_entries(deps_entry_t::asset_types::runtime) : empty;
-    for (const auto& entry : fx_entries)
+    for (int i = 1; i < m_fx_definitions.size(); ++i)
     {
-        if (!process_entry(m_fx_dir, m_fx_deps.get(), entry))
+        const auto& fx_entries = m_portable ? m_fx_definitions[i]->get_deps().get_entries(deps_entry_t::asset_types::runtime) : empty;
+        for (const auto& entry : fx_entries)
         {
-            return false;
+            if (!process_entry(m_fx_definitions[i]->get_dir(), entry))
+            {
+                return false;
+            }
         }
     }
 
@@ -523,8 +529,6 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
     }
 
     pal::string_t additional_deps_serialized = init.additional_deps_serialized;
-    pal::string_t fx_name = init.fx_name;
-    pal::string_t fx_ver = init.fx_ver;
 
     if (additional_deps_serialized.empty())
     {
@@ -544,7 +548,7 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
             {
                 trace::verbose(_X("Using specified additional deps.json: '%s'"), 
                     additional_deps_path.c_str());
-                    
+
                 m_additional_deps_files.push_back(additional_deps_path);
             }
             else
@@ -555,30 +559,38 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
         }
         else
         {
-            // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
-            append_path(&additional_deps_path, _X("shared"));
-            append_path(&additional_deps_path, fx_name.c_str());
-            append_path(&additional_deps_path, fx_ver.c_str());
-
-            // The resulting list will be empty if 'additional_deps_path' is not a valid directory path
-            std::vector<pal::string_t> list;
-            pal::readdir(additional_deps_path, _X("*.deps.json"), &list);
-            for (pal::string_t json_file : list)
+            for (int i = 1; i < m_fx_definitions.size(); ++i)
             {
-                pal::string_t json_full_path = additional_deps_path;
-                append_path(&json_full_path, json_file.c_str());
-                m_additional_deps_files.push_back(json_full_path);
+                // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
+                pal::string_t additional_deps_path_fx = additional_deps_path;
+                append_path(&additional_deps_path_fx, _X("shared"));
+                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_name().c_str());
+                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_found_version().c_str());
 
-                trace::verbose(_X("Using specified additional deps.json: '%s'"), 
-                    json_full_path.c_str());
+                // The resulting list will be empty if 'additional_deps_path_fx' is not a valid directory path
+                std::vector<pal::string_t> list;
+                pal::readdir(additional_deps_path_fx, _X("*.deps.json"), &list);
+                for (pal::string_t json_file : list)
+                {
+                    pal::string_t json_full_path = additional_deps_path_fx;
+                    append_path(&json_full_path, json_file.c_str());
+                    m_additional_deps_files.push_back(json_full_path);
+
+                    trace::verbose(_X("Using specified additional deps.json: '%s'"),
+                        json_full_path.c_str());
+                }
             }
         }
     }
 
+    // The rid graph is obtained from the root framework
+    int root_framework = m_fx_definitions.size() - 1;
+    auto rids = m_fx_definitions[root_framework]->get_deps().get_rid_fallback_graph();
+
     for (pal::string_t json_file : m_additional_deps_files)
     {
         m_additional_deps.push_back(std::unique_ptr<deps_json_t>(
-            new deps_json_t(true, json_file, m_fx_deps->get_rid_fallback_graph())));
+            new deps_json_t(true, json_file, rids)));
     }
 }
 
@@ -615,8 +627,6 @@ bool deps_resolver_t::resolve_probe_dirs(
     pal::string_t non_serviced;
 
     std::vector<deps_entry_t> empty(0);
-    const auto& entries = m_deps->get_entries(asset_type);
-    const auto& fx_entries = m_portable ? m_fx_deps->get_entries(asset_type) : empty;
 
     pal::string_t candidate;
 
@@ -660,6 +670,8 @@ bool deps_resolver_t::resolve_probe_dirs(
         return true;
     };
 
+    // Add app entries
+    const auto& entries = get_deps().get_entries(asset_type);
     for (const auto& entry : entries)
     {
         if (!add_package_cache_entry(entry, m_app_dir))
@@ -669,7 +681,7 @@ bool deps_resolver_t::resolve_probe_dirs(
     }
 
     // If the deps file is missing add known locations.
-    if (!m_deps->exists())
+    if (!get_deps().exists())
     {
         // App local path
         add_unique_path(asset_type, m_app_dir, &items, output, &non_serviced, core_servicing);
@@ -692,11 +704,16 @@ bool deps_resolver_t::resolve_probe_dirs(
         }
     }
     
-    for (const auto& entry : fx_entries)
+    // Add fx package locations to fx_dir
+    for (int i = 1; i < m_fx_definitions.size(); ++i)
     {
-        if (!add_package_cache_entry(entry, m_fx_dir))
+        const auto& fx_entries = m_portable ? m_fx_definitions[i]->get_deps().get_entries(asset_type) : empty;
+        for (const auto& entry : fx_entries)
         {
-            return false;
+            if (!add_package_cache_entry(entry, m_fx_definitions[i]->get_dir()))
+            {
+                return false;
+            }
         }
     }
 
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -583,10 +583,7 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
         }
     }
 
-    // The rid graph is obtained from the root framework
-    int root_framework = m_fx_definitions.size() - 1;
-    auto rids = m_fx_definitions[root_framework]->get_deps().get_rid_fallback_graph();
-
+    auto rids = get_root_framework(m_fx_definitions).get_deps().get_rid_fallback_graph();
     for (pal::string_t json_file : m_additional_deps_files)
     {
         m_additional_deps.push_back(std::unique_ptr<deps_json_t>(
@@ -707,7 +704,7 @@ bool deps_resolver_t::resolve_probe_dirs(
     // Add fx package locations to fx_dir
     for (int i = 1; i < m_fx_definitions.size(); ++i)
     {
-        const auto& fx_entries = m_portable ? m_fx_definitions[i]->get_deps().get_entries(asset_type) : empty;
+        const auto& fx_entries = m_fx_definitions[i]->get_deps().get_entries(asset_type);
         for (const auto& entry : fx_entries)
         {
             if (!add_package_cache_entry(entry, m_fx_definitions[i]->get_dir()))
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -6,6 +6,7 @@
 #include "utils.h"
 #include "libhost.h"
 #include "args.h"
+#include "fx_definition.h"
 #include "fx_ver.h"
 #include "fx_muxer.h"
 #include "trace.h"
@@ -158,9 +159,9 @@ void get_all_sdk_versions(
 }
 
 /**
- * When the framework is not found, display detailed error message
- *   about available frameworks and installation of new framework.
- */
+* When the framework is not found, display detailed error message
+*   about available frameworks and installation of new framework.
+*/
 void handle_missing_framework_error(
     host_mode_t mode,
     const pal::string_t& fx_name,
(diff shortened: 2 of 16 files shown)


src/corehost/cli/deps_resolver.cpp
```diff
@@ -555,30 +559,38 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
...
-                append_path(&json_full_path, json_file.c_str());
-                m_additional_deps_files.push_back(json_full_path);
-
-                trace::verbose(_X("Using specified additional deps.json: '%s'"), 
-                    json_full_path.c_str());
+                // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
+                pal::string_t additional_deps_path_fx = additional_deps_path;
+                append_path(&additional_deps_path_fx, _X("shared"));
+                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_name().c_str());
+                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_found_version().c_str());
```
steveharter: //review note: in the original code this was the version from the config (fx_ver), not the found version (which seems incorrect in the original code)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1195,86 +1237,111 @@ int fx_muxer_t::read_config_and_execute(
...
     }
-    if (!runtime_config.empty() && (!pal::realpath(&runtime_config) || !pal::file_exists(runtime_config)))
-    {
-        trace::error(_X("The specified runtimeconfig.json [%s] does not exist"), runtime_config.c_str());
-        return StatusCode::InvalidConfigFile;
-    }
 
-    pal::string_t config_file, dev_config_file;
+    // Read config
+    // The fx_definitions_owned variable controls object lifetime
```
steveharter: Review note: unique_ptr is a smart pointer which will call delete when popped off the stack. It is not possible to pass std::vector<std::unique_ptr<fx_definition_t>> to other methods because unique_ptr can't be copied. Also we don't want to use shared_ptr instead (which does allow the pointer to be shared) as the ownership is simply stack-based. So another vector was created to pass around the raw pointers (fx_definitions). Another variant would be just to use fx_definitions (not fx_definitions_owned) and call delete[] when the method is done, but that is a bit more complex and error prone due to exception etc.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -354,6 +382,170 @@ public void Roll_Forward_On_No_Candidate_Fx_Fails_If_No_Higher_Version_Is_Availa
...
+                .CaptureStdOut()
+                .Execute()
+                .Should()
+                .Pass()
+                .And
+                .HaveStdOutContaining("9999.0.0")
+                .And
+                .HaveStdOutContaining("9999.0.1")
+                .And
+                .HaveStdOutContaining("7777.0.0")
```
eerhardt: Do you want to verify these have the right names?

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -486,10 +702,101 @@ private void SetRuntimeConfigJson(string destFile, string version, int? rollFwdO
...
+
+            string depsFile = Path.Combine(builtSharedUberFxDir, "Microsoft.UberFramework.deps.json");
+
+            File.WriteAllText(depsFile, depsjson.ToString());
+
+            // Copy the test assembly
+            string fileSource = Path.Combine(builtSharedFxDir, testAssembly + ".dll");
+            string fileDest = Path.Combine(builtSharedUberFxDir, testAssembly + ".dll");
+            FileInfo fileInfo = new FileInfo(fileSource);
+            fileInfo.CopyTo(fileDest);
```
eerhardt: I don't think you need to use `FileInfo` here.  Can't you just say:

`File.Copy(fileSource, fileDest)`

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -354,6 +382,170 @@ public void Roll_Forward_On_No_Candidate_Fx_Fails_If_No_Higher_Version_Is_Availa
...
+                .Copy();
+
+            var dotnet = fixture.BuiltDotnet;
+            var appDll = fixture.TestProject.AppDll;
+
+            string runtimeConfig = Path.Combine(fixture.TestProject.OutputDirectory, "SharedFxLookupPortableApp.runtimeconfig.json");
+            SetRuntimeConfigJson(runtimeConfig, "7777.0.0", null, useUberFramework: true);
+
+            // Add versions in the exe folders
+            AddAvailableSharedFxVersions(_exeSharedFxBaseDir, "9999.0.0");
```
eerhardt: Do we need/want to delete these things out when the tests finish?

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
steveharter: I will test again with names; I was getting an fluentassertions exception for some reason when including the exact message, like it couldn't parse correctly (it did have single quotes).

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
steveharter: I think that is a good practice, but we are inconsistent currently. I'll take a look.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
steveharter: OK

src/corehost/cli/libhost.h
```diff
@@ -100,54 +107,66 @@ class corehost_init_t
...
+
+            m_fx_names.push_back(fx_definitions[i]->get_name());
+            m_fx_names_cstr.push_back(fx_definitions[i]->get_name().c_str());
+            m_fx_dirs.push_back(fx_definitions[i]->get_dir());
+            m_fx_dirs_cstr.push_back(fx_definitions[i]->get_dir().c_str());
+            m_fx_requested_versions.push_back(fx_definitions[i]->get_requested_version());
+            m_fx_requested_versions_cstr.push_back(fx_definitions[i]->get_requested_version().c_str());
+            m_fx_found_versions.push_back(fx_definitions[i]->get_found_version());
+            m_fx_found_versions_cstr.push_back(fx_definitions[i]->get_found_version().c_str());
+        }
```
nguerrera: I think it would be easier to maintain if the parallel c string arrays were made with make_cstr_arr, just as with the clr_keys/values.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: May be best to talk in person to make sure we're on the same page, but here goes:
1. That means we would have to call delete() on the vector, correct? Also, when using an indexer on a vector of unique pointers, the copy ctor is attempted to be called on the unique pointer, which by design isn't supported which results in a compiler error
2. Move the vector from [where?] to corehost_init_t? Note that the array is also passed around elsewhere as reference information (no ownership)
3. We need to dynamically create the fx_definitions (we don't know how many there are), but yes we could create them, copy by value, and then call delete on the original immediately. There are also embedded types (runtime_config_t, deps_json_t) that would also need to make sure support default copy ctor (or add one). Doesn't seem as efficient.

src/corehost/cli/libhost.h
steveharter: I'll take a look thanks.

src/corehost/cli/fxr/fx_muxer.cpp
nguerrera: > That means we would have to call delete() on the vector, correct?

No, but it does means that the thing that owns the vector must be kept alive at least as long as corehost_init_t, which can be non-obvious so (1) probably isn't the best solution.

> Note that the array is also passed around elsewhere as reference information (no ownership)

I don't see why that would be a problem.

> Also, when using an indexer on a vector of unique pointers, the copy ctor is attempted to be called on the unique pointer, which by design isn't supported which results in a compiler error.

What gives you this impression? `vector::operator[]`  returns a reference.

Below is a demo of (2) with pass-by-reference and indexing too.

``` C++
#include <vector>
#include <memory>
#include <iostream>

using namespace std;

class fx_definition_t
{
public:
  fx_definition_t(const string& name) : m_name(name) 
  {
    cout << "construct fx_definition_t " << m_name.c_str() << endl; 
  }

  ~fx_definition_t() 
  { 
    cout << "destruct fx_definition_t" << m_name.c_str() << endl; 
  }

  const string& name() const { return m_name;  }

private:
  string m_name;
};

typedef vector<unique_ptr<fx_definition_t>> fx_definition_vector_t;

class corehost_init_t 
{
public:
  corehost_init_t(fx_definition_vector_t&& fx_definitions)
    : m_fx_definitions(std::move(fx_definitions)) 
  {
    cout << "construct corehost_init_t" << endl;
  }

  ~corehost_init_t()
  {
    cout << "destruct corehost_init_t" << endl;
  }

  const fx_definition_vector_t& fx_definitions() const 
  { 
    return m_fx_definitions; 
  }

private:
  fx_definition_vector_t m_fx_definitions;
};

// can pass vector of unique_ptr by reference
void dump_fx_definitions(const fx_definition_vector_t& fx_definitions)
{
  // can index into vector of unique_ptr
  for (size_t i = 0; i < fx_definitions.size(); i++)
  {
    cout << "visit fx_definition_t " << fx_definitions[i]->name().c_str() << endl;
  }
}

void execute_app(const corehost_init_t& init)
{
  dump_fx_definitions(init.fx_definitions());
}

int read_config_and_execute()
{
  fx_definition_vector_t fx_definitions;
  fx_definitions.emplace_back(new fx_definition_t("A"));
  fx_definitions.emplace_back(new fx_definition_t("B"));

  {
    corehost_init_t init(std::move(fx_definitions)); 
    execute_app(init);
  }

  cout << "init took over fx_definitions, so its empty here. size=" << fx_definitions.size() << endl;
  return 0;
}

int main() 
{
  return read_config_and_execute();
}
```

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -439,16 +444,12 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
...
-        if (!pal::directory_exists(fx_dir))
-        {
-            pal::string_t fx_version = specified_fx_version.empty() ? config.get_fx_version() : specified_fx_version;
-            handle_missing_framework_error(mode, config.get_fx_name(), fx_version, fx_dir, own_dir);
-            return false;
-        }
-
-        expected = fx_dir;
+        // The hostpolicy is required to be in the root framework's location
+        int root_framework = fx_definitions.size() - 1;
```
eerhardt: Would it be good to define a helper method for fx_definitions.GetRootFramework() that returned `fx_definitions[fx_definitions.size() - 1]`?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -408,27 +411,29 @@ pal::string_t get_deps_file(
...
-    const pal::string_t& fx_dir,
+    const std::vector<fx_definition_t*>& fx_definitions,
     const pal::string_t& app_candidate,
     const pal::string_t& specified_deps_file,
     const pal::string_t& specified_fx_version,
     const std::vector<pal::string_t>& probe_realpaths,
-    const runtime_config_t& config,
     pal::string_t* impl_dir)
 {
+    bool portable = fx_definitions[0]->get_runtime_config().get_portable();
```
eerhardt: `fx_definitions[0]` being the "app" is kind of weird.  It isn't really a framework.

I'm not sure what to do about this.  Maybe not naming these things "fx"?  Wanted to get your thoughts on it.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -481,22 +482,23 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
...
         {
-            trace::error(_X("Failed to run as a standalone app because there is no framework specified in '%s'. If this should be a portable app instead, specify the appropriate framework in that file."), config.get_path().c_str());
+            trace::error(_X("Failed to run as a standalone app because there is no framework specified in '%s'. If this should be a portable app instead, specify the appropriate framework in that file."), fx_definitions[0]->get_runtime_config().get_path().c_str());
         }
     }
     return false;
 }
 
-fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& version_list,
+fx_ver_t resolve_framework_version(const std::vector<fx_ver_t>& version_list,
```
eerhardt: Why this change?

eerhardt: I don't see a corresponding change in the `.h` file.  This doesn't seem necessary, is it?

---
In reply to: [153834891](https://github.com/dotnet/core-setup/pull/3460#discussion_r153834891) [](ancestors = 153834891)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1195,86 +1237,111 @@ int fx_muxer_t::read_config_and_execute(
...
     // 1. Command line argument (--roll-forward-on-no-candidate-fx)
     // 2. Runtimeconfig json file ('rollForwardOnNoCandidateFx' property)
     // 3. DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX env var
     // The conflicts will be resolved by following the priority rank described above (from 1 to 3).
     // The env var condition is verified in the config file processing
-    int roll_fwd_on_no_candidate_fx_val = config.get_roll_fwd_on_no_candidate_fx();
     if (!roll_fwd_on_no_candidate_fx.empty())
     {
-        roll_fwd_on_no_candidate_fx_val = pal::xtoi(roll_fwd_on_no_candidate_fx.c_str());
+        config.set_get_roll_fwd_on_no_candidate_fx(pal::xtoi(roll_fwd_on_no_candidate_fx.c_str()));
```
eerhardt: (nit) `set_get_` is kind of a strange name.  Why not just `set_`?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1195,86 +1237,111 @@ int fx_muxer_t::read_config_and_execute(
...
+        // Obtain frameworks\platforms
+        auto version = fx_version_specified;
+        while (!config.get_fx_name().empty() && !config.get_fx_version().empty())
+        {
+            fx_definition_t* fx = resolve_fx_dir(mode, config, own_dir, version);
+            if (fx == nullptr)
+            {
+                pal::string_t searched_version = fx_version_specified.empty() ? config.get_fx_version() : fx_version_specified;
+                handle_missing_framework_error(mode, config.get_fx_name(), searched_version, pal::string_t(), own_dir);
+                return CoreHostLibMissingFailure;
```
eerhardt: This isn't really the right return code, right?  We are probing other shared frameworks, which will never have `CoreHostLib`.  Can we define a new return code for this?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -692,11 +704,16 @@ bool deps_resolver_t::resolve_probe_dirs(
         }
     }
     
-    for (const auto& entry : fx_entries)
+    // Add fx package locations to fx_dir
+    for (int i = 1; i < m_fx_definitions.size(); ++i)
     {
-        if (!add_package_cache_entry(entry, m_fx_dir))
+        const auto& fx_entries = m_portable ? m_fx_definitions[i]->get_deps().get_entries(asset_type) : empty;
```
eerhardt: How could we ever get in here if `m_portable` is false?

src/corehost/cli/hostpolicy.cpp
```diff
@@ -100,8 +100,28 @@ int run(const arguments_t& args)
     pal::pal_clrstring(probe_paths.native, &native_dirs_cstr);
     pal::pal_clrstring(probe_paths.resources, &resources_dirs_cstr);
 
-    pal::pal_clrstring(resolver.get_fx_deps_file(), &fx_deps);
-    pal::pal_clrstring(resolver.get_deps_file() + _X(";") + resolver.get_fx_deps_file(), &deps);
+    if (resolver.get_fx_definitions().size() >= 2)
+    {
+        // Use the first fx to define FX_DEPS_FILE
```
eerhardt: I'm not sure this is right.  I think it might make more sense to be the `root` framework.  There is code that expects this to be `Microsoft.NETCore.App`.  See 

[https://github.com/dotnet/cli/blob/cc80822c16b955a1f664f8a65ed2c76c3c52a824/src/Microsoft.DotNet.Cli.Utils/Muxer.cs#L17-L24](https://github.com/dotnet/cli/blob/cc80822c16b955a1f664f8a65ed2c76c3c52a824/src/Microsoft.DotNet.Cli.Utils/Muxer.cs#L17-L24)

or any of these places:

[https://github.com/search?q=org%3Adotnet+FX_DEPS_FILE&type=Code](https://github.com/search?q=org%253Adotnet+FX_DEPS_FILE&type=Code)

src/corehost/cli/libhost.h
```diff
@@ -76,7 +81,11 @@ static_assert(offsetof(host_interface_t, host_mode) == 14 * sizeof(size_t), "Str
 static_assert(offsetof(host_interface_t, tfm) == 15 * sizeof(size_t), "Struct offset breaks backwards compatibility");
 static_assert(offsetof(host_interface_t, additional_deps_serialized) == 16 * sizeof(size_t), "Struct offset breaks backwards compatibility");
 static_assert(offsetof(host_interface_t, fx_ver) == 17 * sizeof(size_t), "Struct offset breaks backwards compatibility");
-static_assert(sizeof(host_interface_t) == 18 * sizeof(size_t), "Did you add static asserts for the newly added fields?");
+static_assert(offsetof(host_interface_t, fx_names) == 18 * sizeof(size_t), "Struct offset breaks backwards compatibility");
```
eerhardt: You may want to test out `1.0` and `1.1` apps just to make sure you didn't break them.

src/corehost/cli/libhost.h
```diff
@@ -100,54 +107,66 @@ class corehost_init_t
...
+            m_fx_names_cstr.push_back(fx_definitions[i]->get_name().c_str());
+            m_fx_dirs.push_back(fx_definitions[i]->get_dir());
+            m_fx_dirs_cstr.push_back(fx_definitions[i]->get_dir().c_str());
+            m_fx_requested_versions.push_back(fx_definitions[i]->get_requested_version());
+            m_fx_requested_versions_cstr.push_back(fx_definitions[i]->get_requested_version().c_str());
+            m_fx_found_versions.push_back(fx_definitions[i]->get_found_version());
+            m_fx_found_versions_cstr.push_back(fx_definitions[i]->get_found_version().c_str());
+        }
+
+        runtime_config_t::get_properties(combined_properties, &m_clr_keys, &m_clr_values);
```
eerhardt: This method doesn't really feel like it belongs on `runtime_config_t`.  It is a simple loop over `combined_properties` and filling up `m_clr_keys` and `m_clr_values`.  It feels like those 3 lines of code could just be here.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Thanks for the detailed reply. The code now has a single vector, and there was no need to use std::move(). Passing the vector my reference was sufficient.

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
steveharter: I added the fx name

src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
steveharter: Add deletion, including to pre-existing code

src/corehost/cli/libhost.h
steveharter: Done. Thanks

src/corehost/cli/fxr/fx_muxer.cpp
nguerrera: Ah, for some reason I had in mind that corehost_init_t was going to hold on to the vector of pointers, which is where the move idea came from.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Done

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: [0] is not a framework, but it contains the framework information (via runtimeconfig and deps) for the first framework, so kindof. I did add a get_app() helper method to fx_definitions_t to at least make it clear we're talking about the app's framework information.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Removed that change

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -568,31 +570,34 @@ fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& vers
     return most_compatible;
 }
 
-pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode,
-    const pal::string_t& own_dir,
+fx_definition_t* resolve_fx_dir(
```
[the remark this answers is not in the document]
steveharter: Sure. ->resolve_fx

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Fixed (copy-paste issue) Thanks

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Good point; note the existing code used this as well when the framework wasn't found (in addition to when the hostpolicy wasn't found). I added FrameworkMissingFailure value

src/corehost/cli/deps_resolver.cpp
steveharter: Removed the check. I believe the code used to start with `0` instead of `1` thus the check

src/corehost/cli/hostpolicy.cpp
steveharter: Thanks. Changed to root

src/corehost/cli/libhost.h
steveharter: Hmm not sure why it ended up being a separate method. Removed it

src/corehost/cli/libhost.h
steveharter: Tested on 1.0.8 and 1.1.2 runtime with simple console "Hello World" app; basically new hostfxr.dll loaded the older hostpolicy.dll successfully

PR #3464, opened 2017-11-28
Reset package list for 2.0.5 servicing

safern:
cc: @weshaggard @eerhardt

### DIFF
--- a/src/pkg/packaging/dir.proj
+++ b/src/pkg/packaging/dir.proj
@@ -176,6 +176,8 @@
 
     <ItemGroup Condition="'$(BuildAllPackages)' == 'true'">
       <!-- The list of packages we are not servicing -->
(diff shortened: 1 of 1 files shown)


src/pkg/packaging/dir.proj
```diff
@@ -172,8 +172,6 @@
 
     <ItemGroup>
       <!-- The list of packages we are servicing -->
-      <PackageProjects Include="$(ProjectDir)src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
-      <PackageProjects Include="$(ProjectDir)src\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj" />
     </ItemGroup>
 
     <ItemGroup Condition="'$(BuildAllPackages)' == 'true'">
```
eerhardt: I think you should add the above list to this item group.

safern: Oh , you're right. Thanks. Just fixed it.

PR #3465, opened 2017-11-28
Integrate new symbol uploader build task.

mikem8361:
Symbol publishing will only happen for the pipeline builds if PB_SymbolServerPath and
PB_SymbolServerPAT are defined. PB_SymbolExpirationInDays is how long the symbol server
will keep. Default is 30 days. -1 means forever.

### DIFF
--- a/buildpipeline/Core-Setup-Publish.json
+++ b/buildpipeline/Core-Setup-Publish.json
@@ -234,7 +234,7 @@
         "solution": "$(PB_SourcesDirectory)\\publish\\publish.proj",
         "platform": "$(PB_TargetArchitecture)",
         "configuration": "$(BuildConfiguration)",
-        "msbuildArguments": "/p:Configuration=$(BuildConfiguration) $(PB_CommonMSBuildArgs) /p:PublishType=$(PB_PublishType) /p:SignType=$(PB_SignType) /p:NuGetFeedUrl=$(NUGET_FEED_URL) /p:NuGetSymbolsFeedUrl=$(NUGET_SYMBOLS_FEED_URL) /p:NuGetApiKey=$(NUGET_API_KEY) /p:AzureAccountName=$(PB_AzureAccountName) /p:ContainerName=$(PB_ContainerName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:ChecksumAzureAccountName=$(PB_ChecksumAzureAccountName) /p:ChecksumContainerName=$(PB_ChecksumContainerName) /p:ChecksumAzureAccessToken=$(PB_ChecksumAzureAccessToken) /p:PackagesUrl=$(PB_PackagesUrl) /p:SymbolPackagesUrl=$(PB_SymbolPackagesUrl) /p:TransportFeedAccessToken=$(PB_TransportFeedAccessToken) /p:OfficialPublish=true /p:GitHubUser=$(PB_GitHubUser) /p:GitHubEmail=$(PB_GitHubEmail) /p:GitHubAuthToken=$(GITHUB_PASSWORD) /p:VersionsRepoOwner=$(PB_VersionsRepoOwner) /p:VersionsRepo=$(PB_VersionsRepo) /p:VersionsRepoPath=build-info/dotnet/$(PB_RepoName)/$(SourceBranch) /p:Finalize=true /p:DotNetToolDir=$(DotNetToolDir) /p:EmbedIndexToolDir=$(EmbedIndexToolDir) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish.log",
+        "msbuildArguments": "/p:Configuration=$(BuildConfiguration) $(PB_CommonMSBuildArgs) /p:PublishType=$(PB_PublishType) /p:SignType=$(PB_SignType) /p:NuGetFeedUrl=$(NUGET_FEED_URL) /p:NuGetSymbolsFeedUrl=$(NUGET_SYMBOLS_FEED_URL) /p:NuGetApiKey=$(NUGET_API_KEY) /p:AzureAccountName=$(PB_AzureAccountName) /p:ContainerName=$(PB_ContainerName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:ChecksumAzureAccountName=$(PB_ChecksumAzureAccountName) /p:ChecksumContainerName=$(PB_ChecksumContainerName) /p:ChecksumAzureAccessToken=$(PB_ChecksumAzureAccessToken) /p:PackagesUrl=$(PB_PackagesUrl) /p:SymbolPackagesUrl=$(PB_SymbolPackagesUrl) /p:TransportFeedAccessToken=$(PB_TransportFeedAccessToken) /p:OfficialPublish=true /p:GitHubUser=$(PB_GitHubUser) /p:GitHubEmail=$(PB_GitHubEmail) /p:GitHubAuthToken=$(GITHUB_PASSWORD) /p:VersionsRepoOwner=$(PB_VersionsRepoOwner) /p:VersionsRepo=$(PB_VersionsRepo) /p:VersionsRepoPath=build-info/dotnet/$(PB_RepoName)/$(SourceBranch) /p:Finalize=true /p:DotNetToolDir=$(DotNetToolDir) /p:SymbolServerPath=$(PB_SymbolServerPath) /p:SymbolServerPAT=$(PB_SymbolServerPAT) /p:SymbolExpirationInDays=$(PB_SymbolExpirationInDays) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish.log",
         "clean": "false",
         "maximumCpuCount": "false",
         "restoreNugetPackages": "false",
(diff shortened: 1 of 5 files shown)


mikem8361:
The buildtools PR https://github.com/dotnet/buildtools/pull/1821 is needed by my changes. It adds the build task package to the optional tooling list.

publish/publish-symbols.proj
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <Import Project="$(PackagesDir)/$(PublishSymbolsPackage.ToLower())/$(PublishSymbolsPackageVersion)/build/PublishSymbols.targets" Condition="'$(SymbolServerPath)'!=''" />
+
+  <Target Name="Build"
```
weshaggard: Were can I have a look at this PublishSymbols target?

publish/publish-type.proj
```diff
@@ -72,32 +72,17 @@
     </ItemGroup>
   </Target>
 
-  <Target Name="FindDownloadedPackagesForSigning">
+  <Target Name="PublishSymbols"
```
weshaggard: We don't want to actually publish the symbols publicly at this point of the publish cycle we only want to push them to the blob feed. We will do a final publish of the full stack build where we will then publish symbols.

publish/publish-type.proj
```diff
@@ -9,7 +9,7 @@
     <BuildDependsOn>
       ValidateProperties;
       DownloadFilesFromContainer;
-      SignSymbolPackages
+      PublishSymbols
```
weshaggard: Does the new target actually handle signing the symbols correctly now or is that not needed any longer?

publish/publish.proj
```diff
@@ -215,32 +215,17 @@
...
-    <ItemGroup>
-      <EmbedIndexArgPairs Include="@(SymbolPackagesToEmbedIndex -> '%(Identity) $(PublishDirectory)%(Filename)%(Extension)')" />
-    </ItemGroup>
-
-    <!-- Inline the EmbedIndex.ps1 script used in CoreFX and CoreCLR: directly call EmbedIndex. -->
-    <Exec Command="$(DotNetToolDir)\dotnet $(EmbedIndexToolDir)\tools\EmbedIndex.dll %(EmbedIndexArgPairs.Identity)" />
+    <PropertyGroup>
+      <_SymbolPackagesToPublish>@(SymbolPackagesToPublish)</_SymbolPackagesToPublish>
+    </PropertyGroup>
+    <MSBuild Projects="$(MSBuildThisFileDirectory)\publish-symbols.proj"
```
weshaggard: Instead of doing this I would prefer we split out the symbol publishing step by it self in the publish build definition leg.

config.json
```diff
@@ -78,7 +78,7 @@
       "values": [],
       "defaultValue": ""
     },
-    "BatchRestorePackages": {
+    "Sync": {
```
weshaggard: This seems unrelated.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -488,8 +488,25 @@
     "PB_ToolsRoot": {
       "value": "$(Build.SourcesDirectory)\\tools"
     },
-    "SymbolCatalogCertificateId": {
-      "value": "400"
+    "PB_OptionalToolSource": {
+      "value": "https://devdiv.pkgs.visualstudio.com/_packaging/dotnet-core-internal-tooling/nuget/v3/index.json "
```
weshaggard: @dagood don't we already handle the optional tools download? If not how do we currently get the embed index tool?

dependencies.props
```diff
@@ -53,6 +53,12 @@
     <FeedTasksPackageVersion>2.1.0-prerelease-02221-02</FeedTasksPackageVersion>
   </PropertyGroup>
 
+  <!-- Publish symbol build task package -->
+  <PropertyGroup>
+    <PublishSymbolsPackage>Microsoft.SymbolUploader.Build.Task</PublishSymbolsPackage>
+    <PublishSymbolsPackageVersion>1.0.0-alpha-00001</PublishSymbolsPackageVersion>
```
weshaggard: This is too tightly coupled to the version number in the buildtools repo. We need a better way.

buildpipeline/Core-Setup-Publish.json
dagood: Right now it's only in `Core-Setup-Signing-Validation.json`, I believe the changes in this file are just copying it over. (Extra space is also there.)

mikem8361:
Link to that file:

https://devdiv.visualstudio.com/DevDiv/_git/DotNet-SymStore?path=%2Fsrc%2FMicrosoft.SymbolUploader.Build.Task%2FPackageFiles%2FPublishSymbols.targets&version=GBmikem%2Fuploader&_a=contents

mikem8361:
There is no signing involved or needed anymore.  We will be uploading the individual files in the symbol packages to symweb (via the VSTS symbol server https://microsoft.artifacts.visualstudio.com/defaultcollection/_apis/Symbol/symsrv) and eventually the public facing msdl server too.  EmbedIndex, etc. is not needed anymore.

This means we are switching from the myget symbol server to Microsoft’s.

mikem8361:
It related because it allows sync.cmd to also restore the optional tooling packages like corefx does.

mikem8361:
>>>>@dagood don't we already handle the optional tools download? If not how do we currently get the embed index tool?

As far as I know (Davis can confirm) in core-setup only the sign verification step uses the optional tooling (see buildpipeline\core-setup-signing-validation.json).

mikem8361:
>>>This is too tightly coupled to the version number in the buildtools repo. We need a better way.

I used the same way as the FeedTaskPackage (see the same file). This version is the build task package from the DotNet-SymStore repo. It is unfortunate that my only change to buildtools is to add this package to the optional tooling but that what @dagood wanted instead of override optional.json like in corefx.

mikem8361:
>>>> We don't want to actually publish the symbols publicly at this point of the publish cycle we only want to push them to the blob feed. We will do a final publish of the full stack build where we will then publish symbols.

Where (what target file or target I guess) is the final publish of the full stack build?  I was trying to put this where symbols were published using the old Symbols.target; I may have missed the mark.

I also needed the symbol nuget packages so I put the publish symbols target after they are downloaded and the DownloadDirectory property was set.

I put the PublishSymbols target in publish.proj so the optional tooling packages could be restored before trying to import its target file (that happens in publish-symbols.proj).

I like the idea of have the symbol publish as independent as possible but I didn’t what to duplicate how the symbol packages are downloaded from the feed.

mikem8361:
@weshaggard, @dagood I've redone the changes not to use the optional tooling feed which has simplified things greatly.  I've deleted the buildtools PR and pushed my uploader build task package to the myget buildtools feed.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -488,8 +488,15 @@
...
-    "SymbolCatalogCertificateId": {
-      "value": "400"
+    "PB_SymbolServerPath": {
+      "value": "https://mikemvsts.artifacts.visualstudio.com/DefaultCollection"
+    },
+    "PB_SymbolServerPAT": {
+      "value": null,
+      "isSecret": true
+    }
+    "PB_SymbolExpirationInDays": {
```
dagood: Previous property is missing the `,`, causing this parse error.

publish/publish-type.proj
```diff
@@ -118,4 +89,5 @@
 
   <Import Project="$(ToolsDir)VersionTools.targets" Condition="Exists('$(ToolsDir)VersionTools.targets')" />
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
-</Project>
\ No newline at end of file
+</Project>
+
```
dagood: \<nit> One too many newlines.

dependencies.props
eerhardt: And we also need this version to be auto-updated.

weshaggard: @mikem8361 correct me if I'm wrong but this package is manually created so there isn't any auto-updates for it.

eerhardt: Isn't that even worse?  Where does this package live?

mikem8361:
The package is created from the private VSTS DotNet-SymStore repo. It is manual. It is similar to the Microsoft.DotNet.Build.Tasks.Feed package (see dependency.props.

Since it is still pretty “alpha” I think I’d rather explicitly update the version for now. Once it stabilizes then I can look into how to make it autoupdate.

mikem

eerhardt:
Since this is a closed source package, are we going to need a way to turn it "off" for https://github.com/dotnet/source-build/?  

dagood:
Yeah, I think we have the same problem with the Feed package in `init-tools.msbuild`, too, due to its dependencies. I'd classify both as selective BuildTools restore, filed https://github.com/dotnet/buildtools/issues/1824.

Issue #3455, opened 2017-11-24
Dotnet-Host RPM package upgrade removes symbolic link


rakeshsinghranchi:
## Steps to reproduce
Install dotnet-host-2.0.0 rpm package
Install dotnet-host-2.0.x rpm package

## Expected  behavior
dotnet-host payload should be upgraded
dotnet-host symbolic link should be preserved

## Actual behavior
dotnet-host symbolic link is removed as part of dotnet-host-2.0.0 uninstall step.

## Environment data

CRCinAU:
Duplicating my comment from the referenced issue for how we handle this issue when using a .spec file to build the RPM:

Hrrrm - in the .spec world, you'd do something like this:
```
%preun
if [ $1 = 0 ]; then
   unlink /usr/bin/dotnet
fi
```
This roughly translates to: If we're uninstalling (hence %preun), and we don't have any newer version installed, run the script.

Running a complete script file in what I assume is the %post or %preun sections of the RPM installation - I'm not sure there is an easy way to achieve the same - unless you can pass the same $1 to the script and evaluate it in said script...

PR #3467, opened 2017-11-29
Fix upgrade issue where symbolic link for dotnet-host rpm package gets removed

rakeshsinghranchi:
Fixes : https://github.com/dotnet/core-setup/issues/3455

More details : http://meinit.nl/rpm-spec-prepostpreunpostun-argument-values

### DIFF
--- a/src/pkg/packaging/rpm/scripts/after_remove_host.sh
+++ b/src/pkg/packaging/rpm/scripts/after_remove_host.sh
@@ -3,5 +3,8 @@
 # Copyright (c) .NET Foundation and contributors. All rights reserved.
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
-echo "Removing dotnet host symbolic link"
-unlink /usr/bin/dotnet
+# Run the script only when newer version is not installed. Skip during package upgrade 
+if [ $1 = 0 ]; then
+   echo "Removing dotnet host symbolic link"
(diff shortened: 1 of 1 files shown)


src/pkg/packaging/rpm/scripts/after_remove_host.sh
```diff
@@ -3,5 +3,8 @@
 # Copyright (c) .NET Foundation and contributors. All rights reserved.
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
-echo "Removing dotnet host symbolic link"
-unlink /usr/bin/dotnet
+# Run the script only when newer version is not installed. Skip during package upgrade 
+if [ $1 = 0 ]; then
```
eerhardt: (nit) I think it reads easier if you give a name to the arguments.  That way someone doesn't have to look up what the `$1` arg is.

See 

https://github.com/dotnet/core-setup/blob/eef6a7b1fe6883bee0aa967eb1a434c19f6990f4/src/pkg/packaging/osx/sharedhost/scripts/postinstall#L7-L8

As an example.

rakeshsinghranchi: This is by design. Fedora guidance for spec file specifies the variable as $1 . See https://fedoraproject.org/wiki/Packaging:Scriptlets

PR #3472, opened 2017-11-29
Add condition on BuildCustomTasks to not build in offline build

dseefeld:
With the OfflineBuild, tasks are already present in the tarball, so they don't need to be built again.  Adding a condition on BuildCustomTasks to only build when OfflineBuild isn't set.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -55,7 +55,7 @@
                       Overwrite="True" />
   </Target>
 
-  <Target Name="BuildCustomTasks">
+  <Target Name="BuildCustomTasks" Condition="'$(DotNetBuildOffline)' != 'true'" >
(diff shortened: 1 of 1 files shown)


build.proj
```diff
@@ -55,7 +55,7 @@
                       Overwrite="True" />
   </Target>
 
-  <Target Name="BuildCustomTasks">
+  <Target Name="BuildCustomTasks" Condition="'$(DotNetBuildOffline)' != 'true'" >
```
weshaggard: Is it safe to assume these already exist in offline mode? Should we instead do some other condition based on their existence?

dseefeld: It is save to assume that they already exists.  They're being built into the Tools directory, which gets copied into the tarball.  If they don't exist, there's a problem with the original build.

weshaggard: I guess it isn't clear to me that DotNetbuildOffline is only going to be used in the context of the tar-ball. I would expect it to also work on a git repo that I didn't want to restore things from the network.

PR #3483, opened 2017-12-01
Fix alpine linux RID graph generation

janvorli:
Alpine Linux doesn't have 'linux' in its hierarchy, so the
GenerateRuntimeGraph target doesn't output alpine node to the runtimes
section of the Microsoft.NETCore.App.deps.json.

This change is a quick fix to that issue to unblock my work on enabling building of CLI for Alpine. 
In longer term, we should create a new osgroup named `alpine` or maybe better `linuxmusl` or something
along these lines and make it a fully supported OS group everywhere.


janvorli:
@eerhardt it seems there is a problem with this change on Alpine, please do not merge it until I verify it.

janvorli:
@eerhardt the fix for Alpine rid parent in corefx fixed the issue I was having, so this PR is ready to review / merge.

eerhardt (approved):
Looks good.  Thanks @janvorli.

PR #3484, opened 2017-12-01
Update CoreCLR and Jit version

safern:
cc: @weshaggard


safern:
Merging to unblock the builds.

Issue #3325, opened 2017-10-18
[Proposal] : Enable building distro agnostic Debs and RPMs for Dotnet Runtime and upstream packages


rakeshsinghranchi:
**Problem :**
Building native installer for a new distro ( and/or version ) is extremely difficult as it requires changes in all the repos where Dotnet installers are produced ( Runtime|Aspnetcore|SDK ).

**Description:**
As of today , .NET Core builds Distro and Version specific Debs (Ubuntu 14.04, 16.04, & 17.04, Debian 8 & 9 ) and RPMs for Runtime, SDK and Aspnetcore . Dotnet Core Runtime , SDK and Aspnetcore packages are built in their respective repo. SDK package exhibits dependency on Runtime and Aspnetcore packages. Runtime package carries portable payload and exhibits system dependencies.These system dependencies ( or packages ) are Distro/version dependent ( for .e.g libss1.0.2 in Debian.9 and libssl1.0.0 in debian.8) .  Thus, it is the forcing factor to build distro and version specific runtime deb packages and thus, upstream packages. 

**Currently :** 
dotnet-host **-->**  libc6 
dotnet-hostfxr **-->** depends on dotnet-host  and libc6
dotnet-runtime **-->** depends on dotnet-hostfxr and [ distro and version specific system packages including libicuxx and libssl1.x.x] 
dotnet-sdk **-->** dotnet-runtime, aspnetcore-store

**Proposal :** 
**[new]** dotnet-runtime-deps **-->** depends on system packages [distro specific system packages like libicuxx and libssl1.x.x]
dotnet-host **-->** depends on dotnet-runtime-deps
dotnet-hostfxr --> depends on dotnet-host
dotnet-runtime --> depends on dotnet-hostfxr 
dotnet-sdk --> depends on dotnet-runtime,aspnet-store

**Solution:**
Build dotnet-runtime-deps package for each Distro & Version . Package name should be Distro/Version diagnostic . Dotnet Runtime ( dotnet-host package specifically ) exhibits dependency on this package and rest all upstream packages are built only once for each package type ( one deb and one rpm ) .

This will enable us to add new distro installer by just building the corresponding dotnet-runtime-deps package and there would be no change required in the upstream branches. During package upload to distro repo feeds, common Runtime and upstream packages will be uploaded along with matching dotnet-runtime-deps package. 

**Feed upload to a distro would include for e.g ubuntu14.04 :**

**Runtime:**
[...]

PR #3487, opened 2017-12-02
Enable building distro agnostic runtime Deb packages

rakeshsinghranchi:
This work is based on https://github.com/dotnet/core-setup/issues/3325  for Debian based packages. 

We have been building runtime-deps pacakage for each supported debian based distro. This work enables exhibiting dependency on runtime-deps package and removing system dependencies from other runtime packages ( host, hostfxr and runtime ). Looking at https://github.com/dotnet/core-setup/issues/3325 for benefits.

PR contains : 
1. Pre-defined distro specific config files for building runtime-deps package. This package exhibits dependency on system packages. Note that package name is distro agnostic.
2. dotnet-host package exhibits dependency on runtime-deps package. 
3. dotnet-host,dotnet-hostfxr and dotnet-runtime deb packages will not more exhibit system dependencies. Thus, making them distro agnostic. Also, these packages are built only once and package names do not carry host version in it.
4. All runtime-deps are built in ubuntu14.04 docker image and uses Outputrid to generated distro specific packages. Thus, cleaned up unnecessary distro specific variables from build definition
5. Removed DebRepoFeed upload related variables from build definition


Petermarcu:
@weshaggard this is related to the build times we were talking about today.

rakeshsinghranchi:
cc @leecow , @wli3

eerhardt:
Are we going to make the same changes to the .NET Core SDK .deb installers?

cc @livarcocc

tools-local/setuptools/dotnet-deb-tool/tool/templates/debian/control
```diff
@@ -13,7 +13,7 @@ Homepage: {HOMEPAGE}
 
 Package: {PACKAGE_NAME}
 Architecture: {ARCH}
-Depends: ${{shlibs:Depends}}, ${{misc:Depends}}{DEPENDENT_PACKAGES}
+Depends: {DEPENDENT_PACKAGES}
```
eerhardt: What does this change do?

src/pkg/packaging/deb/dotnet-sharedhost-debian_config.json
```diff
@@ -28,6 +28,11 @@
...
     },
 
+    "debian_dependencies":{
+         "%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME%" : {
+         "package_version": "%RUNTIME_DEPS_VERSION%"}
```
eerhardt: (nit) the whitespacing here seems off.  This should be tabbed over one since it is a child of `"%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME%"`

src/pkg/packaging/deb/package.targets
```diff
@@ -41,8 +46,13 @@
                             BuildDebs;"
           Condition="'$(BuildDebPackage)'=='true'" />
 
+  <!-- Build only on Ubuntu14.04 -->
+  <Target Name="GenerateRuntimeDebs" 
+          DependsOnTargets="$(RuntimeDebsDependsOn)"
+          Condition="'$(PackageTargetRid)' == 'ubuntu.14.04-x64'" />
```
eerhardt: Why encode this into the build scripts?  That makes it hard for someone who only has an ubuntu 16 machine develop and test in this area.

Instead, I think we should have a boolean property here that gets passed in by the build definition.

src/pkg/packaging/deb/package.targets
```diff
@@ -9,6 +9,11 @@
                                     $(PackageTargetRid.StartsWith('debian'))">true</IsDebianBasedDistro>
     <BuildDebPackage Condition="'$(IsDebianBasedDistro)' == true and '$(TargetArchitecture)' == 'x64'">true</BuildDebPackage>
     <DebianConfigJsonName>debian_config.json</DebianConfigJsonName>
+    <RuntimeDebsDependsOn>
```
eerhardt: Why split this into a separate property?  I don't think this provides any value over just putting the DependsOnTargets directly into the Target definition below.  This just adds a level of indirection that the reader has to go look for.

src/pkg/packaging/deb/package.targets
```diff
@@ -90,6 +100,12 @@
       <SharedHostTokenValue Include="%SHARED_HOST_BRAND_NAME%">
         <ReplacementString>$(SharedHostBrandName)</ReplacementString>
       </SharedHostTokenValue>
+     <SharedHostTokenValue Include="%RUNTIME_DEPS_DEBIAN_PACKAGE_NAME%">
+        <ReplacementString>$(RuntimeDependenciesDebPkgName)</ReplacementString>
+      </SharedHostTokenValue>
+      <SharedHostTokenValue Include="%RUNTIME_DEPS_VERSION%">
+        <ReplacementString>$(SharedFrameworkNugetVersion)</ReplacementString>
```
eerhardt: `SharedFrameworkNugetVersion` seems like the wrong variable here.  It should only be used when actually referencing the SharedFramework.  Here we are referencing the runtime-deps debian package.

I think this justifies its own new property, whose value get defaulted to `$(ProductVersion)`.

tools-local/setuptools/dotnet-deb-tool/tool/templates/debian/control
rakeshsinghranchi: This enables building runtime packages (host,hostfxr and runtime ) without exhibiting system dependencies by default as we are moving away all system dependencies to runtime-deps package and consolidating all the system dependencies statically in distro specific runtime-deps config file.

rakeshsinghranchi:
@dotnet-bot test Windows_NT arm Debug Build please

tools-local/setuptools/dotnet-deb-tool/tool/templates/debian/control
eerhardt: I assume you've tested installing these new installers on all our targeted platforms and it still works.  I'm not familiar enough with the way debhelper works to know if this is the right change or not - so I trust you've done the appropriate research here.

rakeshsinghranchi: Yes, I have tested installing all runtime packages on all targeted platforms. @wli3 is testing SDK with the private runtime packages to get another level of test coverage.

wli3:
Looks fine from reading the code. But I guess the only way to make sure it to run CLI against the test packages. Please give me more time to try it

wli3:
@rakeshsinghranchi please go head for merging. I tested with unzip deb, edit DEBIAN/control, repack. I can install it successfully.

PR #3492, opened 2017-12-05
Fix 1.0 dependency for 1.1.6

safern:
Fixing typo which caused to have a non-existing package dependency. 

cc: @weshaggard @livarcocc


safern:
Merged to unblock builds.

PR #3497, opened 2017-12-07
Update CoreFX package version explicitly to 4.6.0-preview1-26006-06

zamont:
We need this fix because we need to consume signed assemblies from that package


zamont:
I will merge this as soon as CI completes.

zamont:
cc: @ahsonkhan if you see the CI on this job completes before I see it, can you merge the PR?

zamont:
This will then require a core-setup master build launch, which you can access from the README

ahsonkhan:
> if you see the CI on this job completes before I see it, can you merge the PR?

Done.

Issue #3510, opened 2017-12-11
System.Globalization.Invariant from config.json is not passed to coreclr


janvorli:
## Steps to reproduce
In corefx repo, set `MicrosoftNETCoreDotNetHostPackageVersion` and `MicrosoftNETCoreDotNetHostPolicyPackageVersion` in `dependencies.props` to a recent version of the dotnet host package, e.g. 2.1.0-preview1-26007-02. 
Then e.g on Linux, run `./build.sh -tests`

## Expected  behavior
Test runs completes ok

## Actual behavior
Invariant.Tests fail due to the globalization not being set to invariant.

janvorli:
CC: @steveharter - it seems it was broken by your change https://github.com/dotnet/core-setup/pull/3460

PR #3512, opened 2017-12-12
Fix issue with not passing configProperties to CLR

steveharter:
The "configProperties" section of the runtimeconfig.json was not being passed to the CLR. This broke a couple weeks ago due to https://github.com/dotnet/core-setup/pull/3460

The fix is to move 2 LOC so that the keys and values are properly initialized.

A test was also modified to ensure the trace output lists an expected CLR variable, along with 'override' support for a higher-level runtimeconfig.json. However, the external fluentAssertions test dependency had an issue with a `=` being searched for in the new test (`"Property TestProperty = AppValue"`), so the core-setup test extensions (CommandResultAssertions.cs ) were modified to pass variables instead of literals to fluentAssertions which took care of the `=` issue.

Fixes https://github.com/dotnet/core-setup/issues/3510

### DIFF
--- a/src/test/TestUtils/Assertions/CommandResultAssertions.cs
+++ b/src/test/TestUtils/Assertions/CommandResultAssertions.cs
@@ -21,114 +21,115 @@ namespace Microsoft.DotNet.CoreSetup.Test
         public AndConstraint<CommandResultAssertions> ExitWith(int expectedExitCode)
         {
             Execute.Assertion.ForCondition(_commandResult.ExitCode == expectedExitCode)
-                .FailWith(AppendDiagnosticsTo($"Expected command to exit with {expectedExitCode} but it did not."));
+                .FailWith("Expected command to exit with {0} but it did not.{1}", expectedExitCode, GetDiagnosticsInfo());
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
         public AndConstraint<CommandResultAssertions> Pass()
         {
             Execute.Assertion.ForCondition(_commandResult.ExitCode == 0)
-                .FailWith(AppendDiagnosticsTo($"Expected command to pass but it did not."));
+                .FailWith("Expected command to pass but it did not.{0}", GetDiagnosticsInfo());
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
         public AndConstraint<CommandResultAssertions> Fail()
         {
             Execute.Assertion.ForCondition(_commandResult.ExitCode != 0)
-                .FailWith(AppendDiagnosticsTo($"Expected command to fail but it did not."));
+                .FailWith("Expected command to fail but it did not.{0}", GetDiagnosticsInfo());
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
         public AndConstraint<CommandResultAssertions> HaveStdOut()
         {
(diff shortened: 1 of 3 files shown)


Petermarcu:
LGTM

src/test/TestUtils/Assertions/CommandResultAssertions.cs
```diff
@@ -21,114 +21,115 @@ public CommandResultAssertions(CommandResult commandResult)
...
-            _commandResult.StdOut.Should().Contain($"Project {skippedProject} ({frameworkFullName}) was previously compiled. Skipping compilation.");
+            _commandResult.StdOut.Should().Contain("Project {0} ({1}) was previously compiled. Skipping compilation.", skippedProject, frameworkFullName);
 
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
         public AndConstraint<CommandResultAssertions> HaveCompiledProject(string compiledProject, string frameworkFullName)
         {
-            _commandResult.StdOut.Should().Contain($"Project {compiledProject} ({frameworkFullName}) will be compiled");
+            _commandResult.StdOut.Should().Contain($"Project {0} ({1}) will be compiled", compiledProject, frameworkFullName);
```
eerhardt: Why did you remove the string interpolation?  In general it is easier to read and maintain than using `{0}`.

src/test/TestUtils/Assertions/CommandResultAssertions.cs
```diff
@@ -21,114 +21,115 @@ public CommandResultAssertions(CommandResult commandResult)
...
         {
             Execute.Assertion.ForCondition(string.IsNullOrEmpty(_commandResult.StdErr))
-                .FailWith(AppendDiagnosticsTo("Expected command to not output to stderr but it was not:"));
+                .FailWith("Expected command to not output to stderr but it was not:{0}", GetDiagnosticsInfo());
             return new AndConstraint<CommandResultAssertions>(this);
         }
 
-        private string AppendDiagnosticsTo(string s)
+
+        private string GetDiagnosticsInfo()
```
eerhardt: I'm not sure this pattern is better.  Having to add `{0}` to the end of each string above seems unnecessary.

Why was this changed?

steveharter: Because the fluentassetions library didn't like the `=` in the literal error message. It basically ended up throwing a exception coming from String.Format().

Fixing this also explains why certain (unexpected) failures would not show the stdout information (throwing the String.Format exception), making it difficult to determine the cause -- the stdout contains some literal values that ended up throwing an exception.

src/test/TestUtils/Assertions/CommandResultAssertions.cs
steveharter: See below

src/test/TestUtils/Assertions/CommandResultAssertions.cs
steveharter: fwiw it's possible there was something in stdout and not the test string's `=` that was causing the issue (both were previously concat'd into one string literal); I didn't debug the issue to that level. Fluent assertions has its own expansion logic, so I believe it was getting tripped up by special characters

PR #3523, opened 2017-12-13
Fix signing build

mikem8361:
Broken by a publish symbol change in PR #3465

### DIFF
--- a/publish/publish-type.proj
+++ b/publish/publish-type.proj
@@ -48,11 +48,8 @@
     <!-- TODO: We should also think about pulling down the checksums or recomputing them -->
     <PropertyGroup>
       <DownloadDirectory Condition="'$(DownloadDirectory)' == ''">$(BinDir)PackageDownload/</DownloadDirectory>
-      <PublishDirectory Condition="'$(PublishDirectory)' == ''">$(BinDir)ForPublishing/</PublishDirectory>
-      <!-- if we're not signing packages, publish directly from the download directory, as we won't be
-	       copying them to the indexed directory -->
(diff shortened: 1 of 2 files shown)


publish/publish.proj
```diff
@@ -222,6 +222,16 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="CopyPackagesForSigning">
```
weshaggard: The copying actually isn't needed any longer. Instead you should just make DownloadDirectory and PublishDirectory above the same.

mikem8361:
I just pushed a lot simpler change with no copying.

publish/publish-type.proj
```diff
@@ -48,11 +48,9 @@
     <!-- TODO: We should also think about pulling down the checksums or recomputing them -->
     <PropertyGroup>
       <DownloadDirectory Condition="'$(DownloadDirectory)' == ''">$(BinDir)PackageDownload/</DownloadDirectory>
-      <PublishDirectory Condition="'$(PublishDirectory)' == ''">$(BinDir)ForPublishing/</PublishDirectory>
-      <!-- if we're not signing packages, publish directly from the download directory, as we won't be
-	       copying them to the indexed directory -->
-      <PublishDirectory Condition="'$(WindowsSdkDir)' == '' OR '$(SignType)' != 'real'">$(DownloadDirectory)</PublishDirectory>
+      <PublishDirectory Condition="'$(PublishDirectory)' == ''">$(DownloadDirectory)</PublishDirectory>
```
weshaggard: We should probably just eliminate PublishDirectory and always use DownloadDirectory. As it stands right now someone could accidentally set them independently which would break things.

PR #3495, opened 2017-12-06
Fix dependency props files import order

dagood:
This allows dependencies.props to pick up properties defined by BuildTools, such as `DotNetPackageVersionPropsPath` being assigned to the downloaded package version props path. Importing this early meant that dependency versions weren't being overridden in orchestrated builds.

Now the order is more like CoreFX's: https://github.com/dotnet/corefx/blob/33ec2994d85935b6c62df0792aef88b285d6c765/dir.props#L52-L82

PR #3533, opened 2017-12-14
Prepend to RestoreSources: BuildTools may set it

dagood:
This property group used to be before the `Build.Common.props` import, and my change that moved it after (https://github.com/dotnet/core-setup/pull/3495/) broke this scenario. Instead of moving it back, change it to match what CoreFX is doing for maintainability.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -52,6 +52,7 @@
       https://www.myget.org/F/nugetbuild/api/v3/index.json;
       https://api.nuget.org/v3/index.json;
       https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json;
+      $(RestoreSources);
     </RestoreSources>
   </PropertyGroup>
 


dir.props
```diff
@@ -52,6 +52,7 @@
       https://www.myget.org/F/nugetbuild/api/v3/index.json;
       https://api.nuget.org/v3/index.json;
       https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json;
+      $(RestoreSources);
```
tannergooding: Should this be prepended, as per the title, rather than appended?

dagood: `RestoreSources` is being prepended *to*. Regardless of wording, this is how CoreFX does it: https://github.com/dotnet/corefx/blob/ea8fe5d29798dbfa2a1ed33341b788b61aa59795/dir.props#L80-L81.

Issue #3469, opened 2017-11-29
Roll forward policy changes


steveharter:
Roll-forward policy changes [have been requested for 2.1](https://github.com/dotnet/designs/issues/3).

A framework version is identified by [major.minor.patch]. This issue only pertains to [major.minor] roll-forward because [patch] roll-forward is a separate setting, and no changes are planned for 2.1 for that.

For compatibility, we always want to run on the [major.minor] of the framework that an app targeted, even if newer [major.minor] versions are available on the local system. However, if the [major.minor] is not available on the local system, the following changes are planned for 2.1:
1. Automatically roll-forward on [minor]
-- In 2.0, this is possible but not enabled by default
2. Automatically roll-forward from release->preview on [minor] and [patch] (if patch roll-forward is enabled)
-- In 2.0, this is not possible
-- Try to roll-forward to only release versions. Roll-forward to closest preview version only when there isn't a compatible release version

Also (unless there is feedback otherwise) we plan to keep the existing semantics of **not** allowing preview->release roll-forward.  When a preview build is specified we only roll forward if the specified preview version doesn't exist, and only roll forward to preview versions that have the same [major.minor.patch] and never to release versions.

We need to determine what settings are available to control these. There is existing support for [roll-forward-on-no-candidate-fx](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md) which has environment variable,  runtimeconfig.json and argument support to control the major and minor roll-forward (patch roll-forward is controlled by a different setting). One quirk with this setting is that it enables [major] roll forward, which is not desired (by default anyway).

One option is to extend the roll-forward-on-no-candidate-fx setting into three values:

0. Off.
1. Roll forward on minor
2. Roll forward on major and minor

The default value would be (1). For the `--roll-forward-on-no-candidate-fx` argument, an additional value can be provided (0, 1 or 2) otherwise the default value of (1) will be applied.

steveharter:
Update: I am currently prototyping this for preview1

Petermarcu:
I believe this all matches my understanding. thanks for writing it up.

piers7:
> "we plan to keep the existing semantics of not allowing preview->release roll-forward"

[...]

PR #3537, opened 2017-12-15
Roll-forward policy changes

steveharter:
1) Automatically roll-forward on [minor] if targeted version is not installed (_previously off by default_)
2) Automatically roll-forward from release->preview on [minor] if there is no viable release version (_previously not possible_)
3) Extend the roll-forward-on-no-candidate-fx setting from two to three values:
0 = Off.
1 = Roll forward on minor
2 = Roll forward on major and minor (_previously this was the semantics for value >= 1_)

Fixes https://github.com/dotnet/core-setup/issues/3469

Current doc for 2.0 is located at https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md; this needs to be updated to 2.1 semantics.

Note there is still an ongoing discussion on whether roll-forward changes from preview->release should be done; this will be handled in a separate PR if we decide to do anything.

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -508,7 +508,7 @@ fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& vers
     fx_ver_t most_compatible = specified;
     if (!specified.is_prerelease())
     {
-        if (roll_fwd_on_no_candidate_fx != roll_fwd_on_no_candidate_fx_option::disabled)
+        if (roll_fwd_on_no_candidate_fx != roll_fwd_on_no_candidate_fx_option::roll_fwd_disabled)
         {
             fx_ver_t next_lowest(-1, -1, -1);
 
@@ -520,7 +520,7 @@ fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& vers
             {
                 if (!ver.is_prerelease() && ver >= specified)
                 {
-                    if (roll_fwd_on_no_candidate_fx == roll_fwd_on_no_candidate_fx_option::minor)
+                    if (roll_fwd_on_no_candidate_fx == roll_fwd_on_no_candidate_fx_option::roll_fwd_minor)
                     {
                         // We only want to roll forward on minor
                         if (ver.get_major() != specified.get_major())
@@ -542,7 +542,7 @@ fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& vers
                 {
                     if (ver.is_prerelease() && ver >= specified)
                     {
-                        if (roll_fwd_on_no_candidate_fx == roll_fwd_on_no_candidate_fx_option::minor)
+                        if (roll_fwd_on_no_candidate_fx == roll_fwd_on_no_candidate_fx_option::roll_fwd_minor)
                         {
                             // We only want to roll forward on minor
                             if (ver.get_major() != specified.get_major())
@@ -680,7 +680,7 @@ fx_definition_t* fx_muxer_t::resolve_fx(
             if (!specified.is_prerelease())
             {
                 // If production and no roll forward use given version.
-                do_roll_forward = (config.get_patch_roll_fwd()) || (config.get_roll_fwd_on_no_candidate_fx() != roll_fwd_on_no_candidate_fx_option::disabled);
+                do_roll_forward = (config.get_patch_roll_fwd()) || (config.get_roll_fwd_on_no_candidate_fx() != roll_fwd_on_no_candidate_fx_option::roll_fwd_disabled);
             }
             else
             {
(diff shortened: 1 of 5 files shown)


src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -609,12 +645,20 @@ fx_definition_t* fx_muxer_t::resolve_fx(
...
+    pal::string_t own_dir_temp = own_dir;
+    remove_trailing_dir_seperator(&own_dir_temp);
+
+    hive_dir.push_back(own_dir_temp);
     if (multilevel_lookup && pal::get_global_dotnet_dirs(&global_dirs))
     {
         for (pal::string_t dir : global_dirs)
         {
-            hive_dir.push_back(dir);
+            // Avoid duplicate of own_dir_temp
```
steveharter: Review note: this avoids doing an unnecessary comparison when the two directories are the same

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -682,11 +726,14 @@ fx_definition_t* fx_muxer_t::resolve_fx(
 
             if (pal::directory_exists(fx_dir))
             {
-                //Continue to search for a better match if available
-                std::vector<fx_ver_t> version_list;
-                version_list.push_back(resolved_ver);
-                version_list.push_back(selected_ver);
-                resolved_ver = resolve_framework_version(version_list, fx_ver, specified, config.get_patch_roll_fwd(), config.get_roll_fwd_on_no_candidate_fx());
+                if (selected_ver != fx_ver_t(-1, -1, -1))
```
steveharter: Review note: this avoids doing an unnecessary comparison against (-1, -1, -1) for the first loop

src/corehost/cli/runtime_config.cpp
```diff
@@ -29,6 +29,16 @@ void runtime_config_t::parse(const pal::string_t& path, const pal::string_t& dev
         m_prerelease_roll_fwd = defaults->m_prerelease_roll_fwd;
         m_roll_fwd_on_no_candidate_fx = defaults->m_roll_fwd_on_no_candidate_fx;
     }
+    else
+    {
+        // Since there is no previous config, this is the app's config, so default m_roll_fwd_on_no_candidate_fx from the env variable.
```
steveharter: This fixes a bug (along with code removed later on): we only want the first config (the app's config) to be defaulted from the env variable. Then the first framework will be defaulted from the app's config, but the framework can override in its config. The next, etc. framework will default from the previous framework but can still override in its config as well.

Most likely framework's won't override this setting, and is not likely a user will change it since it may require admin rights at least if it's in the shared Program Files location.

src/corehost/cli/runtime_config.h
```diff
@@ -11,6 +11,13 @@
 
 typedef web::json::value json_value;
 
+enum roll_fwd_on_no_candidate_fx_option
+{
+    roll_fwd_disabled = 0,
+    roll_fwd_minor,
+    roll_fwd_major_or_minor
+};
+
```
nguerrera: Consider `enum class` (scoped enum) and no `roll_fwd_` prefixes on members.

src/corehost/cli/runtime_config.h
```diff
@@ -26,8 +33,8 @@ class runtime_config_t
     const std::list<pal::string_t>& get_probe_paths() const;
     bool get_patch_roll_fwd() const;
     bool get_prerelease_roll_fwd() const;
-    int get_roll_fwd_on_no_candidate_fx() const;
-    void set_roll_fwd_on_no_candidate_fx(int value);
+    enum roll_fwd_on_no_candidate_fx_option get_roll_fwd_on_no_candidate_fx() const;
```
nguerrera: It's unnecessary to prefix all of the uses of the enum with `enum `.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -501,47 +501,83 @@ fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& vers
     const pal::string_t& fx_ver,
     const fx_ver_t& specified,
     const bool& patch_roll_fwd,
-    const int& roll_fwd_on_no_candidate_fx)
+    const enum roll_fwd_on_no_candidate_fx_option& roll_fwd_on_no_candidate_fx)
```
nguerrera: >&  [](start = 49, length = 2)

It's consistent with the bool just above, but it's kind of odd (IMHO) pass an enum by reference.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -501,47 +501,83 @@ fx_ver_t fx_muxer_t::resolve_framework_version(const std::vector<fx_ver_t>& vers
...
 {
     trace::verbose(_X("Attempting FX roll forward starting from [%s]"), fx_ver.c_str());
 
     fx_ver_t most_compatible = specified;
     if (!specified.is_prerelease())
```
nguerrera: >!specified.is_prerelease() [](start = 8, length = 26)

On [https://github.com/dotnet/core-setup/issues/3469](https://github.com/dotnet/core-setup/issues/3469), you said "When a preview build is specified we only roll forward if the specified preview version doesn't exist" but here it seems like all of the roll forward logic is disabled when a preview is specified. What am I missing?

nguerrera: I'm missing the `else` of course. :)

---
In reply to: [157608241](https://github.com/dotnet/core-setup/pull/3537#discussion_r157608241) [](ancestors = 157608241)

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: I wasn't sure if this was a new convention or not, but I agree. I'll at least change this signature to pass primitives by value

src/corehost/cli/runtime_config.h
steveharter: Done. Thanke

src/corehost/cli/runtime_config.h
steveharter: Done. Thanks

PR #3559, opened 2018-01-02
Update Wix toolset version to 3.10.4

rakeshsinghranchi:


rakeshsinghranchi:
@dotnet-bot test Windows_NT x86 Debug build please

src/pkg/packaging/windows/package.targets
```diff
@@ -22,7 +22,7 @@
             Condition="'$(OSGroup)' == 'Windows_NT' and '$(GenerateMSI)' == 'true'">
       
       <PropertyGroup>
-        <WixVersion Condition="'$(WixVersion)' == ''">3.10.2</WixVersion>
+        <WixVersion Condition="'$(WixVersion)' == ''">3.10.4</WixVersion>
```
eerhardt: How can this ever happen when `$(WixVersion)` is unconditionally set in the `package.props` file that's imported above?

Can we just remove this line?

src/pkg/packaging/windows/GetWix.ps1
```diff
@@ -1,5 +1,5 @@
 ﻿Param(
-    [string]$WixFilename="wix.3.10.2.zip",
+    [string]$WixFilename="wix.3.10.4.zip",
```
eerhardt: Can we just remove the default value, so this version number isn't copied in multiple places? This parameter is always specified when we call it from MSBuild. It would be nice to have a single place (the `package.props` file) that defines the version.  That way they can never be out of sync.

rakeshsinghranchi:
@wli3, did you investigation for this during cli repro update for wix version ?
>On a side note, did we investigate upgrading to 3.11.1 instead? I see that is the recommendation from WiX: https://twitter.com/wixtoolset/status/947586452921114624

src/pkg/packaging/windows/package.targets
```diff
@@ -22,11 +22,10 @@
             Condition="'$(OSGroup)' == 'Windows_NT' and '$(GenerateMSI)' == 'true'">
       
       <PropertyGroup>
-        <WixVersion Condition="'$(WixVersion)' == ''">3.10.2</WixVersion>
         <WixFilename>wix.$(WixVersion).zip</WixFilename>
       </PropertyGroup>
 
-      <Exec Command="powershell -NoProfile -NoLogo $(GetWixScript) -WixVersion $(WixVersion) -OutputDir $(WixToolsDir)" />
+      <Exec Command="powershell -NoProfile -NoLogo $(GetWixScript) -WixVersion $(WixVersion) -WixFilename $(WixFilename) -OutputDir $(WixToolsDir)" />
```
eerhardt: Sorry, I didn't look at this closer last time I reviewed it.

`-WixVersion` isn't a parameter to the `GetWix.ps1` script.  Can you remove that parameter?  I thought that the existing parameter was being passed in for the `[string]$WixFilename="wix.3.10.4.zip"` parameter.

So this line can be:

```
<Exec Command="powershell -NoProfile -NoLogo $(GetWixScript) -WixFilename $(WixFilename) -OutputDir $(WixToolsDir)" />
```

rakeshsinghranchi:
Anytime :)

PR #3572, opened 2018-01-05
Pass manifest details to PushToBlobFeed

dagood:
Passes in manifest info so that `core-setup.xml` will be created with useful manifest information instead of the default `anonymous.xml`. See https://github.com/dotnet/corefx/pull/26192

CoreCLR and CoreFX already had `GitHubRepositoryName` defined, so I added it here, too. The initial reason for the property was to create GitHub raw URLs for SourceLink--if this turns out to be destabilizing in Core-Setup I can use `PB_RepoName` instead.

This change also needs https://github.com/dotnet/buildtools/pull/1852 to fully work, but merge order doesn't matter.

https://github.com/dotnet/core-eng/issues/2335

### DIFF
--- a/buildpipeline/Core-Setup-Publish.json
+++ b/buildpipeline/Core-Setup-Publish.json
@@ -262,7 +262,7 @@
         "solution": "$(PB_SourcesDirectory)\\publish\\publish-type.proj",
         "platform": "$(PB_TargetArchitecture)",
         "configuration": "$(BuildConfiguration)",
-        "msbuildArguments": "/p:Configuration=$(BuildConfiguration) $(PB_CommonMSBuildArgs) /p:PublishType=$(PB_PublishType) /p:SignType=$(PB_SignType) /p:AzureAccountName=$(PB_AzureAccountName) /p:ContainerName=$(PB_ContainerName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:PublishBlobFeedUrl=$(PB_PublishBlobFeedUrl) /p:PublishBlobFeedKey=$(PB_PublishBlobFeedKey) /p:OfficialPublish=true /p:GitHubUser=$(PB_GitHubUser) /p:GitHubEmail=$(PB_GitHubEmail) /p:GitHubAuthToken=$(GITHUB_PASSWORD) /p:VersionsRepoOwner=$(PB_VersionsRepoOwner) /p:VersionsRepo=$(PB_VersionsRepo) /p:VersionsRepoPath=build-info/dotnet/$(PB_RepoName)/$(SourceBranch) /p:DotNetToolDir=$(DotNetToolDir) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish-blob.log",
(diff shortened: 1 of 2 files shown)


dir.props
```diff
@@ -16,6 +16,14 @@
 
   <Import Project="$(MSBuildThisFileDirectory)RepoDirectories.props" />
 
+  <PropertyGroup>
+    <!--
+      This name is used to create a GIT repository URL https://github.com/dotnet/$(GitHubRepositoryName) used to find source code for debugging
+      It is also used to name the build output manifest for orchestrated builds.
+    -->
+    <GitHubRepositoryName Condition="'$(GitHubRepositoryName)' == ''">core-setup</GitHubRepositoryName>
```
weshaggard: Where do we consume this for the Publish manifest stuff?

dagood: That's in https://github.com/dotnet/buildtools/pull/1852: `ManifestName` picks up `GitHubRepositoryName` as a default.

In CoreCLR and CoreFX, `src/publish.proj` passes `GitHubRepositoryName` into `PushToBlobFeed` directly.

eerhardt: Why can't we pass in `$(PB_RepoName)` like we do for `UpdatePublishedVersions`?

dagood: We can, I mentioned this in the description:

> CoreCLR and CoreFX already had `GitHubRepositoryName` defined, so I added it here, too. The initial reason for the property was to create GitHub raw URLs for SourceLink--if this turns out to be destabilizing in Core-Setup I can use `PB_RepoName` instead.

I guess there's the assumption that we'll want to use SourceLink in this repo at some point, and might as well do it the same way as the other repos.

buildpipeline/Core-Setup-Publish.json
```diff
@@ -262,7 +262,7 @@
...
+        "msbuildArguments": "/p:Configuration=$(BuildConfiguration) $(PB_CommonMSBuildArgs) $(PB_BuildOutputManifestArguments) /p:PublishType=$(PB_PublishType) /p:SignType=$(PB_SignType) /p:AzureAccountName=$(PB_AzureAccountName) /p:ContainerName=$(PB_ContainerName) /p:AzureAccessToken=$(PB_AzureAccessToken) /p:PublishBlobFeedUrl=$(PB_PublishBlobFeedUrl) /p:PublishBlobFeedKey=$(PB_PublishBlobFeedKey) /p:OfficialPublish=true /p:GitHubUser=$(PB_GitHubUser) /p:GitHubEmail=$(PB_GitHubEmail) /p:GitHubAuthToken=$(GITHUB_PASSWORD) /p:VersionsRepoOwner=$(PB_VersionsRepoOwner) /p:VersionsRepo=$(PB_VersionsRepo) /p:VersionsRepoPath=build-info/dotnet/$(PB_RepoName)/$(SourceBranch) /p:DotNetToolDir=$(DotNetToolDir) /flp:v=detailed;LogFile=$(PB_SourcesDirectory)\\publish-blob.log",
```
eerhardt: Would it make sense if these manifests were built during the normal build as well, not just during publish?

That way we can use it in other places - ex. source-build, or even in normal dev box scenarios.

dir.props
eerhardt: Ok - that makes sense - we should keep all 3 repos as consistent as possible/reasonable.

buildpipeline/Core-Setup-Publish.json
dagood: It would, but the way things are factored right now this would either require some decent sized refactoring or some scenario-specific work. So that non-BuildTools repos can create manifests easily, the manifest code is in the `Feed` package and built into the `PushToBlobFeed` task. That task and the `Feed` package in general depend on some non-source-build-friendly libs.

The refactoring way is to create a new independent MSBuild task package that is included by both main BuildTools and Feeds. Creating these is pretty painful with the current BuildTools structure.

Adding additional manifest tasks to the main BuildTools package would be the way to get the dev + source-build scenarios working quickly, but that would be specific to the non-orchestrated-build scenarios so I haven't done it.

Filed https://github.com/dotnet/buildtools/issues/1853

Issue #3428, opened 2017-11-17
Publishing step in official build need to be refactored


weshaggard:
As part of our orchestrated build work I created a publish-type.proj that forks the publishing into two pieces, see https://github.com/dotnet/core-setup/pull/3404. We need to finishing refactoring all the publish steps to be configurable and not duplicated into multiple publish workflows.

weshaggard:
individual legs seem to still publish to myget directly see https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1173516. Looks like https://github.com/dotnet/core-setup/blob/fa0062baae260c1591a7d71c45f7c241634b8bec/publish/publish.proj#L282.

weshaggard:
After some discussion with others we have decided to remove the deb publish steps so we should do that as part of the publish clean-up in this issue.

weshaggard:
As part of separating the steps we should consider doing the symbol publishing in a parallel definition to try and reduce the time we index.

mikem8361:
@weshaggard, can I go ahead and commit this PR as is? Or should I turn off the pdb conversion?

weshaggard:
>@weshaggard, can I go ahead and commit this PR as is? Or should I turn off the pdb conversion?

Move forward and commit your PR turning on the pdb conversion.

weshaggard:
bumping to post 2.0

weshaggard:
As part of this we also need to ensure that we are using a build id specific storage account for gathering the assets. Right now if we do a stable build it ends up overwriting older builds. cc @mmitche

dagood:
Closing, this has changed a lot (for the better) since the last activity, especially with the migration to Arcade.

PR #3574, opened 2018-01-07
Update symbol publish package to 1.0.0-alpha-00015.

mikem8361:
Enable Portable PDB to Window PDB conversion.

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -219,9 +219,9 @@
 
   <Target Name="SetupPublishSymbols" Condition="'$(SymbolServerPath)'!=''" >
     <PropertyGroup>
-      <SymbolExpirationInDays Condition="'$(SymbolExpirationInDays)'=='' and '$(SymbolExpirationDate)'==''">5</SymbolExpirationInDays>
+      <SymbolExpirationInDays Condition="'$(SymbolExpirationInDays)'=='' and '$(SymbolExpirationDate)'==''">1</SymbolExpirationInDays>
       <SymbolVerboseLogging>true</SymbolVerboseLogging>
-      <ConvertPortablePdbsToWindowsPdbs>false</ConvertPortablePdbsToWindowsPdbs>
+      <ConvertPortablePdbsToWindowsPdbs>true</ConvertPortablePdbsToWindowsPdbs>
     </PropertyGroup>
     <ItemGroup>
       <SymbolPackagesToPublish Include="$(DownloadDirectory)**\*.symbols.nupkg" />
--- a/dependencies.props
+++ b/dependencies.props
@@ -56,7 +56,7 @@
   <!-- Publish symbol build task package -->
   <PropertyGroup>
     <PublishSymbolsPackage>Microsoft.SymbolUploader.Build.Task</PublishSymbolsPackage>
-    <PublishSymbolsPackageVersion>1.0.0-alpha-00014</PublishSymbolsPackageVersion>
+    <PublishSymbolsPackageVersion>1.0.0-alpha-00015</PublishSymbolsPackageVersion>
   </PropertyGroup>
 
   <ItemGroup>


publish/publish.proj
```diff
@@ -219,9 +219,9 @@
 
   <Target Name="SetupPublishSymbols" Condition="'$(SymbolServerPath)'!=''" >
     <PropertyGroup>
-      <SymbolExpirationInDays Condition="'$(SymbolExpirationInDays)'=='' and '$(SymbolExpirationDate)'==''">5</SymbolExpirationInDays>
+      <SymbolExpirationInDays Condition="'$(SymbolExpirationInDays)'=='' and '$(SymbolExpirationDate)'==''">1</SymbolExpirationInDays>
       <SymbolVerboseLogging>true</SymbolVerboseLogging>
-      <ConvertPortablePdbsToWindowsPdbs>false</ConvertPortablePdbsToWindowsPdbs>
+      <ConvertPortablePdbsToWindowsPdbs>true</ConvertPortablePdbsToWindowsPdbs>
```
weshaggard: Do we know what this does to the time of official builds?

mikem8361: With some adjustments to the caching on the input streams, the last total job time was 44 minutes. The publish leg as 37 minutes, 21 minutes of that time was the symbol publishing and PDB conversion.  Normally the  job total time is about 30 minutes so I thought 44 minutes was reasonable.

weshaggard: Just so I understand the PDB conversion adds 10-15 mins to the overall publish leg? That still feels like a large amount relatively speaking. I think we may want to do that in parallel with the other publishing to get the overall time to 30 mins.

mikem8361: I don’t understand how we would make it parallel with the other publishing. Can more than one leg be run at the same time on different machines? Sounds like more of  a pipeline build/infrastructure change.

/cc: @dagood

dagood: Yeah, an example of this is publishing vs. symbol publishing in CoreFX:
https://github.com/dotnet/corefx/blob/master/buildpipeline/DotNet-Trusted-Publish.json
https://github.com/dotnet/corefx/blob/master/buildpipeline/DotNet-Trusted-Publish-Symbols.json

They each have a pipeline with the same dependencies so both definitions are queued at the same time:
https://github.com/dotnet/corefx/blob/baf4a44b949d006e4c48f7dd1520b4701f96db5b/buildpipeline/pipeline.json#L334-L378

mikem8361: I think doing all that for the core-setup repo is out of scope of the work I was planning to do. How will all of this be affected by the orchestrated build? Wes says there will be one publish leg for packages and hopefully one publish symbols leg for all everything that is part of that.  

/cc: @lt72

mikem8361: I still want to move forward with my core-setup testing which I hope can include the PDB conversion.  Can we live with the extra 15 minutes for now?

dagood: > How will all of this be affected by the orchestrated build? Wes says there will be one publish leg for packages and hopefully one publish symbols leg for all everything that is part of that.

This is true, but the orchestrated build doesn't replace isolated builds, which is what you're modifying here.

mikem8361:
I know that that it doesn’t replace isolated builds and each project/repo will still have packages and symbol packages to publish. The orchestrated build will change/reduce the packages that each individual repos need to publish or do I misunderstand that?

The coreclr and corefx repos also need to be converted to the new uploader build task and breaking apart the core-setup publishing is work that needs to be done. I’m not sure how much of all that I can sign up for.  Wasn’t the core-setup work already planned by Wes or Davis?

lt72:
@Wes @dagood: would it be possible for either of you or somebody more expert on our build to take over this last step?
We would like to get some mileage on the conversion code and above all switch over to the production endpoint for uploading the symbols soon, and if Mike has to learn yet another piece it could take longer than desirable...

weshaggard:
@mikem8361 @lt72 I've added a note to https://github.com/dotnet/core-setup/issues/3428 about potentially splitting this part out when we refactor the publishing steps in core-setup.

dagood:
> The orchestrated build will change/reduce the packages that each individual repos need to publish or do I misunderstand that?

My understanding is it won't affect it at all. Ideally the outputs are exactly the same except for the package dependencies.

> would it be possible for either of you or somebody more expert on our build to take over this last step?

Created https://github.com/dotnet/core-eng/issues/2369 for prioritization discussion and tracking if someone on core-eng can help.

Issue #3511, opened 2017-12-11
'std::out_of_range'   what():  basic_string::replace: __pos (which is 18446744073709551612) > this->size() (which is 8)


volak:
I had this segfault happen when running `dotnet myassembly.dll` today - and it was bizzare enough that I feel it should be an issue here but the fix is pretty easy so I'll document that too and you can close if you feel its not worth investigating.

Basically I was running the `aspnetcore:2.0` docker image with a web project of mine, and doing `docker run` resulted in the segfault:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::replace: __pos (which is 18446744073709551612) > this->size() (which is 8)
Aborted (core dumped)
```

Stack trace from gdb:

```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f83301d93fa in __GI_abort () at abort.c:89
#2  0x00007f8330af00ad in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007f8330aee066 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007f8330aee0b1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f8330aee2c9 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007f8330b16b85 in std::__throw_out_of_range_fmt(char const*, ...) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007f8330b30052 in std::string::replace(unsigned long, unsigned long, char const*, unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007f832ff75fed in append_probe_realpath(std::string const&, std::vector<std::string, std::allocator<std::string> >*, std::string const&) () from /usr/share/dotnet/host/fxr/2.0.3/libhostfxr.so
[...]

PR #3583, opened 2018-01-09
Fix issue with --additionalprobingpath parsing

steveharter:
The code to parse `|arch|/|tfm|` from the additional probing paths was using string::find_last_of("|arch|/|tfm|") which returns the last occurrence of ***any*** character (so the last occurrence of `|`, `a`, etc). This was not the intent as the whole string was to be compared, not any character. This can cause incorrect string substitutions and an access violation in some cases when a negative position was calculated.

Note there is a test to cover the positive result (GivenThatICareAboutPortableAppActivation.Muxer_Activation_With_Templated_AdditionalProbingPath_Succeeds) but there is no test for this negative scenario, and no new test added.

Note: I'm not sure why the original code wanted the *last* occurrence; I changed it to just the first occurrence.

Fixes https://github.com/dotnet/core-setup/issues/3511


steveharter:
@eerhardt the code made it this far because the string starts and ends with the pipe "|" which is not a normal character...

I do think it is worth porting to 2.0.0 due to the severity. I'll create the port bug

PR #3596, opened 2018-01-16
Improve the error message when hostpolicy can't be found

steveharter:
When there is no hostpolicy.dll in a standalone app, the existing error messages are confusing because they imply the runtimeconfig.json file must exist and have a framework section which is not true in standalone.

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -485,12 +485,12 @@ bool fx_muxer_t::resolve_hostpolicy_dir(
     {
         if (!pal::file_exists(get_app(fx_definitions).get_runtime_config().get_path()))
         {
-            trace::error(_X("Failed to run as a standalone app because there is no '%s' file. If this should be a portable app instead, add that file specifying the appropriate framework."),
+            trace::error(_X("Failed to run as a standalone app. If this should be a portable app, add the %s file specifying the appropriate framework."),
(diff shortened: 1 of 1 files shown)


src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -485,12 +485,12 @@ bool fx_muxer_t::resolve_hostpolicy_dir(
     {
         if (!pal::file_exists(get_app(fx_definitions).get_runtime_config().get_path()))
         {
-            trace::error(_X("Failed to run as a standalone app because there is no '%s' file. If this should be a portable app instead, add that file specifying the appropriate framework."),
+            trace::error(_X("Failed to run as a standalone app. If this should be a portable app, add the %s file specifying the appropriate framework."),
```
eerhardt: I'm wondering if instead of `standalone` and `portable`, we should be using the same terms as used in documentation:  [https://docs.microsoft.com/en-us/dotnet/core/deploying/](https://docs.microsoft.com/en-us/dotnet/core/deploying/)

steveharter: We can either keep it like this for this issue and create a new issue for the rename, or change it now.

There's at least 3 other locations in this same file that use the standalone\portable terminology when logging.

I don't really like "framework dependent" because the word "shared" is not present - a "self contained" app still depends upon the "framework"...

Issue #3458, opened 2017-11-27
Expose method to return the full path of a given dll name


jkotalik:
I need the ability to locate a native dll which will be in the runtime store before loading dotnet or the coreclr. The method could either return the full path of the specific dll or a property bag of all of the dlls locations.
```c++
PropertyBag GetDllLocations() {

}
or 
string GetDllLocation(string dllName) {

}
```

@steveharter and I discussed offline too. The method would need to be externally exposed on hostfxr.

steveharter:
See also main feature https://github.com/dotnet/core-setup/issues/3457

steveharter:
@jkotalik after thinking about this for a bit, would a synchronous callback function work instead? The callback would be invoked after all information is received and before the the coreclr initialzed and pass in information we agree on.

The advantage is that the extra loop through the entire hostfx->hostpolicy chain will be avoided.

e.g. something like this:
hostfxr_main_with_callback(const int argc, const pal::char_t* argv[], callback_fn* callback_fn)

jkotalik:
That sounds fine to me. What would the signature of the callback_fn be? The specific dll or the property bag?

steveharter:
> What would the signature of the callback_fn be? The specific dll or the property bag?

We'll need to work this out. I understand you don't want things to change if the coreclr interface changes. Currently the coreclr is initialized with "trusted platform assemblies" which is a simple semi-colon (windows) or colon (unix) delimited list.

jkotalik:
> Currently the coreclr is initialized with "trusted platform assemblies" which is a simple semi-colon (windows) or colon (unix) delimited list.
If you return the semicolon delimited string, we can parse that for the specific assembly we need. Seems to be the most stable solution.

jkotalik:
Actually, is it possible to return a vector<string> of the dll locations?

steveharter:
The "trusted platform assemblies" include all known dlls (from the deps files). So see for yourself what the proposed callback will provide, set `COREHOST_TRACE=1`, run your app, and then look for `TRUSTED_PLATFORM_ASSEMBLIES` - you should see a long list of assemblies, each with a full path.

steveharter:
@jkotalik actually the callback should only pass primitive types across the dll boundary - strings, ints, etc to avoid potential issues with having different memory layouts caused by different version\implementation of std:: classes, struct alignment compiler flags, etc. So not a vector.

steveharter:
[...]

PR #3620, opened 2018-01-23
Add native API to hostpolicy that returns the native probe dirs for ASP.NET

steveharter:
Adds a native API to hostfxr.dll that returns the native search paths for a given app. In order to obtain all valid directories (including servicing directories), it must call into the hostpolicy.dll for the current runtime and can't for example simply be implemented just in hostfxr.dll.

Fixes https://github.com/dotnet/core-setup/issues/3458

ASP.NET will call this to address scenarios where they need to pre-load a native dll. ASP.NET will cycle through these returned paths and attempt to load the given native dll manually.

Notes:
- Refactored the fx_muxer.cpp execute-related methods to separate exec, CLI and this new API mode.
- A new hostpolicy.dll entry point was added to address this API mode. It is designed to be used by future API calls in the future (if they just need to return a string).
- The name of the API \ 'command' is not passed to this new hostpolicy.dll entry point. Instead, the name is passed as part of the host_interface_t serialization to allow for future flexibility including adding additional arguments but while keeping the entry point the same.

cc @jkotalik

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -1146,10 +1146,18 @@ std::vector<host_option> fx_muxer_t::get_known_opts(bool exec_mode, host_mode_t 
     return known_opts;
 }
 
-int fx_muxer_t::parse_args_and_execute(
+int fx_muxer_t::parse_args(
     const pal::string_t& own_dir,
     const pal::string_t& own_dll,
-    int argoff, int argc, const pal::char_t* argv[], bool exec_mode, host_mode_t mode, bool* is_an_app)
+    int argoff,
+    int argc,
+    const pal::char_t* argv[],
+    bool exec_mode,
+    host_mode_t mode,
+    bool* is_an_app,
+    int* new_argoff,
+    pal::string_t& app_candidate,
+    opt_map_t& opts)
 {
     *is_an_app = true;
 
@@ -1157,7 +1165,6 @@ int fx_muxer_t::parse_args_and_execute(
 
     // Parse the known arguments if any.
     int num_parsed = 0;
-    std::unordered_map<pal::string_t, std::vector<pal::string_t>> opts;
     if (!parse_known_args(argc - argoff, &argv[argoff], known_opts, &opts, &num_parsed))
     {
         trace::error(_X("Failed to parse supported options or their values:"));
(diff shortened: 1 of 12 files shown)


jkotalik:
cc/ @muratg @davidfowl

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1210,17 +1214,7 @@ int fx_muxer_t::parse_args_and_execute(
...
-        vec_argv.reserve(argc - cur_i + 1); // +1 for dotnet
-        vec_argv.push_back(argv[0]);
-        vec_argv.insert(vec_argv.end(), argv + cur_i, argv + argc);
-        new_argv = vec_argv.data();
-        new_argc = vec_argv.size();
-    }
-
-    // Transform dotnet [exec] [--additionalprobingpath path] [--depsfile file] [dll] [args] -> dotnet [dll] [args]
-    return read_config_and_execute(own_dir, app_candidate, opts, new_argc, new_argv, mode);
+    return 1;
```
eerhardt: I thought the convention was to return `0` on success.

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -178,3 +257,57 @@ SHARED_API int32_t hostfxr_resolve_sdk(
...
+//    buffer_size
+//      The size of the buffer argument in pal::char_t units.
+//
+//    required_buffer_size
+//      If the apis fails because of insufficient buffer size,
+//      required_buffer_size is set to the minimium buffer size
+//      necessary to contain the result.
+//
+// Return value:
+//   1 on success, otherwise failure
```
eerhardt: I think this should be 0 on success.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -175,7 +187,27 @@ int run(const arguments_t& args)
...
+            trace::error(_X("get-native-search-directories failed to find NATIVE_DLL_SEARCH_DIRECTORIES property"));
+            exit_code = HostApiFailed;
+        }
+        else
+        {
+            assert(out_host_command_result != nullptr);
+            pal::clr_palstring(property_values[1], out_host_command_result);
+        }
+
+        return exit_code;
```
eerhardt: Should return `0` for success.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Yes. Thanks

src/corehost/cli/fxr/hostfxr.cpp
steveharter: Done

src/corehost/cli/hostpolicy.cpp
steveharter: Done

src/corehost/cli/hostpolicy.cpp
```diff
@@ -273,16 +308,22 @@ int run(const arguments_t& args)
...
 
     return exit_code;
 }
 
 SHARED_API int corehost_load(host_interface_t* init)
 {
     trace::setup();
-    
+
+    // Re-initialize global state in case of re-entry
```
eerhardt: Do we have a re-entry test case?

steveharter: Yes, the unit test that was added calls back in a second time because its buffer was too small the first time

PR #3621, opened 2018-01-23
Reference Microsoft.NETCore.Targets

### LINKED ISSUE #2716: 1.0 Cryptography native shim is being published in a 2.0 app
## Steps to reproduce

```XML
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
  </ItemGroup>
</Project>
```

dotnet restore -r osx.10.12-x64
dotnet publish -r osx.10.12-x64

*Alternatively* if you call `dotnet store -r osx.10.12-x64` on a manifest that references Newtonsoft.Json/9.0.1, you get the same behavior in the outputted runtime store.

## Expected  behavior
No "1.0" native libraries should be published in the publish folder (or in the runtime store).  This is a "2.0" app, so all the framework assemblies should be our "2.0" assemblies.

## Actual behavior
There is a 1.0 crypto native shim library being published:

```
"runtime.osx.10.10-x64.runtime.native.System.Security.Cryptography/1.0.1": {
        "type": "package",
        "native": {
          "runtimes/osx.10.10-x64/native/System.Security.Cryptography.Native.dylib": {}
        }
      }, 
```

This library is not required, and doesn't work with our 2.0 framework.

Technically, nothing should try and load the assembly, so I'm not sure what, if any, downstream effects this will have.  But the assembly should not be getting published.

## Environment data

```
> dotnet --info
.NET Command Line Tools (2.0.0-preview2-006492)

Product Information:
 Version:            2.0.0-preview2-006492
 Commit SHA-1 hash:  7ab3d5b8d8

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   F:\dotnet\sdk\2.0.0-preview2-006492\

  Version  : 2.0.0-preview2-25407-01
  Build    : 40c565230930ead58a50719c0ec799df77bddee9
```

## Notes

The reason this is happening is because `Newtonsoft.Json/9.0.1` doesn't reference `NETStandard.Library`.  Instead, it references the individual System.XXX packages it requires.  Thus a bunch of our .NET Core 1.0 packages are being referenced.  Normally, this is OK because we've implemented conflict resolution to resolve when 1.0 and 2.0 assemblies are named the same.  However, between 1.0 and 1.1 we renamed/refactored this native Cryptography assembly.  So the 2.0 framework doesn't have this assembly any more, so there is no conflict to resolve.  That means there is nothing stopping this 1.0 assembly from being published.

This doesn't happen when publishing a self-contained app for Windows - only on Linux and macOS - since we don't have native shims on Windows.

/cc @bartonjs @ericstj

ericstj:
We want to make sure that Microsoft.NETCore.App has a reference to Microsoft.NETCore.Targets
so that the latest version is used and none of the old pre-2.0 runtime-specific packages get
brought into the package graph.

We were trying to do this before but it was failing because we weren't directly referencing the package.

In the future we could change this target to use the SDK assets file raising task then read the version
that was actually restored instead of forcing us to keep a direct reference to it.

Fixes  #2716

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -35,7 +35,7 @@
       <Version>$(MicrosoftNETCorePlatformsPackageVersion)</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.Targets">
-      <Version>2.0.0</Version>
+      <Version>$(MicrosoftNETCoreTargetsPackageVersion)</Version>
     </PackageReference>
     <PackageReference Include="NETStandard.Library">
       <Version>$(NETStandardLibraryPackageVersion)</Version>
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -34,6 +34,9 @@
     <PackageReference Include="Microsoft.NETCore.Platforms">
       <Version>$(MicrosoftNETCorePlatformsPackageVersion)</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.NETCore.Targets">
+      <Version>2.0.0</Version>
+    </PackageReference>
     <PackageReference Include="NETStandard.Library">
       <Version>$(NETStandardLibraryPackageVersion)</Version>
     </PackageReference>
--- a/dependencies.props
+++ b/dependencies.props
@@ -18,6 +18,7 @@
 
   <PropertyGroup>
     <MicrosoftNETCorePlatformsPackageVersion>2.1.0-preview2-26127-04</MicrosoftNETCorePlatformsPackageVersion>
+    <MicrosoftNETCoreTargetsPackageVersion>2.0.0</MicrosoftNETCoreTargetsPackageVersion>
     <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.5.0-preview2-26127-04</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
     <MicrosoftPrivateCoreFxUAPPackageVersion>4.6.0-preview2-26127-04</MicrosoftPrivateCoreFxUAPPackageVersion>
     <MicrosoftNETCoreRuntimeCoreCLRPackageVersion>2.1.0-preview2-26129-01</MicrosoftNETCoreRuntimeCoreCLRPackageVersion>


src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -34,6 +34,9 @@
     <PackageReference Include="Microsoft.NETCore.Platforms">
       <Version>$(MicrosoftNETCorePlatformsPackageVersion)</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.NETCore.Targets">
+      <Version>2.0.0</Version>
```
eerhardt: I think you should define a property for this to follow the existing pattern.  That way we can change it with dependency flow.  And also, we can have maestro update this version just like the other package references above.

ericstj: We don't intend to change it.  We produced one last version that was 2.0.0 with an empty runtime.json.

eerhardt: Are we still building it in the master branch?

I still think we should create the property to follow the existing pattern.  Even if the property is only ever hard-coded to `2.0.0`

ericstj: Yes, it's building in master but doesn't need to.  There are no changes since 2.0.  I brought it back from deletion to build it just once for 2.0.0.  We can clean it up in corefx-master now.

eerhardt:
@weshaggard - how is this going to work in source-build?  especially if we stop building `Microsoft.NETCore.Targets` from corefx.

ericstj:
FWIW this 2.0.0 dependency *already exists* on the private NETCoreapp package that corefx produces.  Also, this particular usage (emitting the dependency) doesn't require any resolution of the package at build time: it's just downstream folks who try to restore NETCore.App who would need the package.

eerhardt:
A mantra with the source-build effort is to not ship something that you didn't build from source.  The high-level reasoning is that distro maintainers should be able to change anything and everything in our product for servicing reasons.  See https://github.com/dotnet/source-build/issues/187 for a lot of discussion on this topic.

So taking a dependency on a previously shipped package sort of falls into this crack.  Of course we have this same issue elsewhere, so I wouldn't block this.  But it would be good to know how this affects source-build.

ericstj:
We could always just leave it in corefx and build it at the stable version for source builds.  Its ugly but it would work.  Or I could continue to version it and build it even though it doesn't change.

ericstj:
@weshaggard what do you think about @eerhardt's question around versioning of the targets package?

weshaggard:
>@weshaggard what do you think about @eerhardt's question around versioning of the targets package?

I would prefer we not build it live if it isn't changing any longer. We have a number of pre-built packages that we currently depend on. I think the worst case scenario is we build a dummy package but I don't think there is any way we get out of the business of having a set of pre-built packages, at least not anytime soon.

eerhardt:
Agreed.  It's a larger issue than just NETCore.Targets.  Just something we need to be aware of for `source-build`.

PR #3645, opened 2018-01-28
Fix additional-deps feature with native and resource dlls

steveharter:
Fix issue with --additional-deps not processing native and resource dlls.

It appears this was a copy-paste bug from a similar area of code that processes managed assemblies.

Fixes https://github.com/dotnet/core-setup/issues/3634

After this is the main branch, this will be ported to the 2.1.0 branch and a new issue will be created for a 2.0.x port.


steveharter:
@devinvisible the issue was not ported to 2.0.x because it the port was not approved; it is only fixed in the 2.1 branch.

PR #3655, opened 2018-01-30
[release/2.1] Add ProductVersion to build output manifest

dagood:
Move to the release/2.1 BuildTools branch to get around the current BuildTools upgrade blocker, then pass `ProductVersion` into the build output manifest.

The copy to latest post-build step will pick up on this and be able to finalize core-setup product builds without deriving anything from the blob artifact paths. https://github.com/dotnet/core-eng/issues/2407

This should be ported to master once BuildTools upgrades are unblocked there.


dependencies.props
```diff
@@ -77,7 +77,7 @@
       <CurrentRef>$(WCFCurrentRef)</CurrentRef>
     </RemoteDependencyBuildInfo>
     <RemoteDependencyBuildInfo Include="BuildTools">
-      <BuildInfoPath>$(BaseDotNetBuildInfo)buildtools/$(DependencyBranch)</BuildInfoPath>
+      <BuildInfoPath>$(BaseDotNetBuildInfo)buildtools/2.1</BuildInfoPath>
```
eerhardt: `$(DependencyBranch)` should be changed above right?  we don't want to take `master` corefx and coreclr builds into core-setup/release/2.1.

dagood: True, I should try to set this up so it makes sense. Right now auto-updates aren't enabled, but I'd like to set them up so we keep updating BuildTools at least. (Only buildtools because 2.1 auto-upgrade isn't doable yet: https://github.com/dotnet/core-eng/issues/2435.)

Unfortunately BuildTools doesn't have the same structure on dotnet/versions as the rest right now, thanks to how VSTS calculates `SourceBranchName`. Filed https://github.com/dotnet/buildtools/issues/1891. Probably a quick fix (calculating in inline powershell), so I'll take a look at this now while nobody's depending on the 2.1 build-info.

PR #3656, opened 2018-01-31
Building non-Windows builds using the CLI's MSBuild

joperezr:
cc: @eerhardt @weshaggard 

Updating buildtools and switching non-Windows builds so that they instead get built using the MSBuild that lives inside the CLI instead of the one we manually restore from buildtools.


dir.props
```diff
@@ -436,8 +436,8 @@
     <PortableNuGetMode>true</PortableNuGetMode>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <!-- Never use the NuGet fallback folder that comes with the SDK we use to build.
+  <PropertyGroup Condition="'$(MSBuildProjectName)' != 'crossgen'">
```
eerhardt: ?? I'm not sure this is the right thing to do.  We should NEVER be using the fallback folder.

joperezr: The crossgen build is part of buildtools. When restoring buildtools we will build crossgen against the same runtime that will be used to run msbuild in order to crossgen the whole repo. For that to work, we need to be able to restore the packages for this crossgen project comming from buildtools, and the dependencies that we need for this new version of the CLI are not in myget, so that is why we just use the fallback folder for this project only, since it will contain the missing packages.

init-tools.sh
```diff
@@ -147,11 +147,11 @@ if [ ! -e $__BUILD_TOOLS_PATH ]; then
 fi
 
 echo "Initializing BuildTools..."
-echo "Running: $__BUILD_TOOLS_PATH/init-tools.sh $__scriptpath $__DOTNET_CMD $__TOOLRUNTIME_DIR" >> $__init_tools_log
+echo "Running: $__BUILD_TOOLS_PATH/init-tools.sh $__scriptpath $__DOTNET_CMD $__TOOLRUNTIME_DIR $__PACKAGES_DIR" >> $__init_tools_log
 
 # Executables restored with .NET Core 2.0 do not have executable permission flags. https://github.com/NuGet/Home/issues/4424
 chmod +x $__BUILD_TOOLS_PATH/init-tools.sh
-$__BUILD_TOOLS_PATH/init-tools.sh $__scriptpath $__DOTNET_CMD $__TOOLRUNTIME_DIR >> $__init_tools_log
+$__BUILD_TOOLS_PATH/init-tools.sh $__scriptpath $__DOTNET_CMD $__TOOLRUNTIME_DIR $__PACKAGES_DIR >> $__init_tools_log
```
eerhardt: >__PACKAGES_DIR [](start = 82, length = 14)

Should we be keeping our Windows and non-Windows init-tools in sync?

joperezr: For this particular change it doesn't matter much, since Windows is still using full msbuild, but this is definitelly going to be addressed soon once we move to using the CLI to build on Windows as well. However if we do that change now, it might cause some side effects on Windows builds, so I would rather keep this focused to non-Windows for now, given that as I said, the next workitem is to do the same for Windows.

dir.props
eerhardt: Can't we just set this property in that project then?  Or even better - crossgen.csproj shouldn't be importing the repo's dir.props.  The dir.props is only for projects that are actually in this repo.

---
In reply to: [165137167](https://github.com/dotnet/core-setup/pull/3656#discussion_r165137167) [](ancestors = 165137167)

joperezr: crossgen is an SDK project, and given that there is a Directory.Build.props on core-setup, then dir.props get's loaded automatically, and there is no easy way of disabling this other than making it not be a SDK project which I believe is not what we want. I will explore that suggestion of trying to instead set the property manually, but given that dir.props gets imported automatically, it might be imported late.

eerhardt: `dir.props` will get imported first and set the property to `true`.  Then your project will set the property and overwrite it to `false`.  As the project, you always have full control to set things exactly as you want.

weshaggard: I agree we should not special case crossgen project here but instead set the property in that project.

dir.props
```diff
@@ -452,7 +452,7 @@
     <MSBuildPassThroughPropertyList Condition="'$(OutputRid)' != ''">$(MSBuildPassThroughPropertyList) /p:OutputRid=$(OutputRid)</MSBuildPassThroughPropertyList>
   </PropertyGroup>
 
-  <!-- Use Roslyn Compilers to build, if we are not using the .NET SDK to build -->
-  <Import Project="$(RoslynPropsFile)" Condition="'$(UsingNETSdkCompiler)' != 'true' and '$(OsEnvironment)'!='Windows_NT' and Exists('$(RoslynPropsFile)') and '$(UseRoslynCompilers)'!='false'" />
-  <Import Project="$(RoslynPropsFile)" Condition="'$(UsingNETSdkCompiler)' != 'true' and '$(OsEnvironment)'=='Windows_NT' and Exists('$(RoslynPropsFile)')" />
+  <!-- Use Roslyn Compilers to build -->
+  <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'!='true' and Exists('$(RoslynPropsFile)') and '$(UseRoslynCompilers)'!='false' and '$(RoslynIncompatibleMsbuildVersion)' != 'true'" />
+  <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'=='true' and Exists('$(RoslynPropsFile)') and '$(RoslynIncompatibleMsbuildVersion)' != 'true'" />
```
weshaggard: Are we planning to do away with this import now we are moving to the CLI msbuild? I wouldn't expect we need this anylonger.

dir.props
joperezr: I tried that locally and it does seem to work so I'll make the change in buildtools and update this PR.

dir.props
joperezr: we will eventually throw it once we also build using the CLI on Windows, but we still need it for now.

weshaggard: We still need it even for the RunningOnUnix == true condition?

joperezr: Yes, because the RoslynPropsFile is set to the props file inside the microsoft.netcore.compilers package. Roslyn.Common.props will set the RoslynPropsFile accordingly depending on if you are running on unix or on windows.

PR #3660, opened 2018-02-02
Add sample snaps for .NET Core based applications

rakeshsinghranchi:


eerhardt:
@rakeshsinghranchi - Sorry for the late review, but I was just made aware of this work.

I'm not 100% certain this is the right place for these files.  They feel like they would be better served being in a "Doc" repo, or maybe even `dotnet/core`.

@weshaggard - thoughts?

eerhardt:
I think my main concern is that these files are in `src/pkg` right next to our actual installer logic.  If these are "samples" they don't belong under `src/pkg`.  Only actual product code should be in there.  Someone could be confused that these are actually being used to build our product.

rakeshsinghranchi:
These are supposed to be in dev/snaps branch. I did revert the initial check-in. Seems @leecow PR brought these back : https://github.com/dotnet/core-setup/commit/0539dcba6f135172c0bd5eec4059c682072e98a1 .

eerhardt:
My main concern is putting things in a product repo that aren't part of the product. It adds to the confusion of anyone looking through the repo.

Can we put these things in a feature branch, or a fork, until they are ready to be part of the product?

PR #3665, opened 2018-02-02
Update CoreClr/CoreFx and mark packages as stable for 1.1.7

wtgodbe:
@safern PTAL
@weshaggard I think Santi is OOF today, would you mind taking a look as well?


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -76,7 +76,7 @@
       <Version>1.0.2</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.Runtime.CoreCLR">
-      <Version>1.0.12-servicing-26117-01</Version>
```
safern: You need to update DotNetHostPolicy here as well to the 1.0.0 equivalent. I think it’s 1.0.11?

wtgodbe: Wouldn't it be 1.0.10? https://github.com/dotnet/core-setup/pull/3664/files#diff-c0a524ec6ca51973079c79a8e15af029R324

safern: Yes. 1.0.10

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -70,13 +70,13 @@
       <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
-      <Version>1.0.8</Version>
+      <Version>1.0.10</Version>
```
wtgodbe: @safern is there a reason this wasn't updated to 1.0.9 last time?

safern: I might have missed it.

PR #3679, opened 2018-02-08
Fixing Redhat 7 official builds by copying the tasks dependencies to …

joperezr:
cc: @weshaggard @ericstj @eerhardt 

Fixing Redhat 7 bulids by copying the local tasks dependencies to the local dir. Now that MSBuild is no longer on the Tools directory, the Tools directory was not part of the probing path any longer for loading assemblies. That means that for this other tasks we now require to copy the right dependencies to the local tools dir.

### DIFF
--- a/dir.targets
+++ b/dir.targets
@@ -22,6 +22,7 @@
     </PropertyGroup>
 
       <ItemGroup>
+        <CustomTaskDependencies Include="$(BuildToolsTaskDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
         <CustomTaskDependencies Include="$(BuildToolsTaskDir)Newtonsoft.Json.dll" />
         <CustomTaskDependencies Include="$(BuildToolsTaskDir)Microsoft.DotNet.PlatformAbstractions.dll" />
         <CustomTaskDependencies Include="$(PackagesDir)microsoft.extensions.dependencymodel\1.1.1\lib\$(DependencyModelTFM)\Microsoft.Extensions.DependencyModel.dll" />
--- a/src/pkg/packaging/deb/package.targets
+++ b/src/pkg/packaging/deb/package.targets
@@ -103,12 +103,6 @@
       </SharedHostTokenValue>
     </ItemGroup>
 
-    <!-- Copying dependencies of ReplaceFileContents task to local folder -->
-    <ItemGroup>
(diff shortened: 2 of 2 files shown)


ericstj:
I'm a bit surprised that we need an explicit step to do this.  Shouldn't it be copied by default since it is a reference?

joperezr:
It doesn't, not sure why. I have a local repro and the only files that ended up on the tools/local folder are:
```
 24 -rwxrw-rw- 1 root root  22648 Jul 20  2017 Microsoft.DotNet.PlatformAbstractions.dll
 56 -rwxrw-rw- 1 root root  54264 Feb 10  2017 Microsoft.Extensions.DependencyModel.dll
628 -rwxrw-rw- 1 root root 639488 Jun 18  2017 Newtonsoft.Json.dll
 48 -rw-r--r-- 1 root root  47616 Feb  8 20:05 core-setup.tasks.dll
 16 -rw-r--r-- 1 root root  14596 Feb  8 20:05 core-setup.tasks.pdb
```

it used to work just fine because MSBuild was on the Tools directory so it was part of the probing path.

ericstj:
Look at the log of build of core-setup.tasks.csproj on the redhat machine and see what it does with the Microsoft.DotNet.Build.CloudTestTasks.dll Reference.

joperezr:
Found out why it wasn't being copied to the tools/local directory, it wasn't includded on the item that was being copied as dependencies, but I fixed that on the last commit.

ericstj:
Why do we have to declare this twice?  The project itself tries to set its [OutputPath to LocalToolRuntimePath](https://github.com/dotnet/core-setup/blob/d5e9580461cf73577f4882ce8755a038c907abfb/tools-local/tasks/core-setup.tasks.csproj#L12) which should be the same as [LocalBuildToolsTaskDir](https://github.com/dotnet/core-setup/blob/be8d8e3486b2bf598ed69d39b1629a24caaba45e/RepoDirectories.props#L26) used in this targets file.  I bet that OutputPath setting isn't "sticking"... I'd really prefer if we could remove this redundancy.  I guarantee the next person to add a dependency to this project is going to forget to update this list.  @chcosta might know why we have both of these, looks like he did it initially.

I'm ok with merging this as-is and we have a follow up to remove the redundancy.  First priority is to get good official builds going again.

joperezr:
Sounds good, I’ll merge this now to fix official builds and log an issue to fix the redundancy.

joperezr:
Thanks for ther review Eric

PR #3685, opened 2018-02-11
Update File description for runtime binaries ( host,hostfxr and hostpolicy ) 

rakeshsinghranchi:
Fixes https://github.com/dotnet/core-setup/issues/2817

### DIFF
--- a/src/corehost/build.proj
+++ b/src/corehost/build.proj
@@ -55,10 +55,19 @@
 
     <!-- Generate Version files -->
     <ItemGroup>
-      <HostFiles Include="dotnet;hostfxr;hostpolicy" />
+      <HostFiles Include="dotnet">
+       <AssemblyName>.NET Core Runtime - $(HostVersion)</AssemblyName>
+      </HostFiles>
+      <HostFiles Include="hostfxr">
+        <AssemblyName>.NET Core Host Resolver - $(HostResolverVersion)</AssemblyName>
+      </HostFiles>
+      <HostFiles Include="hostpolicy">
+        <AssemblyName>.NET Core Host Policy - $(HostPolicyVersion)</AssemblyName>
+      </HostFiles>
     </ItemGroup>
+    
     <MSBuild Projects="$(MSBuildProjectFullPath)"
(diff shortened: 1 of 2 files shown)


rakeshsinghranchi:
@steveharter and @leecow   - PTAL . Thanks.

leecow:
Can you give an example of what this change will look like to the user?

rakeshsinghranchi:
For Daily build: 
.NET Core Runtime - 2.1.0-preview2-26211-0
.NET Core Host Resolver - 2.1.0-preview2-26211-0
.NET Core Runtime - 2.1.0-preview2-26211-0

For RTM : 
.NET Core Runtime - 2.1.0
.NET Core Host Resolver - 2.1.0
.NET Core Runtime - 2.1.0

rakeshsinghranchi:
dotnet.exe
filedescription_dotnet

hostfxr.dll
filedescription_host

hostpolicy.dll
filedescription_hostpolicy

rakeshsinghranchi:
cc @steveharter

danmosemsft:
Is that what we call dotnet.exe, ".NET Core Runtime" ? That isn't our concept of a runtime, it's more like an enhanced host?

Petermarcu:
I agree with Dan, dotnet.exe may be better as ".NET Core Host", which loads the .NET Core Version Selector" and the .NET Core Host Policy", which selects and loads the runtime. Doesn't seem right to call the host, the runtime because it could load any number of runtimes, sdk's and versions of them. The versions will not match the runtime in many cases and will confuse.

rakeshsinghranchi:
@leecow / @steveharter - can you look at the updated change and approve again. Thanks.

src/corehost/build.proj
```diff
@@ -55,10 +55,19 @@
 
     <!-- Generate Version files -->
     <ItemGroup>
-      <HostFiles Include="dotnet;hostfxr;hostpolicy" />
+      <HostFiles Include="dotnet">
+       <AssemblyName>.NET Core Host</AssemblyName>
```
weshaggard: Is assembly name the correct name for this metadata? It looks more like a description. The name might cause some confusion.

leecow:
I agree that the description for dotnet.exe should be ".NET Host". The others look good.

rakeshsinghranchi:
>I agree that the description for dotnet.exe should be ".NET Host". The others look good.

@leecow , Just to confirm File description is ".NET **Core** Host"

leecow:
Yes, sorry - ".NET Core Host"

PR #3698, opened 2018-02-16
[release/2.1] Configure auto-update from product builds

dagood:
I enabled standard release/2.0.0 here because it's also enabled in corefx: https://github.com/dotnet/corefx/pull/27180. Not sure if maybe disabling (for both repos?) might make more sense.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -60,7 +60,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!--
     <RemoteDependencyBuildInfo Include="CoreFx">
       <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/$(DependencyBranch)</BuildInfoPath>
       <CurrentRef>$(CoreFxCurrentRef)</CurrentRef>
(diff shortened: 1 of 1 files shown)


dependencies.props
```diff
@@ -113,6 +112,7 @@
       <ElementName>NETStandardLibraryPackageVersion</ElementName>
       <PackageId>NETStandard.Library</PackageId>
     </XmlUpdateStep>
+    <!--
```
weshaggard: As an aside we can likely remove that WCF dependencies now that UWP has been removed in master. https://github.com/dotnet/core-setup/commit/3d9abcca488b38f87ee59e776f98915c0c12f729.

dagood: Cool, I'll go ahead and clean this up here.

dagood:
It looks like we don't have Tizen builds being produced for release/2.1--CI is missing the runtime packages. If I remember correctly, Tizen builds are triggered by the dotnet/versions build-info, so it makes sense that the release/2.1 trigger isn't set up yet (if the intention is to set it up for 2.1). We just recently [started having core-setup build-info files for release/2.1](https://github.com/dotnet/versions/tree/master/build-info/dotnet/core-setup/release/2.1).

@weshaggard Is this mergeable before that's set up, or should I remove the auto-update commit for now?

weshaggard:
@hqueue @hseok-oh do you guys have plans to setup release/2.1 builds?

@dagood please move forward with merging the PR when ready. Don't block on the Tizen leg.

Issue #3546, opened 2017-12-19
Support "framework wins over app" or "lower layer wins" during major\minor roll-forward


steveharter:
This is a core-setup feature to address https://github.com/dotnet/corefx/issues/24090 in a generalized way by supporting a lower layer (e.g. NetCoreApp) to win \ be selected when there is already a copy of that assembly in a higher layer (e.g. the app's folder). In 2.0, the higher layer always wins.

The issue [24090](https://github.com/dotnet/corefx/issues/24090) was addressed for 2.1 preview 1 by duplicating and renaming the assembly. However, it does not solve the issue generally, so this core-setup issue was created to track that. The issue can occur again, including with frameworks other than NetCoreApp such as upcoming Asp.Net framework support which will use the recent multi-layer framework support.

For example, with the new multi-layer framework support the following layers could exist:
- App (the highest layer)
- ASP.NET
- NetCoreApp (the lowest layer)

The three scenarios\outcomes determining which layer should win when a given assembly exists in more than one layer:
**i.** Lower layer should win. This is the scenario as described in [24090](https://github.com/dotnet/corefx/issues/24090) repro steps where an assembly was added to framework package is newer than the app's
**ii.** Higher layer should win. For example, the app wants to bring in a newer OOB assembly like System.Collections.Immutable. This could change into scenario (i) when an app's copy of an OOB assembly becomes older than the current framework's assembly during a roll-forward scenario.
**iii.** Unknown who should win. For example, the app has locally build framework assemblies (not package references).

Currently (in 2.0) the only rule for when a given assembly exists in more than one layer:
**1.** The highest layer wins

The proposed supplementary rule for 2.1:
**2.** If a [major] or [minor] roll-forward occurs for a given framework, that framework wins over any other higher layer(s)

Note this rule does not apply to just patch roll-forward as patches are not supposed to add\remove assemblies and because the higher layer likely has a higher versioned assembly (by major\minor).

The rationale for rule 2 is because the given assembly will likely be newer than the layers above, since the layers above likely shipped at an earlier time and its copy of the assembly is likely older. In addition, it is likely the given assembly is new to that lower layer, and the higher layers could not yet be updated to remove their copy of that assembly.

Because these two rules do not work in all cases, the following can be used as workarounds by the person trying to run the app:
a) Install the targeted framework. No [major][minor] roll-forwards occur unless the targeted framework does not exist.
[...]

PR #3704, opened 2018-02-17
Compare assembly and file versions on minor roll forward

steveharter:
This makes DependencyModel changes so that assemblyVersion and fileVersion can be written to the deps.json file for each runtime asset and adds conflict resolution to hostpolicy.dll that uses that metadata during roll-forward cases.

Fixes https://github.com/dotnet/core-setup/issues/3546

The actual metadata is not added to the .deps.json files yet. That will be done once this PR is in and https://github.com/dotnet/sdk/issues/1847 is resolved.

Changes to conflict resolution are documented at https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/assembly-conflict-resolution.md.

One conflict test has been commented out until the SDK changes are made (the functionality was verified manually) . Additional tests may be added at that time as well.

### DIFF
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -29,16 +29,16 @@ pal::string_t deps_json_t::get_optional_property(
     const json_object& properties,
     const pal::string_t& key) const
 {
-    pal::string_t path;
+    pal::string_t value;
 
     const auto& iter = properties.find(key);
 
     if (iter != properties.end())
     {
-        path = iter->second.as_string();
+        value = iter->second.as_string();
     }
 
-    return path;
+    return value;
 }
 
 pal::string_t deps_json_t::get_optional_path(
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -25,7 +25,7 @@ const deps_entry_t& deps_json_t::try_ni(const deps_entry_t& entry) const
     return entry;
 }
 
-pal::string_t deps_json_t::get_optional_path(
+pal::string_t deps_json_t::get_optional_property(
     const json_object& properties,
     const pal::string_t& key) const
 {
@@ -36,11 +36,20 @@ pal::string_t deps_json_t::get_optional_path(
     if (iter != properties.end())
     {
         path = iter->second.as_string();
+    }
 
-        if (_X('/') != DIR_SEPARATOR)
-        {
-            replace_char(&path, _X('/'), DIR_SEPARATOR);
-        }
+    return path;
+}
+
+pal::string_t deps_json_t::get_optional_path(
+    const json_object& properties,
+    const pal::string_t& key) const
+{
+    pal::string_t path = get_optional_property(properties, key);
+
+    if (path.length() > 0 && _X('/') != DIR_SEPARATOR)
+    {
+        replace_char(&path, _X('/'), DIR_SEPARATOR);
     }
 
     return path;
@@ -206,7 +215,6 @@ bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, c
 
 bool deps_json_t::process_runtime_targets(const json_value& json, const pal::string_t& target_name, const rid_fallback_graph_t& rid_fallback_graph, rid_specific_assets_t* p_assets)
 {
-    version_t empty;
     rid_specific_assets_t& assets = *p_assets;
     for (const auto& package : json.at(_X("targets")).at(target_name).as_object())
     {
@@ -226,7 +234,32 @@ bool deps_json_t::process_runtime_targets(const json_value& json, const pal::str
                 if (pal::strcasecmp(type.c_str(), deps_entry_t::s_known_asset_types[i]) == 0)
                 {
                     const auto& rid = file.second.at(_X("rid")).as_string();
-                    deps_asset_t asset(file.first, empty, empty);
+
+                    version_t assembly_version, file_version;
+                    const auto& properties = file.second.as_object();
+
+                    pal::string_t assembly_version_str = get_optional_property(properties, _X("assemblyVersion"));
+                    if (assembly_version_str.length() > 0)
+                    {
+                        version_t::parse(assembly_version_str, &assembly_version);
+                    }
+
+                    pal::string_t file_version_str = get_optional_property(properties, _X("fileVersion"));
+                    if (file_version_str.length() > 0)
+                    {
+                        version_t::parse(file_version_str, &file_version);
+                    }
+
+                    deps_asset_t asset(file.first, assembly_version, file_version);
+
+                    trace::info(_X("Adding runtimeTargets %s asset %s rid=%s assemblyVersion=%s fileVersion=%s from %s"),
+                        deps_entry_t::s_known_asset_types[i],
+                        asset.relative_path.c_str(),
+                        rid.c_str(),
+                        asset.assembly_version.as_str().c_str(),
+                        asset.file_version.as_str().c_str(),
+                        package.first.c_str());
+
                     assets.libs[package.first].rid_assets[rid][i].push_back(asset);
                 }
             }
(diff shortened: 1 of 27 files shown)


src/managed/Microsoft.Extensions.DependencyModel/CollectionExtensions.cs
```diff
@@ -36,5 +36,22 @@ private static IEnumerable<string> GetAssets(IEnumerable<RuntimeAssetGroup> grou
                 .Where(a => string.Equals(a.Runtime, runtime, StringComparison.Ordinal))
                 .SelectMany(a => a.AssetPaths);
         }
+
+        public static IEnumerable<RuntimeFile> GetDefaultRuntimeFiles(this IEnumerable<RuntimeAssetGroup> self) => GetRuntimeFiles(self, string.Empty);
```
steveharter: Review note: I'm not happy with this naming. These are mirrors of existing methods like `GetDefaultRuntimeAssets` but return RuntimeFile class instead of a string

src/test/Microsoft.Extensions.DependencyModel.Tests/JsonAssetions.cs
```diff
@@ -31,7 +31,7 @@ public JsonAssetions(JToken token)
             var token = Subject[expected];
             Execute.Assertion
                 .ForCondition(token != null)
-                .FailWith($"Expected {Subject} to have property '" + expected + "'");
+                .FailWith("Expected {0} to have property '{1}'", Subject, expected);
```
eerhardt: Why this change?  How about converting it all to interpolated string usage?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -398,6 +398,46 @@ private List<string> ReadPropertyNames(JsonTextReader reader)
...
+
+                runtimeFiles.Add(new RuntimeFile(path, assemblyVersion, fileVersion));
+            }
+
+            reader.CheckEndObject();
+
+            return runtimeFiles;
+        }
+
         private List<RuntimeTargetEntryStub> ReadTargetLibraryRuntimeTargets(JsonTextReader reader)
```
eerhardt: I think you need to update `runtimeTargets` as well.

Try building an app that references `System.Data.SqlClient/4.4.0` and you'll see what I mean:

```JSON
"System.Data.SqlClient/4.4.0": {
        "dependencies": {
          "Microsoft.Win32.Registry": "4.4.0",
          "System.Security.Principal.Windows": "4.4.0",
          "System.Text.Encoding.CodePages": "4.4.0",
          "runtime.native.System.Data.SqlClient.sni": "4.4.0"
        },
        "runtime": {
          "lib/netstandard2.0/System.Data.SqlClient.dll": {}
        },
        "runtimeTargets": {
          "runtimes/unix/lib/netstandard2.0/System.Data.SqlClient.dll": {
            "rid": "unix",
            "assetType": "runtime"
          },
          "runtimes/win/lib/netstandard2.0/System.Data.SqlClient.dll": {
            "rid": "win",
            "assetType": "runtime"
          }
        },
        "compile": {
          "ref/netstandard2.0/System.Data.SqlClient.dll": {}
        }
      },
```

src/corehost/cli/version.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef __VERSION_H__
+#define __VERSION_H__
+
+#include "pal.h"
+#include "utils.h"
+
+struct version_t
```
eerhardt: It would be good to document why we now have 2 "version" types.

Are we sure we can't use the existing fx_ver class?

src/managed/Microsoft.Extensions.DependencyModel/RuntimeAssetGroup.cs
```diff
@@ -22,8 +31,35 @@ public RuntimeAssetGroup(string runtime, IEnumerable<string> assetPaths)
...
+        public IReadOnlyList<RuntimeFile> RuntimeFiles
+        {
+            get
+            {
+                if (_runtimeFiles != null)
+                {
+                    return _runtimeFiles;
+                }
+
+                return _assetPaths.Select(path => new RuntimeFile(path, null, null)).ToArray();
```
eerhardt: It feels wrong for a property to call `ToArray()` every time it is invoked.

src/test/Microsoft.Extensions.DependencyModel.Tests/JsonAssetions.cs
steveharter: Previously, if a test constraint fails, it resulted in an exception due to the way fluentassertions handles args, so it would just show the exception instead of the test assertion info:
```
System.FormatException : Input string was not in a correct format.

System.Text.StringBuilder.FormatError()
System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
System.String.Format(String format, Object[] args)
FluentAssertions.Execution.MessageBuilder.FormatArgumentPlaceholders(String failureMessage, Object[] failureArgs) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\MessageBuilder.cs: line 66
FluentAssertions.Execution.MessageBuilder.Build(String message, Object[] messageArgs, String reason, ContextDataItems contextData) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\MessageBuilder.cs: line 42
FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args) in C:\projects\fluentassertions-vf06b\Src\Core\Execution\AssertionScope.cs: line 190
Microsoft.Extensions.DependencyModel.Tests.JsonAssetions.HaveProperty(String expected) in D:\git\coresetup15\src\test\Microsoft.Extensions.DependencyModel.Tests\JsonAssetions.cs: line 32
```

src/corehost/cli/version.h
steveharter: I added comments to version_t something like this:
```
// This type is made to match semantics of System.Version used by tooling and differs from ver_t:
// -- version_t does not require the third segment
// -- version_t does not support the "pre" label behavior
// -- Different terminology; fx_ver_t(major, minor, patch, build) vs version_t(major, minor, build, revision)
```

src/managed/Microsoft.Extensions.DependencyModel/RuntimeAssetGroup.cs
steveharter: Normally this is not hit; it's for compatibility - it has two modes dependency on the ctor invoked. If the "old" ctor is invoked with string but requests RuntimeFile then this would be hit (and vise-versa if the new ctor invoked).

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
steveharter: Thanks

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextExtensions.cs
```diff
@@ -19,6 +19,15 @@ public static IEnumerable<string> GetDefaultNativeAssets(this DependencyContext
...
+        public static IEnumerable<RuntimeFile> GetDefaultNativeRuntimeFiles(this DependencyContext self)
+        {
+            if (self == null)
+            {
+                throw new ArgumentNullException(nameof(self));
+            }
+            return self.RuntimeLibraries.SelectMany(library => library.GetDefaultNativeRuntimeFiles(self));
+        }
+
         public static IEnumerable<string> GetRuntimeNativeAssets(this DependencyContext self, string runtimeIdentifier)
```
eerhardt: Aren't these breaking changes?

src/corehost/cli/deps_format.cpp
```diff
@@ -228,7 +226,8 @@ bool deps_json_t::process_runtime_targets(const json_value& json, const pal::str
                 if (pal::strcasecmp(type.c_str(), deps_entry_t::s_known_asset_types[i]) == 0)
                 {
                     const auto& rid = file.second.at(_X("rid")).as_string();
-                    assets.libs[package.first].rid_assets[rid].by_type[i].vec.push_back(file.first);
+                    deps_asset_t asset(file.first, empty, empty);
```
eerhardt: doesn't this need to be updated for the `runtimeTargets` support?

steveharter: Yes, thanks.

steveharter:
@eerhardt @nguerrera I addressed all feedback. Can you take a look please. Thanks

src/corehost/cli/deps_format.cpp
```diff
@@ -36,11 +36,20 @@ pal::string_t deps_json_t::get_optional_path(
     if (iter != properties.end())
     {
         path = iter->second.as_string();
+    }
 
-        if (_X('/') != DIR_SEPARATOR)
-        {
-            replace_char(&path, _X('/'), DIR_SEPARATOR);
-        }
+    return path;
```
eerhardt: (nit) it seems like `path` is not a great name for this variable.  How about `value`?

PR #3717, opened 2018-02-21
Remove cli-deps feed and publishing of deb-tool

weshaggard:
cli-deps feed has been removed so cleaning up that feed as
well as removing the publishing of the deb-tool that published
to it.

cc @rakeshsinghranchi @eerhardt

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -24,7 +24,6 @@
     <BuildDependsOn Condition="$(PublishType.Contains('nopublishtype'))">
       $(BuildDependsOn);
       PublishDebFilesToDebianRepo;
-      PublishDebToolPackageToFeed;
       PublishFinalOutput;
     </BuildDependsOn>
   </PropertyGroup>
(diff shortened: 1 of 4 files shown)


publish/publish.proj
```diff
@@ -282,19 +282,6 @@
                     Overwrite="$(OverwriteOnPublish)"/>
   </Target>
 
-  <Target Name="PublishDebToolPackageToFeed"
```
eerhardt: I think this still needs to go somewhere.  Can we put it on dotnet-core or dotnet-test or something until we find a real home for it?

/cc @livarcocc @wli3

publish/publish.proj
```diff
@@ -282,19 +282,6 @@
...
-    <ItemGroup>
-      <DebToolPackages Include="$(PackageOutputPath)/dotnet-deb-tool.*.nupkg" />
-    </ItemGroup>
-    <PropertyGroup>
-      <NuGetPushCommand>$(DotnetToolCommand) nuget push --source $(CliNuGetFeedUrl) --api-key $(CliNuGetApiKey) --timeout $(NuGetPushTimeoutSeconds)</NuGetPushCommand>
-    </PropertyGroup>
-    <ExecWithRetriesForNuGetPush Command="$(NuGetPushCommand) %(DebToolPackages.Identity)"
-                                 IgnoredErrorMessagesWithConditional="@(IgnorableErrorMessages)" />
-  </Target>
-
```
nguerrera: Can you publish it to the blob feed?

publish/publish.proj
nguerrera: We don't use dotnet-core anymore (by design) so blob feed would be preferable.

weshaggard: I can publish it but I honestly would prefer to remove the ad-hoc publishing here and force the issue of moving it to a common location. 

From a quick look it doesn't appear as though the cli has picked up a new version of this tool for 9 months since https://github.com/dotnet/cli/commit/4992b6399e990d8ff710a94a63d936c9ea85ee30. So unless you guys feel it is necessary I would like to kill this ad-hoc publishing.

nguerrera: I suppose if the version we use is on the blob feed (it is now), that's ok.

weshaggard: If we get stuck in a bind then we can always manually publish it.

eerhardt: > From a quick look it doesn't appear as though the cli has picked up a new version of this tool for 9 months since dotnet/cli@4992b63

Hmm, that's interesting.  @rakeshsinghranchi made a change just a few months ago to the tool that I thought CLI needed. See https://github.com/dotnet/core-setup/pull/3487#pullrequestreview-82546308

@wli3?

eerhardt: We can kill the publishing now (since it hasn't been needed for 9 months).  But we should get someone on the hook for putting this tool in the proper place.

weshaggard:
@dotnet-bot test this please

publish/publish.proj
weshaggard: Created https://github.com/dotnet/core-setup/issues/3718 assigned to @rakeshsinghranchi

Issue #3706, opened 2018-02-20
Tizen release/2.1 builds aren't running


weshaggard:
@hqueue @hseok-oh various repo's are failing in the release/2.1 branches for Tizen as the it doesn't look like Tizen is building matching bits for those branches. 

https://github.com/dotnet/core-setup/pull/3698#issuecomment-366335516
https://github.com/dotnet/coreclr/pull/16405#issuecomment-366310355

ahsonkhan:
Also in master: https://github.com/dotnet/core-setup/pull/3736#issuecomment-368890923

weshaggard:
@hqueue @hseok-oh any updates on this? Should we remove tizen from CI temporarily

hqueue:
@weshaggard Sorry we are late to this issue and confused with other Tizen CI issues. 

@alpencolt @gbalykov  It seems that we need  Tizen leg for release/2.1. Can you please take a look at this ?

/cc: @lemmaa

alpencolt:
@hqueue we will enable it.
@weshaggard could you add me and @gbalykov to copy in all Tizen related issues

weshaggard:
OK should we only tag @gbalykov and @alpencolt for tizen related issued now or continue tagging @hqueue  and @hseok-ok as well? I'm planning add this to our owner list at https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/issue-guide.md.

hqueue:
@weshaggard I thinks @alpencolt and @gbalykov will be a first contact point.

And I think @wscho77 will be necessary to be added to the list.

For me(@hqueue) and @hseok-oh, if there is no limitation, keep us in the list for a while and remove us later.

@alpencolt @gbalykov  @wscho77 what do you think of it ?

alpencolt:
Thank you.
I don't mind, other guys I think as well:)

weshaggard:
I've added a section here https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/issue-guide.md#community-partner-experts that lists the points of contact. I've added @alpencolt and @gbalykov already but you guys please feel free to maintain that list and add as many folks as you like. That will give us one location to keep track of contacts.

weshaggard:
This looks to be fixed so closing.

PR #3738, opened 2018-02-27
[release/2.1] Update BuildTools to 2.1.0-preview1-02523-01

dagood:
This version includes a fix to unblock dependency auto-update flow after GitHub stopped allowing weak cryptographic standards: dotnet/core-eng#2734.

This problem hasn't affected builds because 2.1 is built in ProdCon, where Core-Setup's toolset isn't used to publish to dotnet/versions.


dagood:
Tizen failure is known issue: https://github.com/dotnet/core-setup/issues/3706

Issue #3642, opened 2018-01-26
Provide explicit control for an app regarding multiple framework knobs


steveharter:
As part of reviewing the feature https://github.com/dotnet/core-setup/issues/3457, it was determined that the app's deps.json file should have explicit control over settings framework may specify in it's runtimeconfig.json. For example, whether the framework should roll-forward.

Assuming the following layering:
1. App (e.g. has runtimeconfig.json settings affecting roll-forwad of ASP.NET)
2. ASP.NET (e.g. has runtimeconfig.json settings affecting roll-forwad of NetCoreApp)
3. NetCoreApp (has no runtimeconfig.json settings because there is no lower framework)

It is possible ASP.NET turns on the ability of NetCoreApp to roll-forward on [major] releases by specifying `rollForwardOnNoCandidateFx = 2` in its runtimeconfig.json but if that behavior is not wanted by the app, they should have the option of disabling that. Currently, if the app specifies `rollForwardOnNoCandidateFx = 1` to turn [major] roll-forward off that only affects ASP.NET, not NetCoreApp.

The normal hierarchy for most knobs is (from default to explicit)
a) Default value determined by the framework (e.g. roll-forward on [Minor])
b) Environment variable override (e.g. DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX)
c) Each layer's runtimeconfig.json, starting with app (e.g. rollForwardOnNoCandidateFx). Lower layers override.
e) A -- command line value (e.g. --roll-forward-on-no-candidate-fx)

The proposal is to insert a (d) option:
d) Another section in runtimeconfig.json can can specify knobs per-framework (or lower-level)

This would override each framework's value in (c) but still allow for a command-line option (e) to have the last say.

Example of current runtimeconfig.json:
```
{
    "runtimeOptions": {
        "framework": {
            "name": "Microsoft.AspNetCore.All",
            "version": "1.0.0"
        },
        "configProperties": {
            "System.GC.Server": true,
            "System.GC.Concurrent": true
        },
        "additionalProbingPaths": [
            "C:\\Users\\sharter\\.dotnet\\store\\|arch|\\|tfm|",
            "C:\\Users\\sharter\\.nuget\\packages",
            "C:\\Program Files (x86)\\Microsoft SDKs\\NuGetPackagesFallback",
            "C:\\Program Files\\dotnet\\sdk\\NuGetFallbackFolder"
        ],
        "applyPatches": false,
[...]

PR #3741, opened 2018-02-28
Support 'additionalFrameworks' section in app's runtimeconfig.json

steveharter:
This adds support for the app to control and have last say for the following settings for each lower level framework:
1) "rollForwardOnNoCandidateFx"
2) "applyPatches"
3) "version" (the targeted framework version)

Fixes https://github.com/dotnet/core-setup/issues/3642

I will update the cli [doc ](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md)by EOW.

Sample addition in an app's runtimeconfig.json:
```
{
    "runtimeOptions": {
        "additionalFrameworks": [
            {
                "name": "Microsoft.AspNetCore.App",
                "version": "1.0.2",
                "rollForwardOnNoCandidateFx": "0",
            },
            {
                "name": "Microsoft.AspNetCore.All", 
                "version": "2.0.0",
                "applyPatches": true
            }
        ]
}
```

### DIFF
--- a/src/corehost/cli/args.h
+++ b/src/corehost/cli/args.h
@@ -14,7 +14,6 @@ struct probe_config_t
 {
     pal::string_t probe_dir;
     bool patch_roll_fwd;
-    bool prerelease_roll_fwd;
     const deps_json_t* probe_deps_json;
     int fx_level;
 
--- a/src/corehost/cli/fx_definition.h
+++ b/src/corehost/cli/fx_definition.h
@@ -24,7 +24,7 @@ public:
     const pal::string_t& get_found_version() const { return m_found_version; }
     const pal::string_t& get_dir() const { return m_dir; }
     const runtime_config_t& get_runtime_config() const { return m_runtime_config; }
-    void parse_runtime_config(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* previous_config, const runtime_config_t* app_config);
+    void parse_runtime_config(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* higher_layer_config, const runtime_config_t* app_config);
 
     const pal::string_t& get_deps_file() const { return m_deps_file; }
     void set_deps_file(const pal::string_t value) { m_deps_file = value; }
--- a/src/corehost/cli/fx_definition.h
+++ b/src/corehost/cli/fx_definition.h
@@ -24,7 +24,7 @@ public:
     const pal::string_t& get_found_version() const { return m_found_version; }
     const pal::string_t& get_dir() const { return m_dir; }
     const runtime_config_t& get_runtime_config() const { return m_runtime_config; }
-    void parse_runtime_config(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* defaults);
+    void parse_runtime_config(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* previous_config, const runtime_config_t* app_config);
 
     const pal::string_t& get_deps_file() const { return m_deps_file; }
     void set_deps_file(const pal::string_t value) { m_deps_file = value; }
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -1273,7 +1273,7 @@ int read_config(
         get_runtime_config_paths_from_arg(runtime_config, &config_file, &dev_config_file);
     }
 
-    app.parse_runtime_config(config_file, dev_config_file, nullptr);
+    app.parse_runtime_config(config_file, dev_config_file, nullptr, nullptr);
     if (!app.get_runtime_config().is_valid())
     {
         trace::error(_X("Invalid runtimeconfig.json [%s] [%s]"), app.get_runtime_config().get_path().c_str(), app.get_runtime_config().get_dev_path().c_str());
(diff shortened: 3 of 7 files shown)


src/corehost/cli/args.h
```diff
@@ -14,7 +14,6 @@ struct probe_config_t
 {
     pal::string_t probe_dir;
     bool patch_roll_fwd;
-    bool prerelease_roll_fwd;
```
steveharter: Review note: this was not used, so I remove it

src/corehost/cli/runtime_config.cpp
```diff
@@ -201,14 +280,60 @@ bool runtime_config_t::ensure_parsed()
...
+                            {
+                                break;
+                            }
+
+                            m_additional_frameworks[fx_overrides.m_fx_name] = fx_overrides.m_fx;
+                        }
+                    }
+
+                    // Use the framework specified if it exists in the additionalFramework
+                    app_config = this;
```
steveharter: Review note: by falling through to the code below, this means the "additionalFrameworks" section can be used to override the current (first) framework the app specifies. We could prevent that from occurring if we want to limit redundant cases, as now the "framework" section can also be used to specify the other settings besides 'version' - applyPatches, etc

src/corehost/cli/runtime_config.cpp
```diff
@@ -124,8 +137,74 @@ bool runtime_config_t::parse_opts(const json_value& opts)
     m_portable = true;
 
     const auto& fx_obj = framework->second.as_object();
+
     m_fx_name = fx_obj.at(_X("name")).as_string();
-    m_fx_ver = fx_obj.at(_X("version")).as_string();
+
+    // Step #3: Use the values in the app's "framework" section
+    bool rc = parse_framework(fx_obj);
```
steveharter: Review note: we could just read "version" here instead instead of calling into `parse_framework`. However by calling into `parse_framework` we allow the other values (applyPatches, etc) to be specified and make the schema consistent with "additionalFrameworks"

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1326,19 +1326,22 @@ int fx_muxer_t::read_config_and_execute(
...
-    auto config = app->get_runtime_config();
+    auto app_config = app->get_runtime_config();
 
     // 'Roll forward on no candidate fx' is set to 1 (roll_fwd_on_no_candidate_fx_option::minor) by default. It can be changed through:
-    // 1. Command line argument (--roll-forward-on-no-candidate-fx). Only defaults the app's config.
+    // 1. Command line argument (--roll-forward-on-no-candidate-fx).
     // 2. Runtimeconfig json file ('rollForwardOnNoCandidateFx' property), which is used as a default for lower level frameworks if they don't specify a value.
-    // 3. DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX env var. Only defaults the app's config.
-    // The conflicts will be resolved by following the priority rank described above (from 1 to 3).
+    // 3. Runtimeconfig json file ('rollForwardOnNoCandidateFx' property in "framework" section:).
```
eerhardt: I would think the order between 2 & 3 would be flipped.  The 'rollForwardOnNoCandidateFx' property in the "framework" section should 'win' over the global one.

src/corehost/cli/runtime_config.h
```diff
@@ -18,45 +19,110 @@ enum class roll_fwd_on_no_candidate_fx_option
...
+    bool has_fx_ver;
+    bool has_patch_roll_fwd;
+    bool has_roll_fwd_on_no_candidate_fx;
+
+    pal::string_t fx_ver;
+    bool patch_roll_fwd;
+    roll_fwd_on_no_candidate_fx_option roll_fwd_on_no_candidate_fx;
+};
+
+class runtime_config_t;
```
eerhardt: Is this necessary?

src/corehost/cli/fx_definition.h
```diff
@@ -24,7 +24,7 @@ class fx_definition_t
     const pal::string_t& get_found_version() const { return m_found_version; }
     const pal::string_t& get_dir() const { return m_dir; }
     const runtime_config_t& get_runtime_config() const { return m_runtime_config; }
-    void parse_runtime_config(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* defaults);
+    void parse_runtime_config(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* previous_config, const runtime_config_t* app_config);
```
eerhardt: What does `previous_config` mean? Is there a better name for this?

src/corehost/cli/runtime_config.h
```diff
@@ -18,45 +19,110 @@ enum class roll_fwd_on_no_candidate_fx_option
...
+        return (has_roll_fwd_on_no_candidate_fx ? &roll_fwd_on_no_candidate_fx : nullptr);
+    }
+    void set_roll_fwd_on_no_candidate_fx(roll_fwd_on_no_candidate_fx_option value)
+    {
+        has_roll_fwd_on_no_candidate_fx = true;
+        roll_fwd_on_no_candidate_fx = value;
+    }
+
+private:
+    bool has_fx_ver;
```
eerhardt: Can you just check if `fx_ver` is empty instead of needing to track another variable?

src/corehost/cli/runtime_config.cpp
```diff
@@ -201,14 +280,60 @@ bool runtime_config_t::ensure_parsed()
...
         const auto& json = root.as_object();
         const auto iter = json.find(_X("runtimeOptions"));
         if (iter != json.end())
         {
-            parse_opts(iter->second);
+            rc = parse_opts(iter->second);
+
+            if (rc)
+            {
+                if (app_config == nullptr)
```
eerhardt: This check and proceeding comment are confusing.  if `app_config == nullptr` and `// For the app, populate the additionalFrameworks`.

Is this supposed to be `if (app_config != nullptr)`?

src/corehost/cli/fx_definition.h
steveharter: It means the config from the next higher layer which is used to default the current. So nullptr for the app, and the app's config for the first framework.

src/corehost/cli/runtime_config.h
steveharter: Yes it's a forward declaration since runtime_config_t in parse(...) uses its own class

src/corehost/cli/runtime_config.h
steveharter: I was being consistent with the pattern. So assigning a "" to version would mean it would try to set a "" version during defaulting (which may lead to a desired error later on). If I check for empty, it wouldn't do that.

src/corehost/cli/runtime_config.cpp
steveharter: It is correct, but yes perhaps confusing. If app_config is nullptr, then there is app_config yet, so we must be the app.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: OK thanks. I will fix the comment - the comment is wrong, but the current semantics are what you describe.

src/corehost/cli/runtime_config.h
eerhardt: didn't parse(..) use its own class before this change?  How did that work?

The original code was:

```C++
class runtime_config_t
{
public:
    runtime_config_t();
    void parse(const pal::string_t& path, const pal::string_t& dev_path, const runtime_config_t* defaults);
```

steveharter: You're correct; that forward decl was not necessary. Removing.

steveharter:
Merging this; Tizen is not working due to build issues

Issue #3476, opened 2017-11-30
Additional-deps feature doesn't validate deps.json file


steveharter:
## Steps to reproduce
Create the folder c:\ReproDepsIssue\shared\Microsoft.NETCore.App\2.0.0
Add a foo.deps.json file with garbage text to that folder
Set DOTNET_ADDITIONAL_DEPS=c:\ReproDepsIssue
Set COREHOST=TRACE=1
Run an app that targets 2.0.0

## Expected  behavior
An error of some sort and app failure.

Note that there is an error when using a malformed deps.json for an app or framework (just the additional-deps ignores errors)

## Actual behavior
The parsing error is ignored; silent failure. In addition, the trace output contains `Using specified additional deps.json: 'c:\ReproDepsIssue\shared\Microsoft.NETCore.App\2.0.0\foo.deps.json'
which is misleading because it did not (could not)use the file.

PR #3743, opened 2018-02-28
Fail when additional deps files are not valid

steveharter:
Previously a bad deps.json file that was included via the "additional deps files" feature was treated as a silent failure, and the app was allowed to run (or try to run). This PR changes that so the app fails to launch with an error such as:
```
A JSON parsing exception occurred in [d:\git\repros\ValidateAdditionalDeps\shared\Microsoft.NETCore.App\2.0.0\foo.deps.json]: * Line 1, Column 2 Syntax error: Malformed token
Error initializing the dependency resolver: An error occurred while parsing: d:\git\repros\ValidateAdditionalDeps\shared\Microsoft.NETCore.App\2.0.0\foo.deps.json
```
Fixes https://github.com/dotnet/core-setup/issues/3476

Also fixed was to use the "requested" version of the framework when looking up the additional deps, when using the mode where additional deps specifies a directory following framework naming convention such as `c:\foo\shared\Microsoft.NETCoreApp\2.1.0`. Previously, it was using the "found" version of the framework. Originally the code was using "requested", was changed to "found" in 2.1preview a few weeks ago, and now this PR changes it back to "requested". The "requested" version should be used as that is what the app targets, and deployed the additional deps folders using that version in the path.

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -633,11 +633,11 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
         {
             for (int i = 1; i < m_fx_definitions.size(); ++i)
             {
-                // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
+                // We'll search deps files in 'base_dir'/shared/fx_name/fx_requested_ver
                 pal::string_t additional_deps_path_fx = additional_deps_path;
                 append_path(&additional_deps_path_fx, _X("shared"));
                 append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_name().c_str());
-                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_found_version().c_str());
+                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_requested_version().c_str()); // Use requested version as that is what the app deployed with
 
                 // The resulting list will be empty if 'additional_deps_path_fx' is not a valid directory path
                 std::vector<pal::string_t> list;
--- a/src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
@@ -128,61 +128,6 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.LightupApp
                 .HaveStdOutContaining("Hello LightupClient");
         }
 
-        // Attempt to run the app with a bad lightup deps.json specified.
-        [Fact]
-        public void Muxer_activation_of_LightupApp_With_Bad_JsonFile_Fails()
-        {
-            var fixtureLib = PreviouslyBuiltAndRestoredLightupLibTestProjectFixture
-                .Copy();
-
-            var fixtureApp = PreviouslyBuiltAndRestoredLightupAppTestProjectFixture
-                .Copy();
-
-            var dotnet = fixtureApp.BuiltDotnet;
-            var appDll = fixtureApp.TestProject.AppDll;
-            var libDll = fixtureLib.TestProject.AppDll;
-
-            // Get the version number of the SharedFX we just built since that is the version
-            // going to be specified in the test's runtimeconfig.json.
-            var builtSharedFXVersion = Path.GetFileName(dotnet.GreatestVersionSharedFxPath);
-
-            // Create the M.N.App specific folder where lightup.deps.json can be found.
-            var baseDir = fixtureApp.TestProject.ProjectDirectory;
-            var customLightupPath = Path.Combine(baseDir, "shared");
-
-            // Delete any existing artifacts
-            if (Directory.Exists(customLightupPath))
-            {
-                Directory.Delete(customLightupPath, true);
-            }
-
-            customLightupPath = Path.Combine(customLightupPath, "Microsoft.NETCore.App");
-            customLightupPath = Path.Combine(customLightupPath, builtSharedFXVersion);
-
-            // Create the folder to which lightup.deps.json will be copied to.
-            Directory.CreateDirectory(customLightupPath);
-
-            // Copy the lightup.deps.json
-            var libDepsJson = fixtureLib.TestProject.DepsJson;
-            string path = Path.Combine(customLightupPath, Path.GetFileName(libDepsJson));
-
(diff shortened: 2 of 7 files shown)


src/corehost/cli/deps_resolver.cpp
```diff
@@ -633,11 +633,11 @@ void deps_resolver_t::resolve_additional_deps(const hostpolicy_init_t& init)
         {
             for (int i = 1; i < m_fx_definitions.size(); ++i)
             {
-                // We'll search deps files in 'base_dir'/shared/fx_name/fx_ver
+                // We'll search deps files in 'base_dir'/shared/fx_name/fx_requested_ver
                 pal::string_t additional_deps_path_fx = additional_deps_path;
                 append_path(&additional_deps_path_fx, _X("shared"));
                 append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_name().c_str());
-                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_found_version().c_str());
+                append_path(&additional_deps_path_fx, m_fx_definitions[i]->get_requested_version().c_str()); // Use requested version as that is what the app deployed with
```
eerhardt: How does this work with patch roll-forward?  If the app was built for `2.0.0`, but only `2.0.5` was installed on the machine, will the additional deps feature break?

eerhardt: I'm thinking of the existing use-cases today where ASP.NET installs the `additionalDeps` folder.  See 
`C:\Program Files\dotnet\additionalDeps`.

The app developer doesn't create the folder structure.  The whole point of the additional deps feature is that the app developer doesn't know anything about these additional dependencies.  They are injected by the host (like VS or Azure).  So if an Azure machine only have `2.0.5` installed, and a developer deploys their web app to that machine, the additional deps feature won't work anymore.  The `runtimeconfig.json` will have `2.0.0` (because that's what the SDK injects by default, no matter the runtime used).  This means this code will look for `2.0.0` and it won't be there.

/cc @pakrym

src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
```diff
@@ -128,6 +128,61 @@ public void Muxer_activation_of_LightupApp_WithLightupLib_Succeeds()
                 .HaveStdOutContaining("Hello LightupClient");
         }
 
+        // Attempt to run the app with a bad lightup deps.json specified.
```
pakrym: Test for roll forward fix?

src/corehost/cli/deps_resolver.cpp
pakrym: We would have to keep shipping additionaldeps for every patch version of shared runtime we ever produced even if we don't ship the runtime any more in case someone got rolled forward of his original requested version. Sounds pretty broken.

steveharter: I'll verify 2.0 semantics and get back. For now, I assume we want to emulate 2.0 semantics.

pakrym: Thinking more about it if the alternative is rolling to 2.1 and getting into mixed version graph state locking to requested version might not be that bad, considering that by default everyone gets `2.0` as a target runtime.

We might also want to update additionalDeps generation process to produce two versions of deps files - for `2.0` and the latest runtime.

steveharter: The behavior for 2.0.x is to use the "requested" version for the additional deps folder. i.e. 2.0.x uses the additional deps folder that exactly matches the version of the framework specified in the runtimeconfig.json, and does not use the "found" folder name of any patch (or minor) roll-forward.

@eerhardt - when we talked, you hard a different behavior. To emulate true 2.0.x behavior, try disabling (e.g. prefix with an underscore) out any `c:\Program Files\dotnet\host\fxr\2.1.*` folders as we always pick the latest fxr.

Here's my repro steps to verify:

0) Disable any 2.1* folders in `C:\Program Files\dotnet\host\fxr\` as mentioned above
1) Create the folder `c:\ReproDepsIssue\shared\Microsoft.NETCore.App\2.0.0`
2) Add a foo.deps.json file there (empty is fine)
3) Set `DOTNET_ADDITIONAL_DEPS=c:\ReproDepsIssue`
4) Set `COREHOST=TRACE=1`
5) Ensure you have 2.0.5 (the latest 2.0.x) release or a 2.0.x release other than 2.0.0 so we get a patch roll-forward
6) Run an app that targets 2.0.0 (has version=2.0.0 in its runtimeconfig.json) and redirect stderr to make it easier to search (e.g. `dotnet exec myapp.dll 2>out.txt`)
7) Search the stderr for "Using D:\git\repros\console4\bin\Debug\netcoreapp2.0\console4.deps.json deps file

You should see the text: "Using specified additional deps.json: 'c:\ReproDepsIssue\shared\Microsoft.NETCore.App\2.0.0\foo.deps.json'" meaning that it looked in the 2.0.0 folder even though a roll-forward occurred from 2.0.0 to 2.0.5

Based on this, I'm inclined to keep the "requested" semantics (as the PR is currently). However if we do that then I need to change the 2.1 hostfxr.dll to pass the "requested" version (and not "found") to the older 2.0.x hostpolicy.dll based on the issue in repro step 0 above.

src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
steveharter: Yeah agree we should have one. I'll see if this is feasible

src/corehost/cli/deps_resolver.cpp
eerhardt: I just tested it on a machine that only ever had `2.0.3` installed on it, and it works like you describe above - the host only looks for an additionalDeps file in `2.0.0`, since that is what is in the runtimeconfig.json.  I wonder if that was intentional or not, as it doesn't seem correct in my mind.

Note - even though I only installed `2.0.3` on the machine, it appears that ASP.NET installs both `2.0.0` and `2.0.3` folders into the `additionalDeps` folder.

steveharter:
@eerhardt @pakrym any other comments? Summary:
1) Fixes the original issue of not failing when there's a bad json
2) Reverts additional-deps semantics to 2.0.x behavior of using the "requested" folder (not "found")
3) Adds a test that covers both cases above
4) Logged new issue https://github.com/dotnet/core-setup/issues/3754 to discuss options going forward to deal with lightup + roll forward

PR #3766, opened 2018-03-02
Don't publish to deb feed

wtgodbe:
Stop publishing to Cli Deps feed in 2.0.0

CC @weshaggard


wtgodbe:
@weshaggard is this OK to merge?

PR #3767, opened 2018-03-02
Update Buildtools version

wtgodbe:
Hoping to unblock updating the Versions repo for 2.0.0.

CC @dagood


dagood:
Looks like the MyGet dotnet-core feed cleanup is causing problems here with old servicing branch builds, as somewhat expected:

> ...\src\pkg\projects\Microsoft.NETCore.App\src\Microsoft.NETCore.App.depproj : error NU1603: Microsoft.NETCore.App depends on Microsoft.Private.CoreFx.NETCoreApp (>= 4.4.4-servicing-26212-01) but Microsoft.Private.CoreFx.NETCoreApp 4.4.4-servicing-26212-01 was not found. An approximate best match of Microsoft.Private.CoreFx.NETCoreApp 4.5.0-dev-di-25909-02 was resolved.

I'll file a core-eng issue.

dagood:
In the meantime, @wtgodbe can you add the central blob feed `https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json`? Since that's the way of the future.

dagood:
My mistake--that is a manually upgraded version from https://github.com/dotnet/core-setup/commit/9775b21d35d2861996de558b38223a5484666b8b, probably not an old package. You probably know where it's expected to be better than I do--let me know if I can help.

wtgodbe:
@dagood hopefully we should be able to find that package once a clean CoreFx build comes through.

PR #3769, opened 2018-03-03
Turn off symbol exports to prevent dynamic collisions

steveharter:
This turns off symbol export and prevents dotnet.exe, hostfxr and hostpolicy from colliding with the same types which can cause various memory issues etc. when they become out-of-sync over time.

It seems I did not have to create an export list, as re-using the existing `SHARED_API` define with `__attribute__((__visibility__("default")))` enabled per-method symbols.

fixes https://github.com/dotnet/corefx/issues/27571

Verified on OSX per the linked issue repro steps, and also verified by diffing via `nm -g` (which returns the external\global symbols) on the hostfxr.dylib to make sure the cross-compiled code does not show up. In addition, the existing symbols we want to be public (e.g. `hostfxr_main` are exported without mangling.

### DIFF
--- a/src/corehost/cli/dll/CMakeLists.txt
+++ b/src/corehost/cli/dll/CMakeLists.txt
@@ -12,6 +12,8 @@ else()
     add_compile_options(-fPIC)
 endif()
 
+add_compile_options(-fvisibility=hidden)
+
 include(../setup.cmake)
 
 # Include directories
--- a/src/corehost/cli/exe/apphost/CMakeLists.txt
+++ b/src/corehost/cli/exe/apphost/CMakeLists.txt
@@ -5,6 +5,8 @@ cmake_minimum_required (VERSION 2.6)
 project(apphost)
 set(DOTNET_HOST_EXE_NAME "apphost")
 
+add_compile_options(-fvisibility=hidden)
+
 # Add RPATH to the apphost binary that allows using local copies of shared libraries
 # dotnet core depends on for special scenarios when system wide installation of such 
 # dependencies is not possible for some reason.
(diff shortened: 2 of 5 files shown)


steveharter:
test OSX x64 Release Build

nguerrera:
I don't know enough about how this works to review this.

nguerrera:
The nm validation makes me think it's right, but I'll let someone else sign off.

steveharter:
Committing this; the OSX failure is related to infrastructure.

src/corehost/cli/dll/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ else()
     add_compile_options(-fPIC)
 endif()
 
+add_compile_options(-fvisibility=hidden)
```
eerhardt: Instead of duplicating this in all 5 CMakeLists, can you instead define it in a common place?  Ex. https://github.com/dotnet/core-setup/blob/master/src/settings.cmake.

pakrym:
Do we have tests that run latest dotnet+corehost against previously released versions of shared rutimes?

src/corehost/cli/dll/CMakeLists.txt
steveharter: That shared file appears to fairly generic so I kept the setting in each project, close to the fPIC setting which is related

steveharter:
>> Do we have tests that run latest dotnet+corehost against previously released versions of shared rutimes?

None automated - I added https://github.com/dotnet/core-setup/issues/3777. We perform manual validation when the "contract" changes between hostfxr and hostpolicy, but that could use an automated test, which would also have some coverage of these symbol-based issues.

pakrym:
@steveharter do you foresee any issues in the case where hostfxr gets updated but dotnet.exe doesn't?

steveharter:
@pakrym 
> do you foresee any issues in the case where hostfxr gets updated but dotnet.exe doesn't?

No there shouldn't be an issue with that. dotnet.exe by design is very simple and really only job is to find the latest hostfxr and forward to it. Since the dotnet.exe file is global, we expect there will be cases where the file is in use and can't be updated, so we try to avoid having any new code there.

PR #3794, opened 2018-03-07
Update Buildtools

wtgodbe:
CC @dagood, @weshaggard


wtgodbe:
@weshaggard getting a green build of CoreClr at least depends on this going in, due to the failure here - https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1447357. @dagood can you point us to the github issue for this?

weshaggard:
>@weshaggard getting a green build of CoreClr at least depends on this going in, due to the failure here

Yes I understand you need to make these changes in coreclr to get it green but why does the core-setup change need to go in to make coreclr green?

dagood:
Oh, I misread that. I also don't see a reason that this has to merge to get a CoreCLR build out (or CoreFX for that matter).

wtgodbe:
This doesn't need to go in to make CoreClr green, but it should hopefully have a green CI once we get builds out of CoreClr and CoreFx.

wtgodbe:
We have a bit of a deadlock here - This won't be green until we get auto-update PRs merged from CoreClr & CoreFx, but we can't get those Auto-PRs in until this is merged, since the TLS upgrade won't be present, so https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=1461652 will continue to fail. @weshaggard @dagood do you think I should just merge this to unblock the auto-PRs?

dagood:
Hmm. I suggest adding an auto-update commit to this PR by using `/t:UpdateToRemoteDependencies` locally. (It should give you a suggested commit message to match what it usually does.) We'd still have to run another CoreCLR or CoreFX build after merging this PR to see if the "real" Maestro-triggered auto-PRs work though.

wtgodbe:
This now has the dependencies updated as well

wtgodbe:
Just added the blob feed

dir.props
```diff
@@ -35,6 +35,7 @@
     <!-- list of nuget package sources passed to NuGet -->
     <RestoreSources>
       https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;
+      https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
       https://dotnet.myget.org/F/cli-deps/api/v3/index.json;
```
eerhardt: The `cli-deps` feed was deleted.  If you need to make anymore changes to this PR, can you also remove this?  But you don't have to reset CI just for this one change.

PR #3861, opened 2018-03-20
[1.1] Update buildtools and cli version to same ones as CoreFX and CoreCLR

janvorli:
Having separate CLI version greatly increases the cost of adding new
distros, so we should keep coreclr, corefx and core-setup using the
same CLI and buildtools version for their tools.

It also unifies the init-tools.sh / cmd with the other repos.


pkg/Tools-Override/_
```diff
@@ -0,0 +1 @@
+
```
weshaggard: we should just remove the tools-override directory and part of the script.

pkg/init-tools.cmd
```diff
@@ -65,10 +65,20 @@ echo Running: "%BUILD_TOOLS_PATH%init-tools.cmd" "%~dp0" "%DOTNET_CMD%" "%TOOLRU
...
 if not [%INIT_TOOLS_ERRORLEVEL%]==[0] (
-	echo ERROR: An error occured when trying to initialize the tools. Please check '%INIT_TOOLS_LOG%' for more details.
-	exit /b %INIT_TOOLS_ERRORLEVEL%
+  echo ERROR: An error occured when trying to initialize the tools. 1>&2
+  goto :error
 )
 
-:: Create sempahore file
+:: Always copy over the Tools-Override
+xcopy /vy "%~dp0Tools-Override\"* "%~dp0Tools" 1>nul
```
weshaggard: Just remove this part.

pkg/init-tools.sh
```diff
@@ -105,18 +117,29 @@ if [ ! -e $__INIT_TOOLS_DONE_MARKER ]; then
...
     echo "Running: $__BUILD_TOOLS_PATH/init-tools.sh $__scriptpath $__DOTNET_CMD $__TOOLRUNTIME_DIR" >> $__init_tools_log
     $__BUILD_TOOLS_PATH/init-tools.sh $__scriptpath $__DOTNET_CMD $__TOOLRUNTIME_DIR >> $__init_tools_log
     if [ "$?" != "0" ]; then
-        echo "ERROR: An error occured when trying to initialize the tools. Please check '$__init_tools_log' for more details."1>&2
+        echo "ERROR: Could not restore build tools correctly." 1>&2
+        display_error_message
         exit 1
     fi
+
+    cp Tools-Override/* Tools/
```
weshaggard: Just remove this part.

pkg/Tools-Override/_
janvorli: Even though that makes it different from the corefx version?

weshaggard: It just feels pointless to add this empty file. You could also add a exists check in the scripts perhaps.

weshaggard: I don't have a strong opinion but I could definitely see someone looking at this in the future and asking what it is for and even try deleting it.

janvorli: Ok, the exists check sounds good. I'll do that.

janvorli:
@weshaggard I've made the folder optional.

rakeshsinghranchi:
Thanks @janvorli !

PR #3868, opened 2018-03-21
Mark packages as stable for 1.1.8

wtgodbe:
CC @crummel @weshaggard


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -76,7 +76,7 @@
       <Version>1.0.2</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.Runtime.CoreCLR">
-      <Version>1.0.12</Version>
+      <Version>1.0.13</Version>
```
wtgodbe: I think there will be a merge conflict here after I merge auto update PR

weshaggard:
Remember to update the 1.0 package dependencies in Microsoft.NETCore.App.pkgproj as well see https://github.com/dotnet/core-setup/commit/12d23faa0152360759f215f5b35bc2b96e14169f#diff-e2c52444a9da4a5eae65d36e5d3e6aba as an example.

wtgodbe:
@weshaggard what needs to be updated there? I believe we only updated System.Console in 1.0.0, which isn't present in the pkgproj

weshaggard:
>@weshaggard what needs to be updated there? I believe we only updated System.Console in 1.0.0, which isn't present in the pkgproj

Your correct there isn't anything that needs updated there this go around, unless we end up adding some of the new distro stuff.

wtgodbe:
@dotnet-bot test this please

pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
safern: You need to update Microsoft.NETCore.DotNetHostPolicy 1.0.0 dependency to 1.0.11 here as it always gets built for 1.0.0

wtgodbe: Updated

safern:
>I believe we only updated System.Console in 1.0.0, which isn't present in the pkgproj

It is not present in the pkgproj because we had not serviced it before, this is the first time we service it and we just bumped its dependency in the closure for 1.0.0 here:
https://github.com/dotnet/core-setup/blob/release/1.0.0/pkg/projects/Microsoft.NETCore.App/project.json#L14, so you need to add System.Console in the pkgproj and point it to the version we're servicing.

wtgodbe:
@safern done

Issue #3720, opened 2018-02-22
Support framework-dependent apphost


steveharter:
This is a placeholder to support a native executable (e.g. `myapp.exe`) that will find and use the shared framework. There are several scenarios, including global tools, that desire this functionality in the 2.1 timeframe.

Currently, a self-contained app always has a native executable but cannot use the shared framework, and a framework-dependent app is always `dotnet.exe`.

The likely implementation will consist of a native executable (apphost.exe renamed to <app>.exe) that can either run in standalone or shared mode depending on the presence (or lack of) of runtimeconfig.json settings that specify the shared framework. For Linux\OSX, a new environment variable is likely required to specify the location of the shared framework to use; on Windows this is not necessary due to the well-known "Program Files" location.

Related scenarios: https://github.com/dotnet/cli/issues/6237, https://github.com/dotnet/core-setup/issues/196

steveharter:
cc @srivatsn for your scenario (offline)

Petermarcu:
@steveharter should we have a design meeting this week?

steveharter:
@Petermarcu yes I am going to set one up this week. The basic proposal for the host is:
1) Modify the current host code to support a framework-dependent apphost
2) Add a new environment variable (say %DOTNET_PATH%)
3) This environment variable will be used by apphost to find the hostfxr and in turn find the frameworks

Additional options (not required for initial functionality; but want to assume forward-compatibility):
a) Add support for runtimeconfig.json (dotnetPath) and command line (--dotnet_path) which will override the environment variable
b) Embed the default path into the apphost for the appropriate distro, so that the environment variable doesn't have to be set for the "happy path". The environment variable (and runtimeconfig\command-line) would override this.

itadapter:
@steveharter looks like your approach with env var will solve the problems we have now with adoption of core with legacy dev groups. what do you think is a realistic ETA for shipped binary so regular user can start benefiting?

steveharter:
@itadapter can you explain your scenario with legacy dev groups?

ETA is pending; we'd like to get basic support in for 2.1, but no guarantees due to tight schedule.

itadapter:
[...]

Issue #3777, opened 2018-03-05
Add tests to verify hostfxr with older hostpolicy


steveharter:
This issue is to add automated tests to verify a newer version of hostfxr running against older versions of hostpolicy.

This can break because of incorrect "contract" semantics, or because of unix\osx symbol-based export and merging (https://github.com/dotnet/corefx/issues/27571).

pakrym:
I would also add a set of tests that verify old dotnet.exe with new hostfxr and old runtime. I know we don't think that anything can break between dotnet and hostfxr but better safe then sorry.

karelz:
Test work item which didn't fit due to other high-pri features - marking as Post-ZBB

steveharter:
Planning to port this to 2.1.x when the 2.2 tests are ready

PR #3888, opened 2018-03-23
Add support for framework-dependent apphost

steveharter:
New functionality for the apphost (`<myapp>.exe`) including:
1) Ability to have an apphost use the shared framework by specifying the framework information in the runtimeconfig.json
2) Ability to have a non-default location for the shared framework via new `%DOTNET_ROOT%` environment variable. This environment variable is only used by apphost. The default locations otherwise are `%ProgramFiles%`  (Windows), `/usr/share/dotnet` (Unix), `/usr/local/share/dotnet` (OSX)
3) Ability for an apphost to be renamed. Note, the exe still contains the original assembly name, so that does not need to (and cannot) be renamed.
4) Ability for an apphost to specify (through a simple runtimeconfig.json file) where the application files are located (e.g. app.dll, .runtimeconfig.json, deps.json, other dependencies). This works with self-contained as well, meaning all of the self-contained files including the runtime do not need to be located next to the exe anymore.

Fixes https://github.com/dotnet/core-setup/issues/3720

Additional documentation will be provided describing the new functionality and runtimeconfig.json format.

Note for Windows Wow64 support (x86 on 64-bit), the default location is `%ProgramFiles(x86)%` and the new environment variable is `%DOTNET_ROOT(x86)%`

Automated tests are limited until we get tooling support.

The apphost now uses the casablanca json parser for startupconfig.json. I may work on removing that dependency and using simple file I\O instead to save on exe size (details pending)

cc @wli3

### DIFF
--- a/src/corehost/cli/fxr/framework_info.cpp
+++ b/src/corehost/cli/fxr/framework_info.cpp
@@ -28,20 +28,6 @@ bool compare_by_name_and_version(const framework_info &a, const framework_info &
     const pal::string_t& fx_name,
     std::vector<framework_info>* framework_infos)
 {
-    // No FX resolution for standalone apps
-    if (mode == host_mode_t::standalone)
-    {
-        trace::verbose(_X("Standalone mode detected. Not gathering shared FX locations"));
-        return;
-    }
-
-    // No FX resolution for mixed apps
-    if (mode == host_mode_t::split_fx)
-    {
-        trace::verbose(_X("Split/FX mode detected. Not gathering shared FX locations"));
-        return;
-    }
-
     std::vector<pal::string_t> global_dirs;
     bool multilevel_lookup = multilevel_lookup_enabled();
 
--- a/src/corehost/cli/fxr/framework_info.cpp
+++ b/src/corehost/cli/fxr/framework_info.cpp
@@ -28,6 +28,13 @@ bool compare_by_name_and_version(const framework_info &a, const framework_info &
     const pal::string_t& fx_name,
     std::vector<framework_info>* framework_infos)
 {
+    // No FX resolution for mixed apps 
+    if (mode == host_mode_t::split_fx)
+    {
+        trace::verbose(_X("Split/FX mode detected. Not gathering shared FX locations"));
+        return;
+    }
+
     std::vector<pal::string_t> global_dirs;
     bool multilevel_lookup = multilevel_lookup_enabled();
 
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -10,6 +10,7 @@
 #include "fx_definition.h"
 #include "fx_muxer.h"
 #include "fx_ver.h"
+#include "host_startup_info.h"
 #include "libhost.h"
 #include "pal.h"
 #include "runtime_config.h"
@@ -26,7 +27,7 @@ void handle_missing_framework_error(
     const pal::string_t& fx_name,
     const pal::string_t& fx_version,
     const pal::string_t& fx_dir,
-    const pal::string_t& own_dir)
+    const pal::string_t& dotnet_root)
 {
     std::vector<framework_info> framework_infos;
     pal::string_t fx_ver_dirs;
@@ -37,10 +38,10 @@ void handle_missing_framework_error(
     }
     else
     {
-        fx_ver_dirs = own_dir;
+        fx_ver_dirs = dotnet_root;
     }
 
-    framework_info::get_all_framework_infos(mode, own_dir, fx_name, &framework_infos);
+    framework_info::get_all_framework_infos(mode, dotnet_root, fx_name, &framework_infos);
 
     // Display the error message about missing FX.
     if (fx_version.length())
@@ -239,13 +240,13 @@ bool resolve_hostpolicy_dir_from_probe_paths(const pal::string_t& version, const
 * Return name of deps file for app.
 */
 pal::string_t get_deps_file(
-    bool portable,
+    bool is_framework_dependent,
     const pal::string_t& app_candidate,
     const pal::string_t& specified_deps_file,
     const fx_definition_vector_t& fx_definitions
 )
 {
-    if (portable)
+    if (is_framework_dependent)
     {
         // The hostpolicy is resolved from the root framework's name and location.
         pal::string_t deps_file = get_root_framework(fx_definitions).get_dir();
@@ -258,7 +259,7 @@ pal::string_t get_deps_file(
     }
     else
     {
-        // Standalone app's hostpolicy is from specified deps or from app deps.
+        // Self-contained app's hostpolicy is from specified deps or from app deps.
         return !specified_deps_file.empty() ? specified_deps_file : get_deps_from_app_binary(app_candidate);
     }
 }
@@ -268,7 +269,7 @@ pal::string_t get_deps_file(
 */
 bool fx_muxer_t::resolve_hostpolicy_dir(
     host_mode_t mode,
-    const pal::string_t& own_dir,
+    const pal::string_t& dotnet_root,
     const fx_definition_vector_t& fx_definitions,
     const pal::string_t& app_candidate,
     const pal::string_t& specified_deps_file,
@@ -276,10 +277,10 @@ bool fx_muxer_t::resolve_hostpolicy_dir(
     const std::vector<pal::string_t>& probe_realpaths,
     pal::string_t* impl_dir)
 {
-    bool portable = get_app(fx_definitions).get_runtime_config().get_portable();
+    bool is_framework_dependent = get_app(fx_definitions).get_runtime_config().get_is_framework_dependent();
 
     // Obtain deps file for the given configuration.
-    pal::string_t resolved_deps = get_deps_file(portable, app_candidate, specified_deps_file, fx_definitions);
+    pal::string_t resolved_deps = get_deps_file(is_framework_dependent, app_candidate, specified_deps_file, fx_definitions);
 
     // Resolve hostpolicy version out of the deps file.
     pal::string_t version = resolve_hostpolicy_version_from_deps(resolved_deps);
@@ -297,7 +298,7 @@ bool fx_muxer_t::resolve_hostpolicy_dir(
 
     // Get the expected directory that would contain hostpolicy.
     pal::string_t expected;
-    if (portable)
+    if (is_framework_dependent)
     {
         // The hostpolicy is required to be in the root framework's location
         expected.assign(get_root_framework(fx_definitions).get_dir());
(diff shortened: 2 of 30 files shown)


src/corehost/common/pal.windows.cpp
```diff
@@ -198,16 +182,36 @@ bool pal::get_default_servicing_directory(string_t* recv)
...
     dirs->push_back(dir);
     return true;
 }
 
+bool pal::get_default_installation_dir(pal::string_t* recv)
+{
+    pal::char_t* program_files_dir;
+    if (pal::is_running_in_wow64())
+    {
+        program_files_dir = _X("ProgramFiles(x86)");
```
steveharter: This fixes an existing issue with multi-level lookup not working with x86 in wow mode (it would try to look in the x64's Program Files directory). The same wow check is used for DOTNET_ROOT(x86)

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -150,6 +151,77 @@ public void Running_Publish_Output_Standalone_EXE_By_Renaming_dotnet_exe_Fails()
...
+            Directory.CreateDirectory(Path.GetDirectoryName(appExe));
+            File.Move(sourceAppExePath, appExe);
+
+            // Create the startupConfig.json
+            string startupConfigFileName = Path.GetFileNameWithoutExtension(appExe) + ".startupconfig.json";
+            string startupConfigPath = Path.Combine(currentOutDir, startupConfigFileName);
+            SetStartupConfigJson(startupConfigPath, relativeNewPath);
+
+            // TODO: Use FS.Chmod when build utility project is converted to csproj.
+            // See https://github.com/NuGet/Home/issues/4424
```
eerhardt: This issue has been fixed for a long time.  Is this still necessary?

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
```diff
@@ -150,6 +151,77 @@ public void Running_Publish_Output_Standalone_EXE_By_Renaming_dotnet_exe_Fails()
...
+                .CaptureStdOut()
+                .Execute()
+                .Should()
+                .Pass()
+                .And
+                .HaveStdOutContaining("Hello World");
+        }
+
+        [Fact]
+        public void Running_Publish_Output_Standalone_EXE_With_Startupconfig_Succeeds()
```
eerhardt: Do you also want a test that sets the environment variable?

eerhardt:
> (fyi may need to re-target to the release branch)

That's not the process.  The process is get the code checked into master, and then port the change to the release branch.

src/corehost/common/pal.h
```diff
@@ -223,6 +226,8 @@ namespace pal
     bool load_library(const string_t* path, dll_t* dll);
     proc_t get_symbol(dll_t library, const char* name);
     void unload_library(dll_t library);
+
+    bool is_running_in_wow64();
```
eerhardt: I think this belongs under `#if WIN32`

src/corehost/cli/libhost.h
```diff
@@ -364,6 +383,14 @@ struct hostpolicy_init_t
             init->host_command = input->host_command;
         }
 
+        if (input->version_lo >= offsetof(host_interface_t, host_info_host_path) + sizeof(input->host_info_host_path))
+        {
+            init->host_info.host_path = input->host_info_host_path;
+            init->host_info.dotnet_root = input->host_info_dotnet_root;
+            init->host_info.app_path = input->host_info_app_path;
+            // For the backwards compat case, this will be later initialized with argv[0]
```
eerhardt: Do we have tests for the backwards compat at all?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -972,26 +970,27 @@ void append_probe_realpath(const pal::string_t& path, std::vector<pal::string_t>
...
 {
     std::vector<host_option> known_opts = { { _X("--additionalprobingpath"), _X("<path>"), _X("Path containing probing policy and assemblies to probe for") } };
-    if (get_all_options || exec_mode || mode == host_mode_t::split_fx || mode == host_mode_t::standalone)
+    if (get_all_options || exec_mode || mode == host_mode_t::split_fx || mode == host_mode_t::apphost)
     {
-        known_opts.push_back({ _X("--depsfile"), _X("<path>"),  _X("Path to <application>.deps.json file") });
-        known_opts.push_back({ _X("--runtimeconfig"), _X("<path>"), _X("Path to <application>.runtimeconfig.json file") });
+        known_opts.push_back({ _X("--depsfile"), _X("<path>"), _X("Path to <application>.deps.json file")});
+        known_opts.push_back({ _X("--runtimeconfig"), _X("<path>"), _X("Path to <application>.runtimeconfig.json file")});
+        known_opts.push_back({ _X("--dotnet-path"), _X("<path>"), _X("Path to the installed Shared Framework to use to run the application.")});
```
eerhardt: I'm not sure I understand this.  I thought we weren't doing this.

Also, isn't this too late by this time?  This is in hostfxr right?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -972,26 +970,27 @@ void append_probe_realpath(const pal::string_t& path, std::vector<pal::string_t>
...
     }
 
-    if (get_all_options || mode == host_mode_t::muxer)
+    if (get_all_options || mode == host_mode_t::muxer || mode == host_mode_t::apphost )
     {
-        known_opts.push_back({ _X("--fx-version"), _X("<version>"), _X("Version of the installed Shared Framework to use to run the application.") });
-        known_opts.push_back({ _X("--roll-forward-on-no-candidate-fx"), _X(""), _X("Roll forward on no candidate shared framework is enabled") });
-        known_opts.push_back({ _X("--additional-deps"), _X("<path>"), _X("Path to additonal deps.json file") });
+        // If mode=host_mode_t::apphost, these are only used when the app is framework-dependent.
```
eerhardt: How will this ever get into this code for the apphost?  If you invoke it with no args, it should run the app.  If you invoke it with anything like `--help`, it should pass `--help` on to the app.

I don't think these are possible for apphost.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1152,6 +1151,15 @@ int fx_muxer_t::read_config_and_execute(
     }
 
     auto app_config = app->get_runtime_config();
+    bool is_framework_dependent = app_config.get_is_framework_dependent();
+
+    // These settings are only valid for framework-dependent apps
+    if (is_framework_dependent)
+    {
+        fx_version_specified = get_last_known_arg(opts, opts_fx_version, _X(""));
```
eerhardt: Similar to above - I don't think it is possible for apphost to take ANY command line arguments.  The host arguments can only ever work with the `dotnet.exe` muxer.

src/corehost/corehost.cpp
```diff
@@ -65,44 +66,90 @@ bool is_exe_enabled_for_execution(const pal::string_t& own_path)
...
+    if (library_exists_in_dir(app_root, LIBFXR_NAME, out_fxr_path))
     {
-        trace::info(_X("Resolved fxr [%s]..."), fxr_path.c_str());
-        return fxr_path;
+        trace::info(_X("Resolved fxr [%s]..."), out_fxr_path->c_str());
+        out_dotnet_root->assign(app_root);
+        return true;
     }
 
-    trace::error(_X("A fatal error occurred, the required library %s could not be found at [%s]"), LIBFXR_NAME, own_dir.c_str());
```
eerhardt: Do we still write an error in this case?

src/corehost/common/pal.h
steveharter: I elected to have the declaration available globally

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
steveharter: I'll give it a try without

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: Yes I forgot to remove this. Thanks

src/corehost/cli/fxr/framework_info.cpp
```diff
@@ -28,20 +28,6 @@ bool compare_by_name_and_version(const framework_info &a, const framework_info &
...
 {
-    // No FX resolution for standalone apps
-    if (mode == host_mode_t::standalone)
-    {
-        trace::verbose(_X("Standalone mode detected. Not gathering shared FX locations"));
-        return;
-    }
-
-    // No FX resolution for mixed apps
-    if (mode == host_mode_t::split_fx)
```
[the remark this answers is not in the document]
steveharter: I added it back until we remove split_fx

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: The code just above this supports `--depsfile` and `--runtimeconfig` options for apphost. I just verified against 2.0.6 that they can be used with apphost. The changes here also add the other three valid options below (for framework-dependent apphost anyway)

src/corehost/cli/libhost.h
steveharter: Unfortunately, no. I previously created https://github.com/dotnet/core-setup/issues/3777 to track adding such tests

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
steveharter: I'll add one

src/corehost/cli/fxr/fx_muxer.cpp
eerhardt: What if the app wanted to take an argument named the same thing?  It would seem impossible to support.

It seems incorrect to process arguments at all in the apphost.

src/test/HostActivationTests/GivenThatICareAboutStandaloneAppActivation.cs
steveharter: I'll add a basic test for the env var.

Note the env var is only used for apphost + framework-dependent I've only done manual testing for that because there is no tooling for that yet and the workarounds at this point are ugly (publish for both self-contained and framework-dependent). 

So I will add a test adds a simple check to make sure a self-contained apphost tries to use DOTNET_ROOT, but the actual app launch fails due to other (file layout) reasons.

src/corehost/corehost.cpp
steveharter: The error would be alone the lines of "A fatal error occurred, the folder [%s] does not exist", which is not useful depending on the context. Adding a more detailed error.

src/corehost/cli/fxr/fx_muxer.cpp
steveharter: There is a chicken-egg situation currently where we parse the args before we determine if the app is self-contained or framework-dependent, so it's a bit of refactoring (if we want) to remove the 3 extra args from parsing when under apphost + self-contained.

steveharter: I don't plan on changing 2.0.x functionality; the original code does include standalone\apphost:
```
if (get_all_options || exec_mode || mode == host_mode_t::split_fx || mode == host_mode_t::standalone) 
{
    known_opts.push_back({ _X("--depsfile"), _X("<path>"),  _X("Path to <application>.deps.json file") }); 
    known_opts.push_back({ _X("--runtimeconfig"), _X("<path>"), _X("Path to <application>.runtimeconfig.json file") }); 
```

and this "feature" (if so) makes sense to also include the other 3 (which would only do anything in framework-dependent).

PR #3905, opened 2018-03-27
Add linux-arm64 to OfficialBuildRID

eerhardt:
We added an official build for linux-arm64, but we didn't update the OfficialBuildRID list.

/cc @janvorli @sdmaclea


Petermarcu:
Lgtm

PR #3920, opened 2018-03-28
Mark packages as stable for 2.0.7 & bump versions

wtgodbe:
CC @weshaggard @safern

### DIFF
--- a/BranchInfo.props
+++ b/BranchInfo.props
@@ -4,7 +4,7 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>7</PatchVersion>
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
-    <PreReleaseLabel></PreReleaseLabel>
+    <PreReleaseLabel>servicing</PreReleaseLabel>
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
     <ReleaseBrandSuffix></ReleaseBrandSuffix>
     <Channel>release/2.0.0</Channel>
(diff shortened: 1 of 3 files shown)


BranchInfo.props
```diff
@@ -3,8 +3,8 @@
     <MajorVersion>2</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>7</PatchVersion>
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
-    <PreReleaseLabel>servicing</PreReleaseLabel>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
+    <PreReleaseLabel></PreReleaseLabel>
```
safern: No need to change the PreReleaseLabel, this one should always be servicing, and it will be used only when StabilizePackageVersion == false.

wtgodbe: @safern fixed

wtgodbe:
@safern that was done by https://github.com/dotnet/core-setup/pull/3918

src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
```diff
@@ -16,13 +16,13 @@
       <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
-      <Version>1.0.10</Version>
+      <Version>1.0.11</Version>
```
safern: You'll need to update System.Console here as well for 1.0.0 and 1.1.0 to the corresponding serviced versions on each branch. Sorry I missed that but I realized we where servicing System.Console afterwards. Could you do that on a separate PR?

Issue #3910, opened 2018-03-27
Add tests to verify framework-dependent apphost


steveharter:
Placeholder to add tests once the SDK support building a framework-dependent apphost added via https://github.com/dotnet/core-setup/issues/3720

Tests previously added for self-contained apphost (since that is supported by tooling)
1) Ability to rename executable
2) Use of `<app>.startupconfig.json`
3) Negative usage of DOTNET_ROOT

Tests to include:
1) Basic roll-forward and probing tests, similar to what we already have
2) Ability to use command-line args (some or all of `--additionalprobingpath`  `--depsfile`, `--runtimeconfig`, `--fx-version`, `--roll-forward-on-no-candidate-fx`, `--additional-deps`)
3) Positive usage of `DOTNET_ROOT`

steveharter:
Update: Test #3 (Positive usage of DOTNET_ROOT) was added for https://github.com/dotnet/core-setup/commit/6ff3b8f91e2b5fb34e834bb64f421fdb50d6d72d but without the SDK support (manually creating the apphost)

PR #3922, opened 2018-03-28
Fix issue with apphost deps resolver

steveharter:
Fix issue introduced by https://github.com/dotnet/core-setup/pull/3888

The app root was getting set to the incorrect directory, causing deps resolution to fail to load additional assemblies.

I added this scenario to the test issue at https://github.com/dotnet/core-setup/issues/3910

cc @wli3

### DIFF
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -90,7 +90,7 @@ bool parse_arguments(
             trace::error(_X("Failed to locate managed application [%s]"), args.managed_application.c_str());
             return false;
         }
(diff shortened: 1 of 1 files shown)


eerhardt:
I really wish we would invest in tests for this functionality.

steveharter:
>> I really wish we would invest in tests for this functionality.

Yes. It is mostly timing. It's probably 2 days or so of "throw-away" test infrastructure work if we want to do that now, which would throw away once tooling supports this mode.

Also I believe this particular issue also shows a lack of coverage in the existing self-contained tests.

PR #3924, opened 2018-03-28
Add System.Console to netcoreapp.props files

wtgodbe:
@safern PTAL

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
+++ b/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
@@ -57,6 +57,9 @@
     <NETCoreApp11Dependency Include="System.ComponentModel.Annotations">
       <Version>4.3.0</Version>
     </NETCoreApp11Dependency>
+    <NETCoreApp10Dependency Include="System.Console">
+      <Version>4.3.1</Version>
+    </NETCoreApp10Dependency>
     <NETCoreApp11Dependency Include="System.Diagnostics.DiagnosticSource">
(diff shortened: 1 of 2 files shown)


src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
```diff
@@ -57,6 +57,9 @@
     <NETCoreApp11Dependency Include="System.ComponentModel.Annotations">
       <Version>4.3.0</Version>
     </NETCoreApp11Dependency>
+    <NETCoreApp10Dependency Include="System.Console">
```
safern: This should be NETCoreApp11Dependency

safern:
Also in 1.1 branch I don't see System.Console included in the NETCoreApp project.json file

https://github.com/dotnet/core-setup/blob/release/1.1.0/pkg/projects/Microsoft.NETCore.App/project.json

You will need to update that before servicing, if not the System.Console will not be picked up in the closure in NETCoreApp v1.1.8

safern:
> Also in 1.1 branch I don't see System.Console included in the NETCoreApp project.json file

I see you're doing this in: https://github.com/dotnet/core-setup/pull/3868/files#diff-d42418420888e3555f05ef5c16792861R20

Sorry about that.

src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
wtgodbe: My bad - copy/paste error. Fixed.

PR #3938, opened 2018-03-30
DependencyContextLoader should check Assembly.CodeBase

### LINKED ISSUE #3308: Consider using Assembly.CodeBase instead of Assembly.Location when looking for deps file
Original issue for context: https://github.com/aspnet/Mvc/issues/6613

With shadow copy enabled, `Assembly.Location` doesn't point to the path containing the deps file. Consequently `DependencyContextLoader` can no longer discover the deps file: https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.Extensions.DependencyModel/DependencyContextLoader.cs#L118. Using `Assembly.CodeBase` should reliably discover files in this case.

eerhardt:
When running with shadow copy on the .NET Framework, Assembly.Location can be different than Assembly.CodeBase.  We should check both in the DependencyContextLoader to support unit tests running with shadow copy.

Fix #3308

I also took the liberty to finally fix the .sln file.


src/managed/Microsoft.Extensions.DependencyModel/DependencyContextLoader.cs
```diff
@@ -126,6 +126,41 @@ private DependencyContext LoadAssemblyContext(Assembly assembly, IDependencyCont
...
+            return depsJsonFileExists ?
+                depsJsonFile :
+                null;
+        }
+
+        private static string GetNormalizedCodeBasePath(Assembly assembly)
+        {
+            try
+            {
+                return new Uri(assembly.CodeBase).LocalPath;
```
pranavkm: You could also do `if (Uri.TryCreate(assembly.CodeBase, UriKind.Absolute, out var uri) && uri.IsLocal) { return uri.LocalPath; }` to avoid the `try-catch`

eerhardt: Done.

PR #3956, opened 2018-04-02
Handle msbuild package breaking changes

weshaggard:
We need to move our task library to consume
netstandard2.0 and net46 as min targets now as
msbuild has reset their min frameworks to those.

@eerhardt mind taking a quick look please?


tools-local/tasks/core-setup.tasks.csproj
```diff
@@ -82,5 +82,7 @@
     <TargetingPackReference Include="Microsoft.Build.Framework" />
     <TargetingPackReference Include="Microsoft.Build.Utilities.v4.0" />
   </ItemGroup>
+
+  <Import Project="$(IntermediateOutputPath)/core-setup.tasks.csproj.nuget.g.targets" Condition="Exists('$(IntermediateOutputPath)/core-setup.tasks.csproj.nuget.g.targets')" />
```
eerhardt: ?? What's this needed for?

It would be really nice to use an SDK-based project here....

eerhardt:
I assume you will also port this to `master`?  (Which is the "normal" code flow)

tools-local/tasks/core-setup.tasks.csproj
weshaggard: This is need to get the netstandard refs. Over lunch I was thinking the same think you mentioned which is to turn this into a sdk project. Let me give that a go.

weshaggard:
>I assume you will also port this to master? (Which is the "normal" code flow)

Yes I will get changes into master and release/2.1. Right now I'm trying to get source-build working in release/2.1 branch which is why I targeted that.

eerhardt:
> <Project Sdk="Microsoft.NET.Sdk">

I like this :100:

---
Refers to: tools-local/tasks/core-setup.tasks.csproj:1 in f496353. [](commit_id = f496353c5202efb285799635dc6155ba6b22e5a2, deletion_comment = False)

weshaggard:
Now I just have to track down the missing dependencies. I was hoping to eliminate the special copying but maybe it is still needed. I will investigate.

RepoDirectories.props
```diff
@@ -23,7 +23,7 @@
     <BuildToolsTaskDesktopDir>$(ToolsDir)net46/</BuildToolsTaskDesktopDir>
     <BuildToolsTaskDir>$(BuildToolsTaskCoreDir)</BuildToolsTaskDir>
     <BuildToolsTaskDir Condition="'$(MSBuildRuntimeType)' != 'Core'">$(BuildToolsTaskDesktopDir)</BuildToolsTaskDir>
-    <LocalBuildToolsTaskDir>$(LocalToolRuntimePath)</LocalBuildToolsTaskDir>
+    <LocalBuildToolsTaskDir Condition="'$(MSBuildRuntimeType)' == 'Core'">$(LocalToolRuntimePath)netstandard20/</LocalBuildToolsTaskDir>
```
eerhardt: Is the condition necessary?  What if a new `MSBuildRuntimeType` is used?  This would be left empty.

weshaggard: I guess I can just default it.

Issue #3944, opened 2018-03-30
Remove startupconfig.json support, support relative embedded app path in apphost


steveharter:
As a follow-up to https://github.com/dotnet/core-setup/pull/3888, it was decided to remove the `startupconfig.json` support and instead support having the existing embedded dll name support a relative path instead. This makes it more trustworthy.

Currently, the dll name (which is embedded into the apphost at build time) can only be a dll name, so this will change that to allow a relative path plus the dll name.

The relative path is to use `/` for path delimiters, which will get converted to `\` on Windows.

@livarcocc @nguerrera @wli3

PR #3975, opened 2018-04-03
Remove startupconfig and add relative path support

steveharter:
Allow the embedded app dll to contain a relative path. A max of 1024 UTF-8 is still present.

Fixes https://github.com/dotnet/core-setup/issues/3944

cc @wli3


src/corehost/corehost.cpp
```diff
@@ -226,25 +185,38 @@ int run(const int argc, const pal::char_t* argv[])
...
+    if (pos_path_char != pal::string_t::npos)
+    {
+        requires_v2_hostfxr_interface = true;
+    }
+
+    app_path.assign(get_directory(host_path));
+    append_path(&app_path, embedded_app_name.c_str());
+    if (!pal::realpath(&app_path))
+    {
+        trace::error(_X("The application to execute does not exist: '%s'."), app_path.c_str());
```
steveharter: Review note: this is the same message that already exists in hostfxr; we now also check for a valid file in apphost since we need a valid path to look for a local hostfxr.dll.

src/corehost/corehost.cpp
```diff
@@ -271,7 +243,7 @@ int run(const int argc, const pal::char_t* argv[])
         return StatusCode::InvalidArgFailure;
     }
 
-    app_root.assign(host_path);
+    app_root.assign(get_directory(host_path));
```
steveharter: This fixes a prior but benign bug; the app_root is not used with muxer but we still try to calculate it for consistency.

eerhardt:
>     constexpr int EMBED_MAX = (EMBED_SZ > 1025 ? EMBED_SZ : 1025); // 1024 DLL name length, 1 NUL

Note that we only give 1k length for the DLL name.  But now this is a relative path.  Is this going to be long enough to meet all scenarios?

---
Refers to: src/corehost/corehost.cpp:42 in 772518d. [](commit_id = 772518dcd4fd86c651ebb5d2eb59169549b0448a, deletion_comment = False)

eerhardt:
>  * Detect if the apphost executable is allowed to load and execute a managed assembly.

All the below comments should be updated for the new approach, as it is no longer a "DLL filename".

---
Refers to: src/corehost/corehost.cpp:24 in 772518d. [](commit_id = 772518dcd4fd86c651ebb5d2eb59169549b0448a, deletion_comment = False)

eerhardt:
> bool is_exe_enabled_for_execution(const pal::string_t& host_path, pal::string_t* app_dll)

The `host_path` param isn't used in this method, is it?

---
Refers to: src/corehost/corehost.cpp:39 in 772518d. [](commit_id = 772518dcd4fd86c651ebb5d2eb59169549b0448a, deletion_comment = False)

wli3:
I tested with the private build. Both single dll project and multiple dll project can work

steveharter:
>> The host_path param isn't used in this method, is it?

No it wasn't. Removed. Thanks.. Why no compiler warning\error? :(

>> All the below comments should be updated for the new approach, as it is no longer a "DLL filename".

Done

>> I think my only real concern is 1k enough space for the relative path.

Yes this is a concern and was discussed with SDK team members. We could bump it to 2048 or so. However considering the path is optional, plus relative, it was generally excepted during review that 1024 should be fine. If we run into issues in the future, we could either change to 2048 (in a backwards-compatible manner) or add a new 2048 section. cc @nguerrera

PR #3979, opened 2018-04-04
Fix Linux Mint installer package name with correct version ( release/2.1)

rakeshsinghranchi:
Linux Mint installer packages have incorrect version mentioned.


rakeshsinghranchi:
@leecow , can you confirm if we need this for release/2.1.0 ? Created this PR in case we decide to use dedicated feed for Linux mint packages. From our earlier discussions, we want to use the matching Ubuntu feeds. Let me know if that plan still holds good.

leecow:
@rakeshsinghranchi - using the corresponding Ubuntu feeds is still the plan.

Issue #3980, opened 2018-04-04
Stop building mint packages


Petermarcu:
We don't need them. We should just use the Ubuntu packages. We already leverage the ubuntu feed for the mint packages anyways and they have the same dependencies.

Petermarcu:
@leecow

PR #3981, opened 2018-04-04
Stop building Linux Mint packages

rakeshsinghranchi:
Fixes https://github.com/dotnet/core-setup/issues/3980


Petermarcu:
Are there links and other things that need to be fixed up with this change? I think we link to them on the readme for example.

Petermarcu:
I don't think we need this in Preview2 but yes, we need to make sure it gets into release/2.1 right after that.

rakeshsinghranchi:
> Are there links and other things that need to be fixed up with this change? I think we link to them on the readme for example.
No, Linux Mint packages are not in Readme page. 
>I don't think we need this in Preview2 but yes, we need to make sure it gets into release/2.1 right after that.
Labelled this issue as post-ZBB for 2.1.0 milestone. @leecow  suggested it will be taken care of porting from master to release/2.1 branch as centralized effort for all post-ZBB fixes .

Petermarcu:
yeah, it should flow to release automatically. I see links to these packages for prodcon: https://github.com/dotnet/versions/tree/master/build-info/dotnet/product/cli/release/2.1

I don't know where those come from, we can probably just delete those rows.

rakeshsinghranchi:
Looking at the history for readme file,it looks like readme file is regenerated with every build.If it is auto-generated, assumption is that it is auto-deleted once removed from core-setup. 

 @dagood can confirm.

Petermarcu:
Yeah, I just made a PR to drop the mint rows. Dont worry about that one.

Petermarcu:
This is unrelated to the Tizen build so I'm going to merge.

PR #3983, opened 2018-04-04
Fix core-setup.task project for source-build

weshaggard:
Need to restore with the overridden restore sources so
restoring during batch restore.

Update to use convention for NugetPackage versions so
it can be overridden with versions.

cc @eerhardt

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -128,6 +128,7 @@
 
   <!-- list of projects to perform batch restore -->
   <ItemGroup>
+    <SdkRestoreProjects Include="$(MSBuildThisFileDirectory)tools-local/tasks/core-setup.tasks.csproj" />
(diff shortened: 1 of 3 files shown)


eerhardt:
Why are only 3 checks running in this branch?

dir.props
```diff
@@ -128,6 +128,7 @@
 
   <!-- list of projects to perform batch restore -->
   <ItemGroup>
+    <SdkRestoreProjects Include="$(MSBuildThisFileDirectory)tools-local/tasks/core-setup.tasks.csproj" />
```
eerhardt: (nit) below uses `\`, here uses `/`.

eerhardt:
I assume this will go into `master` as well.

Issue #4007, opened 2018-04-10
dotnet-host 2.1.0-preview2-26406-04-1 is not backwards compatible with 2.0.x applications: Failed to resolve library symbol hostfxr_main_startupinfo


natemcmaster:
I'm seeing multiple reports that 2.0.x runtimes no longer launch due to failures in loading libhostfxr.so

https://github.com/travis-ci/travis-ci/issues/9467
https://twitter.com/bradwilson/status/983782235777024000

**Repro**
On a clean machine, configure feeds as described on https://www.microsoft.com/net/download/linux-package-manager/ubuntu16-04/sdk-current
`sudo apt-get install dotnet-sdk-2.1.104`
`dotnet --info`

**Expected**
Stable releases of .NET Core should still work and only install stable versions of the `dotnet-host` package.

**Actual**
The installation brings down the prerelease `dotnet-host` package, version 2.1.0-preview2-26406-04-1, which issues this warning before exiting with and error code.
```
Failed to resolve library symbol hostfxr_main_startupinfo, error: /usr/share/dotnet/host/fxr/2.0.0/libhostfxr.so: undefined symbol: hostfxr_main_startupinfo
```

**Workaround**
Downgrade the `dotnet-host` package to an older version
```
sudo apt install dotnet-host=2.0.6-1
```

@steveharter @Petermarcu @leecow - preview2 impacting

cc @bradwilson @akoeplinger

pakrym:
Is caused by  [this ](https://github.com/dotnet/core-setup/blob/24ba120b0ca1ba60a24a79b7dc95866cd3e9f799/src/corehost/common/pal.unix.cpp#L85) check. On windows returning null is allowed and it's checked in the condition [here](https://github.com/dotnet/core-setup/blob/d6802d0c09f845cc26020691f155320a1fb37a67/src/corehost/corehost.cpp#L274) but on Linux PAL would log an error.

bording:
Installing `dotnet-sdk-2.1.104` also ends up installing the `dotnet-runtime-deps-2.1.0-preview2-26406-04` package.

I don't think you should be publishing preview packages on the same feed as the release packages.

steveharter:
So there are two issues here:
1) We are installing a preview version of the host when we shouldn't
2) There is a log message this is causing downstream issues. However, it is new to me that writing to stderr will cause `csc.exe` to fail since the actual process return value here should be success. Plus we write to stderr for informational cases when COREHOST_TRACE=1.

[...]

PR #4010, opened 2018-04-10
Make logging consistent across platforms for pal::get_symbol

steveharter:
Remove the `trace::error` under Linux\OSX when calling `pal::get_symbol` when no symbol was found. Windows does not log the error in this case and it is not necessarily an error when this occurs.

I did verify that the uses of get_symbol do verify against nullptr for the error cases.

Another change was to add explicit comparisons to `nullptr` although that is not causing any issues.

This is a partial fix for https://github.com/dotnet/core-setup/issues/4007. We still need to determine why preview bits of hostfxr are installed with released bits of SDK.

### DIFF
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -79,12 +79,7 @@ bool pal::load_library(const string_t* path, dll_t* dll)
 
 pal::proc_t pal::get_symbol(dll_t library, const char* name)
 {
-    auto result = dlsym(library, name);
-    if (result == nullptr)
-    {
-        trace::error(_X("Failed to resolve library symbol %s, error: %s"), name, dlerror());
-    }
-    return result;
+    return dlsym(library, name);
 }
 
 void pal::unload_library(dll_t library)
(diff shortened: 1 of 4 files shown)


src/corehost/common/pal.unix.cpp
```diff
@@ -79,12 +79,7 @@ bool pal::load_library(const string_t* path, dll_t* dll)
 
 pal::proc_t pal::get_symbol(dll_t library, const char* name)
 {
-    auto result = dlsym(library, name);
-    if (result == nullptr)
-    {
-        trace::error(_X("Failed to resolve library symbol %s, error: %s"), name, dlerror());
```
eerhardt: would it make sense to at least `trace::log` this?  Or whatever the 'verbose' logging is here.  That way it still is debuggable with COREHOST_TRACE.

steveharter: Yes. Added logging to both Windows and Linux\OSX.

The logging here is not necessary since each error cases does their own logging, although that logging is not as precise as doing it here, plus as you mentioned it may help debugging (esp non-error cases), so I added it.

PR #4014, opened 2018-04-11
Avoid logging to stderr when pal::get_symbol fails.

steveharter:
Port the fix from https://github.com/dotnet/core-setup/commit/9a776be820ff4ae83476c7e9e97514ecd8f3ae1d. Only one line from that fix is ported to release/2.1 (trace::error to trace::info) as that is all that is needed to fix the underlying issue.

For issue https://github.com/dotnet/core-setup/issues/4007

### DIFF
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -82,7 +82,7 @@ pal::proc_t pal::get_symbol(dll_t library, const char* name)
     auto result = dlsym(library, name);
     if (result == nullptr)
     {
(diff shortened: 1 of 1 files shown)


eerhardt:
Do we need this?  I thought `master` will be integrated into `release/2.1` one more time before the next release.  @weshaggard @Petermarcu - is that true?

steveharter:
@eerhardt there is a request to get this fixed before release

eerhardt:
I don't understand that request, but OK.

weshaggard:
>I don't understand that request, but OK.

We are going to spin one last build before the RC payload.

PR #4036, opened 2018-04-17
Add linux-musl build leg

weshaggard:
@eerhardt @janvorli 

I expect this to fail until we have a published coreclr/corefx build for linux-musl.


weshaggard:
>What do you think about adding linux-musl to the netci.groovy file? That way we can delete both the offiical build and CI legs for alpine in one shot, but we will still have the linux-musl CI leg.

I will just rename the alpine CI legs to linux-musl once we remove alpine.

PR #4024, opened 2018-04-14
[WIP]: Enable building new distros  ( opensuse42.3,fedora27,ubuntu18.04)

rakeshsinghranchi:

PR #4043, opened 2018-04-18
[1.1] Update distro support

ianhays:
- Resolves some errors for Ubuntu1804, Fedora27, and OpenSuse423
- Adds support for Debian9 and Fedora28
- Adds cmd arg to skip running the tests
- dotnet-host-build build.sh made to use its own tools restoration logic copied from corefx/coreclr instead of using the remote dotnet-install script from CLI
- Adds portable-linux .dotnetstage0 support for new distros
- Updates build_projects init-tools.sh to be the same as the one used in corefx and coreclr
- Update CoreFX and CoreCLR package versions across all projects
- Remove EnsureSuccessful stipulation from tar generation on linux. Tar can complete with an error code but still produce a correct package (warnings are treated as errors).
- Adds buildpipeline entries for Debian9 and Fedora28 as well as functional DockerFiles based on the ones in the microsoft/dotnetnetcore dockerhub archive
- Disable tests for all new distros to unblock official builds.

Continuation of https://github.com/dotnet/core-setup/pull/4024


ianhays:
Merging to unblock pipebuild full testing.

Issue #4055, opened 2018-04-20
Move to netcoreapp2.2 TFM 


weshaggard:
We have moved the package/installer versions to 2.2 in master but we haven't moved to netcoreapp2.2 TFM yet this issue is tracking that work.

PR #4060, opened 2018-04-23
Move to netcoreapp2.2 as TFM

weshaggard:
cc @ericstj @eerhardt 

fixes https://github.com/dotnet/core-setup/issues/4055

### DIFF
--- a/src/test/Assets/TestProjects/HostApiInvokerApp/HostApiInvokerApp.csproj
+++ b/src/test/Assets/TestProjects/HostApiInvokerApp/HostApiInvokerApp.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>$(NETCoreAppFramework)</TargetFramework>
     <OutputType>Exe</OutputType>
     <RuntimeFrameworkVersion>$(MNAVersion)</RuntimeFrameworkVersion>
   </PropertyGroup>
--- a/src/test/TestUtils/TestProjectFixture.cs
+++ b/src/test/TestUtils/TestProjectFixture.cs
@@ -53,11 +53,12 @@ namespace Microsoft.DotNet.CoreSetup.Test
             string dotnetInstallPath = null,
             string currentRid = null,
             string builtDotnetOutputPath = null,
-            string framework = "netcoreapp2.1")
+            string framework = null)
         {
             ValidateRequiredDirectories(repoDirectoriesProvider);
 
             _testProjectName = testProjectName;
+            _framework = framework ?? Environment.GetEnvironmentVariable("MNA_TFM");
 
             _exeExtension = exeExtension ?? RuntimeInformationExtensions.GetExeExtensionForCurrentOSPlatform();
             _sharedLibraryExtension = sharedLibraryExtension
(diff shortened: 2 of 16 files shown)


src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -53,11 +53,13 @@ public class TestProjectFixture
...
             string currentRid = null,
             string builtDotnetOutputPath = null,
-            string framework = "netcoreapp2.1")
+            string framework = null)
         {
             ValidateRequiredDirectories(repoDirectoriesProvider);
 
             _testProjectName = testProjectName;
+            _framework = framework ?? Environment.GetEnvironmentVariable("MNA_TFM");
+            Console.WriteLine($"TPF: MNA_TFM={_framework}");
```
eerhardt: Do you intend this to be written all the time?  Or should it be removed before check in?

src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -297,6 +298,7 @@ private void ValidateRequiredDirectories(RepoDirectoriesProvider repoDirectories
             }
 
             storeArgs.Add($"/p:MNAVersion={_repoDirectoriesProvider.MicrosoftNETCoreAppVersion}");
+            storeArgs.Add($"/p:NETCoreAppFramework={_framework}");
```
ericstj: I think you end up passing this in twice, here and at line 280.

src/test/TestUtils/TestProjectFixture.cs
weshaggard: Yeah that was for debugging I will remove.

src/test/TestUtils/TestProjectFixture.cs
weshaggard: I believe I still need to set the property so that it is correctly set in the msbuild property, but perhaps not I didn't verify that directly I just ensured we passed it everywhere we pass MNAVersion.

src/test/Assets/TestProjects/HostApiInvokerApp/HostApiInvokerApp.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>$(NETCoreAppFramework)</TargetFramework>
```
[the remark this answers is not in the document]
weshaggard: Yes originally I thought they imported but later found out they didn't. I still kind of like keeping set here so we can easily find all the projects that we expect to be set to this value in the repo.

natemcmaster:
We tried to update aspnet to this, but all netcoreapp2.1 projects break. Is this expected? It looks to me like the dependencies of the package aren't defined correctly. Is this just another case where you're not supposed to use Microsoft.NETCore.App 2.x with netcoreapp2.(x-1)? (even though NuGet allows it)

https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App/2.2.0-preview1-26425-05

image

weshaggard:
It is expected that the 2.2 package only supported netcoreapp2.2. If you still want to target netcoreapp2.1 you need to stay with the 2.1 version of the package. This was the same in 2.1 although we did put a place older dependency that has nothing, as you can see in the dependency list there. We could do that here for 2.1 if it was interesting but it would still not be functional.

natemcmaster:
Why not just remove the .NETCoreApp2.0 TFM altogether? (and 1.0 and 1.1 for that matter) NuGet would correctly say "sorry, you can't do that" when trying to install the 2.2 package for netcoreapp2.1.

weshaggard:
I actually planned to remove the 1.0/1.1 dependencies but I've not gotten to it yet. I will also remove the 2.0 dependencies. I just filed https://github.com/dotnet/core-setup/issues/4071 against myself so I don't forget.

natemcmaster:
Awesome, thanks. We'll have to start the work to switch our projects to netcoreapp2.2 then. In the past, we've had issues with VS and dotnet-CLI when new TFMs are added, so it may take longer than a normal core-setup update.

PR #4062, opened 2018-04-24
Remove Alpine 3.6 builds

weshaggard:
Remove alpine3.6 builds not that we have linux-musl to replace
them more generically.

Rename the alpine CI legs to be linux-musl instead.

cc @eerhardt @janvorli


init-tools.sh
```diff
@@ -93,14 +93,12 @@ if [ ! -e $__DOTNET_PATH ]; then
                 if [ -e /etc/os-release ]; then
                     source /etc/os-release
                     if [[ $ID == "alpine" ]]; then
-                        # remove the last version digit
-                        VERSION_ID=${VERSION_ID%.*}
-                    __PKG_RID=alpine.$VERSION_ID
+                        __PKG_RID=linux-musl.$VERSION_ID
```
janvorli: The $VERSION_ID should not be here.

weshaggard:
This is for master right now, but it is unclear to me how asp.net pulls these in so it may just be fine. @natemcmaster do you know what you guys depend upon from the runtime/cli side when it comes to alpine? Do you guys just resolve via nuget or depend on a particular name being present in the blob storage?

init-tools.sh
weshaggard: good call. I will fix.

natemcmaster:
aspnet is downloading dotnet-runtime-$ver-alpine.3.6-x64.tar.gz from the blob store. We basically plan to follow your lead here, so if you pull the alpine 3.6 builds, we will too.

Petermarcu:
Just want to avoid build breaks so I would be good to be coordinated.

natemcmaster:
Thanks. I'll prep a PR to aspnet to remove alpine 3.6. Will this also get ported to 2.1 for RTM?

weshaggard:
>Will this also get ported to 2.1 for RTM?

Yes the plan is to port this to 2.1 for RTM.

weshaggard:
test ci please

natemcmaster:
aspnet is ready to drop this too https://github.com/aspnet/Universe/pull/1107. Let me know when this makes it back to 2.1 branches.

init-tools.sh
```diff
@@ -93,14 +93,12 @@ if [ ! -e $__DOTNET_PATH ]; then
                 if [ -e /etc/os-release ]; then
                     source /etc/os-release
                     if [[ $ID == "alpine" ]]; then
-                        # remove the last version digit
-                        VERSION_ID=${VERSION_ID%.*}
-                    __PKG_RID=alpine.$VERSION_ID
+                        __PKG_RID=linux-musl
```
eerhardt: We are going to have to bump the CLI version right?  We don't currently have a linux-musl CLI for `v2.1.2`.  Or did you make a "bootstrap" CLI for it?

src/sharedFramework/sharedFramework.proj
```diff
@@ -123,8 +123,7 @@
       <SharedFrameworkDepsFile>$(SharedFrameworkNameAndVersionRoot)\$(SharedFrameworkName).deps.json</SharedFrameworkDepsFile>
       <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='Windows_NT'">win</RuntimeGraphGeneratorRuntime>
       <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='OSX'">osx</RuntimeGraphGeneratorRuntime>
-      <RuntimeGraphGeneratorRuntime Condition="$(OutputRID.StartsWith('alpine'))">alpine</RuntimeGraphGeneratorRuntime>
-      <RuntimeGraphGeneratorRuntime Condition="'$(RuntimeGraphGeneratorRuntime)'==''">linux</RuntimeGraphGeneratorRuntime>      
+      <RuntimeGraphGeneratorRuntime Condition="'$(RuntimeGraphGeneratorRuntime)'==''">linux</RuntimeGraphGeneratorRuntime>
```
eerhardt: Do we really want the whole `linux` runtime graph in the shared fx .deps.json file?  It probably won't cause issues, but it is bloat/unnecessary.

The way this works is that it reads all the `runtime.json` files in all the packages, and scans all the RIDs looking for RIDs that import this specified RID.  So in this case, all the RIDs that import `linux` will get put in the .deps.json file, like `ubuntu`, `rhel`, etc.  However, since we are building for `linux-musl`, it seems like we should only be including RIDs that import `linux-musl`.

weshaggard: I looked at this and personally felt it was probably OK to just include the entire runtime graph. Is there any harm beyond a size?

init-tools.sh
weshaggard: I've not yet but I plan to.

natemcmaster:
By the way, I think we'll need to change dotnet/cli too. I was looking at the dotnet/cli build to see if I needed to update it to pull aspnet's linux-musl-x64 tar, and it appears dotnet/cli is still currently pulling dotnet-runtime-$ver-alpine.3.6-x64.tar.

src/sharedFramework/sharedFramework.proj
eerhardt: I don't *think* there is any harm, but following that logic we should put the whole graph in all the time - put Windows RIDs in for Linux, Linux in for Windows, etc.

init-tools.sh
eerhardt: Which do you plan to do?  I assume bump the CLI version....

weshaggard: No what I did was copy the alpine cli in the 2.1.2 and named in linux-musl.

weshaggard:
test ci please

weshaggard:
>By the way, I think we'll need to change dotnet/cli too. I was looking at the dotnet/cli build to see if I needed to update it to pull aspnet's linux-musl-x64 tar, and it appears dotnet/cli is still currently pulling dotnet-runtime-$ver-alpine.3.6-x64.tar.

the master cli is still building alpine and not linux-musl yet. At any rate if needed we can temporary continue to pull the alpine sdk until linux-musl is fully plumbed through the system.

Issue #4047, opened 2018-04-19
Host behavior on roll-forward when app-local assembly + file version match framework should be to prefer the shared framework.


ericstj:
When rolling an application forward to a new shared framework and examining versions of assemblies we need to prefer the framework over the app if all other things are equal.

This will make things consistent with the way that 1.x behaved when you referenced a package of the same exact version (prefer the shared framework) as well as 2.x behavior: where conflict resolution prefers the shared framework when all else is the same.

The shared framework copy of the dll is *better* since its been cross-gen'ed and is shareable between multiple apps running on the shared framework.

ahsonkhan:
Is this blocking release just like https://github.com/dotnet/corefx/issues/29218? Should we tag it as such?

cc @karelz

karelz:
@ahsonkhan sure, the label won't hurt, although I don't think anyone is paying attention to it just yet.

cc @steveharter

steveharter:
This is really a performance issue, should not be "blocking release" IMO. Not to say we shouldn't fix it for 2.1, however.

This requires a tweak to the existing code, as the default behavior is app wins with everything else being equal.

karelz:
BTW: blocking-release != blocking-RC1
It just means must-have for 2.1 RTW

steveharter:
Note that this issue is essentially "rule # 2" in the original design\proposal at https://github.com/dotnet/core-setup/issues/3546:

_2. If a [major] or [minor] roll-forward occurs for a given framework, that framework wins over any other higher layer(s)_

but was not implemented as it was deemed not necessary at the time because we were implementing rule # 3 which uses the assembly\file version metadata in the deps.json.

ericstj:
No this is not that rule.  I'm not saying that the framework always wins.  I'm further specifying the versioning rule such that it matches the behavior of the in-build conflict resolution.  Simply: If the versions are equal on a roll-forward, framework should win.

steveharter:
The matrix of metadata vs. non-metadata, assuming minor\major roll-forward:
a) No metadata on either side. Currently app\higher layer wins
b) Metadata on one side. Currently the side with metadata wins
c) Metadata on both sides. Currently the app\higher layer wins in a tie (assemblyVersion and fileVersion are the same)

For (c) we favor the lower layer in a tie.

For (a) do we want to do also favor a lower layer in a tie? Note that this will be unlikely as we should have metadata for 2.1+

Petermarcu:
Is this critical for RC?

[...]

PR #4076, opened 2018-04-26
On minor roll-forward framework wins over app when same versions in deps.json

steveharter:
In order to prevent issues on roll-forward with loading an incompatible (older) assembly from the app, we now select the framework's (newer) assembly on minor\major roll forward.

Fixes https://github.com/dotnet/core-setup/issues/4047

This will be ported to 2.1 release pending shiproom approval.

### DIFF
--- a/src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
@@ -1049,9 +1049,6 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.MultilevelSharedFxLooku
             string appDepsJson = Path.Combine(fixture.TestProject.OutputDirectory, "SharedFxLookupPortableApp.deps.json");
             AddImmutableAssemblyToDepsJson(appDepsJson);
 
-            // Expected: the uber framework's version of System.Collections.Immutable is used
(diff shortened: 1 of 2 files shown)


src/test/HostActivationTests/GivenThatICareAboutMultilevelSharedFxLookup.cs
```diff
@@ -1023,6 +1023,58 @@ public void Additional_Deps_Lightup_Folder_With_Roll_Forward_And_Bad_JsonFile()
...
+            string fxAssemblyPath = Path.Combine(_exeSharedFxBaseDir, "9999.0.0", "System.Collections.Immutable.dll");
+            string appAssembly = Path.Combine(fixture.TestProject.OutputDirectory, "System.Collections.Immutable.dll");
+            File.Copy(fxAssemblyPath, appAssembly);
+
+            // Modify the app's deps.json to add System.Collections.Immmutable
+            string appDepsJson = Path.Combine(fixture.TestProject.OutputDirectory, "SharedFxLookupPortableApp.deps.json");
+            AddImmutableAssemblyToDepsJson(appDepsJson);
+
+            // Expected: the uber framework's version of System.Collections.Immutable is used
+            string additionalDepsAssemblyPath = Path.Combine(fixture.TestProject.OutputDirectory, "System.Collections.Immutable");
```
eerhardt: Is the `additionalDepsAssemblyPath` variable used?  It doesn't appear to be.

eerhardt:
The Alpine legs are supposed to be turned off, but for some reason they are still running.  You can ignore that leg.

Issue #3889, opened 2018-03-23
The Dependency resolver should select the most compatible [Patch] version instead of only the exact 'requested' version


clguimanMSFT:
## Steps to reproduce
Starting with netcore 2.1 runtime the logic to find deps files has changed. The runtime will only search for the deps file in a folder matching with the app's requested framework version. Previously it would just use whatever folder it found.

The changed was done in this PR: https://github.com/dotnet/core-setup/pull/3743/files 

## Expected  behavior

The runtime should still look in the requested version folder, but if it can't find it, it should fallback and search in the <fx major version>.<fx minor version> folder.

This is useful for scenarios where assemblies are loaded at runtime from a previously deployed folder structure (without having control over the app's targeted framework version). With the new behavior, users will need to deploy the new folder structure each time a new patch version is released. My proposal would limit the folder structure release only when a minor framework version is released.

## Environment data

steveharter:
That 2.1 PR mentioned above reverted behavior to the 2.0 behavior which is to use the "requested" version, not the "found" version. So this issue existed in 2.0 as well AFAIK

clguimanMSFT:
I don't know what the intended behavior was in 2.0, but it was enough to just have one folder ("2.0") and this one was always picked up when searching for the deps file. It's possible some other logic that decides what fx version to use might have changed in the 2.1 runtime

steveharter:
Perhaps a 2.0 runtimeconfig.json specified "2.0", and not "2.0.0" for the framework version? I'll check when I have time.

clguimanMSFT:
this is the app's runtimeconfig.json:
`{
  "runtimeOptions": {
    "tfm": "netcoreapp2.0",
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "2.0.0"
    },
    "configProperties": {
      "System.GC.Server": true
    }
  }
}`

steveharter:
Whether it's the same or not I suppose is not a concern, because we want a "reverse roll forward"-like approach which is different than "requested" vs. "found". Based on the offline discussion today, we'll need to clarify the requirements.

karelz:
The is late design change request, which needs solid design first, adding Post-ZBB

steveharter:
Adding link to [design doc](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/additional-deps.md) which is still a work-in-progress

karelz:
Fixed in release/2.1 in PR #4094.

Issue #3884, opened 2018-03-22
additionalDeps downgrades shared runtime library


pakrym:
## Steps to reproduce
1. `dotnet new web`
2. `mkdir deps`
3. set `yourdir\deps\shared\Microsoft.NETCore.App\2.1.0-preview2-26308-01\a.deps.json` (modify runtime framework version to be the one referenced in `.dotnet\x64\shared\Microsoft.AspNetCore.App\2.1.0-preview2-xxxxx\Microsoft.AspNetCore.App.runtimeconfig.json` app references)
```
{
  "runtimeTarget": {
    "name": ".NETCoreApp,Version=v2.1.0",
    "signature": "5e814c43a4bb318be73c818ca300cacedf418e91"
  },
  "compilationOptions": {},
  "targets": {
    ".NETCoreApp,Version=v2.1.0": {
      "Microsoft.AspNetCore.Hosting.Abstractions/2.0.0": {
        "dependencies": {
          "Microsoft.AspNetCore.Hosting.Server.Abstractions": "2.0.0",
          "Microsoft.AspNetCore.Http.Abstractions": "2.0.0",
          "Microsoft.Extensions.Configuration.Abstractions": "2.0.0",
          "Microsoft.Extensions.DependencyInjection.Abstractions": "2.0.0",
          "Microsoft.Extensions.FileProviders.Abstractions": "2.0.0",
          "Microsoft.Extensions.Hosting.Abstractions": "2.0.0",
          "Microsoft.Extensions.Logging.Abstractions": "2.0.0"
        },
        "runtime": {
          "lib/netstandard2.0/Microsoft.AspNetCore.Hosting.Abstractions.dll": {}
        }
      }
    }
  },
  "libraries": {
    "Microsoft.AspNetCore.Hosting.Abstractions/2.0.0": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-836NSTzCM1r1tgCRw92y2/TOKVI8HC0ahh+SDer4oMNwuk7LYgSayjQ3rYHDpQDAaGCMF8vFjFphhRECc9Xtog==",
      "path": "microsoft.aspnetcore.hosting.abstractions/2.0.0",
      "hashPath": "microsoft.aspnetcore.hosting.abstractions.2.0.0.nupkg.sha512"
    }
  }
}

```
4. Set `$env:DOTNET_ADDITIONAL_DEPS="yourdir\deps"`
5. dotnet run

## Expected  behavior
[...]

PR #4084, opened 2018-04-26
Add "no downgrade" and "rollback" features to additional deps

steveharter:
This addresses 2 additional-deps features that were reviewed prior. These will also go into 2.1 pending shiproom approval:
https://github.com/dotnet/core-setup/issues/3884
https://github.com/dotnet/core-setup/issues/3889

One feature is “no downgrade”, here’s some detail:

Currently we process deps.json entries in this order today: app, additional-deps, frameworks
    - Where first-in wins
    - Thus the additional-deps “downgrades” the framework on collision (except for major\minor roll forward – see below)

Proposal is we change to this order: app, frameworks, additional-deps
    - Plus we allow additional-deps to upgrade the previous app\framework whether or not there was any roll-forward (based on version numbers in the deps.json files)

The other feature is "roll-backwards. Here we limit to patch version only (no major/minor roll-backward)

cc @pakrym, @clguimanMSFT, @karelz

### DIFF
--- a/src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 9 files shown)


src/test/HostActivationTests/GivenThatICareAboutLightupAppActivation.cs
```diff
@@ -128,6 +128,82 @@ public void Muxer_activation_of_LightupApp_WithLightupLib_Succeeds()
                 .HaveStdOutContaining("Hello LightupClient");
         }
 
+        // Success case with roll-backwards support
+        [Fact]
+        public void Muxer_activation_of_LightupApp_WithLightupLib_and_Roll_Backwards_Succeeds()
```
pakrym: Test for prerelease to release rollback

pakrym: Test that we don't roll over minor.

steveharter: Sure. I will add that, however I'll need to refactor the "shared fx" test code (from `GivenThatICareAboutMultilevelSharedFxLookup.cs`) into a common class so I can deploy and target a specific preview\non-preview framework version (instead of the current one like the current test).

steveharter:
@pakrym I added the two tests you asked for.

Issue #4081, opened 2018-04-26
RuntimeEnvironment.RuntimeArchitecture on ARM64 says "x64" instead of "arm64"


steveharter:
The current code assumes if running in 64-bit environment, then it is "x64".

See https://github.com/dotnet/cli/issues/9118

At a minimum this displays incorrect information via `dotnet --info`

wli3:
https://github.com/dotnet/core-setup/issues/4083

Petermarcu:
We need to get this fixed. Its important that information and data coming from these machines is accurate. It will greatly assist long term diagnosability.

Petermarcu:
https://github.com/dotnet/core-setup/blob/9dff3ce12574699e6d4d31722047f775d92d05fd/src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs#L26

Petermarcu:
This is fixed in master and 2.1 now.

PR #4085, opened 2018-04-26
Updating arch support for arm

Petermarcu:
This adds support for arm and arm64 rids to address https://github.com/dotnet/core-setup/issues/4081

### DIFF
--- a/src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
+++ b/src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
@@ -24,7 +24,7 @@ namespace Microsoft.DotNet.PlatformAbstractions
 #if NET45
             return Environment.Is64BitProcess ? "x64" : "x86";
 #else
-            return RuntimeInformation.OSArchitecture.ToString().ToLowerInvariant();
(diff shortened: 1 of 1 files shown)


Petermarcu:
I haven't been able to build and test locally on different arch's so marking this a no merge.

src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs
```diff
@@ -36,11 +37,7 @@ public static string GetRuntimeIdentifier()
 
         private static string GetRIDArch()
         {
-            if (!string.IsNullOrEmpty(RuntimeArchitecture))
-            {
-                return $"-{RuntimeArchitecture.ToLowerInvariant()}";
-            }
-            return string.Empty;
+            return $"-{RuntimeArchitecture}";
```
kasper3: nit: based on https://github.com/dotnet/corefx/pull/28578#discussion_r177939345, `'-' + RuntimeArchitecture` or `string.Concat("-", RuntimeArchitecture)` would be slightly cheaper here

Petermarcu:
Finished verifying this on my Pi and confirmed we now have the correct behavior.

Issue #4071, opened 2018-04-25
Remove all but latest TFM from Microsoft.NETCore.App dependencies


weshaggard:
After https://github.com/dotnet/core-setup/pull/4060 moved to netcoreapp2.2 we should make it explicit that we no longer support the older TFMs with the latest package so we should remove netcoreapp1.0/1.1/2.0 dependencies from the package.

ericstj:
We missed https://github.com/dotnet/core-setup/blob/7362f3bc1dca0ad537871f4f33e4bfce66c9ed19/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj#L33-L37

weshaggard:
I agree we should remove that as well. Out of curiosity how did you find this? Did it cause some issue somewhere?

ericstj:
@livarcocc was updating the SDK and compile was failing.  They hadn’t retargeted.

weshaggard:
Fixed the last issue with #4167

PR #4089, opened 2018-04-27
Remove older TFMs from Microsoft.NETCore.App

weshaggard:
Microsoft.NETCore.App will only support the target framework
that matches its major/minor version after this change.

Example: 2.2.x will only support netcoreapp2.2

Fixes https://github.com/dotnet/core-setup/issues/4071

@ericstj PTAL


ericstj:
@weshaggard we missed this guy:
https://github.com/dotnet/core-setup/blob/7362f3bc1dca0ad537871f4f33e4bfce66c9ed19/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj#L33-L37

/cc @livarcocc

Issue #4053, opened 2018-04-20
ThirdPartyNotices.txt in installer/packages is missing entries


RussKeldorph:
_From @gregg-miskelly on April 20, 2018 0:38_

Repro steps:
1. Download https://www.nuget.org/packages/runtime.linux-x64.Microsoft.NETCore.App/2.1.0-preview2-26406-04
2. Look at THIRD-PARTY-NOTICES.TXT

Expected:
This should call out all of CoreCLR's Third Party Notices. An example missing one is libunwind8. I believe the complete list can be found at https://github.com/dotnet/coreclr/blob/release/2.1/THIRD-PARTY-NOTICES.TXT

Actual:
[THIRD-PARTY-NOTICES.TXT](https://github.com/dotnet/coreclr/files/1930160/THIRD-PARTY-NOTICES.TXT)

There is a similar problem when installing the .NET SDK (and maybe runtime?) - there is a ThirdPartyNotices.txt (laid down to /usr/share/dotnet/ThirdPartyNotices.txt and c:\Program Files\dotnet\ThirdPartyNotices.txt), but this is also missing many items from this repo's Third-Party-Notices, including libunwind8 again.

_Copied from original issue: dotnet/coreclr#17690_

RussKeldorph:
_From @danmosemsft on April 20, 2018 4:46_

@RussKeldorph seems like 2.1 fix is needed

RussKeldorph:
@danmosemsft Sounds like you understand this better than I do.  Can you give more details on exactly what needs to happen?  Does coreclr just need to add its THIRD-PARTY-NOTICES.TXT to its package(s) [all of them? which path?] and upstack can take it from there?

I'm sure I could figure it out eventually, but given the urgency, if someone knows exactly what needs to happen, I'd appreciate a hint.

@eerhardt @weshaggard @janvorli

RussKeldorph:
_From @eerhardt on April 20, 2018 14:19_

There looks to be a few things wrong here.

We also aren't keeping the corefx third-party-notices up to date.  It appears the way this works is that there is a duplicated license file in core-setup that contains all the 3rd party notices from all 3 core repos.  However, it hasn't been updated in almost 2 years.  See https://github.com/dotnet/core-setup/blob/master/resources/ThirdPartyNotices.txt.  This is the file that is included in the installers and tarballs.

[...]

PR #4090, opened 2018-04-27
Update 3PN merging in coreclr and corefx notices

weshaggard:
Fixes https://github.com/dotnet/core-setup/issues/4053

PTAL @leecow @richlander


weshaggard:
I did miss that. I will update the installers to use the same file as the packages which is the one in the root of the repo.

PR #4020, opened 2018-04-12
Enable building upgradeable linux packages with updated runtime-deps dependency moved from host to runtime package

rakeshsinghranchi:
This enables producing upgradeable major.minor installers for Linux. This enables acquiring latest major.minor.release packages using Linux package managers. 

For 2.1 release , these are the package names: 
**dotnet-host**
**dotnet-runtime-deps-2.1**
**dotnet-hostfxr-2.1**
       --depends on dotnet-host (>= major.minor.patch-revision)
**dotnet-runtime-2.1**
    --depends on dotnet-hostfxr-2.1 (>= major.minor.patch-revision)
    -- depends on dotnet-runtime-deps-2.1 (>= major.minor.patch-revision)

This PR takes care of the following : 
1. Updating runtime-deps dependency from dotnet-host to dotnet-runtime package
2. Produce upgradeable Debs and RPMs for Runtime. 
3. Build will continue to produce the old style packages for now. Switching to major upgradeable package will be done later. @leecow  has more details

PR #4104, opened 2018-04-30
[release/2.1] Build upgradeable linux packages port

weshaggard:
Port of #4020 

cc @rakeshsinghranchi @Petermarcu @leecow


Petermarcu:
Wes, do you think we are good on this one with that CI failure given that its likely related to removing those packages as part of the musl move?

weshaggard:
test Linux-musl x64 Debug Build

The linux-musl leg should pass so we should understand that before merging.

weshaggard:
test Linux-musl x64 Debug Build

Issue #3852, opened 2018-03-19
`Error resolving full path []` when running dotnet


MichaelSimons:
## Steps to reproduce
1.  Get a Nano Server machine
2.  Run `dotnet --verion` or any other cmd

**Alternate Repro**
1.  From a windows 10 machine running in Windows Container mode run `docker run --rm microsoft/dotnet-nightly:2.1-sdk dotnet help`

## Expected  behavior
dotnet help should be displayed

## Actual behavior
Before the dotnet cmd is run the following is displayed:

```
C:\>dotnet --version
Error resolving full path []
Error resolving full path []
2.1.300-preview2-008354
```

## Environment data
Using dotnet cli 2.1.300-preview2-008354

**Note:**
1. This is a regression from preview1.
1. This regression has happened before - see https://github.com/dotnet/core-setup/issues/1507.  Please run your tests on Nano Server.  It is an important supported platform that has a number of differences.

steveharter:
This appears to be an issue with the first couple lines of `stderr` sometimes writing to `stdout`. I haven't noticed any changes in core-setup related to buffering, etc.

>> This is a regression from preview1.

Is this issue a proven regression from netcore (with the same docker build), or could it have been a newer docker runtime\bug?

steveharter:
Update: I ran against 2.0.6 and got similar results, although the exact "Error resolving full path" was not shown

Do this:
```
set COREHOST_TRACE=1
dotnet
```

Should result in the first two lines being
```
Tracing enabled
--- Invoked dotnet [version: 2.0.6, commit hash: 74b1c703813c8910df5b96f304b0f2b78cdf194d] main = {
dotnet
}
```
and then the expected help.

This is the same repro as "Error resolving full path", as it shows the first two lines being sent to `stdout` instead of `stderr`.

At this point it appears either to be a Docker issue with stdout\stderr or an issue with needing an extra flush for stdout\stderr

steveharter:
I believe I know what changed.

The call to output "Error resolving full path" is coming from a background worker thread that writes "breadcrumbs" for servicing. That background thread does a file_exists() that now calls into a common method to implement that, which will write to stderr in certain cases. Previously file_exists() did not call that common method and thus did not write to stderr.

The short-term fix is to change file_exists so it won't write to stderr.

The long-term fix is to determine why a different thread running in Docker has issues with stderr\stdout.

richlander:
This is still an issue ...

```console
[...]

PR #4110, opened 2018-05-01
Avoid logging to stderr if core_servicing location is empty

steveharter:
Addresses https://github.com/dotnet/core-setup/issues/3852 by not logging to stderr when the call to realpath() fails when resolving the servicing location. 

The issue is the servicing location "\Program Files (x86)" does not exist, and the call to realpath() fails. Note if that directory is manually created, the error also goes away.

Verified by reproducing the original problem, and then replacing the hostpolicy.dll in the Docker container with the "fixed" version. Then the "Error resolving full path []" message was not displayed.

Will ask shiproom for permission to add to 2.1.0.


steveharter:
test 
OSX x64 Release Build

Issue #4068, opened 2018-04-25
Build failures in 'official/core-setup/release/1.1.0/' - '20180424.08'


jcagme:
@dotnet-mc-bot commented on [Tue Apr 24 2018](https://github.com/dotnet/core-eng/issues/3359)

There were a set of failures during this build. Here is a summary of these:
* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1625890
  - **Agent:** DDVSOLINUXAGE013
  - **Error log:** [39;49mBuild failed: System.AggregateException: One or more errors occurred. (The type initializer for 'Crypto' threw an exception.) ---> System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ...

* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1625891
  - **Agent:** DDVSOLINUXAGE016
  - **Error log:** [39;49mBuild failed: System.AggregateException: One or more errors occurred. (The type initializer for 'Crypto' threw an exception.) ---> System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ...

* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1625893
  - **Agent:** DDVSOLINUXAGE025
  - **Error log:** [39;49mBuild failed: System.AggregateException: One or more errors occurred. (The type initializer for 'Crypto' threw an exception.) ---> System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ...

* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1625900
  - **Agent:** DDVSOLINUXAGE029
  - **Error log:** dpkg-shlibdeps: warning: debian/dotnet-hostfxr-1.1.9-servicing-001724-00/usr/share/dotnet/host/fxr/1.1.9-servicing-001724-00/libhostfxr.so contains an unresolvable reference to symbol dlerror: it's probably a plugin...

@jcagme, @markwilkie

steveharter:
@weshaggard @rakeshsinghranchi @ianhays any thoughts on these 1.1 issues? Some Crypto (OpenSsl?) issues as well as a hostfxr issue.

weshaggard:
cc @wtgodbe @bartonjs 

They don't look familiar to me I have to wonder if some image was update which broken openssl.

ianhays:
[...]

PR #4111, opened 2018-05-01
[1.1] Misc fixes

ianhays:
- Fix bugs in debian9 dockerfile
- Remove libcurl3 dependency from shared framework package
- Fix Package restore error for new distros:

The new distros added for 1.1 currently have an issue where the tools restoration for the build/packaging stage is pulling in an old version of System.Security.Cryptograph.X509Certifications (and a few others) that directly reference the runtime-specific meta-package component runtime.native.System.Security.Cryptography.OpenSsl. This is an issue because the 4.3.0 version being restored does not include support for the new distros, so we're seeing dll missing errors on those platforms.

There are a few possible long-term solutions to this problem from the CoreFX repo:
1. Ship a new version of NETStandard.Library that contains an updated version of the managed packages that reference the Crypto runtime component.
2. Modify Microsoft.NETCore.Targets to include runtime package references for the crypto native packages. This would resolve the issue for any project on a new distro that references both the Targets package as well as the NetStandardLibrary package, as the newer runtime specific package would win out during package restoration.

This commit doesn't aim to do those things but works around the issue in core-setup by directly referencing the OpenSsl native package from the projects that actually use the cryptographic methods that rely upon the OpenSsl native package.

Might resolve https://github.com/dotnet/core-setup/issues/4068

cc: @weshaggard 

@wtgodbe would it be possible to run a full pipebuild on this changeset?


ianhays:
superseded by https://github.com/dotnet/core-setup/pull/4132

PR #4115, opened 2018-05-01
[release/2.1] Update to rc1 sdk

weshaggard:
I accidentally deleted my branch which closed out https://github.com/dotnet/core-setup/pull/4107 so I'm reopening a PR.

cc @Petermarcu @steveharter

PR #4122, opened 2018-05-03
[master] Update to 2.1.300-rtm-008823 sdk

weshaggard:


weshaggard:
linux-x64 build will fail until we get a fix for the directory enumeration issue into the sdk, see https://github.com/dotnet/core-setup/pull/4115#issuecomment-386650493 for details.

weshaggard:
Move to a newer version of the SDK that should have the file-system fix needed to fix the linux build issues.

PR #4127, opened 2018-05-04
Revert "Update to 2.1 RC1 verison of SDK tools"

weshaggard:
This reverts commit 412cc708c416d808e9be2191d31ea956b63bbe11.

Moving back to older SDK until we can fix official build

FYI @Petermarcu @mmitche 

Official builds are failing https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=1660133&_a=summary

I suspect it is because of the file system issue manifesting in a different way, not sure yet. I don't have enough time to fix it today so I'm reverting so we don't block prodcon builds from succeeding in the mean time.


weshaggard:
I believe I found a fix for the official build issue so lets not revert just yet. If that fix doesn't work and we get blocked we can merge this PR.

weshaggard:
Closing this PR as https://github.com/dotnet/core-setup/pull/4128 fixed the official build so we don't need to revert the SDK change.

PR #4128, opened 2018-05-05
Remove file names from PackagePath folder

weshaggard:
We are hitting https://github.com/NuGet/Home/issues/6351, which
was resolved as a by-design change so we need to adjust to
handle it.

Paths for filenames that don't have extensions turned into
<filename>\<filename> as the file got appended to it. Without
an extension nuget assumes the target path is a directory,
for better or worse.

cc @Petermarcu here is another SDK regression caused by a nuget change.


weshaggard:
This is blocking official builds and won't get exercised in CI so merging. I will monitor official builds after this change goes through.

PR #4135, opened 2018-05-09
Use current version of DependencyModel

### LINKED ISSUE #4082: Update to use latest SDK for 2.1 RTW
In order to write the version metadata to the framework deps.json fils, core-setup will need to take a newer SDK build that has that functionality once RC has shipped.

The latest [RC build](https://github.com/dotnet/versions/tree/master/build-info/dotnet/product/cli/release/2.1) does have the functionality to write the version metadata, but core-setup will not (should not) be updated until after RC.

cc @weshaggard @livarcocc @dleeapho 

cc @natemcmaster ASP.NET will need to do this also in order to support roll-forward properly.

See https://github.com/dotnet/corefx/issues/29032

steveharter:
Update the core-setup.tasks.dll to a newer Microsoft.Extensions.DependencyModel so that it properly generates the version metadata in Microsoft.NETCore.App.deps.json. The version of Microsoft.Extensions.DependencyModel selected is the same as the SDK's (2.1.300-rc1-008673) reference which adds support to use the newer DependencyModel.

Fixes #4082

### DIFF
--- a/tools-local/tasks/core-setup.tasks.csproj
+++ b/tools-local/tasks/core-setup.tasks.csproj
@@ -14,9 +14,7 @@
     <PackageReference Include="NuGet.ProjectModel">
       <Version>$(NugetProjectModelPackageVersion)</Version>
     </PackageReference>
-    <PackageReference Include="Microsoft.Extensions.DependencyModel">
-      <Version>1.1.1</Version>
-    </PackageReference>
+    <ProjectReference Include="..\..\src\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj"/>
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' != 'net46'">
     <Reference Include="$(BuildToolsTaskCoreDir)Microsoft.DotNet.Build.CloudTestTasks.dll" />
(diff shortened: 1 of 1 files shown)


tools-local/tasks/core-setup.tasks.csproj
```diff
@@ -14,9 +14,7 @@
     <PackageReference Include="NuGet.ProjectModel">
       <Version>$(NugetProjectModelPackageVersion)</Version>
     </PackageReference>
-    <PackageReference Include="Microsoft.Extensions.DependencyModel">
-      <Version>1.1.1</Version>
-    </PackageReference>
+    <ProjectReference Include="..\..\src\managed\Microsoft.Extensions.DependencyModel\Microsoft.Extensions.DependencyModel.csproj"/>
```
weshaggard: You cannot take a project reference here, because it will cause a cycle (i.e. we cannot build a project in core-setup until these build tasks are built). 

Instead you need to update to a later package version.

weshaggard:
The linux leg is failing due to the https://github.com/dotnet/core-setup/pull/4115#issuecomment-386650493 and can be ignore for now.

weshaggard:
@dagood @dseefeld this might cause some issues in source-build.

dagood:
I don't think there are any immediate problems since either way this is going to be part of prebuilts. It is strange that SDK 2.1.300-rc1-008673 depends on 2.1.0-preview2-**26306-03**, though... the Core-Setup preview2 release on the NuGet Gallery is [2.1.0-preview2-**26406-04**](https://www.nuget.org/packages/Microsoft.Extensions.DependencyModel/2.1.0-preview2-26406-04). This will make some commits harder to build when we get around to implementing a retention policy. I'd  actually have expected the ProdCon rc1 SDK to go even further and use the 2.1.0-rc1 version also on the gallery.

PR #4138, opened 2018-05-10
Enable Microbuild Signing in Core-Setup 2.0.0

wtgodbe:
For issue https://github.com/dotnet/core-eng/issues/3393

This will also require passing `PB_UseEsrpSigning` from the 2.0.0 pipebuild

@weshaggard @eerhardt @dagood PTAL, especially at publish.proj

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -149,8 +149,10 @@
   </Target>
 
   <Target Name="PublishCoreHostPackagesToFeed"
-          DependsOnTargets="CheckIfAllBuildsHavePublished"
-          Condition="'@(_MissingBlobNames)' == '' AND '$(NuGetFeedUrl)' != ''">
+          DependsOnTargets="CheckIfAllBuildsHavePublished;DownloadCoreHostPackagesFromAzure;SignPackages;DoPushCoreHostPackagesToFeed"
+          Condition="'@(_MissingBlobNames)' == '' AND '$(NuGetFeedUrl)' != ''"/>
+
+  <Target Name="DownloadCoreHostPackagesFromAzure">
     <Error Condition="'$(NuGetApiKey)' == ''" Text="Missing required property NuGetApiKey" />
     <Error Condition="'$(AzureAccessToken)' == ''" Text="Missing required property 'AzureAccessToken'" />
     <Error Condition="'$(AzureAccountName)' == ''" Text="Missing required property 'AzureAccountName'" />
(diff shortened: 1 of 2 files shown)


publish/publish.proj
```diff
@@ -185,6 +187,23 @@
...
     
+  </Target>
+
+  <Target Name="GetPackagesToSign">
+    <ItemGroup>
+      <FilesToSign Include="@(_DownloadedStandardPackages)">
+        <Authenticode>NuGet</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+    <Message Importance="High" Text="Attempting to sign package '%(FilesToSign.Identity)'" />
```
eerhardt: Might want to add an error here if there are no files to sign.

publish/publish.proj
```diff
@@ -185,6 +187,23 @@
...
+  </Target>
+
+  <Target Name="SignPackages"
+          Condition="'$(SkipSigning)' != 'true' and '$(SignType)' != 'public'"
+          DependsOnTargets="GetPackagesToSign;SignFiles">
+  </Target>
+
+  <Target Name="DoPushCoreHostPackagesToFeed">
```
eerhardt: (nit) Can we drop the `Do` in the name?  `PushCoreHostPackagesToFeed` seems good enough to me.

dagood: `Do` also makes me read the rest as if it's a question. 😄

I think adding in `Signed` would make the sequence look good:

`PublishCoreHostPackagesToFeed` depends on:
`GetPackagesToSign;SignPackages;PushSignedCoreHostPackagesToFeed`.

publish/publish.proj
dagood: A concern might be if there could be a servicing release where no packages are built in the first place, but I don't know why that would happen.

dagood: Trying to think of an edge case, maybe a Debian package itself (or something else this repo builds) could need some servicing fix that doesn't affect the shared framework?

Seems like enough of a stretch that catching "no packages downloaded" problems is more useful than avoiding a respin in some weird servicing scenario down the line.

eerhardt: Yes, given the history of suddenly having unsigned official builds, we've been adding these kinds of checks into the signing in order to try to catch issues sooner.

e.g. https://github.com/dotnet/core-setup/blob/3b80e812b14cd7bbdc87c6e4916523bed0de9a0b/signing/sign.proj#L54

weshaggard: Yeah I think it is reasonable to assume we have at least one package to sign.

wtgodbe:
Updated to address feedback

PR #4142, opened 2018-05-11
Add support for clang 4.0 cross build

gbalykov:
cc @alpencolt @Dmitri-Botcharnikov


kasper3:
Note that CoreCLR and CoreFX builds with clang v3.x, v4, v5 and v6. It would be much profitable if we extract out these "detect clang + llvm + lldb" mechanics into one script and place it in a common location (dotnet/buildtools) to avoid duplicating the effort and be in this desync situation all the time.

gbalykov:
@Petermarcu, could this pull request be merged?

gbalykov:
cc @weshaggard

weshaggard:
@gbalykov do you think this will address the CI failure we are seeing in https://github.com/dotnet/core-setup/pull/4157?

gbalykov:
@weshaggard, yes, looks like it fixed them

Issue #4113, opened 2018-05-01
Stabilize 2.1 versioning and branding for release


leecow:
Placeholder for this work to be done closer to final release time. 

- installer UI
- built assets

cc @Petermarcu @weshaggard

weshaggard:
@leecow do you know what all needs to be updated? Do we just need to empty out https://github.com/dotnet/core-setup/blob/master/BranchInfo.props#L15? If so I'd like to do this based on our stable package build parameter so we don't need to make a commit to switch these.

leecow:
No, I don't know the details of how this is done. cc @rakeshsinghranchi in case he's listening.

rakeshsinghranchi:
Yes, just empty out the ReleaseBrandSuffix property for installer UI.

weshaggard:
Support fixed with https://github.com/dotnet/core-setup/pull/4145.

PR #4145, opened 2018-05-11
[release/2.1] Remove brand suffix for stable builds

weshaggard:
This removes the brand suffix from the installers for
stable builds.

cc @rakeshsinghranchi @leecow @mmitche


weshaggard:
Fixes https://github.com/dotnet/core-setup/issues/4113

PR #4172, opened 2018-05-29
Bump version to 2.1.1

wtgodbe:
@mmitche @weshaggard PTAL

### DIFF
--- a/src/pkg/packaging/dir.proj
+++ b/src/pkg/packaging/dir.proj
@@ -174,6 +174,8 @@
 
     <ItemGroup Condition="'$(BuildAllPackages)' == 'true'">
       <!-- The list of packages we are not servicing -->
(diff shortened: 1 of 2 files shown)


src/pkg/packaging/dir.proj
```diff
@@ -174,6 +174,8 @@
 
     <ItemGroup Condition="'$(BuildAllPackages)' == 'true'">
       <!-- The list of packages we are not servicing -->
+      <PackageProjects Include="$(ProjectDir)src\managed\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
```
weshaggard: You also need to remove them from the other item group, but keep it around with just the comment.

Issue #4183, opened 2018-06-01
Stop building old style runtime linux installers


rakeshsinghranchi:
Two changes are required:
1.	Stop producing old style packages
2.	Remove “-upgrade” suffix from new style packages.

PR #4184, opened 2018-06-01
Stop building old style linux installers and rename new style packages

rakeshsinghranchi:
Fixes : https://github.com/dotnet/core-setup/issues/4183


natemcmaster:
Do you have a build number with this change that we can use to update dotnet/cli?

cc @livarcocc

Issue #4169, opened 2018-05-26
Framework-dependent apphost can't sit side-by-side with application dll.


peterhuene:
## Steps to reproduce

1. Have a directory structure like this:
 ```
.
├── foo (renamed apphost, embeds './foo.dll' as path to application)
├── foo.deps.json (for foo.dll)
├── foo.dll
└── foo.runtimeconfig.json (for foo.dll)
```

2. Run the apphost: `./foo`

## Expected  behavior

The application runs in "apphost" mode and attempts to find a shared framework to run the application dll with.

If a shared framework is found, it runs the application.  If not, it prints a helpful diagnostic.

## Actual behavior

The apphost executes in split mode, printing out `dotnet` help.

```
$ COREHOST_TRACE=1 ./foo
Tracing enabled
--- Invoked apphost [version: 2.1.0-rtm-26515-03, commit hash: caa7b7e2bad98e56a687fb5cbaf60825500800f7] main = {
./foo
}
The managed DLL bound to this executable is: 'foo.dll'
Using default installation location [/usr/local/share/dotnet] as runtime location.
Reading fx resolver directory=[/usr/local/share/dotnet/host/fxr]
Considering fxr version=[2.0.6]...
Considering fxr version=[2.1.0-rtm-26510-03]...
Considering fxr version=[2.0.0]...
Considering fxr version=[2.1.0-rc1]...
Considering fxr version=[2.1.0-preview2-26404-04]...
Detected latest fxr version=[/usr/local/share/dotnet/host/fxr/2.1.0-rtm-26510-03]...
Resolved fxr [/usr/local/share/dotnet/host/fxr/2.1.0-rtm-26510-03/libhostfxr.dylib]...
Invoking fx resolver [/usr/local/share/dotnet/host/fxr/2.1.0-rtm-26510-03/libhostfxr.dylib] v2
Host path: [/Users/peterhuene/tmp/foo/bin/Debug/netcoreapp2.1/osx-x64/foo]
Dotnet path: [/usr/local/share/dotnet]
App path: [/Users/peterhuene/tmp/foo/bin/Debug/netcoreapp2.1/osx-x64/foo.dll]
Tracing enabled
--- Invoked hostfxr v2 [commit hash: b2a23fb9737779fc19146b775f1135a740f456f3] main
Checking if CoreCLR path exists=[/usr/local/share/dotnet/libcoreclr.dylib]
[...]

PR #4191, opened 2018-06-04
Fix issue with framework-dependent apphost launched from own directory

steveharter:
If a framework-dependent apphost is launched from its own directory, it fails because the code to detect the mode picks the legacy "split fx" mode instead of "apphost" (which was the old "standalone" mode before the  framework-dependent apphost feature was added). 

The framework-dependent apphost feature added in 2.1 did not change the semantics of this area of code (but should have). The fix is to assume if the CLR is not installed locally then assume the mode is "apphost". I am not sure if anyone uses the split_fx mode anymore, but I kept the code as compatible as possible in either case; any use of split_fx in 2.1.x would be a roll-forward from 1.x.

Fixes https://github.com/dotnet/core-setup/issues/4169

Also this will need to be ported to 2.1.x. I will create a new issue for that.

cc @jeffschwMSFT

### DIFF
--- a/src/corehost/cli/libhost.cpp
+++ b/src/corehost/cli/libhost.cpp
@@ -51,18 +51,28 @@ void get_runtime_config_paths(const pal::string_t& path, const pal::string_t& na
 
 host_mode_t detect_operating_mode(const host_startup_info_t& host_info)
 {
-    if (coreclr_exists_in_dir(host_info.dotnet_root) || pal::file_exists(host_info.app_path))
+    bool has_local_coreclr = coreclr_exists_in_dir(host_info.dotnet_root);
+    if (has_local_coreclr || pal::file_exists(host_info.app_path))
     {
-        pal::string_t own_deps_json = host_info.dotnet_root;
-        pal::string_t own_deps_filename = host_info.get_app_name() + _X(".deps.json");
-        pal::string_t own_config_filename = host_info.get_app_name() + _X(".runtimeconfig.json");
-        append_path(&own_deps_json, own_deps_filename.c_str());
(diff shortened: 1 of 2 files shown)


src/corehost/cli/libhost.cpp
```diff
@@ -51,18 +51,28 @@ void get_runtime_config_paths(const pal::string_t& path, const pal::string_t& na
...
         }
-        return ((pal::file_exists(own_deps_json) || !pal::file_exists(own_config_filename)) && pal::file_exists(host_info.app_path)) ? host_mode_t::apphost : host_mode_t::split_fx;
+
+        if (has_local_coreclr)
+        {
+            // Name of runtimeconfig file; since no path it is checked for in current working directory
+            pal::string_t config_in_cwd = host_info.get_app_name() + _X(".runtimeconfig.json");
+
+            // Detect standalone apphost or legacy split mode (specifying --depsfile and --runtimeconfig)
+            return (pal::file_exists(deps_in_dotnet_root) || !pal::file_exists(config_in_cwd)) && pal::file_exists(host_info.app_path) ? host_mode_t::apphost : host_mode_t::split_fx;
```
eerhardt: `pal::file_exists(host_info.app_path)` is checked twice.  Can that be moved to a local variable as well?

steveharter: I refactored the code to make it easier to understand; the check is only done once now. Also, the trace message is only included in the standalone\split check now since it didn't really add anything otherwise

eerhardt: I like the new refactoring.  thanks!

PR #4196, opened 2018-06-05
Update CoreFx Dependencies for 2.0.0

wtgodbe:
Consume new CoreFx version for non-stable Core-Setup build. We ran CoreFx builds for all 3 release branches, but not CoreClr builds - therefore Core-Setup 1.x.x took no updates as the only CoreFx package we published was the PackageBaseline. As such, this should be all that is needed for 2.0.0.

@safern please confirm


wtgodbe:
Actually, there are a bunch of newly serviced 1.1.0 packages for CoreFx - this will require an update for that as well.

PR #4204, opened 2018-06-07
Build Microsoft.NETCore.App and all of the DotNetHost packages for 1.1.9 servicing.

eerhardt:
We made a mistake and started building **all** packages for release/1.1.  This causes problems because there are still UWP packages in this repo, which we don't want to build.

Making it so just the packages we need are built.


eerhardt:
CI is expected to fail in the release branch.  Merging.

PR #4206, opened 2018-06-07
Take CoreClr update & bump to 2.0.9

wtgodbe:
CC @weshaggard 

@safern PTAL

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
+++ b/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
@@ -22,7 +22,7 @@
       <Version>1.1.8</Version>
     </NETCoreApp11Dependency>
     <NETCoreApp11Dependency Include="Microsoft.NETCore.Runtime.CoreCLR">
-      <Version>1.1.9-servicing-26607-01</Version>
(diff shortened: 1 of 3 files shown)


src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.1.props
```diff
@@ -22,7 +22,7 @@
       <Version>1.1.8</Version>
     </NETCoreApp11Dependency>
     <NETCoreApp11Dependency Include="Microsoft.NETCore.Runtime.CoreCLR">
-      <Version>1.1.8</Version>
+      <Version>1.1.9-servicing-26607-01</Version>
```
weshaggard: Lets not do this change yet. We should only do this once we ship the 1.1 stable, we don't want to tie the shipping of these 2 release together.

wtgodbe: I'm confused here, why build CoreClr if we don't want Core-Setup to consume it for this release?

wtgodbe: Actually never mind, I get what you're saying. Will remove

PR #4235, opened 2018-06-14
Mark 2.0.9 stable & bump versions

wtgodbe:
CC @weshaggard @safern

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
+++ b/src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
@@ -16,13 +16,13 @@
       <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
-      <Version>1.0.11</Version>
+      <Version>1.0.12-servicing-005061-00</Version>
     </NETCoreApp10Dependency>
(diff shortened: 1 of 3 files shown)


src/pkg/projects/Microsoft.NETCore.App/netcoreapp1.0.props
```diff
@@ -16,13 +16,13 @@
       <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="Microsoft.NETCore.DotNetHostPolicy">
-      <Version>1.0.11</Version>
+      <Version>1.0.12-servicing-005061-00</Version>
```
weshaggard: We cannot have unstable package references here.

wtgodbe: Will it affect the build if I run a 2.0.0 build with the expected stable version, before that package exists?

weshaggard: It shouldn't. I do expect your current change to cause an issue though because nuget will not let you pack a stable package with prerelease dependencies.

wtgodbe: @weshaggard what about NETCore.Targets & NETCore.Platforms?

weshaggard: Those need to be stable as well.

PR #4237, opened 2018-06-14
Mark backwards compatibility packages as stable

wtgodbe:
@weshaggard PTAL


weshaggard:
Is this going to cause you to respin the 2.0 stable builds? If so please clean-up the earlier builds otherwise this will fail.

wtgodbe:
This won't cause me to re-spin builds.

PR #4238, opened 2018-06-14
Add host versioning tests

steveharter:
Add tests that verify the apphost and hostfxr are backwards and forward compatible.

This will catch potential future issues with binary contracts between apphost\dotnet->hostfxr->hostpolicy and issues associated with unix\osx symbol-based export merging.

The apphost is used in a standalone mode (instead of dotnet muxer in shared mode) to make the tests easier to write and maintain. Only versions 2.0.x and 2.1 are currently verified (not 1.0 or 1.1 because they didn't use apphost).

Fixes https://github.com/dotnet/core-setup/issues/3777

cc @jeffschwMSFT

### DIFF
--- a/src/test/Assets/TestProjects/StandaloneApp20/StandaloneApp20.csproj
+++ b/src/test/Assets/TestProjects/StandaloneApp20/StandaloneApp20.csproj
@@ -1,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+	<AssemblyName>StandaloneApp</AssemblyName>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <RuntimeIdentifier>$(TestTargetRid)</RuntimeIdentifier>
+    <RuntimeFrameworkVersion>2.0.7</RuntimeFrameworkVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+  </ItemGroup>
+
+</Project>
--- a/src/test/HostActivationTests/GivenThatICareAboutHostVersionCompatibility.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutHostVersionCompatibility.cs
@@ -1,0 +1,179 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
+using Microsoft.DotNet.Cli.Build.Framework;
+using Microsoft.DotNet.CoreSetup.Test;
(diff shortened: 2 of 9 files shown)


src/test/Assets/TestProjects/StandaloneApp20/StandaloneApp20.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+	<AssemblyName>StandaloneApp</AssemblyName>
```
eerhardt: Why not keep the assembly name the same as the project?  I think that would avoid confusion later, wouldn't it?

steveharter: Because the old and new versions of the apphost are swapped depending on the test, and we need to have the sane embedded apphost.name

src/test/HostActivationTests/GivenThatICareAboutHostVersionCompatibility.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            try 
+            {
+                publishFixture
+                    .EnsureRestoredForRid(publishFixture.CurrentRid, repoDirectories.CorehostPackages)
+                    .PublishProject(runtime: publishFixture.CurrentRid);
+            }
+            catch (BuildFailureException)
+            {
+                // Some RIDs didn't exist in 2.0+ so if we get an error publishing, ignore.
```
eerhardt: Would it be more appropriate to block on the list of RIDs up front? That way we know exactly which RIDs don't work, and for the RIDs we do expect to work, they always will work.

steveharter: I did consider that, but since the list of RIDs can change over time (in master branch), I want to avoid having to change the code whenever a new RID comes out that is not present in 2.0\2.1

eerhardt: You could white-list it - here are the list of RIDs that should work.  If the current RID isn't in that list, then skip it.

steveharter:
test

steveharter:
test OSX x64 Release Build

ALL_BUILD.vcxproj
```diff
@@ -0,0 +1,187 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
steveharter: Yes. It looks like a build was cancelled and left a bunch of temp files

src/test/HostActivationTests/GivenThatICareAboutHostVersionCompatibility.cs
steveharter: I whitelisted Ubuntu 16. Don't really see the need to test all variants anyway.

steveharter:
Test Tizen armel Release Build

PR #4241, opened 2018-06-15
Download, Sign, & Re-upload Core-Setup packages during pipeline

wtgodbe:
This PR implements Microbuild Package Signing in the Core-Setup pipeline. We need an additional, post-build job to do this in 1.x because 1.x's pipeline just takes the last build to finish, and finalizes the build from there. However, Microbuild signing needs to be done on Windows, so if the last pipeline build to finish were a non-Windows build, we'd be out of luck. This PR adds the following files:

- `syncAzure.proj` - syncs blobs from the Azure container holding the current build's packages. For this to work, I needed to restore a newer version of Buildtools than was already present in Core-Setup. This is because we need to download packages from a relative blob path `(dotnet/release/1.1.0/Binaries/{BuildNumber}`), which isn't supported in 1.1.0 buildtools.
- `publish.proj` - handles signing & publishing of packages
- `Core-Setup-Sign-And-Publish.json` - defines the new job that does signing & publishing. This runs after every other build has finished.

The workflow for the publish job is as follows:

1. Install the signing plugin
2. Run the new "WriteAzureBlobInfoFile" Target, which calculates the current build number, container name, and container channel (`release/1.1.0` in this case), and writes them to a .props file.
3. Run init-tools.cmd
4. Run `syncAzure.proj` to download the packages from the current build
5. Run `publish.proj` to re-publish the just-downloaded packages to a folder named `unsigned`, alongside their current location in the Azure container
6. Run `publish.proj` to Sign the packages
7. Run `publish.proj` to upload the newly-signed packages, overwriting the unsigned packages that were originally there. This leaves us with signed packages in the expected location, and unsigned packages in a folder named `unsigned` alongside them.
8. Run `publish.proj` to copy the packages to their own directory separate from the symbol packages, for myget publishing
9. Publish packages to myget

I've tested the workflow locally and it seems to work as expected.

This will also require the following changes in the pipebuild itself:

- Factor out and pass `CloudDropAccessToken` and `CloudDropAccountName` to child builds
- Pass `PB_UseEsrpSigning=true` to child builds
- Pass `PB_PublishType=myget` to child builds

@weshaggard @MattGal @chcosta @eerhardt @dagood PTAL
CC @mmitche @karelz


wtgodbe:
I'll also need to port this to 1.0.0.

wtgodbe:
I'll also need to get signing permission from barsd for the new job.

pkg/init-tools.cmd
```diff
@@ -8,10 +8,25 @@ set DOTNET_PATH=%TOOLRUNTIME_DIR%\dotnetcli\
...
 if [%BUILDTOOLS_SOURCE%]==[] set BUILDTOOLS_SOURCE=https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json
 set /P BUILDTOOLS_VERSION=< "%~dp0BuildToolsVersion.txt"
+set /P SYNCTOOLS_VERSION=< "%~dp0SyncToolsVersion.txt"
 set BUILD_TOOLS_PATH=%PACKAGES_DIR%Microsoft.DotNet.BuildTools\%BUILDTOOLS_VERSION%\lib\
+
 set PROJECT_JSON_PATH=%TOOLRUNTIME_DIR%\%BUILDTOOLS_VERSION%
 set PROJECT_JSON_FILE=%PROJECT_JSON_PATH%\project.json
 set PROJECT_JSON_CONTENTS={ "dependencies": { "Microsoft.DotNet.BuildTools": "%BUILDTOOLS_VERSION%" }, "frameworks": { "netcoreapp1.0": { } } }
+
+set SYNC_JSON_PATH=%PACKAGES_DIR%
```
weshaggard: Why do we need a new separate build tools version? Can to support not be added to the other build tools version?

buildpipeline/pipeline.json
```diff
@@ -137,6 +137,28 @@
           }
         }
       ]
+    },
+    {
+      "Name": "Sign And Publish Packages",
```
weshaggard: Is your plan to stop the publishing from the individual legs now as just do it here? I think that would be a good idea but I didn't see anything obvious that was doing that.

pkg/init-tools.cmd
wtgodbe: Do you mean porting https://github.com/dotnet/buildtools/commit/505527f53e4a0cf0836f6760aa1f11b2f2eb3d1f#diff-2ac12292d334d19fc842d030a958c0eb to 1.1.0 buildtools & relying on the new version with that change? I don't see any reason why that wouldn't work.

buildpipeline/pipeline.json
wtgodbe: No, we still need the individual legs to publish to Azure in order for this new job to find the packages to sign. I was under the impression that the plan from our old email thread was to just have the signed packages available in Azure, and publish to nuget from there

wtgodbe: Or do you think we should disable myget publishing by the individual legs, and have this job do the myget publishing instead?

weshaggard: I realize they need to publish to azure but IIRC what happens today is whatever leg finishes last publishes to myget and given you are creating a sync point now we should be able to stop doing that and just have your leg publish to myget.

pkg/init-tools.cmd
weshaggard: Yes port whatever change you need, that will be way better then trying to create a parallel build-tools.

wtgodbe: https://github.com/dotnet/buildtools/pull/2070

buildpipeline/pipeline.json
wtgodbe: Fair point, I'll work on removing the myget publish from the Finalize step & adding it to the new pipeline job.

wtgodbe:
@weshaggard PTAL at the latest commit

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.DotNet.Cli.Build
     public class AzurePublisher
     {
         private static readonly string s_dotnetBlobRootUrl = "https://dotnetcli.blob.core.windows.net/dotnet/";
-        private static readonly string s_dotnetBlobContainerName = "dotnet";
+        private static readonly string s_dotnetBlobContainerName = Environment.GetEnvironmentVariable("CONTAINER_NAME");
```
eerhardt: I think doing `Environment.GetEnvironmentVariable("CONTAINER_NAME") ?? "dotnet";` might be best here - keep the fallback in case it isn't supplied.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.DotNet.Cli.Build
     public class AzurePublisher
     {
         private static readonly string s_dotnetBlobRootUrl = "https://dotnetcli.blob.core.windows.net/dotnet/";
```
eerhardt: What about adding a override for this hard-coded URL?

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -164,6 +159,27 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
...
+                "<Project ToolsVersion=\"14.0\" DefaultTargets=\"Build\" xmlns=\"http://schemas.microsoft.com/developer/msbuild/2003\">",
+                "\t<PropertyGroup>",
+                "\t\t<ContainerName>" + Environment.GetEnvironmentVariable("CONTAINER_NAME") + "</ContainerName>",
+                "\t\t<Channel>" + Channel + "</Channel>",
+                "\t\t<BuildNumber>" + SharedFrameworkNugetVersion + "</BuildNumber>",
+                "\t</PropertyGroup>",
+                "</Project>"
+            };
+
+            System.IO.File.WriteAllLines(Dirs.RepoRoot + "\\AzureBlob.props", lines);
```
eerhardt: It usually isn't good to write files into the repo during the build.

Why can't these variables be passed on the command-line when we invoke the `syncAzure.proj` file?

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
wtgodbe: Will do

build_projects/dotnet-host-build/PublishTargets.cs
wtgodbe: I could pass ContainerName & Channel from the command line, but that would duplicate what's in BranchInfo.txt. The real issue is BuildNumber, though - this is calculated from CommitCountString, which I wouldn't really have access to from the command line.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
eerhardt: OK, this is fine.

build_projects/dotnet-host-build/PublishTargets.cs
```diff
@@ -137,12 +138,6 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
                         "opensuse.42.3.x64.version"
                     };
 
-                    BuildTargetResult feedResult = c.BuildContext.RunTarget(nameof(PublishTargets.PublishCoreHostPackagesToFeed));
```
eerhardt: Does `PublishCoreHostPackagesToFeed` still get used/called?  If not, can it be removed?

PR #4263, opened 2018-06-19
Mark packages as stable for 1.0.12

wtgodbe:
CC @weshaggard @safern


wtgodbe:
See also https://github.com/dotnet/core-setup/pull/4268 for non-stabilizing

PR #4264, opened 2018-06-19
Mark packages as stable for 1.1.9

wtgodbe:
@weshaggard @safern PTAL (especially at stable.package.props)

Don't merge until we resolve the non-stable build issue (hostfxr installer pinned to 1.1.2) for Core-Setup 1.1.0 (cc @eerhardt)

PR #4267, opened 2018-06-20
Mark more packages as stable for 1.1.9

wtgodbe:
CC @weshaggard this should resolve the errors in the Maestro build for core-setup.

Also @dagood, these files aren't getting touched by the auto-update PRs - is there an easy way to fix that?


dagood:
You should be able to add to this list to update more:

https://github.com/dotnet/core-setup/blob/801629db94a5ef63a805ca1b8e72d9f4100af915/build_projects/update-dependencies/UpdateFilesTargets.cs#L90-L93

wtgodbe:
@weshaggard @ianhays @eerhardt this is failing to find CoreClr packages for fedora.23-x64, opensuse.13.2-x64, and ubuntu.16.10-x64. Looks like CoreClr didn't produce any packages for these RIDs, and hasn't since at least last September - is this what we expect to happen here?

wtgodbe:
And why wouldn't we be seeing the same failures in https://github.com/dotnet/core-setup/pull/4264?

wtgodbe:
Actually, looks like it's https://github.com/wtgodbe/core-setup/blob/d3268bf31dc088b4b9dfbb7f4e4ba8fbbb14f37d/tools/independent/RuntimeGraphGenerator/project.json#L26-L44 and https://github.com/wtgodbe/core-setup/blob/d3268bf31dc088b4b9dfbb7f4e4ba8fbbb14f37d/build_projects/dotnet-host-build/project.json#L37-L56. Should I just remove the offending RIDs from those lists?

eerhardt:
Yes, those deprecated Linux distro versions should be removed.

wtgodbe:
@dotnet-bot test this please

PR #4273, opened 2018-06-20
Update Locked HostFXR Installer version to 1.1.9

wtgodbe:
@weshaggard @vivmishra @eerhardt PTAL - we should get this in ASAP to unblock CLI

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/HostVersion.cs
+++ b/build_projects/shared-build-targets-utils/Utils/HostVersion.cs
@@ -127,6 +127,8 @@ namespace Microsoft.DotNet.Cli.Build
 
         // This method returns the locked hostfxr version based on the flag fExplicitHostFXRMSIVersion and the current platform.
         // For MSI (Windows) generation we specify a newer version for handling issue #1574 and for non-Windows platform we return the LockedHostFxrVersion.
+        // We are currently locked at 1.1.9 because we updated the HostFXR Platform Installer with new RIDs during the 1.1.9 release. At such a time when we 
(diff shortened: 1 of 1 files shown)


build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -133,7 +133,7 @@ public VerInfo GetLockedHostFXRPlatformInstallerVersion()
 
             if (fExplicitHostFXRMSIVersion && CurrentPlatform.Current == BuildPlatform.Windows)
             {
-                version = new VerInfo(1, 1, 2, "", "", "", CommitCountString);
```
eerhardt: Should we also flip `fExplicitHostFXRMSIVersion` to `false`?  that way the MSI version stays in step with `LockedHostFxrVersion`.

vivmishra: The explicit version was set to prevent HostFXR from getting new versions when there is minimal churn in HostFXR. 
Assuming there is new change in hotsfxr now, I think setting the value to 1.1.9 seems right. 

We don't want to revert to LockedHostFxrVersion as IsLocked is not yet locked. It has most probably not set to true ever.

wtgodbe: So we should keep `fExplicitHostFXRMSIVersion=true` ?

vivmishra: Right.

Issue #4099, opened 2018-04-30
Improve host test usability


steveharter:
- The host tests cannot be debugged via breakpoint (easily)
- Even if a breakpoint can be hit in the test, it will not flow into the native code (because it launches another process - dotnet.exe)
- The tests cannot be started or run from the Tests tab in Visual Studio due to msbuild dependencies
- When a test project is open in Visual Studio, it locks files resulting in errors when re-building the tests from command prompt

nguerrera:
> Easily re-running tests. The tests cannot be started or run from the Tests tab in Visual Studio due to msbuild dependencies. Also it is (was?) not possible to run an individual test suite (class) or method like is possible in corefx

It looks like it is just the dependency on environment variables that prevents this. I have individual tests running in test explorer right now using this:

https://gist.github.com/nguerrera/e679a0d6a06ee9f9610896d35851f367

Can we change the test code to figure out these values without using environment variables?

PR #4274, opened 2018-06-20
Fix test output from from FAIL to OK for tests expected to fail

steveharter:
Part of improving test usability via https://github.com/dotnet/core-setup/issues/4099


steveharter:
@weshaggard do you know why on Linux-msul test ran? And not the rest?

Issue #4098, opened 2018-04-30
Hosts tests use too much disk space


steveharter:
The tests routinely copy the entire framework, leaving behind 8+ GB of disk space on every run, and thus requiring a `git clean -xfd` every so often to free up the disk space.

The tests should clean up after a successful run, and continue to be idempotent.

PR #4297, opened 2018-06-22
Clean up disk space after test runs

steveharter:
Fixes https://github.com/dotnet/core-setup/issues/4098

- Added Dispose to the TestProjectFixture to delete any directories it created.
- Applied the XUnit IClassFixture pattern to each test class that has shared state to dispose its test fixtures.
- Applied Dispose pattern to each test class that had instance-level test fixtures or temp directories in order to dispose test fixtures and delete directories.
- Removed unnecessary methods
- For debugging, added PRESERVE_TEST_RUNS=1 environment variable in order to prevent deletion of test run (not set by default). This leaves ~11GB around but is useful when a test fails in order to re-run EXEC commands it from command line.

cc @jeffschwMSFT

### DIFF
--- a/src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
+++ b/src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
@@ -2,15 +2,13 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Threading;
 using Microsoft.DotNet.InternalAbstractions;
 using Xunit;
 
 namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.MultilevelSDKLookup
 {
-    public class GivenThatICareAboutMultilevelSDKLookup
+    public class GivenThatICareAboutMultilevelSDKLookup : IDisposable
     {
-        private static readonly Mutex id_mutex = new Mutex();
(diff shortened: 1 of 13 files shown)


src/test/HostActivationTests/GivenThatICareAboutMultilevelSDKLookup.cs
```diff
@@ -2,15 +2,13 @@
...
 using System.IO;
 using System.Linq;
-using System.Threading;
 using Microsoft.DotNet.InternalAbstractions;
 using Xunit;
 
 namespace Microsoft.DotNet.CoreSetup.Test.HostActivation.MultilevelSDKLookup
 {
-    public class GivenThatICareAboutMultilevelSDKLookup
+    public class GivenThatICareAboutMultilevelSDKLookup : IDisposable
```
eerhardt: Why do these tests have a different pattern than the other tests?  All the other tests use an `IClassFixture<SharedState>`, but this one doesn't.

steveharter: The classes that have global state (that used to be in statics) use IClassFixture, including SDKLookup and SharedFxLookup . The ones that used instance state just implement IDisposable.

However I do think some of the ones that use instance state could used shared state, but that would be an optimization that could come later.

PR #4301, opened 2018-06-25
New hostfxr API to support new MSBuild SDK resolver scenarios

nguerrera:
Prerequisite of https://github.com/dotnet/cli/pull/9547

I don't think the core-setup side will change much, but I will keep this marked as work-in-progress until I get through the TODO list on the CLI side.

Two new API are added:

### hostfxr_resolve_sdk2

Deprecates hostfxr_resolve_sdk  and adds two new capabilities:

1. Return the path to global.json if global.json is used
   - This allows the msbuild sdk resolver to customize its error handling and logging appropriately. It will also be used to enable msbuild to fall back to most recent *compatible* SDK when the most recent SDK chosen without global.json is incompatible with the current msbuild version.

2. Add disallow_prerelease flag
   - When set, this will prevent prerelease SDKs from being resolved unless a prerelease version was specified in global.json. This will be used to prevent prerelease VS installations from impacting side-by-side releases.

### hostfxr_get_available_sdks

Returns all available SDKs in sorted order by version.

Allows MSBuild SDK resolver to fall back to latest compatible SDK when there is no global.json and latest SDK is incompatible with current version of MSBuild.


src/corehost/cli/fxr/sdk_info.cpp
```diff
@@ -7,9 +7,38 @@
...
+    //    * dotnet dir has version 4.0, 5.0, 6.0
+    //    * global dir has 5.0
+    //    * 6.0 is incompatible with calling msbuild
+    //    * 5.0 is compatible with calling msbuild
+    //
+    // MSBuild should select 5.0 from dotnet dir (matching probe order) in muxer
+    // and not 5.0 from global dir.
+    if (a.version == b.version)
+    {
+        assert(a.hive_depth != b.hive_depth);
```
nguerrera: We're crashing on mac somewhere in here and it may be this assert. I need to remember to bring in my mac to debug.

nguerrera:
@steveharter @eerhardt This is ready for review despite WIP label. We shouldn't merge until dotnet/cli#9547 is approved, but it would be good to get the feedback now on the core-setup side. Thanks.

src/corehost/cli/fxr/sdk_info.h
```diff
@@ -8,18 +8,22 @@
 
 struct sdk_info
 {
-    sdk_info(pal::string_t path, fx_ver_t version)
-        : path(path)
-        , version(version) { }
+    sdk_info(pal::string_t base_path, pal::string_t full_path, fx_ver_t version, int32_t hive_depth)
```
steveharter: nit: the original code here should probably be `const &` for the string and version (pass in by ref then assign fields by value) to avoid an extra copy

src/corehost/cli/fxr/sdk_resolver.cpp
```diff
@@ -64,7 +64,7 @@ pal::string_t resolve_cli_version(const pal::string_t& global_json)
     return retval;
 }
 
-pal::string_t resolve_sdk_version(pal::string_t sdk_path, bool parse_only_production, pal::string_t global_cli_version)
+pal::string_t resolve_sdk_version(pal::string_t sdk_path, bool disallow_prerelease, pal::string_t global_cli_version)
```
steveharter: Pass `disallow_prerelease` by reference? The calling functions pass by value, but don't have the check for `if (specified.is_prerelease())` like below

eerhardt:
Do we have any tests for these APIs?

steveharter:
>> Do we have any tests for these APIs?

@nguerrera see `GivenThatICareAboutNativeHostApis` for an example of how I added a test for the native API I added. A little kludgy but does invoke the native API through PInvoke.

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -274,6 +276,167 @@ SHARED_API int32_t hostfxr_resolve_sdk(
...
+//
+// Return value:
+//   0 on success, otherwise failure
+//   0x8000809b - SDK could not be resolved (SdkResolverResolveFailure)
+// 
+// String encoding:
+//   Windows     - UTF-16 (pal::char_t is 2 byte wchar_t)
+//   Unix        - UTF-8  (pal::char_t is 1 byte char)
+//
+SHARED_API int32_t hostfxr_resolve_sdk2(
```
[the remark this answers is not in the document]
nguerrera: > Given the material difference between hostfxr_resolve_sdk and this method is that the location of global.json is returned

The flags (disallow_prerelease) are also material.

> did we consider adding a 'return the path to the global.json' instead of adding a new overload?

I have considered this, but my concerns with that are:

1. Perf: Looking for global.json requires probing on disk from working directory up to root, and I don't want to do this once for the resolution and again for the logging. I would also like to be able to log the path on success paths too because it is common that users do not realize that they have a leftover global.json and that has made.

2. There's a possible race between "resolve the SDK", and "where is global.json" such that the global.json returned by the second query is not the same as the one that was actually used during resolution.

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -274,6 +276,167 @@ SHARED_API int32_t hostfxr_resolve_sdk(
...
+//      Callback invoke to return the list of SDKs by their directory paths.
+//
+// Return value:
+//   0 on success, otherwise failure
+//
+// String encoding:
+//   Windows     - UTF-16 (pal::char_t is 2 byte wchar_t)
+//   Unix        - UTF-8  (pal::char_t is 1 byte char)
+//
+SHARED_API int32_t hostfxr_get_available_sdks(
```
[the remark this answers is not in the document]
nguerrera: I am intentionally giving Visual Studio the freedom to apply some policy of its own here. I agree with your general principle in the abstract, but there's a trade off at play in this case: which is worse -- teaching tooling  too much about hostfxr internals or teaching hostfxr too much aboout tooling internals? 

The external policy that applies in the MSBuild case that does not apply in the mainline dotnet case is that we have to consider whether MSBuild is too old to use a given SDK:

* When you say `dotnet build`, then we pick a full .NET Core SDK including msbuild itself and the rest of the toolset. There is no incompatibility possible.
* When you say `msbuild`, then we pick only the SDK's msbuild tasks and targets and use the toolset from the calling msbuild. The risk is that the tasks and targets are incompatible with the toolset.

We could technically implement this compat check here, but the layering feels wrong to me. Also, regardless of which layering is more sound in theory, it is cost-effective to keep the things that are specific to the VS case out of this repo and in managed code.

Another issue with moving the code to determine if an SDK is compatible with a given msbuild version here, is that we would we either need to either rediscover the compatibility on each call (parsing the minimumMSBuildVersion file out of each SDK several times over the lifetime of a VS instance), or add some static state to the implementation here.

One thing that has been troubling me is that we don't really need/want to have a formal public surface area for the line between VS and the SDK resolution logic in this repo. The goal is really just to share as much code as possible. In other words, we don't want to change the msbuild resolver and hostfxr every time the policy changes, but I also don't want to be constrained to make further changes if the way we've decided to separate responsibilities ends up being a problem. We don't actually need these API to live in hostfxr.dll. We could consider building them into something else that only ever lives in the msbuild sdk resolver folder and that we do not support from any other caller. @eerhardt and I discussed that and the feeling at the time was that the effort to produce a different library and flow it to the CLI repo wasn't worth it for what started as a single, simple entry point. As the complexity here is increasing, it may be worth reconsidering. In all honesty, I am not all certain that we won't need a hostfxr_resolve_sdk3 at some point.  I think separating these concerns out of hostfxr itself is probably the right direction to go, but there is also still pressure to get this VS functionality in very soon...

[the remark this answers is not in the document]
jeffschwMSFT: I see the tradeoffs and I think you hit the nail on the head.  The higher order thing that I think is missing is the set of principles/rules on why we would or would not include a shared_api.  In some sense this API fits really well because I think that having hostfxr be the sole piece of code that understand how to walk our directory structure and return versions numbers feels like.  On the other hand, how it is then used in the CLI feel like we are exposing an internal detail.
You bring up a good point that even though I think of hostfxr as being the 'host' in the VS scenario msbuild/vs are really the host and having them own the policy feels right.
Long way of saying this API seems fine to me.

src/corehost/cli/fxr/hostfxr.cpp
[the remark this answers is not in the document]
jeffschwMSFT: I would prefer a place where we land either with smaller building blocks (perf and a potential race could be interesting, though do we have any evidence that is a concern?) or a new API that is more expandable.  If we stay with an SDK2 version, I think we should do a bit more to help it be more expandable in the future.  This shared_api is really just a thin wrapper over the resolve_sdk_dotnet_path.  I like the addition of flags.  Can we come up with a strategy that enables us to return more than just the two paths?  Perhaps either a property bag, or have the function callback take a key and a value and then call it multiple times.  I don't think we need to do much future proofing, but a little more to help us as things may evolve in intuitive ways.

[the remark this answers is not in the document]
nguerrera: > Perhaps either a property bag, or have the function callback take a key and a value and then call it multiple times. I don't think we need to do much future proofing, but a little more to help us as things may evolve in intuitive ways.

Very good call.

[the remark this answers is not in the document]
nguerrera: I think I'll go with the multiple callbacks because it's more flexible. We could even wire tracing to it at some point, which is something I've wanted. Right now, you can set COREHOST_TRACE and get the logging to stderr, but it would be nice if you could set a tracing flag on the API and get the verbose diagnostics sent to the msbuild log.

src/corehost/cli/fxr/sdk_resolver.cpp
nguerrera: I'd like to leave this as is.

I consider the fact that disallow_prerelease is overridden when global.json has a preview version specified to be an implementation detail of this function. As of now, the caller has no need to observe this. I also find in/out parameters confusing.

src/corehost/cli/fxr/sdk_info.cpp
nguerrera: Fixed. Turns out std::sort on Mac is comparing elements with themselves sometimes. That's legal, so I just removed the invalid assert.

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -274,6 +276,167 @@ SHARED_API int32_t hostfxr_resolve_sdk(
...
+// Invoked by MSBuild resolver when the latest SDK used without global.json
+// present is incompatible with the current MSBuild version. It will select
+// the compatible SDK that is closest to the end of this list.
+//
+// Parameters:
+//    exe_dir
+//      The path to the dotnet executable.
+//
+//    result
+//      Callback invoke to return the list of SDKs by their directory paths.
```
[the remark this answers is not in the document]
nguerrera: Done.

src/corehost/cli/fxr/sdk_info.h
nguerrera: Fixed.

nguerrera:
@jeffschwMSFT I've made the change to convert the result callback to something we invoke more than once with key-value pairs that can evolve over time. 

@steveharter I also pushed some changes to address your smaller concerns and fixed the crash on Mac

I am looking at adding unit tests for the shared API now.

src/corehost/cli/fxr/sdk_resolver.cpp
nguerrera: So here was a bug here, but I fixed it by removing disallow_prerelease as a parameter where it does not make sense to ever disallow prerelease. Still prefer to avoid in/out param for this. 

See 03ab2be

nguerrera:
I've added some tests and fixed a bug. I still intend to leave this WIP until the CLI side is approved and I still have some follow up to do on the CLI side. 

I wills squash this when actually merging, but I'm keeping the individual commits so reviewers can easily see what I've done since they last approved.

nguerrera:
@dotnet-bot Test this please (to check that multilevel lookup being disabled doesn't impact the new tests)

nguerrera:
@steveharter @jeffschwMSFT Taking this out of WIP as the feature on the VS side is approved. I have only added tests since you signed off. Can you take a look at the commits since the green check? I will squash this when merging.

src/test/HostActivationTests/GivenThatICareAboutNativeHostApis.cs
```diff
@@ -96,6 +90,142 @@ public void Breadcrumb_thread_does_not_finish_when_app_has_unhandled_exception()
...
+
+            // Without multi-level lookup: get only sdks sorted by ascending version
+            expectedList = string.Join(';', new[]
+            {
+                 Path.Combine(localSdkDir, "0.1.2"),
+                 Path.Combine(localSdkDir, "1.2.3"),
+                 Path.Combine(localSdkDir, "5.6.7-preview"),
+            });
+
+            dotnet.Exec(appDll, new[] { "hostfxr_get_available_sdks", exeDir })
```
vitek-karas: Personally I would prefer to split this into multiple tests. The setup can be either shared, or rerun every time, it's not that expensive. I don't like tests which run the product multiple times - it's really hard to debug.

nguerrera: You're right. I'll fix it.

nguerrera:
Tests are split. Looking into one thing that I spotted while doing that.

nguerrera:
The thing I spotted was a non-issue. I thought for a minute I was copying libhostfx.so for the test to the wrong place, but I am not. @vitek-karas can you look at the last commit that addressed your feedback. Good to squash-and-merge?

vitek-karas:
LGTM

nguerrera:
@dotnet-bot Test Linux x64 Release Build (it's been pending all night)

PR #4302, opened 2018-06-25
Use --no-cache when restoring packages in 1.0.0

wtgodbe:
This allowed me (locally) to restore non-cached CoreClr/CoreFx packages, which unblocks getting signed binaries out for our builds. This can be reverted after a successful build of Core-Setup.

@weshaggard @eerhardt PTAL

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/DotNetCli.cs
+++ b/build_projects/shared-build-targets-utils/Utils/DotNetCli.cs
@@ -25,6 +25,10 @@ namespace Microsoft.DotNet.Cli.Build
         {
             var newArgs = args.ToList();
             newArgs.Insert(0, command);
+            if (command == "restore")
+            {
+                newArgs.Insert(1, "--no-cache");
+            }
 
             if (EnvVars.Verbose)
             {


build_projects/shared-build-targets-utils/Utils/DotNetCli.cs
```diff
@@ -25,6 +25,10 @@ public Command Exec(string command, params string[] args)
         {
             var newArgs = args.ToList();
             newArgs.Insert(0, command);
+            if (command == "restore")
+            {
+                newArgs.Insert(1, "--no-cache");
```
weshaggard: Are you sure this is the only place that we do restore?

wtgodbe: This covers all location other than init-tools.cmd & pack.cmd. A local build with just this change gave me the signed packages we need from CoreClr/CoreFx

eerhardt:
> This can be reverted after a successful build of Core-Setup.

Instead of reverting it, I think it would be better to add a build option for this.  `DOTNET_NO_CACHE_RESTORE` or something.  When that is set, then add the `--no-cache` option.  That way if we ever get into this situation again, we can just set the build option instead of making a code change.

PR #4331, opened 2018-07-04
packaging for FreeBSD

wfurt:
This is similar to https://github.com/dotnet/coreclr/pull/18764
This is essentially copy from Linux version.
With this change (and few more) I can finish source-build to to core-setup on FreeBSD.
crossgen has some troubles so I disabled it for now in my build tree. 

Published runtime can run Hello world. 

big thanks to @eerhardt for helping me sort this out. 

```
source-build2]$ ./build.sh  /p:RootRepo=core-setup /p:TargetRid=freebsd.11-x64
```
rid is casse sensitive and needs to match through all other repos.
(and it does not by default)

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.DotNetAppHost/runtime.FreeBSD.Microsoft.NETCore.DotNetAppHost.props
+++ b/src/pkg/projects/Microsoft.NETCore.DotNetAppHost/runtime.FreeBSD.Microsoft.NETCore.DotNetAppHost.props
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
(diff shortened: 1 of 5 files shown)


eerhardt:
Note: the Tizen build has been failing. You can ignore that failure for now.

am11:
@wfurt, is there a plan to support other BSD-based operating systems at some point? I remember NetBSD team ported .NET Core v1.X in their ports (with some feature discrepancies). From that prespective, whouldn't `BSD` as a group name be most suitable (like `Windows_NT` and `Linux`):

```diff
<RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='Windows_NT'">win</RuntimeGraphGeneratorRuntime> 
<RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='OSX'">osx</RuntimeGraphGeneratorRuntime> 
+ <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='BSD'">bsd</RuntimeGraphGeneratorRuntime> 
<RuntimeGraphGeneratorRuntime Condition="'$(RuntimeGraphGeneratorRuntime)'==''">linux</RuntimeGraphGeneratorRuntime> 
```

Issue #4315, opened 2018-06-30
AV in breadcrumb_writer_t when exception is being thrown in main


pakrym:
This happens in AspNetCore module coreclr hosting scenario when the application that is being executed immediately throws an exception. 

I suspect the following happens: exception is getting thrown from `execute_assembly` call in `hostpolicy.cpp run` method while `breadcrumb_writer_t::write_callback` is still running, deconstructing `writer` local and freeing memory from under running loop and that causes AV.

![image](https://user-images.githubusercontent.com/1697911/42119007-7a1b1c18-7bbc-11e8-8b8d-a84ef252703c.png)

Stack trace:
```
 	KernelBase.dll!00007ffd048150d8()	Unknown
 	hostpolicy.dll!_CxxThrowException(void * pExceptionObject, const _s__ThrowInfo * pThrowInfo) Line 136	C++
 	hostpolicy.dll!__scrt_throw_std_bad_alloc() Line 36	C++
 	hostpolicy.dll!operator new(unsigned __int64 size) Line 53	C++
 	hostpolicy.dll!std::_Default_allocate_traits::_Allocate(const unsigned __int64 _Bytes) Line 53	C++
 	hostpolicy.dll!std::_Allocate_manually_vector_aligned<std::_Default_allocate_traits>(const unsigned __int64 _Bytes) Line 102	C++
 	hostpolicy.dll!std::_Allocate<16,std::_Default_allocate_traits,0>(const unsigned __int64 _Bytes) Line 195	C++
 	hostpolicy.dll!std::allocator<wchar_t>::allocate(const unsigned __int64 _Count) Line 1037	C++
 	hostpolicy.dll!std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >::_Construct_lv_contents(const std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> > & _Right) Line 2394	C++
 	hostpolicy.dll!std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >(const std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> > & _Right) Line 2114	C++
>	hostpolicy.dll!breadcrumb_writer_t::write_callback() Line 49	C++
[...]

PR #4346, opened 2018-07-10
Change main thread to join to breadcrumb thread when exception thrown

steveharter:
A custom host that catches exceptions currently hits an access violation during clr shutdown because of a background thread that references invalid memory.

Fixes https://github.com/dotnet/core-setup/issues/4315

A couple unit tests were added. However, they do not actually verify the AV is not thrown, as that would require significant infrastructure to host C++ native tests. Instead I manually verified that the access violation no longer occurs by:
1) Create a managed app that throws an exception
2) Create a custom host compiled with SEH
3) Invoke hostfxr's hostfxr_main() with a try\catch and launching the managed app from (1)
4) Verify that there is no longer an access violation and that the try\catch works properly

This will need to be ported to the 2.2 branch when that branch is created; a port issue will be logged.

cc @pakrym 
cc @jeffschwMSFT

### DIFF
--- a/src/corehost/cli/breadcrumbs.cpp
+++ b/src/corehost/cli/breadcrumbs.cpp
@@ -7,13 +7,22 @@
 #include "trace.h"
 #include "breadcrumbs.h"
 
-breadcrumb_writer_t::breadcrumb_writer_t(const std::unordered_set<pal::string_t>* files)
+breadcrumb_writer_t::breadcrumb_writer_t(bool enabled, const std::unordered_set<pal::string_t>* files)
     : m_status(false)
+    , m_enabled(enabled)
     , m_files(files)
 {
-    if (!pal::get_default_breadcrumb_store(&m_breadcrumb_store))
+    if (enabled && !pal::get_default_breadcrumb_store(&m_breadcrumb_store))
     {
         m_breadcrumb_store.clear();
+    }
+}
+
+breadcrumb_writer_t::~breadcrumb_writer_t()
+{
+    if (m_enabled)
+    {
+        end_write();
     }
 }
 
@@ -21,6 +30,7 @@ breadcrumb_writer_t::breadcrumb_writer_t(const std::unordered_set<pal::string_t>
 // thread to write breadcrumbs.
 void breadcrumb_writer_t::begin_write()
 {
+    assert(m_enabled);
     trace::verbose(_X("--- Begin breadcrumb write"));
     if (m_breadcrumb_store.empty())
     {
@@ -88,4 +98,3 @@ bool breadcrumb_writer_t::end_write()
     trace::verbose(_X("--- End breadcrumb write %d"), m_status);
     return m_status;
 }
-
--- a/src/corehost/cli/breadcrumbs.cpp
+++ b/src/corehost/cli/breadcrumbs.cpp
@@ -30,23 +30,25 @@ breadcrumb_writer_t::~breadcrumb_writer_t()
 // thread to write breadcrumbs.
 void breadcrumb_writer_t::begin_write()
 {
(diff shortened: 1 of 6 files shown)


steveharter:
Looks like issues with Linux; investigating...

src/corehost/cli/hostpolicy.cpp
```diff
@@ -275,7 +275,7 @@ int run(const arguments_t& args, pal::string_t* out_host_command_result = nullpt
     std::vector<char> managed_app;
     pal::pal_clrstring(args.managed_application, &managed_app);
 
-    breadcrumb_writer_t writer(&breadcrumbs);
+    breadcrumb_writer_t writer(breadcrumbs_enabled, &breadcrumbs);
     if (breadcrumbs_enabled)
```
eerhardt: Any thoughts on removing this check here too?  Then you wouldn't need to assert in `begin_write`.

```C++
breadcrumb_writer_t writer(breadcrumbs_enabled, &breadcrumbs);
writer.begin_write();
```

`begin_write` would check if it is enabled, and skip doing anything if it isn't.

src/test/HostActivationTests/GivenThatICareAboutNativeHostApis.cs
```diff
@@ -52,29 +52,91 @@ public void Muxer_activation_of_Publish_Output_Portable_DLL_hostfxr_get_native_s
...
+                .Execute(fExpectedToFail: true)
+                .Should()
+                .Fail()
+                .And
+                .HaveStdErrContaining("Unhandled Exception: System.Exception: Goodbye World")
+                .And
+                // The breadcrumb thread does not wait since destructors are not called when an exception is thrown.
+                // However, destructors will be called when the caller (such as a custom host) is compiled with SEH Exceptions (/EHa) and has a try\catch.
+                // Todo: add a native host test app so we can verify this behavior.
+                .NotHaveStdErrContaining("Waiting for breadcrumb thread to exit...");
```
eerhardt: Do you want to also assert that there ARE breadcrumb messages getting written?  Just not this one...

src/corehost/cli/hostpolicy.cpp
steveharter: Sure makes it a bit more consistent. Thanks

src/test/HostActivationTests/GivenThatICareAboutNativeHostApis.cs
steveharter: The test is to ensure the destructor is called in the happy path, and not called in the exception case (for the normal host that doesn't have SEH exception catching support), and checking for that string is sufficient.

Also I think adding a probe to see if a physical file would be too intrusive. The breadcrumb files are only written if they do not exist, so that would mean we'd have to delete existing files global files (or change the date on them to an earlier date, since a touch is done if they already exist).

steveharter:
@eerhardt @nguerrera any additional comments? Thanks

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -7,36 +7,48 @@
...
     {
         m_breadcrumb_store.clear();
     }
 }
 
+breadcrumb_writer_t::~breadcrumb_writer_t()
+{
+    if (m_enabled)
+    {
+        end_write();
```
nguerrera: Do we need to be concerned about an exception happening in here?

I am not an expert on this.

https://akrzemi1.wordpress.com/2011/09/21/destructors-that-throw/

steveharter: There are no expected exceptions in the breadcrumb dtor; the AV that was occurring in end_write should no longer occur.

PR #4347, opened 2018-07-10
Import build.common.props in publish.proj

wtgodbe:
This should allow us to correctly import the Microbuild targets, meaning packages will get signed correctly in official builds.

CC @dagood @weshaggard


pkg/publish.proj
```diff
@@ -8,6 +8,7 @@
   <Import Project="$(MSBuildThisFileDirectory)dir.props" />
   <Import Project="$(ToolsDir)PublishContent.targets" />
   <Import Project="$(ToolsDir)versioning.targets" />
+  <Import Project="$(ToolsDir)Build.Common.Props" />
```
weshaggard: This will work for our current scenarios but if we ever end up trying to sign somewhere else we might want to move this import into dir.props instead.

PR #4358, opened 2018-07-11
Set property IntermediateOutputPath for microbuild signing

wtgodbe:
My test build for signing failed because `IntermediatesDirectory` (which gets set to `IntermediateOutputPath`) was unset: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/Default/_build/results?buildId=1847834&view=logs

This defaults that property. CoreFx/CoreClr set the property to essentially `bin\obj` (https://github.com/dotnet/coreclr/blob/2d0167f7002643dc1bd77602805439fb889d39bf/dir.props#L104), and it does not appear to be used in any special way by the `SignFiles` target.

@weshaggard PTAL

### DIFF
--- a/pkg/publish.proj
+++ b/pkg/publish.proj
@@ -14,11 +14,12 @@
 
   <PropertyGroup>
     <PackageDownloadDirectory>$(MSBuildThisFileDirectory)packages\AzureTransfer\</PackageDownloadDirectory>
+    <!-- SignFiles needs IntermediateOutputPath & OutDir to be set -->
     <IntermediateOutputPath>$(MSBuildThisFileDirectory)bin\obj</IntermediateOutputPath>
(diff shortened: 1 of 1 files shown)


pkg/publish.proj
```diff
@@ -14,10 +14,12 @@
 
   <PropertyGroup>
     <PackageDownloadDirectory>$(MSBuildThisFileDirectory)packages\AzureTransfer\</PackageDownloadDirectory>
+    <!-- SignFiles needs IntermediateOutputPath & OutDir to be set -->
+    <IntermediateOutputPath>$(MSBuildThisFileDirectory)bin\obj</IntermediateOutputPath>
```
weshaggard: I thought it only needed OutDir but I guess we always had IntermediateOutputPath set in other contexts.

PR #4363, opened 2018-07-12
Bump version for PlatformAbstractions & DependencyModel

wtgodbe:
CC @eerhardt @weshaggard


src/Microsoft.DotNet.PlatformAbstractions/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.1.9",
+  "version": "1.1.10",
```
eerhardt: these need to be `1.1.10-*`

eerhardt: Or else they will be stable versions for every build.

PR #4367, opened 2018-07-13
Disable multilevel lookup during build

steveharter:
@vitek-karas ran into an issue where a "newer" version of the SDK was being used to build the tests, causing a test failure. The "newer" version was actually a 15.* version that we used temporary in an old preview build when we were planning on changing the version numbers.

If the global location has a newer SDK or runtime we should not use it during local build to ensure we are building and testing with the expected environment at `Tools\dotnetcli\sdk\<version>` and runtime at `Tools\dotnetcli\shared\Microsoft.NETCore.App\<version>`

Issue is listed as the 3rd item in https://github.com/dotnet/core-setup/issues/4099

Note: there may be test fallout; we at least need to agree to the build change first.

### DIFF
--- a/run.ps1
+++ b/run.ps1
@@ -5,6 +5,9 @@ $restorePackagesPath = Join-Path $PSScriptRoot "packages"
 # We do not want to run the first-time experience.
 $env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 1
 
+# Don't resolve runtime, shared framework, or SDK from other locations to ensure build determinism
+$env:DOTNET_MULTILEVEL_LOOKUP=0
+
 $initTools = Join-Path $PSScriptRoot "init-tools.cmd"
 & $initTools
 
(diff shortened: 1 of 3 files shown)


run.ps1
```diff
@@ -5,6 +5,9 @@ $restorePackagesPath = Join-Path $PSScriptRoot "packages"
 # We do not want to run the first-time experience.
 $env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 1
 
+# Don't resolve runtime, shared framework, or SDK from other locations to ensure build determinism
+$env:DOTNET_MULTILEVEL_LOOKUP=0
```
weshaggard: can you also add this in run.sh to keep things consistent.

src/test/HostActivationTests/GivenThatICareAboutDotnetArgValidationScenarios.cs
```diff
@@ -94,7 +94,7 @@ public void Detect_Missing_Argument_Value()
             dotnet.Exec("--fx-version")
                 .CaptureStdOut()
                 .CaptureStdErr()
-                .Execute()
```
steveharter: This change is unrelated to the environment variable change; however still address test usability via #4099

run.ps1
steveharter: Yes. Thanks

steveharter:
Note there was no test fallout; the tests will now run without multi-level lookup but no current test depended upon that behavior. There will now be no lines in the COREHOST_TRACE=1 output like this (for Windows): _Searching FX directory in [C:\Program Files\dotnet]_

nguerrera:
FYI, I'm adding some tests that depend on mutlilevel lookup being set, but they force it and also force the program files environment variable to a mock location. They should be unaffected by this.

PR #4385, opened 2018-07-19
Crossgen with PartialNGEN enabled

adiaaida:
With this change, we will PartialNGEN all of the NetCoreApp assemblies. This change is a companion to https://github.com/dotnet/buildtools/pull/2103, which enables the consumption of IBC data for corefx. These two changes together will move us to only precompiling functions for which we have collected IBC data.

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -295,7 +295,7 @@
 
     <MakeDir Directories="$(_crossGenSymbolsOutputDirectory)" />
 
-    <Exec Command="$(_crossGenPath) @$(_crossGenSymbolsResponseFile)" WorkingDirectory="$(_clrDirectory)" EnvironmentVariables="COMPlus_PartialNGen=0" />
+    <Exec Command="$(_crossGenPath) @$(_crossGenSymbolsResponseFile)" WorkingDirectory="$(_clrDirectory)" EnvironmentVariables="COMPlus_PartialNGen=1" />
 
     <Touch Files="%(_filesToCrossGen.CrossGenSymbolSemaphorePath)" AlwaysCreate="true" />
   </Target>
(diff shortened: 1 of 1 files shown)


adiaaida:
@dotnet/rap-team

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -263,12 +263,16 @@
       <_crossGenArgs Include="-platform_assemblies_paths $(_crossgenPlatformAssemblies)" />
       <_crossGenArgs Include="-JITPath $(_jitPath)" />
     </ItemGroup>
+    <PropertyGroup>
+      <_partialNgen Condition="'$(OS)' == 'Windows_NT'">1</_partialNgen>
+      <_partialNgen Condition="'$(OS)' != 'Windows_NT'">0</_partialNgen>
```
brianrob: Can you share why we're only enabling partial NGEN for Windows and not UNIX?

adiaaida: On Windows, we will have IBC data, while on linux we will not. When we have ibc data, we want to partial crossgen the entire framework to get the size benefits (so the assemblies with ibc data will be partial ngened, where as the ones without will be IL). On Linux, where we won't have ibc data yet, we don't want to stop crossgening the entire framework, so we don't want to use partial ngen.

adiaaida:
@brianrob to answer your questions:

1. I have not confirmed that the official build will ultimately match the results that I have collected, but I think we will need to check in before we can confirm the results.
2. All assemblies in the framework will be crossgened with PartialNGEN, if they have IBC data or not. If there is no IBC data, it will not crossgen, and the resulting assembly will just be IL.

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
brianrob: Gotcha.  So, this is just a point-in-time thing?

brianrob:
Thanks @adiaaida.  Yes, we should validate the results, but I think it's OK to validate them after they get merged and we get an official build.

adiaaida:
@dotnet-bot test Tizen armel Release Build please

adiaaida:
@brianrob am I good to check this one in?

adiaaida:
@brianrob Are we good here?

brianrob:
I forgot to post a note here.  We are purposely waiting to merge this change until we are able to do some analysis to confirm that we are going to get the wins we want here without taking to much of a hit elsewhere.  We also need to confirm that this change and tiered jitting play nicely together.  I'll report back once we have done this work, but until then, this PR should not get merged.

PR #4394, opened 2018-07-23
Remove BranchName

eerhardt:
I can't find any usages of this property anymore.

### DIFF
--- a/BranchInfo.props
+++ b/BranchInfo.props
@@ -14,7 +14,6 @@
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
     <ReleaseBrandSuffix>Preview 1</ReleaseBrandSuffix>
     <Channel>master</Channel>
-    <BranchName>master</BranchName>
     <ContainerName>dotnet</ContainerName>
     <ChecksumContainerName>$(ContainerName)</ChecksumContainerName>
     <NETCoreAppMaximumVersion>$(MajorVersion).$(MinorVersion)</NETCoreAppMaximumVersion>


BranchInfo.props
```diff
@@ -14,7 +14,6 @@
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
     <ReleaseBrandSuffix>Preview 1</ReleaseBrandSuffix>
     <Channel>master</Channel>
```
weshaggard: Do you think we can get away with using SourceBranch for this as well to eliminate the need for maintaining it across branches?

eerhardt: In theory, the `Channel` can and should be independent of the branch of the repo.  For example, we *should* have a `3.0` channel right now, and the `master` branch should be outputting to that channel.  See https://github.com/dotnet/versions#channels for more info.

In practice, I've almost never seen this done.

weshaggard: I understand the theory but I'm just wonder if we shouldn't give it a default that allows for us to not have to think or worry about it when we create new branches. So if we default it to SourceBranch it will help prevent collisions but folks can potentially still change it to whatever. FWIW there is going to become more of a channel system with the new maestro work happening in arcade.

eerhardt: I'm fine with someone moving this to use `$(SourceBranch)` instead.

eerhardt:
Any concerns on merging this?

BranchInfo.props
weshaggard: Thanks for cleaning up the BranchName... to bad you didn't take the bait and fix the other one as well :)

PR #4397, opened 2018-07-23
Port ReleaseTools changes to release/1.1.0

wtgodbe:

This will also require passing PB_InternalBuild=true from the Internal pipebuild.

@dagood @weshaggard PTAL


wtgodbe:
Updated to address feedback from https://github.com/dotnet/core-setup/pull/4395

PR #4399, opened 2018-07-23
Port ReleaseTools changes to release/2.0.0

wtgodbe:

This will also require passing the following from the internal pipebuild:
 `PB_InternalBuild=true` 
`OverridePackageSource=/p:OverridePackageSource=$(PB_GitDirectory)/AzureCoreFxDownload/Release`
`PB_GitDirectory=/root/coresetup`

And the following from the non-internal pipebuild:
`CLI_NUGET_FEED_URL=https:%2F%2Fdotnet.myget.org/F/cli-deps/api/v2/package`
`NUGET_FEED_URL=https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v2/package`
`NUGET_SYMBOLS_FEED_URL=https:%2F%2Fdotnet.myget.org/F/dotnet-core/symbols/api/v2/package`

@dagood @weshaggard PTAL

### DIFF
--- a/buildpipeline/Core-Setup-Linux-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Linux-Arm-BT.json
@@ -24,6 +24,24 @@
     {
       "enabled": true,
       "continueOnError": true,
+      "alwaysRun": true,
+      "displayName": "Remove Release Tools directory",
+      "timeoutInMinutes": 0,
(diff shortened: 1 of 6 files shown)


wtgodbe:
@dagood yeah I don't love it either, but I couldn't think of a cleaner fix

Issue #4200, opened 2018-06-06
The core_setup README.md has broken "version" links


steveharter:
The 2.x “version” links are broken on https://github.com/dotnet/core-setup#daily-builds

For example, for Windows (x64) when I click on the “version” for the winx64 it tries to go to https://dotnetcli.blob.core.windows.net/dotnet/Runtime/master/latest.sharedfx.win.x64.version which is broke \ missing blob.

Expect: it used to go to the build information I believe, or at least provide more information on the version (e.g. hash).

Did the blob names change or the .version file stop being generated? I don't see any recent README.md changes that that would affect this.

steveharter:
cc @weshaggard

weshaggard:
@wtgodbe could you please make a pass over the readme file and update the links to where things live in the storage accounts?

PR #4406, opened 2018-07-25
Update version links for Readme.md

wtgodbe:
This updates the links for the Version badges to point to `latest.version` for 2.0.0, 2.1.0, and master, as that is the only `.version` file present in the relevant blob stores.

CC @weshaggard @steveharter 

For https://github.com/dotnet/core-setup/issues/4200


weshaggard:
Thanks @wtgodbe I just assigned another issue about a broken link in the readme as well. Could you try to update that as well. Also it'd be good to check all the links to make sure they are updated and work.

PR #4409, opened 2018-07-26
Reupload signed packages to intermediate Azure drop

wtgodbe:
Part of dotnet/core-eng#3872.

We want all package drops to contain signed packages - right now, the intermediate Azure drops wind up containing unsigned packages, which can cause issues during the final Nuget.org push.

@weshaggard @dagood @MattGal @eerhardt PTAL

### DIFF
--- a/publish/publish.proj
+++ b/publish/publish.proj
@@ -154,7 +154,7 @@
   </PropertyGroup>
 
   <Target Name="PublishCoreHostPackagesToFeed"
-          DependsOnTargets="CheckIfAllBuildsHavePublished;DownloadCoreHostPackagesFromAzure;SignPackages;PushSignedCoreHostPackagesToFeed"
+          DependsOnTargets="CheckIfAllBuildsHavePublished;DownloadCoreHostPackagesFromAzure;SignPackages;PushSignedCoreHostPackagesToAzure;PushSignedCoreHostPackagesToFeed"
           Condition="'@(_MissingBlobNames)' == '' AND '$(NuGetFeedUrl)' != ''"/>
 
   <Target Name="DownloadCoreHostPackagesFromAzure">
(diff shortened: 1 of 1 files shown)


publish/publish.proj
```diff
@@ -209,6 +209,26 @@
           DependsOnTargets="GetPackagesToSign;SignFiles">
   </Target>
 
+  <Target Name="PushSignedCoreHostPackagesToAzure">
+    <Message Text="Re-uploading signed CoreHost packages to Azure" />
```
eerhardt: Why don't we sign the packages before we upload them to Azure in the first place?

dagood: We can only sign on Windows, unfortunately.

eerhardt: 👎

weshaggard: lol my sentiment exactly :)

publish/publish.proj
```diff
@@ -209,6 +209,26 @@
...
+
+    <ItemGroup>
+      <ItemsToPublish Include="@(_DownloadedStandardPackages)" />
```
eerhardt: I think we should probably use a different item name than `@(ItemsToPublish)`.  It looks like there are other targets in this file that use that Item name as well, and we could be publishing more than we intend.

So we should either:
1. clear the Item out, and then start adding to it (not my recommendation)
2. Use a unique Item name (my recommendation)

publish/publish.proj
```diff
@@ -209,6 +209,26 @@
...
+    </ItemGroup>
+    <!-- add relative blob path metadata -->
```
eerhardt: I don't think this comment provides value.

Also, these two operations don't need to be split into different `ItemGroup`s.  They can just be inside the same `<ItemGroup>`.

publish/publish.proj
wtgodbe: Good call, I've just pushed an update to name the Item `SignedPackagesToPublish`

publish/publish.proj
wtgodbe: Updated

PR #4421, opened 2018-08-01
Add startup hook design document

sbomer:
This adds a proposal for an early startup hook that allows managed
code to run before the entry point.

@jeffschwMSFT @vitek-karas @swaroop-sridhar @AaronRobinsonMSFT @richlander @jkotas

### DIFF
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -198,7 +198,8 @@ method, so thread state will persist between startup hooks. The
 threading apartment state will be set based on any attributes present
 in the `Main` method of the app, before startup hooks execute. As a
 result, attemps to explicitly set the thread apartment state in a
-startup hook will likely fail.
+startup hook will fail if the requested state is incompatible with the
+app's threading state.
 
 While it may make sense to set global behavior in startup hooks, it is
 not recommended to use the thread state as a communication mechanism
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -20,33 +20,41 @@ point, so that user code doesn't need to be modified.
 ## Proposed behavior
 
 The `DOTNET_STARTUP_HOOKS` environment variable can be used to specify
-a list of managed assemblies and type names that contain `public
-static void Initialize()` methods, which will be called in the order
-specified, before the `Main` entry point:
+a list of managed assemblies that contain a `StartupHook` type with a
+`public static void Initialize()` method, each of which will be called
+in the order specified, before the `Main` entry point
 
+Unix:
 ```
-DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll!StartupHookNamespace.StartupHookType1;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll:/path/to/StartupHook2.dll
 ```
 
-This variable is a list of absolute assembly paths, each with a type
-name following an exclamation mark. The list is delimited by the
+Windows:
+```
+DOTNET_STARTUP_HOOKS=D:\path\to\StartupHook1.dll;D:\path\to\StartupHook2.dll
+```
+
+This variable is a list of absolute assembly paths, delimited by the
 platform-specific path separator (`;` on Windows and `:` on Unix). It
-may not contain any empty entries or a trailing path separator.
+may not contain any empty entries or a trailing path separator. The
+type must be named `StartupHook` without any namespace, and should be
+`internal`.
 
-Setting this environment variable will cause each of the specified
-types' `public static void Initialize()` methods to be called in
-order, synchronously, before the main assembly is loaded. The hooks
-are all called on the same managed thread (the same thread that calls
-`Main`). The environment variable will be inherited by child processes
-by default. It is up to the `StartupHook.dll`s and user code to decide
-what to do about this - `StartupHook.dll` may clear them to prevent
-this behavior globally, if desired.
+Setting this environment variable will cause the `public static void
+Initialize()` method of the `StartupHook` type in each of the
+specified assemblies to be called in order, synchronously, before the
+main assembly is loaded. The hooks are all called on the same managed
+thread (the same thread that calls `Main`). The environment variable
+will be inherited by child processes by default. It is up to the
+`StartupHook.dll`s and user code to decide what to do about this -
+`StartupHook.dll` may clear them to prevent this behavior globally, if
+desired.
 
 Specifically, hostpolicy starts up coreclr and sets up a new
 AppDomain. It then invokes a private method in
 `System.Private.CoreLib`, which will call each
-`StartupHookType.Initialize()` in turn synchronously. This gives
-`StartupHookType` a chance to set up new `AssemblyLoadContext`s, or
+`StartupHook.Initialize()` in turn synchronously. This gives
+`StartupHook` a chance to set up new `AssemblyLoadContext`s, or
 register other callbacks. After all of the `Initialize()` methods
 return, control returns to hostpolicy, which then calls the main entry
 point of the app like usual.
@@ -72,17 +80,21 @@ assemblies that are on the TPA list from a different location. Future
 changes to `AssemblyLoadContext` could make this easier to use by
 making the default load context or TPA list modifiable.
 
+Note that the `StartupHook` type is internal and in the global
+namespace, and the signature of the `Initialize` method is `public
+static void Initialize()`.
+
 ```c#
+internal class StartupHook
+{
+    public static void Initialize()
+    {
+        AssemblyLoadContext.Default.Resolving += SharedHostPolicy.SharedAssemblyResolver.LoadAssemblyFromSharedLocation;
+    }
+}
+
 namespace SharedHostPolicy
 {
-    class SharedHostInitializer
-    {
-        public static void Initialize()
-        {
-            AssemblyLoadContext.Default.Resolving += SharedAssemblyResolver.LoadAssemblyFromSharedLocation;
-        }
-    }
-
     class SharedAssemblyResolver
     {
         public static Assembly LoadAssemblyFromSharedLocation(AssemblyLoadContext context, AssemblyName assemblyName)
@@ -99,7 +111,7 @@ namespace SharedHostPolicy
 ## Error handling details
 
 Problems with the startup hook should be fairly straightforward to
-diagnose. All of these exceptionts will contain the startup hook path
+diagnose. All of these exceptions will contain the startup hook path
 (`System.StartupHookProvider.ProcessStartupHooks`) on the stack
 trace. They fall into the following categories:
 
@@ -126,6 +138,11 @@ trace. They fall into the following categories:
   - Invalid `Initialize` methods (with an incorrect signature - that
     take parameters, have a non-void return type, are not public, or
     are not static) throw an `ArgumentException`.
+
+  - Unhandled exceptions thrown from a startup hook will have the same
+    exception behavior as any other managed exception thrown from
+    `Main` - by default, they will terminate the process and show a
+    stack trace.
 
 ## Guidance and caveats
 
@@ -175,7 +192,7 @@ compiled against a different version.
 
 ### Threading behavior
 
-Each startup hook with run on the same managed thread as the `Main`
+Each startup hook will run on the same managed thread as the `Main`
 method, so thread state will persist between startup hooks. While it
 may make sense to set global behavior in startup hooks, it is not
 recommended to use the thread state as a communication mechanism
@@ -199,3 +216,9 @@ be defined either in the app or within the first hook that uses it:
   need to communicate with each other part of a common plugin
   framework. If necessary, the plugin host could be injected into the
   process with a single startup hook.
+
+### Visibility of `StartupHook` type
+
+The type should be made `internal` to prevent exposing it as API
+surface to any managed code that happens to have access to the startup
+hook dll. However, the feature will also work if the type is `public`.
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -1,72 +1,74 @@
 # Host startup hook
 
-For .NET Core 3+, we want to provide a hook that allows managed code
-to run before the main application's entry point.
+For .NET Core 3+, we want to provide a low-level hook that allows
+injecting managed code to run before the main application's entry
+point. This hook will make it possible for the host to customize the
+behavior of managed applications during process launch, after they
+have been deployed.
 
 ## Motivation
 
-This would allow hosting providers to put configuration and policy in
-a single location, including settings that potentially influence load
-behavior of the main entry point such as the `AssemblyLoadContext`
-behavior. This could be used to set up callbacks for handling
+This would allow hosting providers to define custom configuration and
+policy in managed code, including settings that potentially influence
+load behavior of the main entry point such as the
+`AssemblyLoadContext` behavior. The hook could be used to set up
+tracing or telemetry injection, to set up callbacks for handling
 Debug.Assert (if we make such an API available), or other
-environment-dependent behavior. The hook needs to be separate from the
-entry point, so that user code doesn't need to be modified.
+environment-dependent behavior. The hook is separate from the entry
+point, so that user code doesn't need to be modified.
 
 ## Proposed behavior
 
-Environment variables or `<appname>.runtimeconfig.json` can be used to
-specify a managed assembly and type that contains a `public void
-Initialize()` method.
+The `DOTNET_STARTUP_HOOKS` environment variable can be used to specify
+a list of managed assemblies and optional type names that contain
+`public void Initialize()` methods, which will be called in the order
+specified, before the `Main` entry point:
 
 ```
-DOTNET_MANAGED_HOST_ASSEMBLY=/path/to/ManagedHost.dll
-DOTNET_MANAGED_HOST_TYPE=ManagedHostNamespace.ManagedHostType
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
 ```
 
-```
-{
-    "runtimeOptions": {
-        "managedHostAssembly": "/path/to/ManagedHost.dll",
-        "managedHostType": "ManagedHostNamespace.ManagedHostType"
-    }
-}
-```
+The list is a semicolon-delimited list of assembly paths (absolute, or
+relative to the working directory), each with an optional type name
+(following an exclamation mark). If no type name is specified, the
+default is `StartupHook.Initializer`.
 
-The environment variables, if set, take precedence over the
-`<appname>.runtimeconfig.json` settings. These settings result in
-`ManagedHostType.Initialize()` being called in hostpolicy, before the
-main assembly is loaded. If it read these settings from environment
-variables, they will be inherited by child processes by default. It is
-up to the ManagedHost.dll and user code to decide what to do about
-this - ManagedHost.dll may clear them to prevent this behavior
-globally, if desired.
+Setting this environment variable will cause each of the specified
+types' `public void Initialize()` method to be called before the main
+assembly is loaded. The environment variable will be inherited by
+child processes by default. It is up to the `StartupHook.dll`s and
+user code to decide what to do about this - `StartupHook.dll` may
+clear them to prevent this behavior globally, if desired.
 
 Specifically, hostpolicy starts up coreclr and sets up a new AppDomain
-with `ManagedHost.dll` on the TPA list. It then invokes
-`ManagedHost.Initialize()`. This gives `ManagedHost` a chance to set
-up new `AssemblyLoadContext`s, or register other callbacks. After
-`Initialize()` returns, hostpolicy starts up the main entry point of
-the app like usual.
+with each `StartupHook.dll` on the TPA list. It then invokes a private
+method in `System.Private.CoreLib`, which will call each
+`StartupHookType.Initialize()` in turn. This gives `StartupHookType` a
+chance to set up new `AssemblyLoadContext`s, or register other
+callbacks. After the `Initialize()` methods return, control returns to
+hostpolicy, which starts up the main entry point of the app like
+usual.
 
 Rather than forcing all configuration to be done through a single
 predefined API, this creates a place where such configuration could be
 centralized, while still allowing user code to do its own thing if it
 so desires.
 
-The producer of `ManagedHost.dll` needs to ensure that
-`ManagedHost.dll` is compatible with the dependencies specified in the
+The producer of `StartupHook.dll` needs to ensure that
+`StartupHook.dll` is compatible with the dependencies specified in the
 main application's deps.json, since those dependencies are put on the
-TPA list during the runtime startup, before `ManagedHost.dll` is
-loaded. This means that `ManagedHost.dll` needs to built against the
+TPA list during the runtime startup, before `StartupHook.dll` is
+loaded. This means that `StartupHook.dll` needs to built against the
 same or lower version of .NET Core than the app.
 
 ## Example
 
 This could be used with `AssemblyLoadContext` APIs to resolve
-non-framework dependencies from a shared location, similar to the GAC
-on full framework. Future changes to `AssemblyLoadContext` could make
-this easier to use by making the default load context modifiable.
+dependencies not on the TPA list from a shared location, similar to
+the GAC on full framework. It could also be used to forcibly preload
+assemblies that are on the TPA list from a different location. Future
+changes to `AssemblyLoadContext` could make this easier to use by
+making the default load context or TPA list modifiable.
 
 ```
 namespace SharedHostPolicy
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -72,7 +72,7 @@ assemblies that are on the TPA list from a different location. Future
 changes to `AssemblyLoadContext` could make this easier to use by
 making the default load context or TPA list modifiable.
 
-```
+```c#
 namespace SharedHostPolicy
 {
     class SharedHostInitializer
@@ -106,36 +106,26 @@ trace. They fall into the following categories:
 - Errors detected eagerly, with exceptions thrown before the execution
   of any startup hook.
 
--- Invalid syntax. This will throw an `ArgumentException` with "The
-  syntax of the startup hook variable was invalid.".
+  - Invalid syntax throws an `ArgumentException`.
 
--- Partially qualified path in the startup hook. This will throw an
-   `ArgumentException` with "Absolute path information is required.".
+  - Partially qualified paths in the startup hook throw an
+    `ArgumentException`.
 
 - Exceptions thrown during the call to a given startup hook. Previous
   hooks may have run successfully.
 
--- Missing startup hook assembly. This will throw a
-   `FileNotFoundException` with "Could not load file or assembly
-   '<assemblyPath>'".
+  - Missing startup hook assemblies throw a `FileNotFoundException`.
 
--- Invalid startup hook assembly. This will throw a
-   `BadImageFormatException` with the startup hook path on the stack
-   trace.
+  - Invalid startup hook assemblies throw a `BadImageFormatException`.
 
--- Missing startup hook type. This will throw a `TypeLoadException`
-   with "Could not load type '<specifiedType>' from assembly
-   '<assemblyName>'".
+  - Missing startup hook types throw a `TypeLoadException`.
 
--- Missing `Initialize` method in startup hook. This will throw a
-   `MissingMethodException` with "Method '<specifiedType>.Initialize'
-   not found.".
+  - Missing `Initialize` methods in startup hooks throw a
+    `MissingMethodException`.
 
--- Invalid `Initialize` method (with an incorrect signature - one that
-   takes parameters, has a non-void return type, is not public, or is
-   not static). This will throw an `ArgumentException` with "The
-   signature of the startup hook '<specifiedType>.Initialize was
-   invalid. It must be 'public static void Initialize()'.
+  - Invalid `Initialize` methods (with an incorrect signature - that
+    take parameters, have a non-void return type, are not public, or
+    are not static) throw an `ArgumentException`.
 
 ## Guidance and caveats
 
@@ -145,12 +135,11 @@ need for this kind of power. It should only be used in situations
 where modifying application code is not an option and there is not an
 existing structured dependency injection framework in place. An
 example of such a use case is a hook that injects logging, telemetry,
-or profiling into an existing deployed application at run tim.e
+or profiling into an existing deployed application at runtime.
 
 It is prone to ordering issues when multiple hooks are used, and does
 nothing to attempt to make dependencies of hooks easy to
-manage. Multiple hooks, if used, should be fairly independent of each
-other.
+manage. Multiple hooks should be independent of each other.
 
 ### No built-in solution to ordering issues
 
@@ -163,9 +152,9 @@ framework that gives independently-owned components access to shared
 resources - often dependency injection frameworks will have a
 dependency manager that loads components in a specific order. If this
 kind of behavior is required, a proper dependency injection framework
-should be used instead of multpile startup hooks.
+should be used instead of multiple startup hooks.
 
-## No dependency resolution for non-app assemblies
+### No dependency resolution for non-app assemblies
 
 Another example regarding hook dependencies: the startup hook dll must
 not depend on any assemblies outside of the app's TPA list. If a
@@ -176,7 +165,7 @@ startup hooks. Any startup hook that wants to modify load behavior
 will have to use framework APIs like AssemblyLoadContexts to do this
 manually.
 
-## No conflict resolution for dependencies shared by hooks or the app
+### No conflict resolution for dependencies shared by hooks or the app
 
 If a startup hook decides to do something dangerous like force the
 load of a particular assembly, any later hooks (or the entry point)
@@ -184,7 +173,7 @@ that run in the same AssemblyLoadContext and depend on that assembly
 will use the version that was forcefully loaded, even if they were
 compiled against a different version.
 
-## Threading behavior
+### Threading behavior
 
 Each startup hook with run on the same managed thread as the `Main`
 method, so thread state will persist between startup hooks. While it
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -7,7 +7,7 @@ to run before the main application's entry point.
 
 This would allow hosting providers to put configuration and policy in
 a single location, including settings that potentially influence load
-behavior of the main entry point such as the AssemblyLoadContext
+behavior of the main entry point such as the `AssemblyLoadContext`
 behavior. This could be used to set up callbacks for handling
 Debug.Assert (if we make such an API available), or other
 environment-dependent behavior. The hook needs to be separate from the
@@ -16,8 +16,8 @@ entry point, so that user code doesn't need to be modified.
 ## Proposed behavior
 
 Environment variables or `<appname>.runtimeconfig.json` can be used to
-specify a managed assembly and type that contains an `Initialize`
-method.
+specify a managed assembly and type that contains a `public void
+Initialize()` method.
 
 ```
 DOTNET_MANAGED_HOST_ASSEMBLY=/path/to/ManagedHost.dll
@@ -45,7 +45,7 @@ globally, if desired.
 Specifically, hostpolicy starts up coreclr and sets up a new AppDomain
 with `ManagedHost.dll` on the TPA list. It then invokes
 `ManagedHost.Initialize()`. This gives `ManagedHost` a chance to set
-up new AssemblyLoadContexts, or register other callbacks. After
+up new `AssemblyLoadContext`s, or register other callbacks. After
 `Initialize()` returns, hostpolicy starts up the main entry point of
 the app like usual.
 
@@ -63,9 +63,9 @@ same or lower version of .NET Core than the app.
 
 ## Example
 
-This could be used with AssemblyLoadContext APIs to resolve
+This could be used with `AssemblyLoadContext` APIs to resolve
 non-framework dependencies from a shared location, similar to the GAC
-on full framework. Future changes to AssemblyLoadContext could make
+on full framework. Future changes to `AssemblyLoadContext` could make
 this easier to use by making the default load context modifiable.
 
 ```
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -51,13 +51,14 @@ will be inherited by child processes by default. It is up to the
 desired.
 
 Specifically, hostpolicy starts up coreclr and sets up a new
-AppDomain. It then invokes a private method in
-`System.Private.CoreLib`, which will call each
+AppDomain, passing in the startup hook variable if it was
+set. Hostpolicy then asks the runtime to execute the main method.
+Just before the main method is called, the runtime will call a private
+method in `System.Private.CoreLib`, which will call each
 `StartupHook.Initialize()` in turn synchronously. This gives
 `StartupHook` a chance to set up new `AssemblyLoadContext`s, or
 register other callbacks. After all of the `Initialize()` methods
-return, control returns to hostpolicy, which then calls the main entry
-point of the app like usual.
+return, the runtime calls the main entry point of the app like usual.
 
 Rather than forcing all configuration to be done through a single
 predefined API, this creates a place where such configuration could be
@@ -193,9 +194,14 @@ compiled against a different version.
 ### Threading behavior
 
 Each startup hook will run on the same managed thread as the `Main`
-method, so thread state will persist between startup hooks. While it
-may make sense to set global behavior in startup hooks, it is not
-recommended to use the thread state as a communication mechanism
+method, so thread state will persist between startup hooks. The
+threading apartment state will be set based on any attributes present
+in the `Main` method of the app, before startup hooks execute. As a
+result, attemps to explicitly set the thread apartment state in a
+startup hook will likely fail.
+
+While it may make sense to set global behavior in startup hooks, it is
+not recommended to use the thread state as a communication mechanism
 between startup hooks. Any setup that requires multiple communicating
 hooks should consider using a plugin system instead.
 
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -20,34 +20,34 @@ point, so that user code doesn't need to be modified.
 ## Proposed behavior
 
 The `DOTNET_STARTUP_HOOKS` environment variable can be used to specify
-a list of managed assemblies and optional type names that contain
-`public void Initialize()` methods, which will be called in the order
+a list of managed assemblies and type names that contain `public
+static void Initialize()` methods, which will be called in the order
 specified, before the `Main` entry point:
 
 ```
-DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll!StartupHookNamespace.StartupHookType1;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
 ```
 
 The list is a semicolon-delimited list of assembly paths (absolute, or
-relative to the working directory), each with an optional type name
-(following an exclamation mark). If no type name is specified, the
-default is `StartupHook.Initializer`.
+relative to the working directory), each with a type name following an
+exclamation mark.
 
 Setting this environment variable will cause each of the specified
-types' `public void Initialize()` method to be called before the main
-assembly is loaded. The environment variable will be inherited by
-child processes by default. It is up to the `StartupHook.dll`s and
-user code to decide what to do about this - `StartupHook.dll` may
-clear them to prevent this behavior globally, if desired.
+types' `public static void Initialize()` methods to be called in
+order, synchronously, before the main assembly is loaded. The
+environment variable will be inherited by child processes by
+default. It is up to the `StartupHook.dll`s and user code to decide
+what to do about this - `StartupHook.dll` may clear them to prevent
+this behavior globally, if desired.
 
 Specifically, hostpolicy starts up coreclr and sets up a new AppDomain
 with each `StartupHook.dll` on the TPA list. It then invokes a private
 method in `System.Private.CoreLib`, which will call each
-`StartupHookType.Initialize()` in turn. This gives `StartupHookType` a
-chance to set up new `AssemblyLoadContext`s, or register other
-callbacks. After the `Initialize()` methods return, control returns to
-hostpolicy, which starts up the main entry point of the app like
-usual.
+`StartupHookType.Initialize()` in turn synchronously. This gives
+`StartupHookType` a chance to set up new `AssemblyLoadContext`s, or
+register other callbacks. After all of the `Initialize()` methods
+return, control returns to hostpolicy, which then calls the main entry
+point of the app like usual.
 
 Rather than forcing all configuration to be done through a single
 predefined API, this creates a place where such configuration could be
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -36,12 +36,14 @@ DOTNET_MANAGED_HOST_TYPE=ManagedHostNamespace.ManagedHostType
 The environment variables, if set, take precedence over the
 `<appname>.runtimeconfig.json` settings. These settings result in
 `ManagedHostType.Initialize()` being called in hostpolicy, before the
-main assembly is loaded.
+main assembly is loaded. If it read these settings from environment
+variables, they will be inherited by child processes by default. It is
+up to the ManagedHost.dll and user code to decide what to do about
+this - ManagedHost.dll may clear them to prevent this behavior
+globally, if desired.
 
 Specifically, hostpolicy starts up coreclr and sets up a new AppDomain
-with `ManagedHost.dll` on the TPA list. If it read these settings from
-environment variables, it clears them so that child processes do not
-inherit these settings. It then invokes
+with `ManagedHost.dll` on the TPA list. It then invokes
 `ManagedHost.Initialize()`. This gives `ManagedHost` a chance to set
 up new AssemblyLoadContexts, or register other callbacks. After
 `Initialize()` returns, hostpolicy starts up the main entry point of
@@ -55,7 +57,9 @@ so desires.
 The producer of `ManagedHost.dll` needs to ensure that
 `ManagedHost.dll` is compatible with the dependencies specified in the
 main application's deps.json, since those dependencies are put on the
-TPA list when the runtime is started.
+TPA list during the runtime startup, before `ManagedHost.dll` is
+loaded. This means that `ManagedHost.dll` needs to built against the
+same or lower version of .NET Core than the app.
 
 ## Example
 
--- a/Documentation/design-docs/host-startup-hook.md
+++ b/Documentation/design-docs/host-startup-hook.md
@@ -35,9 +35,10 @@ may not contain any empty entries or a trailing path separator.
 
 Setting this environment variable will cause each of the specified
 types' `public static void Initialize()` methods to be called in
-order, synchronously, before the main assembly is loaded. The
-environment variable will be inherited by child processes by
-default. It is up to the `StartupHook.dll`s and user code to decide
+order, synchronously, before the main assembly is loaded. The hooks
+are all called on the same managed thread (the same thread that calls
+`Main`). The environment variable will be inherited by child processes
+by default. It is up to the `StartupHook.dll`s and user code to decide
 what to do about this - `StartupHook.dll` may clear them to prevent
 this behavior globally, if desired.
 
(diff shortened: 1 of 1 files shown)


jkotas:
@brianrob Does this look like something you would want to use to injecting Performance Controller into existing apps?

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+The environment variables, if set, take precedence over the
+`<appname>.runtimeconfig.json` settings. These settings result in
+`ManagedHostType.Initialize()` being called in hostpolicy, before the
+main assembly is loaded.
+
+Specifically, hostpolicy starts up coreclr and sets up a new AppDomain
+with `ManagedHost.dll` on the TPA list. If it read these settings from
+environment variables, it clears them so that child processes do not
```
jkotas: I think that clearing the variable may or may not be the desired behavior.  It depends on what folks are using this for.

jkotas: `DOTNET_ROOT` environment variable is similar to this - it has similar problems. We are not clearing that one either.

jeffschwMSFT: +1 I think we should leave them as is.  The developer can either use runtimeconfig or manually not inherit the env block on new process launches if desired.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,89 @@
...
+so desires.
+
+The producer of `ManagedHost.dll` needs to ensure that
+`ManagedHost.dll` is compatible with the dependencies specified in the
+main application's deps.json, since those dependencies are put on the
+TPA list when the runtime is started.
+
+## Example
+
+This could be used with AssemblyLoadContext APIs to resolve
```
jkotas: This is the service fabric example. Do we expect that service fabric will use the environment variable or config file setting?

If it is going to use the environment variable, where are they going to set the environment variable? It should not be just set globally so that non-service fabric processes are left unaffected.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,89 @@
...
+
+Rather than forcing all configuration to be done through a single
+predefined API, this creates a place where such configuration could be
+centralized, while still allowing user code to do its own thing if it
+so desires.
+
+The producer of `ManagedHost.dll` needs to ensure that
+`ManagedHost.dll` is compatible with the dependencies specified in the
+main application's deps.json, since those dependencies are put on the
+TPA list when the runtime is started.
```
jeffschwMSFT: I assume that the only additional dependency we would put on the tpa is managedhost.dll.  Is this comment calling out that if managedhost.dll has additional dependencies they need to be satisfied by the original deps.json?

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+DOTNET_MANAGED_HOST_ASSEMBLY=/path/to/ManagedHost.dll
+DOTNET_MANAGED_HOST_TYPE=ManagedHostNamespace.ManagedHostType
+```
+
+```
+{
+    "runtimeOptions": {
+        "managedHostAssembly": "/path/to/ManagedHost.dll",
```
jkotas: We may want to allow the case where the assembly is part of the app. 

It may be the best option for the service fabric case. Service fabric will drop a small ServiceFabricLoader.dll into each app and add it to .runtime.config.json.

This would save them from messing with environment variables, and also make the same app for across different machines.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,89 @@
...
+up new AssemblyLoadContexts, or register other callbacks. After
+`Initialize()` returns, hostpolicy starts up the main entry point of
+the app like usual.
+
+Rather than forcing all configuration to be done through a single
+predefined API, this creates a place where such configuration could be
+centralized, while still allowing user code to do its own thing if it
+so desires.
+
+The producer of `ManagedHost.dll` needs to ensure that
```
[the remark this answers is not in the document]
jkotas: ManagedHost.dll has to be built against same or lower version of .NET Core.

Documentation/design-docs/host-startup-hook.md
yizhang82: In general we'd like to avoid any injection / magic in the build process - so that the user code/package is just a regular package and they can use dotnet CLI commands to build it. Those package would have be launched via service fabric anyway so we have full control over the environment / lookup.

yizhang82: But I do see an option that Service Fabric can inject this DLL at package upload time to become part of the app, but I don't see any significant difference between the two.

jkotas: > In general we'd like to avoid any injection / magic in the build process - so that the user code/package is just a regular package and they can use dotnet CLI commands to build it. 

I was not sure whether you prefer injecting magic at build time or injecting magic at launch time (you need to do one at least). Good to know that you prefer magic at launch time.

Documentation/design-docs/host-startup-hook.md
sbomer: I changed the default to "do not clear" and pointed out the ramifications: https://github.com/dotnet/core-setup/pull/4421/commits/f5a3acfb07592c236062646ba4a6fee9d07164b6.
I think it makes sense to let the managed code decide what to do here.

Documentation/design-docs/host-startup-hook.md
sbomer: Yes, that was the intention - and to point out that the original deps.json is owned by the application. I tried to clarify a bit in the doc.

Documentation/design-docs/host-startup-hook.md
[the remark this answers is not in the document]
sbomer: Pointed this out in the doc more explicitly.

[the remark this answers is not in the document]
yizhang82: Sounds good. What about dependencies to CoreFX? If ManagedHost.dll is built with lower version of CoreFX assemblies and that got unified to the application packaged higher version at runtime, it should be fine.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,93 @@
+# Host startup hook
+
+For .NET Core 3+, we want to provide a hook that allows managed code
+to run before the main application's entry point.
+
+## Motivation
+
+This would allow hosting providers to put configuration and policy in
+a single location, including settings that potentially influence load
+behavior of the main entry point such as the AssemblyLoadContext
```
AaronRobinsonMSFT: nit: All code should be surrounded by back ticks as an indication.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,93 @@
...
+behavior. This could be used to set up callbacks for handling
+Debug.Assert (if we make such an API available), or other
+environment-dependent behavior. The hook needs to be separate from the
+entry point, so that user code doesn't need to be modified.
+
+## Proposed behavior
+
+Environment variables or `<appname>.runtimeconfig.json` can be used to
+specify a managed assembly and type that contains an `Initialize`
```
AaronRobinsonMSFT: Please define the exact signature in either C# (e.g. `public void Initialize()`)

Documentation/design-docs/host-startup-hook.md
AaronRobinsonMSFT: I would like to consider the development process here as well - which should arguably be the most important consideration. What is the least confusing and unexpected for the developer of the app. Magic is fine for Azure, but during development time there should be 0 magic. Environment variables aren't always good if development is happening in VS - unless the project system can help with that and be reliable - I have serious doubts about this based on my experience working with the project system.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,93 @@
...
+    "runtimeOptions": {
+        "managedHostAssembly": "/path/to/ManagedHost.dll",
+        "managedHostType": "ManagedHostNamespace.ManagedHostType"
+    }
+}
+```
+
+The environment variables, if set, take precedence over the
+`<appname>.runtimeconfig.json` settings. These settings result in
+`ManagedHostType.Initialize()` being called in hostpolicy, before the
```
AaronRobinsonMSFT: This is a perfect place for notification. Users very often have no idea what the CLR is doing nor why. If cases like this occur (i.e. both env var _and_ json file has conflicting settings) then we should output something to the console via stderr or some other stream so users know we are going to ignore some setting and defer to another. Having this always printed out should be fine because if users are doing both either (A) they are debugging/in-development or (B) making a mistake.

brianrob:
This does seem like something that could help with injecting the performance controller.  One thing that comes to mind:

This mechanism will probably need to support multiple entries into the start-up hook.  People will use these hooks in different ways, and I can see them wanting to have two or more.

For example, one hook is used to set assembly loading policy, while the other is used to start-up the performance controller or some other telemetry service before Main() is called.  These two cases might be implemented by two different parties (say the assembly loading policy is implemented by the hoster while the performance controller policy is implemented by Microsoft).  Thus, they can't be combined.

FYI @vancem, @valenis

Documentation/design-docs/host-startup-hook.md
yizhang82: Arguably, given that this is single-purpose (you can only set one managed host assembly), in practice this is always controlled by one single platform that is determined by the project / environment / application type, and therefore there are usually a "launcher" involved to launch them. I think this is a stop-gap measure specifically designed for these kind of scenarios.

In the future, one can imagine you can bring multiple of these host assemblies into the app, each determined from different nuget packages, and in that case I agree folks should never use environment variable in this case.

AaronRobinsonMSFT: @yizhang82 I am not sure if you are agreeing or not 😄 My point here is that during development users should not have a concern about their application being launched in a different way or unknown way that causes them to scramble to try and debug the scenario that is impossible for them to simulate locally. If your statement is 'the same host will be injected by the project system/template' so in theory this should be a no-op - yay! However, in my experience users try something locally, deploy and then assumptions about what will happen are broken and that makes me want to cry because it demonstrates a lack of developer focus and creates an environment for developers where they can't succeed.

yizhang82: >> My point here is that during development users should not have a concern about their application being launched in a different way or unknown way that causes them to scramble to try and debug the scenario that is impossible for them to simulate locally

My point is that this is not true today for such platforms. Basically, by adding dependency to service fabric assemblies, you are effectively taking a dependency to the underlying service fabric platform and you cannot run it without service fabric environment - which is not "just load those assemblies".  Think of them as loosely coupled "plugins". Or imagine they are VS extension helper processes that can't run without VS and meanwhile need to load VS assemblies to be able to communicate with VS. 

I can see a scenario where a managed host assembly being injected by the project template, but I think it is important to keep the environment variable option for our production scenarios, and even for development scenarios (where we launch a mini-cluster on user machine which has the full environment). 

For cloud platforms, unfortunately, the biggest challenge is exactly you mentioned - how can you be sure that what you developed locally and make sure it works when deploy into the cloud where the assumptions are different, which is why it is important to do exactly what production does - spinning up a service fabric environment / cluster and launch the processes from there, even for development purposes. Even then it is still quite challenging, now that I'm on the other side of the fence here ;)

Documentation/design-docs/host-startup-hook.md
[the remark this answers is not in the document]
sbomer: Yup, no unification will happen. We're making no attempt to process a deps file for ManagedHost.dll - it's just an assembly. It needs to be compatible with the dependencies in the *app's* deps.json.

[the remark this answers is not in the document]
yizhang82: Imagine the managedhost.dll in the platform is built with .NET Core 3.0 + CoreFX 3.0, and user want to run .NET Core 4.0 apps, then the dependencies in managedhost.dll with CoreFX 3.0 won't resolve correctly to CoreFX 4.0. So this effectively forces the path to *build*  managedhost.dll as part of the app, or inject it with the .deps to unify with CoreFX 4.0?

vancem:
My biggest feedback is the same as @brianrob, which is that users that are injecting their startup code have not reason to believe they are the ONLY ones.     Thus the scheme has to work when more than one system is doing this.   Thus it is not a DLL/Method name, but a LIST of those specifications.  

Thus I would recommend instead having NOT having a tuple, but a syntax.   I would recommend that you have PATH!Type, but I would also recommend making a default type name (e.g. DotNetStartup).  This allows people to only specify the PATH (since they will be building the type specificaly for this purpose, so can conform to whatever convention we wish).

In either case as single string represents the 'beforeStartup' point.  The environment variable/config entry is then a semicolon separated list of these.   

This does bring up the question of ordering of these things (everybody wants to be first).   Certainly we would order them by the order in the semicolon delimited list, but if we get this data from multiple places (some from env vars, some from config), then it is not so clear.    Even if it was clear, it is not obvious that the software updating the config/env vars knows enough to do a good job.  

To address this,  we could have convention if the DLL name (before suffix) ends in a _N* where N is a digit, that it is a desire to be point in that order (lexographic)   Thus X_12.dll comes before A_2.dll ...).   

The main point, however is that multiple 'things' will want to use this hook, and they need to be able to do so without collision.  

Comments?

brianrob:
With respect to what @vancem says about ordering, there is precedent for using a priority value to handle ordering.  GCC does this with usages of ```__attribute__((constructor))``` and ```__attribute__((destructor))``` which are responsible for running before and after entrypoint execution.

From https://gcc.gnu.org/onlinedocs/gcc-4.3.0/gcc/Function-Attributes.html:

>You may provide an optional integer priority to control the order in which constructor and destructor functions are run. A constructor with a smaller priority number runs before a constructor with a larger priority number; the opposite relationship holds for destructors. So, if you have a constructor that allocates a resource and a destructor that deallocates the same resource, both functions typically have the same priority. The priorities for constructor and destructor functions are the same as those specified for namespace-scope C++ objects (see C++ Attributes).

I'd recommend making it part of the config and not part of the DLL name, just because it will be easier to change when needed.

Documentation/design-docs/host-startup-hook.md
AaronRobinsonMSFT: > Basically, by adding dependency to service fabric assemblies, you are effectively taking a dependency to the underlying service fabric platform and you cannot run it without service fabric environment

If this is true, we are in bigger trouble than I thought. The whole point of having an IDE and projects is to have some reliability about debugging and designing a solution. I see how my concern is larger than the current review. Thanks for the unsettling information 😖

Documentation/design-docs/host-startup-hook.md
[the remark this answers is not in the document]
jeffschwMSFT: Refs roll forward, so the .NET Core 3.0 built managedhost.dll will load fine within a .NET Core 4.0 process.  The inverse is not true.

jeffschwMSFT:
@vancem @brianrob we discussed multiple possible entries and decided to go with a similar model to what we have in .NET framework.  https://msdn.microsoft.com/en-us/library/system.appdomainmanager(v=vs.110).aspx  Do you know how the performance scenarios operate today in .NET Framework?  Given this is an advanced scenario it may be that the developer has to wrap these 3rd party providers and call them in the order they choose explicitly.  This is not ideal, but hopefully this is not incredibly common.

yizhang82:
@vancem @brianrob @jeffschwMSFT 
In the future it would be very nice to have nuget package have their own managed host and flow that into deps.json which contains all the combinations, and .NET Core loader can find the correct managed host depending on which assembly is being loaded.

Having said that, we (service fabric) need a stop-gap solution where there is a single application/orchestration platform that launches the user micro-service and allow them to load shared service fabric assemblies from the platform (rather than app-local), as if we are carrying our own GAC for those user services. We are also hoping this feature can be back ported to .NET Core 2.0, so a simpler solution with less dependencies/complexity is preferable at this point.

jkotas:
FWIW, I agree with @vancem that we should allow more than one of these to be specified (as you know I have mentioned it during our discussions).

I do not think we need to have a way to tweak the order. The moment you get multiple of these and they are colliding, you are in trouble.

sbomer:
Thanks all for the feedback!

@brianrob, @vancem I can see why people would want multiple such hooks, and I don't think the native host is the right place. Rather, I'd hope that the ManagedHost.dll can set up further hooks in managed code if that's needed.

This startup hook exists to solve the bootstrapping problem of "how can we (host provider or user code) control, in managed code, how the entry point dependencies are loaded, given that those dependencies are usually loaded before any managed code runs?". This hook makes that possible by separating the concepts of "user entry point" and "first managed code to run", while imposing some constraints on the host implementation (it needs to be compatible with the app's dependencies).

In other words, this setting specifies which managed code runs first, and we're recommending it as a solution for folks who need to solve the bootstrapping problem because they don't control the managed entry point. Anyone who does control the entry point needn't do this - and anyone who uses this setting is expected to understand that it only works in a cooperative environment.

I agree that if our recommendation is ignored (if a bunch of folks go and try to use this hook to control which managed code runs first), we'll hit issues like this. Perhaps we can design this in a way that would allow chained "first managed code" settings (which inevitably has the same problem - who is first of the first?), but I hope we never have to go there.

@vancem How about this: we put the setting in one variable, so it just becomes a path to an assembly and a type. This would make it possible to extend with additional syntax if we ever need to. But in this iteration we only allow one "first managed code" setting.

Thoughts?

sbomer:
@yizhang82 I think the long-term solution that allows managed hosts to be shipped as nuget packages is just to provide a managed hosting library that has to be explicitly initialized from Main. This feature is intended more as the mentioned stop-gap solution for cases where, for whatever reason, the entry point is a bad place to put this.

jkotas:
> This startup hook exists to solve the bootstrapping problem of "how can we (host provider or user code) control, in managed code, how the entry point dependencies are loaded,

This startup hook is very powerful. I had seen people coming to me with all sorts of problems that could have been solved by it. I am pretty sure it will be used for more just the scenario motivating it.

> stop-gap solution

I do not see this as stop-gap solution. If this is a stop-gap solution, we should not build this.

Documentation/design-docs/host-startup-hook.md
sbomer: Agreed!

Documentation/design-docs/host-startup-hook.md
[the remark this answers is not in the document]
sbomer: If your concern is about breaking changes between major versions, you'll run into the same issue for any .NET Core 4.0 app that tries to run against .NET Core 3.0. I'm definitely not suggesting building it with the app - this will likely cause contention issues where each app and framework wants control over the first managed code to run (see the discussion below). My understanding is that the host provider would have a shared runtime install that supports user apps targeting 2.x or 3.x, but not both if there are breaking changes. To support all apps targeting that major version, a managed host dll should then be compiled against the lowest minor version of that major version (or at least not reference any surface area that doesn't exist in lower minor versions).

sbomer:
@jkotas I should say "narrow use case" rather than "stop-gap" in my previous post. But if a hook like this is likely to be used for other reasons, I agree that we need to consider those. Do you have any examples of other problems solved by it, that wouldn't be better solved by adding managed hooks to user code?

ashishnegi:
One requirement : (a) Is there any existing api in .Net Core that gives precedence to a path over app local (as mentioned in deps.json) to load dlls ? This is what GAC does in .Net Framework.

`Scenario`: We have existing deployed coreclr apps that have X.dll in app local. In .Net Framework X.dll is GACed. 

Without (a), upgrading to next Service Fabric runtime (which use this feature) will break existing apps.

This is required for our Linux cluster where customers will be deploying FDD coreclr apps and using dotnet installed by Service Fabric runtime.

For new apps, we can change our SDK to put X.dll in ref folder of nupkg. X.dll will not become part of deps.json. So, this problem does not occur with new apps.

jkotas:
@ashishnegi Why can't you leave the existing service fabric applications alone and not use any of this for them?

jkotas:
> The cluster may have many other non-coreclr apps

How are you dealing with this problem for the non-coreclr apps?

yizhang82:
> How are you dealing with this problem for the non-coreclr apps?

In full desktop everything is loaded from GAC - even though some SDK assemblies are app-local they are still loaded from GAC.

jkotas:
I meant non-.NET apps - e.g. Java apps?

yizhang82:
Java apps don't have this problem today because they are on an entirely different technology stack that is based on our native store/replicator/logger. The "managed" SDK layer is very thin and the majority work is done through JNI into native layer with a well defined ABI. So having everything app local works great there, and native libraries are loaded from platform path (from enviroment variable when process is launched)

The new .NET SDK is also based on the native store/replicator/logger so it is also app-local and mostly p/invoke to native code.

As you can see, the new model is better suited for CoreCLR and all "managed" componets can be app-local. 

However, the context here is the existing "old" .NET SDK where the implementation is mostly managed (yes, we have both managed and native implementation of the same thing, and the native implementation being the new thing) and exist in platform folder (internally we call it runtime folder but I intentionally avoid it here...), and that's where the app-local vs platform folder diverge happens. I can imagine if we built this technology entirely in Java it would have the same problem/requirement.

yizhang82:
And all of .NET customers are on the existing managed "old" SDK- we haven't shipped the new native based SDK yet.

yizhang82:
> It sounds like that you are rolling forward all coreclr apps to the latest .NET Core runtime today. It has potential to break them. If you want to have high confidence in not breaking older coreclr apps, you should keep running them against the runtime that they were built for. We do not guarantee bug-for-bug compatibility between major .NET Core versions.

This is a good point. If we roll forward to .NET Core 3, it might break customer in a subtle fashion that they haven't tested and doing diagnostic in the cluster is going to be much harder than local debugging. So ideally we should find a way to upgrade from within VS before deploy (not open project, if they are still on older version of SDK), so that they can opt into the upgrade and explicitly test their app before deployment. 

Alternatively, if this feature can be back ported to .NET 2.0 (we discussed with @jeffschwMSFT earlier as a possibility) it might make things a bit easier.

jkotas:
So if this is just a compat quirk that will hopefully go away eventually, can you do a magic for it during launch time? For example, fixup .deps.json file in the app to point where you need the bits to come from? The compat quirks do not need to be pretty.

> .NET 2.0 

I assume you mean .NET 2.1. .NET 2.0 is out of support in October: https://www.microsoft.com/net/support/policy

yizhang82:
> So if this is just a compat quirk that will hopefully go away eventually, 

I don't see this as a compact quirk (but I see your perspective). The existing managed SDK will be there for a long time until we can get everybody on board to the native stack based SDK. Before that happens (it may never happen due to various factors such as data upgrade), we need to find a way to evolve the managed SDK and native stack based SDK together and adding new APIs to both (they share the same API surface). In other words, it is somewhat similar to the .NET vs .NET Core situation with two sets of implementation but a common set of API and both evolves at similar pace in terms of shared set of APIs (of course, in a smaller scale and not exactly the same) :)

> can you do a magic for it during launch time? For example, fixup .deps.json file in the app to point where you need the bits to come from? 

Assuming deps.json you can put a full path (haven't looked at it in a while), this might work, but I can see a few concerns:
* We support all kinds of apps - anything that runs can be launched there, regardless of languages. From our perspective it is just a EXE that needs to be run. It's not clear if only .NET Core uses deps.json convention and we might stumble on other stuff
* Having to do surgey to deps.json takes a dependency on deps.json format and the location 
* In containers, we don't directly control what's in the container and we are only telling docker to run the EXE. having to mutate deps.json means we now have to write a shim EXE that does the mutation and then launch the EXE. Not rocket science, though. 

It is doable, but like you said, it isn't pretty :)

Given that we need to support the managed SDK for a long time, I think it is important to have a officially supported way to have a "application-level GAC" and redirect launched apps go probe there. And as folks pointed out earlier, this proposal solves a broad set of problems and is going to be helpful for other .NET Core customer as well. 

> I assume you mean .NET 2.1. .NET 2.0 is out of support in October: 

Sure. Whatever is supported. :)

sbomer:
I don't think this is quite the right mechanism for a general-purpose library initializer. If we make it so, imagine the following:
- User app defines a library initializer
- Host provider defines a library initializer (that sets up a load context to pull assemblies from a shared location)

If the host provider wants dependencies of the user initializer to be loaded from a shared location, we hit the same issue again - the host will want their initializer to run first, so that it can control where the user app's library initializer dependencies get loaded from. This proposal is intended to solve the "who comes first" problem, nothing more.

If we do expand the discussion to general-purpose library initializers, I think they should be designed slightly differently from this feature. I would instead suggest something along the lines of:
- Don't use environment variables for this. Instead use runtimeconfig only (or perhaps assembly-level attributes, if we can read these without fully loading assemblies).
- Define library initializers in the library itself, not in a separate assembly.
- We would also need to consider whether library initialization happens eagerly for all static dependencies, or when the library is first loaded.

However, I think that should be a separate discussion, and I don't want to get into it here unless we decide that this feature needs to solve the library initializer problem in addition to the "who comes first" problem.

I see two paths forward from here:
- We provide this feature to determine who comes first only, and document it as such, making clear that it's not a recommendation for general-purpose library initialization. We can consider backporting it to 2.1 if Service Fabric doesn't find any suitable workarounds for their "runtime depends on applocal types" problem. Currently I think this is the right direction.
- Or, we develop this into a feature that is intended to provide a library initialization hook in coreclr. I'm happy to go this direction too, but I think it solves different problems.

Just my thoughts so far - I appreciate further feedback from you all!

jkotas:
> library initializer

I would stick with calling this startup hook. I think it is the right name for this. I would stop calling this managed host - e.g. change the environment variable to `DOTNET_STARTUP_HOOKS`. The startup hook allows running one or more pieces of code before the apps Main method. It is meant to be used to inject extra configuration into existing apps, without rebuilding the apps. 

I do not expect app developers to be implementing these hooks. App developers have control over the Main method, so they can just call the relevant code in the Main method themselves.

I expect that there will be a very few folks implementing these hooks (folks who build platforms, profilers or telemetry systems). These folks are pretty well aware of issues around initialization order, dependencies, conflicts, polluting the environment, etc. We can document some of the best practices for them if necessary. The simple rule is to do as little as possible as late as possible.

> paths forward from here

We should come up with the right design, implement it in master and validate that it works. Once we have that, we can discuss the port to .NET Core 2.1. Note that this startup hook alone is not sufficient to solve the service fabric problem per https://github.com/dotnet/core-setup/pull/4421#issuecomment-409907950 .

We had number of very narrow scoped features, such as `--additional-deps` or `dotnet store`, that were designed to solve a very specific problem and that cannot be really used for anything else. I would be nice to stop building features like these. They add complexity that does not pay for itself.

sbomer:
Just to clarify, I'm not suggesting that we call it a library initializer - just pointing out the difference between this feature and a library initializer.

> I would stick with calling this startup hook. I think it is the right name for this.

I like this suggestion.

> Note that this startup hook alone is not sufficient to solve the service fabric problem per #4421 (comment) .

It will solve their problem (by letting them ensure that sdk_X.dll loads from the shared location). And it can do so for existing apps if we backport the change. Unless I misunderstood something?

> I would be nice to stop building features like these. They add complexity that does not pay for itself.

Agreed. The way I can see to avoid building this in the first place would be to define the managed entry point as the first managed code that runs, but that potentially has other problems (as you pointed out to me), with tooling expecting that the entry point lives in user code, and existing applications putting the entry point in user-owned code.

yizhang82:
+1 to what @jkotas said. Library initializer is a confusing name. Just to clarify - what we really need is a shared/app-local GAC and this is a more general solution that achieve this by having .NET host run some hook up code and own the assembly resolution process. 

>  Note that this startup hook alone is not sufficient to solve the service fabric problem per #4421 (comment) .

Just to make sure we are on the same page - for existing CoreCLR apps, I'm assuming the AssemblyResolve is going to be kicked in for those existing app-local SDK assemblies in older apps (of older version) and in the custom resolve event we'll load the SDK assembly (of current version) from platform folder instead, so everything still matches with the rest of the platform implementation assemblies. So this would work the same way as the GAC just like full .NET and should "just work". 

The remaining problem, is that those apps are still built against older version of CoreCLR and needs to be rebuilt against .NET Core 3.0. If the fix is back ported, the situation will be a bit better if the app is not a standalone package and we can install latest .NET Core 2.0 with the fix on the cluster. Either way we can issue an warning in cluster upgrade if we detect such situation. 

@ashishnegi Please chime in as well.

vitek-karas:
I don't think that the startup hook alone solves the service fabric problem re sdk_X.dll loads. The app would have a reference to sdk_X.dll in its .deps.json (at least that's how I understand the scenario). This means the sdk_X.dll will make it into TPA. And thus without the ability to fully override the default load context, even startup hooks can't redirect the reference. Or maybe I'm missing something...?

jkotas:
> for existing CoreCLR apps, I'm assuming the AssemblyResolve is going to be kicked in for those existing app-local SDK assemblies in older apps

Not with the design proposed so far, as @vitek-karas pointed out.

yizhang82:
@jkotas @vitek-karas Thanks... I must admit I'm mostly clueless when it comes to CoreCLR loading... So I suppose the workaround is as suggested earlier to muck with the deps.json and update those entries, or force users to upgrade to newer package (which excludes those assemblies from deps.json) + new .NET Core. Neither of these are pretty.

sbomer:
@vitek-karas thanks for clarifying. So loading sdk_X.dll from a shared location would require either removing it from deps.json, or setting up a new load context, and reading deps.json... which requires a modifiable default load context and is also not pretty. So this feature is only useful to Service Fabric in conjunction with some improvements to the ALC design (if we are able to backport them). Right?

vancem:
Given that they have the need to work with existing .NET Core apps and they want quite strong control over loading (they want to override local DLL lookup),  can someone describe why the native hosting APIs are not an option?

     https://github.com/dotnet/docs/blob/master/docs/core/tutorials/netcore-hosting.md

Sure, it is a mildly painful native code exercise, but it does give you lots of control, and works on existing runtimes.   
--------------
I was reasonably surprised that AssemblyLoadContext did not provide 'full control' over assembly load (it only kicks in if the DLL can't be found using the deps file).     Fixing that with some opt-in kind of request, I think is at least worth discussing (Forcing people to use native code is not great).  

However this all seems to be unrelated to the startup hook itself.   I think we were approaching consensus there (Change to be a single variable, allow multiple entries, Environment Variable modules come first).   

I think our strategy here is to try to make a few very powerful hooks (here at startup, and probably also in AssemblyLoadContext for loading), that has as little policy structure as possible.   

Where do we go from here
   1. explore (or indicate why it is problematic), to use hosting API.
   2. Determine if we want to think about expanding AssemblyLoadContext (note I think it is reasonable for this type to have the ability to gain 'full control' without needing to touch files on disk).  
   3. Update doc on concensus on startup feature.

sbomer:
@vancem agreed that the native hosting APIs are another option for someone who wants full control like Service Fabric. The idea behind the startup hook is to make it a bit easier on them by letting them use managed code with a predefined hook (so that they don't have to re-implement logic in our native host, like finding dependencies from deps.json files).

The ALC discussion is one that's going on in parallel to this feature - it's definitely possible that we'll make the default ALC behavior easier to override. In that case the startup hook might provide a good place to do that for hosting providers. The features complement each other.

> I think our strategy here is to try to make a few very powerful hooks (here at startup, and probably also in AssemblyLoadContext for loading), that has as little policy structure as possible.

I think that's a great summary of where we're headed with this feature and future ALC improvements.

> 1. explore (or indicate why it is problematic), to use hosting API.

Perhaps @yizhang82 can provide more info on how well this could work for Service Fabric? I assume it's the complexity involved in writing/deploying cross-plat native code that re-implements logic from our host.

> 2. Determine if we want to think about expanding AssemblyLoadContext (note I think it is reasonable for this type to have the ability to gain 'full control' without needing to touch files on disk).

Agreed, and this discussion is happening in parallel. Feel free to ping me if you'd like me to loop you in.

> 3. Update doc on concensus on startup feature.

I'll be updating the doc soon.

vancem:
> Feel free to ping me if you'd like me to loop you in.

I am interested in knowing what is happening with AssemblyLoadContext.

ashishnegi:
>So this feature is only useful to Service Fabric in conjunction with some improvements to the ALC design (if we are able to backport them). Right?

@sbomer Can you include me, @yizhang82 and @raunakpandya in this ALC design ?
 We would like to know whether this will be helpful for our case, in SCD and FDD apps.

vitek-karas:
Potential idea for the service fabric scenario.

Assumptions:
- The app might have a dependency on for example SFabricSdk.dll - version 2.0.0
- The service fabric has a newer version of SFabricSdk.dll - version 2.1.0 and wants all apps to use it
- The new version is fully backward compatible
- The new version of SDK doesn't have any big changes in terms of its dependencies (like depending on a new major version of a dll or something like that)

It should be possible to use the startup hook described here to inject a piece of code into the app which does this:
`Assembly.LoadFrom(@"Path to SFabricSdk.dll v 2.1.0")`

This will run before the app's main. In fact this should run before any code from the app gets a chance to run (including static .ctors and alike). It will preload the newer version of the SDK into the default load context, effectively ignoring whatever the app brings with it in .deps.json and thus in its TPA.

Later on when the app code actually does run and there's a reference to the SDK, it will resolve in the default context against the already loaded version and end up using the newer version.

I tested this on a small test project (adding code at the beginning of Main, but that should be equivalent to the startup hook). It works just fine as long as the preloaded version if not a new major version. So minor version change to a higher version seems to work just fine.

The downside of this solution is that it force preloads the SDK to all .NET Core apps, regardless if they actually use it or not. This could in theory be worked around by inspecting the TPA, but it shouldn't be such a big problem.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,95 @@
...
+specified, before the `Main` entry point:
+
+```
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
+```
+
+The list is a semicolon-delimited list of assembly paths (absolute, or
+relative to the working directory), each with an optional type name
+(following an exclamation mark). If no type name is specified, the
+default is `StartupHook.Initializer`.
```
AaronRobinsonMSFT: I am not a fan of magic type names. This is such a low level construct the engineering cost to design a hook grossly outweighs the cost of providing a type name. All it does is create more magic and implicitly creates yet another test case.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,95 @@
...
+child processes by default. It is up to the `StartupHook.dll`s and
+user code to decide what to do about this - `StartupHook.dll` may
+clear them to prevent this behavior globally, if desired.
+
+Specifically, hostpolicy starts up coreclr and sets up a new AppDomain
+with each `StartupHook.dll` on the TPA list. It then invokes a private
+method in `System.Private.CoreLib`, which will call each
+`StartupHookType.Initialize()` in turn. This gives `StartupHookType` a
+chance to set up new `AssemblyLoadContext`s, or register other
+callbacks. After the `Initialize()` methods return, control returns to
```
AaronRobinsonMSFT: "After **all** `Initialize()` methods return, ..." Also please be clear that these functions are called in a synchronous fashion.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,95 @@
...
+hostpolicy, which starts up the main entry point of the app like
```
AaronRobinsonMSFT: "which will then call the main entry..."

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,95 @@
...
+tracing or telemetry injection, to set up callbacks for handling
+Debug.Assert (if we make such an API available), or other
+environment-dependent behavior. The hook is separate from the entry
+point, so that user code doesn't need to be modified.
+
+## Proposed behavior
+
+The `DOTNET_STARTUP_HOOKS` environment variable can be used to specify
+a list of managed assemblies and optional type names that contain
+`public void Initialize()` methods, which will be called in the order
```
AaronRobinsonMSFT: Boo. Your example below implies the function must be `static` is that true? Please clarify.

Documentation/design-docs/host-startup-hook.md
sbomer: That's reasonable. I've removed the default type name in the latest iteration.

Documentation/design-docs/host-startup-hook.md
sbomer: Fixed!

Documentation/design-docs/host-startup-hook.md
sbomer: Fixed!

Documentation/design-docs/host-startup-hook.md
sbomer: Good catch, thanks. Fixed!

sbomer:
@jkotas I can use platform-specific path separators (`;` vs `:`) to separate multiple hooks. Is there an alternative to `!` for separating the assembly path and type name that I should use on unix?

jkotas:
I cannot think of any good prior art for this. `!` sounds reasonable to me. It may be nice to look for the `!` from the end so that if the file path happens to contains `!` as well, it is still going to work.

mattwarren:
> I expect that there will be a very few folks implementing these hooks (**folks who build platforms, profilers or telemetry systems**). These folks are pretty well aware of issues around initialization order, dependencies, conflicts, polluting the environment, etc. We can document some of the best practices for them if necessary. The simple rule is to do as little as possible as late as possible.

A while back @noahfalk created a [Profiler Interest Group - Announcements](https://github.com/dotnet/coreclr/issues/15136) issue, if you want to get wider knowledge of this hook, you might want to post something there?

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,96 @@
...
+specified, before the `Main` entry point:
+
+```
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll!StartupHookNamespace.StartupHookType1;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
+```
+
+This variable is a list of absolute assembly paths, each with a type
+name following an exclamation mark. The list is delimited by the
+platform-specific path separator (`;` on Windows and `:` on Unix). It
+may not contain any empty entries or a trailing path separator.
```
danmosemsft: The code can presumably easily tolerate a trailing path separator

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,96 @@
...
+this behavior globally, if desired.
+
+Specifically, hostpolicy starts up coreclr and sets up a new
+AppDomain. It then invokes a private method in
+`System.Private.CoreLib`, which will call each
+`StartupHookType.Initialize()` in turn synchronously. This gives
+`StartupHookType` a chance to set up new `AssemblyLoadContext`s, or
+register other callbacks. After all of the `Initialize()` methods
+return, control returns to hostpolicy, which then calls the main entry
+point of the app like usual.
```
danmosemsft: Is it worth specifying that this is on the same thread? In the unlikely case that the Initialize() sets thread state, we presuambly want that to stably work or not.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,96 @@
...
+default. It is up to the `StartupHook.dll`s and user code to decide
+what to do about this - `StartupHook.dll` may clear them to prevent
+this behavior globally, if desired.
+
+Specifically, hostpolicy starts up coreclr and sets up a new
+AppDomain. It then invokes a private method in
+`System.Private.CoreLib`, which will call each
+`StartupHookType.Initialize()` in turn synchronously. This gives
+`StartupHookType` a chance to set up new `AssemblyLoadContext`s, or
+register other callbacks. After all of the `Initialize()` methods
```
SergeyKanzhelev: Any limitations of this callback? Can new `Thread` be started? Will callback be able to get the startup into and path to app folder to obtain config files and make decision about what features to enable?

AaronRobinsonMSFT: @SergeyKanzhelev - The process will have already been started, so I don't see how the startup info could be useful at this point? Environment variables will be accessible and modifiable if the start up hooks don't want to propagate them to potential child processes.

Starting a new thread makes me nervous - I don't think there would be a limitation, but at that point it seems like the hook is doing more than it should. Do you have a scenario in mind?

SergeyKanzhelev: Startup info and environment variables is a way to configure the callback. Arguments from startup info may help to identify the process so proper configuration source will be used (you may configure for monitoring winform app ONLY started with specific parameters or such). Also it is typically to check for some settings from app folder. So ideally startup hook should know where to read appsettings file from.


Simplest example of task from the startup hook is sending notification to the remote endpoint that process got started. And/or getting some configuration from that remote location.

Example of a thread is to initiate performance counters collection thread or run dedicated thread to batch and send events to the server. Timers are not ideal for this purpose as timers can be affected by thread-pool starvation. So you will loose precision of counters or telemetry from the app that really has a problem.

Another example of thread is to start a implement a delayed initialization of monitoring. Let's say we want to use startup hook for Application Insights (AI). AI SDK can be used as a direct reference from customer code or via runtime injection. If it is used via runtime injection - we may choose to wait and see whether direct reference was taken by an application in `Main` method so we will not initialize any data collection from SDK enabled via runtime.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,96 @@
...
+static void Initialize()` methods, which will be called in the order
+specified, before the `Main` entry point:
+
+```
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll!StartupHookNamespace.StartupHookType1;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
+```
+
+This variable is a list of absolute assembly paths, each with a type
+name following an exclamation mark. The list is delimited by the
+platform-specific path separator (`;` on Windows and `:` on Unix). It
```
vancem: I don't think we are well served by making things platform specific.   Basically it takes something that SHOULD have been platform neutral (this startup variable) and requires users of this feature to make platform specific adaptations.   We should not introduce gratuitous platform differences.  It just leads to unnecessary  complexity for your users. 

The simplest resolution to this is to simply pick one (it becomes the 'dotnet' convention for separating things).  I would recommend ';' because ':' will be needed in most windows paths, but the reverse is not true.    If you REALLY feel strongly you can make the UNIX version support both ';' and ':' but would not bother here.  This is especially true because we expect in the common case it will have only one entry.

davidfowl:
This is extremely similar to the feature that ASP.NET Core has https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/platform-specific-configuration?view=aspnetcore-2.1. Via an environment variable (or configuration) you can specify which assemblies are contain these startup hooks. 

One of the fundamental design decisions we made was that we didn't want to deal with *any* loading policy whatsoever, this means that there were 2 different pieces things that needed to work together to make the end to end work:

1. Somebody has the job of making the assembly loadable. This means the assembly with the hook and it's transitive closure needed to be in the "loading path".
2. ASP.NET Core would load these assemblies by name not by path to ensure the the regular .NET Core assembly loading logic would occur.

That lead to the --additionalDeps feature. The idea being that the only job of the consumer was to lay stuff down on disk in the format that made it loadable by the runtime so no custom code had to be written to handle complex dependency loading.

Documentation/design-docs/host-startup-hook.md
vancem: To answer @SergeyKanzhelev  question, there are no limitiations on what the code can do.   Note however that the code DOES have to work with the runtime / framework  that was launched (bascialy it needs to be compatible).

davidfowl:
Given the amount of problems we've had with additionalDeps in real life usage (and that's a completely declarative approach), I think we need strong guidance for the types of code people should put in these startup hooks. I like the list of assumptions here https://github.com/dotnet/core-setup/pull/4421#issuecomment-411208848 because those are the sort of things that people don't think through when building these plugins.

I'm also concerned about how hard it is to load manually load your transitive closure of assemblies. How much does the runtime help you here (for example, resources, native binaries etc)?

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Proposed behavior
+
+The `DOTNET_STARTUP_HOOKS` environment variable can be used to specify
+a list of managed assemblies and type names that contain `public
+static void Initialize()` methods, which will be called in the order
+specified, before the `Main` entry point:
+
+```
+DOTNET_STARTUP_HOOKS=/path/to/StartupHook1.dll!StartupHookNamespace.StartupHookType1;/path/to/StartupHook2.dll!StartupHookNamespace.StartupHookType2
```
vancem: In its current form, I don't think relative paths are useful for startup hooks because the current directory is typically not under the control of the entity doing the hooking (and will be arbitrary).   

On the other hand, having a hook that is relative to app directory is useful (if the user puts the file in that directory they can give a relative path to it).    Thus we may wish to define the current directory as the app directory (Basically you change to this directory before resolving paths, then change back). 

I am trying to decide if this scenario is actually worthwhile...

Documentation/design-docs/host-startup-hook.md
danmosemsft: Also, care is necessary if you use `;` separator. See https://github.com/dotnet/core-setup/issues/4141 for a break this already caused. `|` may be better and presumably works for Unix also. We would want to accept it for all the host configuration variables.

Documentation/design-docs/host-startup-hook.md
jkotas: >  having a hook that is relative to app directory is useful 

I would agree that this is useful for the config file case. Hard to see the usefulness for the env variable case.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,96 @@
...
+load behavior of the main entry point such as the
+`AssemblyLoadContext` behavior. The hook could be used to set up
+tracing or telemetry injection, to set up callbacks for handling
+Debug.Assert (if we make such an API available), or other
+environment-dependent behavior. The hook is separate from the entry
+point, so that user code doesn't need to be modified.
+
+## Proposed behavior
+
+The `DOTNET_STARTUP_HOOKS` environment variable can be used to specify
```
danmosemsft: What is the behavior if the startup hook cannot be resolved, eg., DOTNET_STARTUP_HOOKS has an invalid path, file does not exist, is not an assembly, or does not contain Initialize(). Can we log something useful to the console? Ie not something like `Failed to initialize CoreCLR` which means we have to debug it.

Documentation/design-docs/host-startup-hook.md
jkotas: > The simplest resolution to this is to simply pick on

I agree with you if this is in the config file.

For environment variables, it is an interesting question whether to be platform native or unified. My experience is that the separators are platform native for environment variables. For example, here is what Java `CLASSPATH` does on Windows vs. Unix:

https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html
https://docs.oracle.com/javase/8/docs/technotes/tools/unix/classpath.html

Are there examples where folks choose to be platform unified for path separators in environment variables?

Documentation/design-docs/host-startup-hook.md
sbomer: > The code can presumably easily tolerate a trailing path separator

It could - see my comment above for the rationale behind the current choice. I'm fine with changing this if there's a good reason.

About the proper path separator - it's not clear that using `|` would better since it is allowed on unix. I think it's fine to specify that the path may not contain a `;` for this low-level hook, but am open to considering a more correct solution. It would probably take some more thought to figure out the right thing to do to handle weird but valid paths on different platforms.

Documentation/design-docs/host-startup-hook.md
sbomer: Clarified this in the update.

Documentation/design-docs/host-startup-hook.md
sbomer: New threads can be started, but I agree with @AaronRobinsonMSFT that it would be best to avoid relying too much on the thread behavior. I updated the doc with some details on the threading behavior and caveats.

Documentation/design-docs/host-startup-hook.md
sbomer: I added some details to he doc about the error behavior of the current implementation (https://github.com/dotnet/coreclr/pull/19486, with tests here: https://github.com/dotnet/core-setup/pull/4465). I'm not sure to what extent these exceptions should be called out in the design vs an implementation detail, but I've described them explicitly for now.

Documentation/design-docs/host-startup-hook.md
jkotas: Hmmm ... maybe we should not be looking at these as paths (they are not actually proper paths because they have type appended), and so it is not worth worrying about being platform native?

Documentation/design-docs/host-startup-hook.md
```diff
@@ -93,4 +94,119 @@ namespace SharedHostPolicy
...
+
+Problems with the startup hook should be fairly straightforward to
+diagnose. All of these exceptionts will contain the startup hook path
+(`System.StartupHookProvider.ProcessStartupHooks`) on the stack
+trace. They fall into the following categories:
+
+- Errors detected eagerly, with exceptions thrown before the execution
+  of any startup hook.
+
+-- Invalid syntax. This will throw an `ArgumentException` with "The
```
AaronRobinsonMSFT: Documenting the messages in the exception isn't a good idea. It encourages people to do weird things and write tests that check error messages - which is very brittle.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -93,4 +94,119 @@ namespace SharedHostPolicy
...
+
+## Guidance and caveats
+
+This hook is meant as a low-level, powerful way to inject code into
+the process at runtime, for use by tool developers who truly have a
+need for this kind of power. It should only be used in situations
+where modifying application code is not an option and there is not an
+existing structured dependency injection framework in place. An
+example of such a use case is a hook that injects logging, telemetry,
+or profiling into an existing deployed application at run tim.e
```
AaronRobinsonMSFT: "runtime."

Documentation/design-docs/host-startup-hook.md
```diff
@@ -93,4 +94,119 @@ namespace SharedHostPolicy
...
+
+It is prone to ordering issues when multiple hooks are used, and does
+nothing to attempt to make dependencies of hooks easy to
+manage. Multiple hooks, if used, should be fairly independent of each
```
AaronRobinsonMSFT: "Multiple hooks should be independent of each other" <- This should be a very strong statement with no ambiguity.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -93,4 +94,119 @@ namespace SharedHostPolicy
...
+For example, if one hook sets global state that introduces logging in
+the process, the new behavior will affect all subsequent hooks in the
+process and the `Main` entry point. Subsequent hooks may attempt to
+modify logging behavior in a way that conflicts with the first hook,
+leading to unexpected results. This kind of problem exists for any
+framework that gives independently-owned components access to shared
+resources - often dependency injection frameworks will have a
+dependency manager that loads components in a specific order. If this
+kind of behavior is required, a proper dependency injection framework
+should be used instead of multpile startup hooks.
```
AaronRobinsonMSFT: "multiple"

Documentation/design-docs/host-startup-hook.md
danmosemsft: Paths with semicolons in seem to be very rare on Windows. In the linked issue it was there accidentally and the real problem was that it needed a debugger to figure that out. So if your error messages/tracing is good, I think semicolon is fine

Documentation/design-docs/host-startup-hook.md
danmosemsft: Whether and how the user receives good error messages is a design consideration IMO

Documentation/design-docs/host-startup-hook.md
sbomer: I think there's a reasonable argument in either direction. Consistency with existing PATH-like variables is nice. OTOH, they're not really paths, so consistency with PATH-like variables may not be essential. Another consideration is that platform-specific separators make it easy to set PATHs on in bash without quotes:
```
PATH=a:b
```
works on unix, whereas
```
PATH=a;b
```
gets split into two commands unless the `"a;b"` is quoted. That's an argument for using platform-specific separators in PATH-like variables... *but* in our case `!` causes problems in bash anyway, even if it's quoted, because bash interprets the `!` as a history expansion prefix:
```
$ DOTNET_STARTUP_HOOKS="path!type"
-bash: !type": event not found
```
Setting this requires single quotes:
```
DOTNET_STARTUP_HOOKS='path!type'
```
I think that pretty much defeats the purpose (from a usability perspective) of having platform-specific separators. So at this point I'd suggest either sticking with `;` or using something else (`|`?) instead of `!` to prevent the need for single quotes on unix. Note that even `|` as a path/type separator would require double quotes.

Documentation/design-docs/host-startup-hook.md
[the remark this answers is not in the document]
sbomer: Yes, I removed it because it simplifies things and encourages the startup hook to be seen as an implicit way to inject dependencies into the process from outside at process launch, rather than a part of the app. We could add the option to runtimeconfig.json if folks need it.

I'm not sure I understand your scenario. The point of making this an environment variable was exactly so that it could be controlled by someone who controls launching of apps, but not the deployed bits. Could you clarify?

Documentation/design-docs/host-startup-hook.md
sbomer: > the current directory is typically not under the control of the entity doing the hooking

The working directory at process launch is controlled by whoever is launching the app, which is the scenario I had in mind.

> if the user puts the file in that directory they can give a relative path to it

I'm hoping that the user will tend not put the startup hook in the app directory. The case I envision is having a "global" startup hook inject code into different .NET processes on launch.

The case where I can see a relative path being useful is the following:
```
DOTNET_STARTUP_HOOKS='./startupHook/hook.dll' ./app/bin/main.dll
```
It could make it easier to launch the process from the command line if the hook dll is in a location close to the current working directory. However, given that this is supposed to be a low-level hook, not one frequently used by non-devs, I think it's fine to require absolute paths in this scenario.

Documentation/design-docs/host-startup-hook.md
sbomer: Ok, I'll keep the documented exceptions but remove the error text from the spec.

Documentation/design-docs/host-startup-hook.md
sbomer: Ok, I'll remove the exception messages from the spec.

Documentation/design-docs/host-startup-hook.md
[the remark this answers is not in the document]
vancem: You are correct that a 'wrapper' tool that launches the app will use the Env variable.   The issue is other scenarios.  Today apps that use AppInsights require that you modify the app to include its telemetry.   It would be a useful feature for them to be able to instrument the app for you (post build) without breaking any signing that happened (you don't want to modify their DLLs), by plopping down their DLL and modifying the app's config.    Thus there are multiple scenarios. 

I want to mull this over some more...

sbomer:
@davidfowl similar issues exist for this startup hook. The hook on its own does not introduce any loading policy, so whoever is deploying the hook/app needs to ensure that their load policies work well together. This feature is more low-level than `--additionalDeps` and doesn't even try to make loading of transitive closures just work. From what I've heard, `--additionalDeps` made this sound easy but was difficult to use in practice. The startup hook makes no claim that this will be easy in the first place. :)

> those are the sort of things that people don't think through when building these plugins

I'm hoping that folks won't see this as a solution to their plugin problems. I think the use case for the startup hooks is a bit more narrow than that, and more work needs to go into providing a proper plugin model.

> I think we need strong guidance for the types of code people should put in these startup hooks

I added a section with some guidance to start with. Please let me know if you think something needs to be added, changed, clarified, etc.!

Documentation/design-docs/host-startup-hook.md
AaronRobinsonMSFT: I was against the idea of having a hard-coded default type name just because it seemed like too much magic. However, if it wasn't a default, but rather the **required** name of the type, I think that is entirely reasonable.

jkotas: > Also I think you will find that we use ';' as a separator in other places already

Do you have a example? I believe that the .NET APIs actually lead you to using platform native path separator in environment variables because of we have `Path.PathSeparator` since forever. In other words, the platform specific shim has been created for you already.

> hard-coded default type name 

+1 The configurability of type name seems to have net negative value. It just requires users to do more typing in vast majority of cases. A required hardcoded type name should be just fine.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+}
+```
+
+## Error handling details
+
+Problems with the startup hook should be fairly straightforward to
+diagnose. All of these exceptionts will contain the startup hook path
+(`System.StartupHookProvider.ProcessStartupHooks`) on the stack
```
vitek-karas: Personally I don't like a presence of a specific stack frame as a diagnostics mechanism - too easy to miss, and somewhat hard to recognize.
It also doesn't always provide enough information.
The exceptions thrown during execution of a startup hook should at the very least include which startup hook failed. Not all of the below might include this now. Also, if the hook itself throws for whatever reason, the information will only exist on the callstack.
.NET already has a similar concept in static .ctors. If there's an exception thrown from one of them we throw TypeInitializationException which wraps the actual exception inside. We could introduce similar StartupHookException as well.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -0,0 +1,201 @@
...
+            string sharedAssemblyPath = // find assemblyName in shared location...
+            if (sharedAssemblyPath != null)
+                return AssemblyLoadContext.Default.LoadFromAssemblyPath(sharedAssemblyPath)
+            return null;
+        }
+    }
+}
+```
+
+## Error handling details
```
vitek-karas: More of a technical consideration: Since this happens so early on, there's no way for the program to setup any kind of exception handling. Please make sure that the experience is meaningful and document it somewhere. Not sure what .NET Core does on top-level unhandled exception, but I know that .NET Framework will trigger Watson and then write it out to console.
I guess it should be the same as what happens for top-level unhandled exception...

Documentation/design-docs/host-startup-hook.md
sbomer: All right, so it sounds like we're leaning towards platform-specific separators, without the type name in the syntax. I'll specify that the type **must** be named "StartupHook" (in the global namespace). This will allow the hook to be set without quotes on unix and windows:
```
DOTNET_STARTUP_HOOKS=/path/to/startupHook.dll:/path/to/startupHook2.dll
```
and on windows:
```
set DOTNET_STARTUP_HOOKS=d:\path\to\startupHook.dll;d:\path\to\startupHook2.dll
```

Documentation/design-docs/host-startup-hook.md
jkotas: > Also it is typically to check for some settings from app folder. So ideally startup hook should know where to read appsettings file from.

Does the hook have a good way to find out the name of the app being executed? I suspect that the hooks may be running too early right now before the app name is configured in CoreCLR.

sbomer: With the current design, there's no built-in way to get the app name, but it would be possible to get the app directory (`AppContext.BaseDirectory`, or `AppContext.GetData("APP_CONTEXT_BASE_DIRECTORY")`). This might be sufficient depending on the configuration mechanism used.

If getting the app name is important we could either:
- Set an additional app variable when initializing coreclr, before calling the startup hooks - maybe `APP_CONTEXT_APP_NAME`, and expose it as `AppContext.AppName`.
- Or, call the startup hooks on a later code path (somewhere in `coreclr_execute_assembly`, instead of from the host), where the app name is already known by the clr. We would then also have to decide how to expose it, likely something like `AppContext.AppName`.

sbomer: I think the first option I described above would be the way to go if we do make the app name available. The reasoning being that for other activation scenarios that don't go through `coreclr_execute_assembly`, like COM, we would still be able to support a startup hook on the current path. This leaves us the flexibility to expose the app name in a way that's independent of the startup hook feature for activation scenarios that do have an app.

sbomer:
@jeffschwMSFT please give this another look when you get a chance.

Documentation/design-docs/host-startup-hook.md
sbomer: I noted that the behavior is the same as for top-level exceptions in the doc, thanks!

Documentation/design-docs/host-startup-hook.md
sbomer: As discussed, `ProcessStartupHooks` in the stack trace isn't the only indication - most of the exception messages will include information that makes it easy to tell which hook is failing. We can still consider wrapping exceptions from the startup hook in a special exception, but I don't think that's necessary.

Documentation/design-docs/host-startup-hook.md
```diff
@@ -51,13 +51,14 @@ will be inherited by child processes by default. It is up to the
 desired.
 
 Specifically, hostpolicy starts up coreclr and sets up a new
-AppDomain. It then invokes a private method in
-`System.Private.CoreLib`, which will call each
+AppDomain, passing in the startup hook variable if it was
```
jeffschwMSFT: Can you add the exact key name here?  https://github.com/dotnet/core-setup/blob/7e88484b2ab88b98088c2f40b620c7593b0fdd16/src/corehost/cli/hostpolicy.cpp#L97

sbomer:
Two more concerns:
1. Unfortunately, it looks like it won't be possible to use the startup hook to forcibly preload app dependencies - `LoadFromAssemblyPath` will resolve the assembly's simple name to the path it was given on the TPA list.
2. Even if we make some changes to the clr assembly loading logic to let `LoadFromAssemblyPath` override TPA paths, moving the hook to `ExecuteMainMethod` (after reading the main method's attributes) means that the main dll's static dependencies will already be loaded by the time the startup hook executes.

The takeaway is that we're back in a place where the startup hook alone won't be useful as a way to modify load behavior of dependencies of main on the TPA list.

@yizhang82 for Service Fabric, it could be used to set a resolving event that applies to non-TPA assemblies. It sounds like this would still enable you to add new API surface for newly published apps by having them reference the new assembly, without putting it on the TPA list or deploying it app-local. Can you confirm that this is your intended use?

jeffschwMSFT:
@sbomer can we share the modified components with @yizhang82?  I was talking with @jkotas about the Service Fabric scenario and it seems they may have a viable way of using this feature to enable their scenario.

sbomer:
@jeffschwMSFT

sbomer:
Talked to @yizhang82 - so far it sounds like we'll enable their scenario as I described. I'm going to work with them to test this, since this area has given us a few unexpected surprises so far...

samsosa:
From a security perspective this could be a major risk factor. Malicious dll injection made easy? Are there any plans to protect against unwanted loading?

AaronRobinsonMSFT:
>Malicious dll injection made easy? 

These kind of vectors already exist in coreclr. See the `CORECLR_PROFILER_PATH` environment variable. A discussion was had regarding this issue.

Issue #4376, opened 2018-07-18
Apps fail when app-local binary is lower than the version provided by a shared runtime


natemcmaster:
## Steps to reproduce
Install ASP.NET Core 2.1.3 and `dotnet run` this project
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.1" />
    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.1.2" />
  </ItemGroup>
</Project>
```
```csharp
using System;
using Microsoft.AspNetCore;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Server.Kestrel.Core;

public class Program
{
    public static void Main(string[] args)
    {
        Console.WriteLine("Kestrel assembly location = " + typeof(KestrelServerOptions).Assembly.Location);
        Console.WriteLine("Kestrel assembly version  = " + typeof(KestrelServerOptions).Assembly.GetName().ToString());
        Console.WriteLine("Deps files = " + AppContext.GetData("APP_CONTEXT_DEPS_FILES") as string);
        
        // run the kestrel server
        WebHost.CreateDefaultBuilder(args).Build().Run();
    }
}
```

## Expected  behavior

I expected the host to rollforward to the latest Microsoft.AspNetCore.App shared runtime (2.1.3), and to prefer the version of Microsoft.AspNetCore.Server.Kestrel.Core.dll it provides (AssemblyVersion=2.1.3.0) because it is higher than the app-local version (Version=2.1.2.0)

## Actual behavior

The host prefers the app-local version (2.1.2.0), and when it rollsforward to the highest patch, the application fails with `FileLoadException` due to an assembly verison downgrade forced upon the other binaries in the Microsoft.AspNetCore.App shared runtime.

```
$ dotnet run
Kestrel assembly location = /Users/namc/.nuget/packages/microsoft.aspnetcore.server.kestrel.core/2.1.2/lib/netcoreapp2.1/Microsoft.AspNetCore.Server.Kestrel.Core.dll
Kestrel assembly version  = Microsoft.AspNetCore.Server.Kestrel.Core, Version=2.1.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60
[...]

PR #4423, opened 2018-08-01
Always compare assembly and file versions for TPA assets

steveharter:
Per the linked issue and offline discussions, we now will now compare assemblyVersion\fileVersion for patch roll-forward cases to support better serviceability and to prevent run-time FileLoadException from occurring due to assembly-to-assembly versions being too low.

Fixes https://github.com/dotnet/core-setup/issues/4376

This fix is expected to be ported to 2.2 and 2.1.x. The change affects `hostpolicy.dll` which is local to each installed version of the runtime (in Microsoft.NETCore.App).

Previously the runtime only compared versions for minor- and major roll-forward scenarios. The reason was because it allowed for the app to have complete control over any duplicated assemblies that exist in the app bin location, to support build-from-source and other explicit scenarios.

The reason this issue manifests for patch-roll forward:
- The current tooling leaves behind "duplicated" assemblies in the app bin folder. Ideally this shouldn't occur (SDK issue TBD) because they are the exact same version as the ones in the targeted framework. This occurs in the linked issue when the application explicitly references the same packages that also exist in the targeted framework
- The linked issue has duplicated assemblies that have different versions than the framework. This does not occur for all frameworks. For example, Microsoft.NETCore.App framework does not normally increment assembly versions for patches, unlike the ASP.NET frameworks. This will cause a run-time exception if the app assembly is loaded first, and then the framework attempts to load the newer version due to an assembly-to-assembly reference.

With this change, if an application wishes to always load their app bin copies of framework assemblies (for example, if they have a build-from-source scenario with new features or bug fixes) then they should prevent Minor\Major roll-forward and patch roll-forward from occurring by modifying their runtimeconfig.json to add `applyPatches:false` and `rollForwardOnNoCandidateFx" : 0` like so:
```
{
  "runtimeOptions": {
    "tfm": "netcoreapp2.1",
    "applyPatches" : false,
    "rollForwardOnNoCandidateFx" : 0,
    "framework": {
      "version": "2.1.0"
    }
  }
}
```

cc @natemcmaster @jeffschwMSFT 

note: this does not support comparing versions with native assemblies and resource assemblies. These are handled differently by the host and the CLR. Currently the host does not make any version check for roll-forward (Major\Minor\Patch) and the app bin location always wins for these files. Supporting comparing versions would require CLR changes to allow for per-file paths (instead of a shared path list like today) and we would also need to address issues on Unix which won't always have a comparable fileVersion for native files. There is no assemblyVersion for native files, so only fileVersion would be used. Additional scenarios (around the expected serviceability of these files) and feedback on the Unix fileVersion are required to address this properly.


src/corehost/cli/deps_resolver.cpp
```diff
@@ -521,31 +521,27 @@ bool deps_resolver_t::resolve_tpa_list(
...
             {
                 return false;
             }
         }
     }
 
-    // If additional deps files were specified that need to be treated as part of the
-    // application, then add them to the mix as well.
-    for (const auto& additional_deps : m_additional_deps)
+    // Probe FX deps entries after app assemblies are added.
```
steveharter: Note I reverted the order of probing additional-deps to how the code was before 2.1. This was changed in 2.1 to support issues with additional-deps and roll-forward. However, since we always compare versions now the change made in 2.1 is no longer necessary.

eerhardt:
> this does not support comparing versions with native assemblies and resource assemblies

How do resource assemblies work in this case?  If the host loads the assembly from a shared fx folder, it better use the resource assemblies that are in the shared fx folder, right?  Or else you could get missing resource exceptions.

joshfree:
@dotnet-bot test this please

steveharter:
>> How do resource assemblies work in this case? If the host loads the assembly from a shared fx folder, it better use the resource assemblies that are in the shared fx folder, right? Or else you could get missing resource exceptions.

Resource assemblies have the same roll-forward problem as native files where the app will always win. So if shared\foo.dll assembly is loaded in the shared framework location because it is newer than the app bin copy, its resource assembly will still be loaded from the app bin location if the resource file exists. If the resource file does not exist there, then it will be loaded from the shared location. So yes it is possible to get a missing resource exception during roll-forward, but only if the app actually had the resource assemblies to begin with.

steveharter:
Pending any other new feedback, this will go in later today.

steveharter:
Committing this; will ask for test verification before creating 2.2 and 2.1.x PRs

PR #4436, opened 2018-08-08
Port fx-dependent apphost fix

steveharter:
Awaiting shiproom approval. Note this may have to be ported to 2.1.x as well pending offline mail discussion.

Fixes https://github.com/dotnet/core-setup/pull/4436 

Original issue: https://github.com/dotnet/core-setup/issues/4169

#### Description 
The issue applies to the new framework-dependent apphost feature that was added in 2.1 which allows the apphost (<myapp>.exe) to use the shared framework.

This feature was added in 2.1.0 to support global tools, which currently does not experience the issue. However, the SDK is adding the capability for users to create their own framework-dependent apphost and that results in an application refusing to launch when the apphost is located in the same directory as the current working directory.

#### Customer Impact 
Without the fix, the SDK will not be able to add generalized support for a framework-dependent apphost in 2.2.

#### Regression? 
 No

#### Risk
Minimal.  The change affects code for 1.x compat when using the 1.x corehost.exe in a “split framework” mode. That mode was used for tooling\testing with a local copy of the coreclr and not with a shared framework. Any potential side-effect from this fix would be a roll-forward from 1.x to 2.1 and using the “split” mode.

### DIFF
--- a/BranchInfo.props
+++ b/BranchInfo.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>1</MinorVersion>
-    <PatchVersion>3</PatchVersion>
+    <PatchVersion>4</PatchVersion>
 
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
 


steveharter:
Closing as we'll merge this into 2.1.x first

PR #4439, opened 2018-08-09
Updating init-tools to be inline with the CoreFX/CoreCLR copy

tannergooding:
The file had gotten out of sync on a number of minor (but important) changes.


tannergooding:
FYI. @eerhardt

Issue #4236, opened 2018-06-14
dotnet --info only finds same architecture, and lists entries twice when called with absolute path.


jkotas:
_From @leecow on June 14, 2018 16:38_

@markusschaber commented on [Fri May 18 2018](https://github.com/dotnet/core-setup/issues/4154)

## Steps to reproduce
call "dotnet --info" or "C:\Absolute\Path\to\dotnet.exe --info"

See https://gist.github.com/markusschaber/480879b83ec9a22077c6c31e3d7f8346 for full repro.

## Expected  behavior
All installed SDKs and runtimes are listed exactly once.

## Actual behavior
When calling without absolute path, only the SDKs and runtimes of the same architecture (e. G. x86 vs x64) are listed.

When calling with absolute path, it also only finds the SDKs and runtimes of the same architecture, but lists all of them twice.

## Environment data
Microsoft Windows [Version 6.1.7601], 64 bit (x64).

---

@markusschaber commented on [Fri May 18 2018](https://github.com/dotnet/core-setup/issues/4154#issuecomment-390234367)

Remark: I filed this report as recommended on gitter (https://gitter.im/dotnet/coreclr) - I'll be on holiday for the next to weeks, with only limited ability to monitor this issue. Thanks for your patience.

---

@steveharter commented on [Mon May 21 2018](https://github.com/dotnet/core-setup/issues/4154#issuecomment-390657990)

On`C:\Absolute\Path\to\dotnet.exe` showing duplicate entries: if the path exactly matches (casing included) a previous entry, then it is not shown. In the example `c:` was used for the path of `dotnet.exe` while the physical disk is `C:` causing two entries to be found. We can attempt to filter ignoring case on Windows, but on Unix we can't since the file system can be case-sensitive.

---

@markusschaber commented on [Wed May 23 2018](https://github.com/dotnet/core-setup/issues/4154#issuecomment-391433151)

Maybe the pathes could be normalized, like the System.IO.Path.GetFullPath() method does on the classic .NET? This could also resolve symlinks etc...

_Copied from original issue: dotnet/coreclr#18466_

steveharter:
@markusschaber my recommendation is to remove the duplicated entries as suggested with realpath.

However, I do not recommend showing both x64 and x86 frameworks in the same list, because the instance of dotnet.exe being used will only look in either the x64 or x86 folder for compatible frameworks, not both. 

[...]

PR #4441, opened 2018-08-09
Avoid searching multilevel dir when same as dotnet root

steveharter:
On Windows only, if an explicit path is specified when launching `dotnet.exe`, and that path is also the multilevel lookup location ('C:\Program Files\dotnet') it is possible for `--info`, `--list-runtimes`, `--list-sdks` and the actual probing for the runtime and SDK to return duplicate entries when there exists casing differences between the actual file system and the path specified with `dotnet.exe`.

This affects the output of the commands (duplicated information) plus adds an unnecessary probing pass when finding the SDK or runtime (performance).

The fix is to perform a case-insensitive compare on Windows.

Fixes https://github.com/dotnet/core-setup/issues/4236

Also found and detected is that the SDK would *always* perform an extra unnecessary probing pass when the path to `dotnet.exe` is also the multilevel location (irregardless of casing). It occurred because of a trailing slash that was checked in the other cases, but not for the core SDK probing. Fixing this should help SDK performance as it will remove a few I\O probing hits.

@nguerrera do you think this should be ported to 2.1.x\2.2? Mostly based on the SDK performance aspect.

No tests were added because it is not possible to change the multilevel location via %ProgramFiles% and %ProgramFiles(x86)% environment variables to a test location. Changing them in the current environment does not actually change the value obtained when reading them (they appear to be special). We could change our internal code to use an indirection (e.g. create a new environment variable) but that would be too intrusive just for this IMO.

### DIFF
--- a/src/corehost/cli/fxr/sdk_resolver.cpp
+++ b/src/corehost/cli/fxr/sdk_resolver.cpp
@@ -179,18 +179,20 @@ bool sdk_resolver_t::resolve_sdk_dotnet_path(const pal::string_t& dotnet_root, c
     std::vector<pal::string_t> global_dirs;
     bool multilevel_lookup = multilevel_lookup_enabled();
 
-    // dotnet_root contains DIR_SEPARATOR appended that we need to remove.
-    assert(!dotnet_root.empty());
-    pal::string_t dotnet_root_temp = dotnet_root;
-    remove_trailing_dir_seperator(&dotnet_root_temp);
-
-    hive_dir.push_back(dotnet_root_temp);
+    pal::string_t dotnet_root_temp;
+    if (!dotnet_root.empty())
+    {
+        // dotnet_root contains DIR_SEPARATOR appended that we need to remove.
+        dotnet_root_temp = dotnet_root;
+        remove_trailing_dir_seperator(&dotnet_root_temp);
+        hive_dir.push_back(dotnet_root_temp);
+    }
 
     if (multilevel_lookup && pal::get_global_dotnet_dirs(&global_dirs))
     {
         for (pal::string_t dir : global_dirs)
         {
-            // Avoid duplicate of dotnet_dir_temp
+            // Avoid duplicate of dotnet_root_temp
             if (!pal::are_paths_equal_with_normalized_casing(dir, dotnet_root_temp))
             {
                 hive_dir.push_back(dir);
--- a/src/corehost/cli/fxr/sdk_resolver.cpp
+++ b/src/corehost/cli/fxr/sdk_resolver.cpp
@@ -179,16 +179,22 @@ bool sdk_resolver_t::resolve_sdk_dotnet_path(const pal::string_t& dotnet_root, c
     std::vector<pal::string_t> global_dirs;
     bool multilevel_lookup = multilevel_lookup_enabled();
 
-    if (!dotnet_root.empty())
-    {
-        hive_dir.push_back(dotnet_root);
-    }
+    // dotnet_root contains DIR_SEPARATOR appended that we need to remove.
+    assert(!dotnet_root.empty());
+    pal::string_t dotnet_root_temp = dotnet_root;
+    remove_trailing_dir_seperator(&dotnet_root_temp);
+
+    hive_dir.push_back(dotnet_root_temp);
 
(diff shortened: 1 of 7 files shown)


src/corehost/cli/fxr/sdk_resolver.cpp
```diff
@@ -179,16 +179,22 @@ bool sdk_resolver_t::resolve_sdk_dotnet_path(const pal::string_t& dotnet_root, c
     std::vector<pal::string_t> global_dirs;
     bool multilevel_lookup = multilevel_lookup_enabled();
 
-    if (!dotnet_root.empty())
-    {
-        hive_dir.push_back(dotnet_root);
-    }
+    // dotnet_root contains DIR_SEPARATOR appended that we need to remove.
+    assert(!dotnet_root.empty());
```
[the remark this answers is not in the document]
nguerrera: IIRC, I think it can be empty if passed in that way from hostfxr_resolve_sdk. I think I even documented that you can do this to use only the multilevel locations. I have to double check the code, though.

nguerrera:
I did manage to test multi-level lookup in #4301 (really need to get that in). I basically pass another variable to the child process asking it to change ProgramFiles.

src/corehost/cli/fxr/sdk_resolver.cpp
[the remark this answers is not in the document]
steveharter: Yes I see the doc for hostfxr_resolve_sdk does say that. I'll add the check back and adjust the check that removes the last slash

steveharter:
>> I did manage to test multi-level lookup in #4301 (really need to get that in). I basically pass another variable to the child process asking it to change ProgramFiles.

OK, I'll look at how you did %ProgramFiles% and add a test for this (if it works as expected)

src/corehost/common/pal.windows.cpp
```diff
@@ -519,3 +519,9 @@ bool pal::is_running_in_wow64()
     }
     return (fWow64Process != FALSE);
 }
+
+bool pal::are_paths_equal_with_normalized_casing(const string_t& path1, const string_t& path2)
+{
+    // On Windows, paths are case-insensitive
+    return (strcasecmp(path1.c_str(), path2.c_str()) == 0);
```
nguerrera: This is:

https://github.com/dotnet/core-setup/blob/24ba120b0ca1ba60a24a79b7dc95866cd3e9f799/src/corehost/common/pal.h#L125

And I believe [_wcsicmp](https://msdn.microsoft.com/en-us/library/k59z8dwe.aspx) behavior varies by locale.

I think it should be [CompareStringOrdinal](https://docs.microsoft.com/en-us/windows/desktop/api/stringapiset/nf-stringapiset-comparestringordinal) with bIgnoreCase=true.

Since we already seem to be using strcasecmp in other places where it should probably be ordinal-ignore-case, I'd be fine with moving this feedback item to an issue to look at it everywhere in the repo.

src/corehost/common/pal.unix.cpp
```diff
@@ -641,3 +641,9 @@ bool pal::is_running_in_wow64()
 {
     return false;
 }
+
+bool pal::are_paths_equal_with_normalized_casing(const string_t& path1, const string_t& path2)
+{
+    // On Unix paths are case-sensitive.
```
nguerrera: The default Mac OS X file-system is case-insensitive.

nguerrera:
> @nguerrera are you OK with this PR as-is, or do you want me to explore using the main hostfxr entrypoint approach for testing reasons?

I'm OK with just manual testing for this.

nguerrera:
Adding @JeremyKuhne for his thoughts on the best way to compare these paths x-plat and from native code.

src/corehost/common/pal.unix.cpp
steveharter: Thanks. New Commit to fix this

src/corehost/common/pal.windows.cpp
steveharter: _wcsicmp uses the default "C locale" which should be sufficient, at least for ASCII. Not sure how Unicode file systems would work, however. Created issue https://github.com/dotnet/core-setup/issues/4459 to track

Issue #4249, opened 2018-06-18
Move the "Finalize build" logic to Core-Setup-Sign-And-Publish.json in 1.x.x


wtgodbe:
Now that we have a pipeline job in 1.x.x that handles finalizing the package publishing, we should move all of the finalization tasks from https://github.com/dotnet/core-setup/blob/release/1.1.0/build_projects/dotnet-host-build/PublishTargets.cs#L74 to that .json file.

CC @eerhardt @weshaggard

PR #4442, opened 2018-08-09
Publish signed packages to myget in 1.1.0 pipeline

wtgodbe:
This attempts to change the workflow for finalizing/publishing in Core-Setup 1.1.0.

Old workflow:

1. Run all build legs - whoever finishes last, runs `finalize`
    - copy published blobs into `latest` container
    - publish packages to myget
    - update versions repo
2. Run `sign-and-publish` leg
    - download unsigned packages from blob storage
    - sign packages
    - reupload them to blob storage

New workflow:

1. Run all build legs
2. Run `sign-and-publish` leg
    - download unsigned packages from blob storage
    - sign packages
    - reupload them to blob storage
    - run `finalize` as above

For https://github.com/dotnet/core-setup/issues/4249

@eerhardt @dagood PTAL - also needs to be ported to 1.0.0

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/Dirs.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Dirs.cs
@@ -17,7 +17,7 @@ namespace Microsoft.DotNet.Cli.Build
 
         public static readonly string Intermediate = Path.Combine(Output, "intermediate");
         public static readonly string PackagesIntermediate = Path.Combine(Output, "packages/intermediate");
-        public static readonly string PackagesNoRID = Path.Combine(RepoRoot, "artifacts", "packages");
+        public static readonly string PackagesNoRID = Path.Combine(RepoRoot, "pkg", "packages", "AzureTransfer", "pkg");
         public static readonly string Packages = Path.Combine(Output, "packages");
         public static readonly string Stage1 = Path.Combine(Output, "stage1");
         public static readonly string Stage1Compilation = Path.Combine(Output, "stage1compilation");
--- a/build_projects/shared-build-targets-utils/Utils/Dirs.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Dirs.cs
@@ -17,7 +17,7 @@ namespace Microsoft.DotNet.Cli.Build
 
         public static readonly string Intermediate = Path.Combine(Output, "intermediate");
         public static readonly string PackagesIntermediate = Path.Combine(Output, "packages/intermediate");
-        public static readonly string PackagesNoRID = Path.Combine(RepoRoot, "pkg", "packages", "AzureTransfer", "pkg");
+        public static readonly string DownloadedPackagesForFinalPublish = Path.Combine(RepoRoot, "pkg", "packages", "AzureTransfer", "pkg");
         public static readonly string Packages = Path.Combine(Output, "packages");
         public static readonly string Stage1 = Path.Combine(Output, "stage1");
         public static readonly string Stage1Compilation = Path.Combine(Output, "stage1compilation");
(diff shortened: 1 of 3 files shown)


build_projects/shared-build-targets-utils/Utils/Dirs.cs
```diff
@@ -17,7 +17,7 @@ public static class Dirs
 
         public static readonly string Intermediate = Path.Combine(Output, "intermediate");
         public static readonly string PackagesIntermediate = Path.Combine(Output, "packages/intermediate");
-        public static readonly string PackagesNoRID = Path.Combine(RepoRoot, "artifacts", "packages");
+        public static readonly string PackagesNoRID = Path.Combine(RepoRoot, "pkg", "packages", "AzureTransfer", "pkg");
```
eerhardt: ?? any assets that are the output of the build need to keep going into the `artifacts` directory.

wtgodbe: This doesn't refer to packages that we've just built, but packages that we've downloaded to re-publish in the finalize step. https://github.com/dotnet/core-setup/blob/4cc6ac5f43f48b3a6fd3ab73763ce9f287c05fa2/build_projects/shared-build-targets-utils/Utils/Dirs.cs#L21 is where we put the produced packages from a build.

wtgodbe: Sure, how about `PackagesForFinalPublish` ?

wtgodbe: And this will only include packages built by Core-Setup (although that includes NetCore & NetCore.App)

dagood: That sounds pretty good, maybe also add Downloaded for `DownloadedPackagesForFinalPublish` to make it super clear they aren't being built on the machine.

dagood: Oh, and I guess I'm assuming this change is because of the switch to use the BuildToolsy download, rather than a CLI C# build project task, and it would take more effort to move the download location to `artifacts/something`. In general I agree this would belong in `artifacts/` otherwise.

wtgodbe:
Updated to address feedback. @eerhardt anything else that looks like it needs to be changed?

wtgodbe:
I believe the failures are due to https://github.com/dotnet/core-setup/commit/542adf337c94879ee1ac33852c660fd01ccdb9e5. CI was green before my most recent commit, which came after the commit above.

dagood:
Jenkins merges from the target branch before running CI on your changes, so you're probably right.

eerhardt:
>     "AzureContainerPackageDirectory": {

Is this used anymore?

---
Refers to: buildpipeline/Core-Setup-Sign-And-Publish.json:276 in 63550b4. [](commit_id = 63550b445fa138f734f775c56d985402453a568f, deletion_comment = False)

build_projects/shared-build-targets-utils/Utils/Dirs.cs
eerhardt: I don't understand why we are putting assets that the product produces into the source tree under the `pkg` folder.  Assets that are produced during the build should go in the `artifacts` folder.  Even if they are downloaded from somewhere else.

wtgodbe: Essentially this is the same as the `packages` folder in CoreFx/CoreClr, except it exists one level down the tree in `pkg`. This directory already exists before these changes, as it's where `init-tools.cmd` puts the packages it restores.

wtgodbe:
@eerhardt no it isn't, removed.

PR #4443, opened 2018-08-09
Bump versions for 1.0.13

wtgodbe:
@safern PTAL


wtgodbe:
@safern do I also need to add `winHttp` to the 2 project.jsons?

pkg/projects/Microsoft.NETCore/project.json
```diff
@@ -26,7 +26,7 @@
     "System.Linq.Expressions": "4.1.0",
     "System.Linq.Parallel": "4.0.1",
     "System.Linq.Queryable": "4.0.1",
-    "System.Net.Http": "4.1.1",
+    "System.Net.Http": "4.1.4-servicing-26809-01",
```
safern: I don't think you have to bump it here, we haven't touched this in 2 years.

PR #4465, opened 2018-08-14
Add host startup hook and tests

sbomer:
See the design here: https://github.com/dotnet/core-setup/pull/4421.

The tests depend on changes in coreclr: https://github.com/dotnet/coreclr/pull/19486, and will fail until a recent coreclr is picked up in core-setup. Do not merge until this happens.


src/corehost/cli/coreclr.cpp
```diff
@@ -52,6 +62,7 @@ bool coreclr::bind(const pal::string_t& libcoreclr_path)
     coreclr_initialize = (coreclr_initialize_fn)pal::get_symbol(g_coreclr, "coreclr_initialize");
     coreclr_shutdown = (coreclr_shutdown_fn)pal::get_symbol(g_coreclr, "coreclr_shutdown_2");
     coreclr_execute_assembly = (coreclr_execute_assembly_fn)pal::get_symbol(g_coreclr, "coreclr_execute_assembly");
+    coreclr_create_delegate = (coreclr_create_delegate_fn)pal::get_symbol(g_coreclr, "coreclr_create_delegate");
```
AaronRobinsonMSFT: I would suggest doing the following instead:

> `coreclr_create_delegate = (declspec(coreclr_create_delegate))pal::get_symbol(g_coreclr, "coreclr_create_delegate");`

The `declspec()` keyword is rather helpful and makes the intent here clearer.

src/test/Assets/TestProjects/StartupHookWithInstanceMethod/StartupHookWithInstanceMethod.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+internal class StartupHook
+{
+    public void Initialize()
+    {
+        Console.WriteLine("Hello from startup hook with instance method!");
```
AaronRobinsonMSFT: This is a failure right? It should throw an exception if called.

src/corehost/cli/coreclr.cpp
```diff
@@ -33,9 +33,19 @@ typedef pal::hresult_t(STDMETHODCALLTYPE *coreclr_execute_assembly_fn)(
     const char* managedAssemblyPath,
     unsigned int* exitCode);
 
+// Prototype of the coreclr_create_delegate function from coreclr.dll
```
jeffschwMSFT: Given the new design is the coreclr_create_delegate function necessary?

src/corehost/error_codes.h
```diff
@@ -5,33 +5,34 @@
...
+    InvalidConfigFile            = 0x80008093,
+    AppArgNotRunnable            = 0x80008094,
+    AppHostExeNotBoundFailure    = 0x80008095,
+    FrameworkMissingFailure      = 0x80008096,
+    HostApiFailed                = 0x80008097,
+    HostApiBufferTooSmall        = 0x80008098,
+    LibHostUnknownCommand        = 0x80008099,
+    LibHostAppRootFindFailure    = 0x8000809a,
+    SdkResolverResolveFailure    = 0x8000809b,
+    CoreClrCreateDelegateFailure = 0x8000809c,
```
jeffschwMSFT: Ditto... if coreclr_create_delegate is not called, then this may not be necessary

src/corehost/cli/coreclr.cpp
sbomer: It's not called anymore - I thought we should keep it here since other hosting scenarios are likely to use it. Or I can remove it and let @AaronRobinsonMSFT check in something similar.

src/test/Assets/TestProjects/StartupHookWithDependency/StartupHookWithDependency.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+internal class StartupHook
+{
+    public static void Initialize()
+    {
+        Console.WriteLine("Hello from startup hook with dependency!");
+
+        // A small operation involving NewtonSoft.Json to ensure the assembly is loaded properly
+        var t = typeof(Newtonsoft.Json.JsonReader);
```
jeffschwMSFT: A few additional tests we should add is around creating a CustomerAssemblyLoad context and one that preloads assemblies (dynamically).  This is one of the core golden path scenarios and we should ensure it is functional.

src/corehost/cli/coreclr.cpp
jeffschwMSFT: I would remove it, since it no longer aligns with the intention of this change.  The work should be shared so that @AaronRobinsonMSFT does not need to reinvent the wheel.

src/test/HostActivationTests/GivenThatICareAboutStartupHooks.cs
```diff
@@ -0,0 +1,622 @@
...
+            // Simple startup hook
+            dotnet.Exec(appDll)
+                .EnvironmentVariable(startupHookVarName, startupHookDll)
+                .CaptureStdOut()
+                .CaptureStdErr()
+                .Execute()
+                .Should()
+                .Pass()
+                .And
+                .HaveStdOutContaining("Hello from startup hook!")
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: Spoke offline. I am still not a fan, but since the core-setup infrastructure doesn't seem to test the return values of the exe this approach is the style of the repo - boo.

src/corehost/cli/coreclr.cpp
sbomer: Removed this function since the new implementation doesn't use it.

src/corehost/cli/coreclr.cpp
sbomer: Removed.

src/test/Assets/TestProjects/StartupHookWithMultipleIncorrectSignatures/StartupHookWithMultipleIncorrectSignatures.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+internal class StartupHook
+{
+    public static int Initialize()
+    {
+        Console.WriteLine("Hello from startup hook returning int!");
```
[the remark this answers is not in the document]
sbomer: The negative test is looking for an exception from the hook provider code, so I don't think it makes a difference if the hook itself throws. If anything I'd expect that to make this test more fragile because it could make the process fail even if the hook is run, giving a false positive.

I added some comments to all of the test projects that should make it more clear how they're being used.

src/test/Assets/TestProjects/StartupHookWithDependency/StartupHookWithDependency.cs
sbomer: As discussed, the preloading scenario won't work due to bootstrapping issues. I don't think there's a good testcase for a custom ALC here either - that seems like it belongs to the plugin model work.

sbomer:
@dotnet-bot test Windows_NT arm Debug Build
@dotnet-bot test Windows_NT x86 Debug Build

PR #4468, opened 2018-08-14
Set PB_OverridePackageSource inside of each build's json

wtgodbe:
It turns out that if we want to pass an override directory relative to where the current build is happening, it needs to be set directly inside the .json corresponding to each individual build, rather than being passed from pipebuild (which, naturally, doesn't know what directories its child builds will be run out of).

This instead sets a velue for `OverridePackageSourc` in dir.props, given that we're running an internal build. It also changes `PB_OverridePackageSource` to `PB_OverrideDockerPackageSource` for Linux builds, as the value previously set for `PB_OverridePackageSource` is correct for the builds which run inside of Docker containers.

I should also change `PB_OverridePackageSource` to `PB_OverrideDockerPackageSource` in the VSTS pipebuild definition once this is merged.

@dagood @dseefeld PTAL

### DIFF
--- a/buildpipeline/Core-Setup-Linux-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Linux-Arm-BT.json
@@ -610,6 +610,9 @@
     },
     "PB_InternalBuild": {
       "value": "false"
+    },
+    "PB_OverridePackageSource": {
+      "value": "/p:OverridePackageSource=$(PB_SourcesDirectory)/AzureCoreFxDownload/Release"
     }
   },
   "demands": [
--- a/buildpipeline/Core-Setup-Linux-Arm-BT.json
+++ b/buildpipeline/Core-Setup-Linux-Arm-BT.json
@@ -281,7 +281,7 @@
       },
       "inputs": {
         "filename": "docker",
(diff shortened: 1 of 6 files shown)


chcosta:
Is this preferable over setting the property in some props file (like https://github.com/dotnet/core-setup/blob/master/dir.props) if `OfficialBuildId != ''`?

wtgodbe:
https://github.com/dotnet/core-setup/blob/release/2.0.0/buildpipeline/Core-Setup-Windows-BT.json#L806

It gets set to `$(Build.SourcesDirectory)\\core-setup`. So, you're right, that would work. However, we only want to set the package override dir during Internal builds, so the property would have to be conditioned on that. We already have stuff checked into the pipeline folder that's conditioned on `PB_InternalBuild=true`, so I don't see why we couldn't do the same in dir.props.

chcosta:
I feel like "IsOfficial" was commonly being used for Buildtools repos but I can only find a reference to it in `Tools\CloudTest.Helix.targets`.  If that's true, then you should probably follow suit.  If not, then "InternalBuild" sounds find to me.  I don't remember if that value gets explicitly set by Pipebuild or somewhere else.

wtgodbe:
`PB_InternalBuild` is what we use in all of the core repos, at least for 2.0.0 & earlier.

Issue #4453, opened 2018-08-12
Make Environment.Version return the actual runtime version


steveharter:
@davidfowl commented on [Mon Jul 16 2018](https://github.com/dotnet/corefx/issues/31099)

Today on a .NET Core 2.1.0 project printing `Environment.Version` prints out `4.0.30319.42000` which is a hard coded constant in the runtime https://github.com/dotnet/coreclr/blob/ce0a2617d40bc217b8e0a2137e340afe585afcb1/src/System.Private.CoreLib/src/System/Environment.cs#L277. This should print out the *actual* runtime version based on the policy applied by the host (like 2.1.1 if that's what runtime version is actually being used).

This also needs to be used as the version logged into the event log when there's an unhandled exception reported by the runtime. Today is a bogus version which isn't very useful. 

---

@weshaggard commented on [Mon Jul 23 2018](https://github.com/dotnet/corefx/issues/31099#issuecomment-407190291)

Today we don't have the "marketing version" (i.e. 2.1.1) embedded anywhere that we can consume so the best we could really do would be to return the file version of corelib. 

---

@jkotas commented on [Mon Jul 23 2018](https://github.com/dotnet/corefx/issues/31099#issuecomment-407194700)

The host knows this version - it can pass it to the runtime as key-value pair.

---

@weshaggard commented on [Mon Jul 23 2018](https://github.com/dotnet/corefx/issues/31099#issuecomment-407197277)

>The host knows this version - it can pass it to the runtime as key-value pair.

I assume it only knows it for applications running on the shared framework correct? What should we return for standalone applications?

---

@jkotas commented on [Mon Jul 23 2018](https://github.com/dotnet/corefx/issues/31099#issuecomment-407202854)

All applications have it in .deps.json file.

---

@weshaggard commented on [Mon Jul 23 2018](https://github.com/dotnet/corefx/issues/31099#issuecomment-407205744)

I assume you are referring to the package version for Microsoft.NETCore.App (or runtime.[rid].Microsoft.NETCore.App) then? If so yes that may work for now but that may not be in the deps file in the future but we could probably ensure some version remains for this purposes. 

---

@abbotware commented on [Mon Jul 23 2018](https://github.com/dotnet/corefx/issues/31099#issuecomment-407274807)

I've been looking for away to get the RuntimeFrameworkVersion, so I tried all the API's that I could find (#31249) and none seemed to return anything useful - Any programmatic way to get the 'marketing version' would be good enough for me.  

[...]

PR #4470, opened 2018-08-14
Add appdomain variable for current framework version

steveharter:
This adds appdomain variable "FX_PRODUCT_VERSION" which returns the product version of Microsoft.NETCore.App, so, for example, "3.0.0-preview1". It works in both framework-dependent and self-contained modes.

Fixes https://github.com/dotnet/core-setup/issues/4453

### DIFF
--- a/src/test/Assets/TestProjects/StandaloneApp/Program.cs
+++ b/src/test/Assets/TestProjects/StandaloneApp/Program.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Reflection;
 
 namespace StandaloneApp
 {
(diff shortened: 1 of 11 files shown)


src/test/Assets/TestProjects/StandaloneApp/Program.cs
```diff
@@ -9,8 +10,19 @@ public static void Main(string[] args)
             Console.WriteLine("Hello World!");
             Console.WriteLine(string.Join(Environment.NewLine, args));
 
+            Console.WriteLine($"Framework Version:{GetFrameworkVersionFromAppDomain()}");
+
             // A small operation involving NewtonSoft.Json to ensure the assembly is loaded properly
             var t = typeof(Newtonsoft.Json.JsonReader);
         }
+		
+        public static string GetFrameworkVersionFromAppDomain()
```
davidfowl: This doesn't have the app domain type?

steveharter: I just cleaned that up. Thanks

steveharter:
@davidfowl do you want this ported to 2.1.x\2.2?

steveharter:
Merging this; I will be OOF until 8/27 so if there is additional feedback I can address when I return

davidfowl:
Yay!

davidfowl:
cc @natemcmaster @mikeharder @DamianEdwards

davidfowl:
@steveharter can we still back port this? I'm not sure it meets the bar but we can try...

jeffschwMSFT:
@davidfowl what is the 2.2 scenario for this work?
cc @swaroop-sridhar

davidfowl:
It's not critical, it mostly having a sane way to know what you are running on. Various diagnostic pages today show the wrong version.

PR #4474, opened 2018-08-15
port fx-dependent apphost fix in hostfxr.dll

steveharter:
Awaiting shiproom approval.

Fixes https://github.com/dotnet/core-setup/issues/4192
Original issue: #4169
#### Description
The issue applies to the new framework-dependent apphost feature that was added in 2.1 which allows the apphost (.exe) to use the shared framework.
This feature was added in 2.1.0 to support new global tools functionality, which currently does not encounter the issue. However, the SDK is adding the capability for users to create their own framework-dependent apphost and that results in an application refusing to launch when the apphost is located in the same directory as the current working directory.
#### Customer Impact
Without the fix, the SDK will not be able to add generalized support for a framework-dependent apphost in 2.2 (global tools, as mentioned prior, is not affected as it does not encounter the issue).
#### Regression?
No
#### Risk
Minimal. The change affects code for 1.x compat when using the 1.x corehost.exe in a “split framework” mode. That mode was used for tooling\testing with a local copy of the coreclr and not with a shared framework. Any potential side-effect from this fix would be a roll-forward from 1.x to 2.1 and using the “split” mode.

### DIFF
--- a/src/corehost/cli/libhost.cpp
+++ b/src/corehost/cli/libhost.cpp
@@ -51,23 +51,31 @@ void get_runtime_config_paths(const pal::string_t& path, const pal::string_t& na
 
 host_mode_t detect_operating_mode(const host_startup_info_t& host_info)
 {
-    if (coreclr_exists_in_dir(host_info.dotnet_root) || pal::file_exists(host_info.app_path))
+    if (coreclr_exists_in_dir(host_info.dotnet_root))
     {
-        pal::string_t own_deps_json = host_info.dotnet_root;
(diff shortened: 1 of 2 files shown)


steveharter:
@jeffschwMSFT (or others with permission) I will be OOF until 8\27. If this is ready to go in before I return, please merge (marked servicing approved without any further comments to delay)

danmosemsft:
Removing servicing-consider as this needs to be set by "little shiproom" and the email thread seems not concluded.

jamshedd:
Approved for 2.1.5

Issue #4314, opened 2018-06-29
Host should respect application manifests defined in assembly


AaronRobinsonMSFT:
## Steps to reproduce
Create an assembly that expects RegFree COM via a app.manifest
The host binary does not respect the app manifest defined in the assembly
COM activation fails since the host's manifest does not reflect the dependent assembly

This is not limited to to just RegFree COM, but to anything placed in the assembly manifest.

See [sdk#1899](https://github.com/dotnet/sdk/issues/1899)

## Expected  behavior
RegFree COM activation should succeed

## Actual behavior
COM activation fails

cc @jeffschwMSFT @luqunl @ericstj

ericstj:
Thanks for opening this @AaronRobinsonMSFT.  I think its important to consider this for both self-contained application which can ship a specialized copy of the host EXE as well as for shared-framework applications which don't contain a specialized host.  In the later case we can't rely on host specialization to embed the windows manifest.

We may be able to have the host call [activation context APIs](https://docs.microsoft.com/en-us/windows/desktop/SbsCs/activation-contexts).  It seems like these would allow the host to create a context for the application that uses the manifest file from the managed DLL and then that would flow normally as threads are created throughout the application.

AaronRobinsonMSFT:
@ericstj Yep. I found the activation context API after we chatted. I am currently looking at testing out this scenario in the [CoreRun project](https://github.com/dotnet/coreclr/tree/master/src/coreclr/hosts/corerun) to determine feasibility.

ericstj:
Cool.  That's a good testing ground as a fix for the host. /cc @steveharter 

I bet you could even call that from individual test cases if you were interested in using it for interop tests of CoreCLR.  I could imagine something like `ActivationHelper.RunWithManifest(() => myTest(), myManifestFile);`

luqunl:
> I bet you could even call that from individual test cases if you were interested in using it for interop tests of CoreCLR.

AFAIK CoreCLR testcase also use CoreRun,  if Aaron change Corerun to support application manfiest, then these testcase should just work, why need `ActivationHelper.RunWithManife`?

ericstj:
It was just an idea if you wanted each test case within an application to define its own manifest file, in case you didn't want to spin up a new process every time you wanted to test with different COM registration.

AaronRobinsonMSFT:
[...]

PR #4476, opened 2018-08-15
Add hosting for COM activation design proposal

AaronRobinsonMSFT:
Hosting proposal for COM activation scenarios on Windows

cc @jkotas @vitek-karas @sbomer @luqunl @jeffschwMSFT


AaronRobinsonMSFT:
cc @richlander

AaronRobinsonMSFT:
Related issues:
#4314
dotnet/coreclr#5506
dotnet/coreclr#18490

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,92 @@
+# COM Activation for .NET Core on Windows
+
+## Purpose
+
+In order to more fully support the vast number of existing .NET Framework in their transition to .NET Core, support of the COM Activation scenario in .NET Core is required. Without this support it is not possible for many .NET Framework consumers to even consider transitioning to .NET Core. The intent of this document is to describe aspects of the activation of a .NET class written for .NET Core within a COM environment. This support includes but is not limited to activation via the [`CoCreateInstance()`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cocreateinstance)API in C/C++, from a [Windows Script Host](https://docs.microsoft.com/en-us/windows/desktop/com/using-com-objects-in-windows-script-host), or as a [Office VSTO Add-In](https://docs.microsoft.com/en-us/visualstudio/vsto/create-vsto-add-ins-for-office-by-using-visual-studio).
+
```
jkotas: `Office VSTO Add-In` comes with number of requirements. We have said earlier that we are not going to worry about it for .NET Core 3.0.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,92 @@
...
+### Requirements
+
+* Discover all installed versions of .NET Core.
+* Load the appropriate version of .NET Core for the class.
+* Instantiate an instance of the class and return a pointer to an [`IUnknown`](https://docs.microsoft.com/en-us/windows/desktop/api/unknwn/nn-unknwn-iunknown) that represents the newly constructed class.
+* Ensure that if an existing CLR is loaded in process, the exising one is used.
```
jkotas: Typo: exising

jkotas: > the existing one is used

What if the COM component needs a newer version than what's loaded already?

jkotas:
@richlander How does this align with customer research that we have been doing?

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,92 @@
...
+
+When `DllGetClassObject()` is called in an activation scenario, the following will occur:
+
+1) Determine additional registration information needed for activation.
+    * **The location and mechanics of accessing this registration information is still TBD**
+    * `Assembly` (**required**) - the [Fully-Qualified Name](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names) of the assembly.
+    * `Class` (**required**) - the full type name (e.g. `MyCompany.MyProduct.MyClass`).
+    * `RuntimeVersion` (optional) - if not supplied the latest CLR found will be used.
+    * `Codebase` (optional) - an absolute path to the assembly to load. If not supplied, the default [`AssemblyLoadContext`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.1) will be used to attempt to find the assembly. If provided, a [`runtimeconfig.json`](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md) file adjacent to the assembly can be used to describe runtime configuration details.
```
jkotas: Which path does the Assembly come from when Codebase is not specified? I do not understand how Codebase can be optional.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,92 @@
...
+1) Using the existing `hostfxr` library, attempt to discover the desired CLR.
+    * A current running CLR will be preferred, otherwise a new CLR instance will be created.
+1) A request to the CLR via [`Activator.CreateInstance()`](https://docs.microsoft.com/en-us/dotnet/api/system.activator.createinstance?view=netcore-2.1) will be used to instantiate an instance of the type in the assembly.
```
jkotas: What if there are multiple COM components in the same process and each come with different dependencies (e.g. each comes with different version of Json.NET)? We may need to load each COM component into its own AssemblyLoadContext to handle this.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Confusion on my part. Will remove.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: That should be: "Ensure that if an existing CLR instance of the correct version is already present in process, that existing one is used."

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: My expectation was to allow the assembly to be discovered as applocal. On deeper reflection though, this would potentially make a registered instance different between applications - which could be confusing.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: >  We may need to load each COM component into its own AssemblyLoadContext to handle this.

That was my intent - I will call this out explicitly.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Requirements
+
+* Discover all installed versions of .NET Core.
+* Load the appropriate version of .NET Core for the class.
+* Instantiate an instance of the class and return a pointer to an [`IUnknown`](https://docs.microsoft.com/en-us/windows/desktop/api/unknwn/nn-unknwn-iunknown) that represents the newly constructed class.
+* Ensure that if an existing CLR instance of the correct version is already present in process, that existing one is used.
```
vitek-karas: Do we know if there's an existing mechanism to query for loaded CLR instances in the process (and more importantly have a way to talk to them in a safe way)? I'm mostly just curious as this feels like a rather tricky bit.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,105 @@
...
+        {
+            ...
+            #if WINDOWS
+            public static ObjectHandle CreateInstanceForCom(string assemblyName, string typeName);
+            #endif
+            ...
+        }
+        ```
+        Note this API would not be exposed outside of `System.Private.CoreLib`.
+    * The loading of the assembly will take place in a new [`AssemblyLoadContext`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.1) for dependency isolation.
```
vitek-karas: This will require much more detail eventually. Things which feel like being problematic:
- What type of resolution logic will the new ALC use? (The potentially upcoming .deps.json based ALC, or something else).
- Do we need / will there be a way to customize the ALC used here.
- How much do we share - how much we enforce isolation in ALCs versus sharing through default.
- What happens if the same assembly is requested twice - do we load it again into a new ALC - or do we maintain a map of COM-loaded assemblies.
- Thinking about scenarios like: Creating several instances of some COM object, with the expectation that they have a way to talk to each other.

vitek-karas:
I assume we don't even want to think about out-of-proc COM, right? Would be nice to explicitly mention in the doc.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: That is a very tricky bit and something I am looking for feedback on. I think the .NET Framework's hosting API handles that logic, but I do not yet know if this support is within the .NET Core code base. @jkotas or @davidwrighton thoughts?

jkotas: Yes, .NET Framework hosting APIs have logic for this. There is nothing like that in .NET Core.

Loading multiple runtimes into a process was possible in theory in .NET Framework. In practice,  most people tried to stay away from it. 

I think we may want to limit .NET Core 3.0 support to having one runtime in a process. I do not think it is worth  opening the multiple runtimes rat-hole just for COM interop.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: This is a good list of concerns. Here are my initial thoughts:

>What type of resolution logic will the new ALC use? (The potentially upcoming .deps.json based ALC, or something else).

I am inclined to use whatever would be used if the server was used as a console application, so yes?

>Do we need / will there be a way to customize the ALC used here.

I think that is the `.deps.json` or `.runtimeconfig.json`. I think extending the ALC beyond the `*.json`s for COM activation is a tricky situation and something we should avoid.

>How much do we share - how much we enforce isolation in ALCs versus sharing through default.

I would prefer to isolate as much as possible. This opinion is based on trying to reduce situations where users can get themselves into intractable problems. However that opinion may be tempered based on what @richlander has to say.

>What happens if the same assembly is requested twice - do we load it again into a new ALC - or do we maintain a map of COM-loaded assemblies.

I am inclined to match the native server behavior and once the assembly is loaded that context is used for all further class activation from that assembly.

>Thinking about scenarios like: Creating several instances of some COM object, with the expectation that they have a way to talk to each other.

Another great question for @richlander. This is a real scenario that needs to be investigated.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,105 @@
+# COM Activation for .NET Core on Windows
+
+## Purpose
+
+In order to more fully support the vast number of existing .NET Framework in their transition to .NET Core, support of the COM Activation scenario in .NET Core is required. Without this support it is not possible for many .NET Framework consumers to even consider transitioning to .NET Core. The intent of this document is to describe aspects of the activation of a .NET class written for .NET Core within a COM environment. This support includes but is not limited to activation via the [`CoCreateInstance()`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cocreateinstance)API in C/C++ or from a [Windows Script Host](https://docs.microsoft.com/en-us/windows/desktop/com/using-com-objects-in-windows-script-host).
```
adityamandaleeka: > vast number of existing .NET Framework in their transition

users?
>

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Sounds good. I will update the requirements.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Requirements
+
+* Discover all installed versions of .NET Core.
+* Load the appropriate version of .NET Core for the class.
+* Instantiate an instance of the class and return a pointer to an [`IUnknown`](https://docs.microsoft.com/en-us/windows/desktop/api/unknwn/nn-unknwn-iunknown) that represents the newly constructed class.
+* Ensure that if an existing CLR instance of the correct version is already present in process, that existing one is used.
+* Support the discrimation of concurrently loaded CLR versions.
```
adityamandaleeka: Sounds like your next version will remove this line, but typo: "discrimation"

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Actually I am bit confused here. If we aren't going to allow multiple instances, then we have the unfortunate situation where the first COM activation sets the precedent for all further activations? I think this could be relaxed to, only 1 version of each runtime will be created from the loaded version of `hostfxr`?

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,107 @@
...
+            public static ObjectHandle CreateInstanceForCom(string assemblyName, string typeName);
+            #endif
+            ...
+        }
+        ```
+        Note this API would not be exposed outside of `System.Private.CoreLib`.
+    * The loading of the assembly will take place in a new [`AssemblyLoadContext`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.1) for dependency isolation.
+1) The class instance will be returned to the caller of `DllGetClassObject()`.
+
+In order to properly support `DllCanUnloadNow()`, .NET Core will be required to expose an API that permits the querying of how many COM activated classes are currently outstanding.
```
luqunl: Do we support unload managed assembly in dotnetcore?

AaronRobinsonMSFT: Good question. I think if we are using the ALC that is possible, but I don't know for sure.

jkotas: We might, but this very hard to get right. .NET Framework just returns false from `DllCanUnloadNow`. I think we should do the same.

AaronRobinsonMSFT: Okay, if there is no appetite for this support, we will document it as always returning false until needed.

Documentation/design-docs/COM-activation.md
sbomer: Regfree COM could have a registered instance different between apps right? Would applocal assembly discovery make sense in that case?

Based on my (limited) understanding of COM, Codebase seems necessary only if the registry info is stored in a location logically shared between multiple applications.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,107 @@
...
+* Check if the activation host has already created a CLR instance of the correct version or create an appropriate instance.
+* Support the discrimination of concurrently loaded CLR versions.
+
+### Environment Matrix
+
+The following list represents an exhaustive activation matrix.
+
+|Server | Client | Current Support |
+| --- | --- | :---: |
+| COM* | .NET Core | Yes |
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: Nope that is the next row. At present, .NET Core can activate a COM server (e.g. native DLL) via the registry _or_ when using `CoreRun` RegFree COM is also available. The .NET Server scenario (e.g. managed DLL) cannot be activated by a COM client (e.g. native EXE) at the moment because there is no way to instantiate a managed class without the CLR being loaded/started.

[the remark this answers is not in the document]
sbomer: Oh, I had read it to mean the scenario you're designing. Thanks!

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+            ...
+            #if WINDOWS
+            public static ObjectHandle CreateInstanceForCom(string assemblyName, string typeName);
+            #endif
+            ...
+        }
+        ```
+        Note this API would not be exposed outside of `System.Private.CoreLib`.
+    * The loading of the assembly will take place in a new [`AssemblyLoadContext`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.1) for dependency isolation.
+1) The class instance will be returned to the caller of `DllGetClassObject()`.
```
luqunl: You should return addrefed IClassFactory instance. This class instance is confusing.

luqunl: what you return is instance of https://github.com/dotnet/coreclr/blob/a9b25d4aa22a1f4ad5f323f6c826e318f5a720fe/src/vm/classfactory.cpp#L30

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+    * `Assembly` (**required**) - the [Fully-Qualified Name](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names) of the assembly.
+    * `Class` (**required**) - the full type name (e.g. `MyCompany.MyProduct.MyClass`).
+    * `Codebase` (**required**) - an absolute path to the assembly to load. If a [`runtimeconfig.json`](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md) file exists adjacent to the assembly that file will be preferred to describe CLR configuration details.
+    * `RuntimeVersion` (optional) - CLR version requirement.
+1) Using the existing `hostfxr` library, attempt to discover the desired CLR.
+    * If a CLR is present, the requested `RuntimeVersion` will be validated against that CLR. If version satisfiability fails, activation will fail.
+    * If a CLR is **not** present, an attempt will be made to create a satisfying CLR instance. Failure to create an instance will result in activation failure.
+1) A request to the CLR will be made via a new method for class activation within a COM environment.
```
luqunl: why do we need this? if native call DllClassObject (), we will return IClassFactory2/IClassFactory, then later it will call https://github.com/dotnet/coreclr/blob/a9b25d4aa22a1f4ad5f323f6c826e318f5a720fe/src/vm/classfactory.cpp#L142 to create object.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+COM activation in this document is currently limited to in-proc scenarios. Scenarios involving out-of-proc COM activation are deferred.
+
+### Requirements
+
+* Discover all installed versions of .NET Core.
+* Load the appropriate version of .NET Core for the class if a .NET Core instance is not running, or validate the currently existing .NET Core instance can satisfy the class requirement.
+* Instantiate an instance of the class and return a pointer to an [`IUnknown`](https://docs.microsoft.com/en-us/windows/desktop/api/unknwn/nn-unknwn-iunknown) that represents the newly constructed class.
```
luqunl: For DllGetClassObject, We should return an instance of EEClassFactory which implements IClassFactory/IClassFactory2.
We don't return newly constructed class..

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Purpose
+
+In order to more fully support the vast number of existing .NET Framework users in their transition to .NET Core, support of the COM activation scenario in .NET Core is required. Without this support it is not possible for many .NET Framework consumers to even consider transitioning to .NET Core. The intent of this document is to describe aspects of COM activation for a .NET class written for .NET Core. This support includes but is not limited to activation scenarios such as the [`CoCreateInstance()`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-cocreateinstance)API in C/C++ or from within a [Windows Script Host](https://docs.microsoft.com/en-us/windows/desktop/com/using-com-objects-in-windows-script-host) instance.
+
+COM activation in this document is currently limited to in-proc scenarios. Scenarios involving out-of-proc COM activation are deferred.
+
+### Requirements
+
+* Discover all installed versions of .NET Core.
```
luqunl: This belongs to vitek-karas app hosting work.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Environment Matrix
+
+The following list represents an exhaustive activation matrix.
+
+| Server | Client | Current Support |
+| --- | --- | :---: |
+| COM* | .NET Core | Yes |
+| .NET Core | COM* | No |
+| .NET Core | .NET Core | No |
```
luqunl: how does these managed(.NET Core/Framework) to manage(.NET Core/Framework) need activation support?

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+| .NET Framework | .NET Core | No |
+| .NET Core | .NET Framework | No |
+
+\* 'COM' is used to indicate any COM environment (e.g. C/C++) other than .NET.
+
+## Design
+
+One of the basic issues with the activation of a .NET class within a COM environment is the loading or discovery of an appropriate CLR instance. The .NET Framework addressed this issue through a shim library (described below). The .NET Core scenario has different requirements and limitations on system impact and as such an identical solution may not be optimal or tenable.
```
luqunl: > The .NET Core scenario has different requirements and limitations on system impact

Could you list some of these difference in this design doc?  so customer can understand better why we design this way.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+        ```
+        Note this API would not be exposed outside of `System.Private.CoreLib`.
+    * The loading of the assembly will take place in a new [`AssemblyLoadContext`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.1) for dependency isolation.
+1) The class instance will be returned to the caller of `DllGetClassObject()`.
+
+The `DllCanUnloadNow()` function will always return `S_FALSE` indicating the shim is never able to be unloaded. This matches .NET Framework semantics and can be adjusted in the future if needed.
+
+#### Class Registration
+
+Registration will depend on the application's [deployment scenario](https://docs.microsoft.com/en-us/dotnet/core/deploying/).
```
luqunl: There is another "class"--- Framework class's registration, who should register these class and where to register these class?

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Yep. Will clarify.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+* [Framework-dependent deployments (FDD)](https://docs.microsoft.com/en-us/dotnet/core/deploying/#framework-dependent-deployments-fdd) - This is most similar to the .NET Framework scenario and registration will be done using the shared framework. 
+* [Self-contained deployments (SCD)](https://docs.microsoft.com/en-us/dotnet/core/deploying/#self-contained-deployments-scd) - Registration will be done using the contained framework.
+
+##### Registry
```
luqunl: Another interesting problem: SxS. There are couples ways that COM activation will accidently introduce SxS
Scenario: .NET Core SxS .NET Framework
CoCreateInstance(CLSID_A, ...)   // CLSID_A exists in both  .Net FrameWork but not exists in .net core
CoCreateInstance(CLSID_B, ...)  //   CLSID_B exists in .Net FrameWork but not exists in .net core
As a result, .NET Framework runtime(CLR.dll) and .NET Core runtime(coreclr.dll) will both launch at same process.

Scenario: .NET Core SxS .NET Core
CoCreateInstance(CLSID_C, ...)   // CLSID_C exists in reg-free 
CoCreateInstance(CLSID_D, ...)  //   CLSID_B exists in register for .NET Core

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: It hasn't been defined at the moment. It is same issue with .NET Core and .NET Core - what is the mechanism for when a .NET Core application activates a class that was defined by .NET Framework?

The point of that matrix is to be complete, we may not support certain scenarios, but I want to call them out.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,132 @@
...
+
+>[`HRESULT DllCanUnloadNow();`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-dllcanunloadnow)
+
+When `DllGetClassObject()` is called in an activation scenario, the following will occur:
+
+1) Determine additional registration information needed for activation.
+    * **The location and mechanics of accessing this registration information is still TBD**
+    * `Assembly` (**required**) - the [Fully-Qualified Name](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names) of the assembly.
+    * `Class` (**required**) - the full type name (e.g. `MyCompany.MyProduct.MyClass`).
+    * `Codebase` (**required**) - an absolute path to the assembly to load. If a [`runtimeconfig.json`](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md) file exists adjacent to the assembly that file will be preferred to describe CLR configuration details.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: I am more and more inclined to think that we should require the `codebase` value. My principle argument is that if you don't have codebase and want it the assembly to be local, then we should push users toward RegFree.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: RegFree context is based on application manifest so I am unclear how we could mix the RegFree vs registry. If the application has a entry in the manifest that is what should be used. I guess a fallback would be to hit the registry, but I am unclear if that should be supported.

AaronRobinsonMSFT: >Scenario: .NET Core SxS .NET Framework

That is exactly why that is present in the activation matrix above.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: It is still a requirement of this feature. These are requirements that this host needs to satisfy.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: I am not as confident we should be using the `EEClassFactory` approach. I am much more inclined to write most of this in managed code.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: I would prefer to avoid that code path and push more of this into managed code. Almost all the needed activation logic above can be written in managed. This would allow us to remove these legacy mechanisms. Another reason would be the longer term desire of potentially having some kind of basic activation on non-Windows. If the new activation logic is written in managed it will just work.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Will mention the issue with conflicting `GUID`s between .NET Framework and .NET Core.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,139 @@
...
+    * `RuntimeVersion` (optional) - CLR version requirement.
+1) Using the existing `hostfxr` library, attempt to discover the desired CLR.
```
vitek-karas: This is actually more than just CLR. We need to have the right frameworks as well. In fact it goes hand in hand. Currently without multiple frameworks the runtime config may specify for example ASP.NET framework, which than depends on the NETCore.App which is the one actually identifying the runtime to use.
In the future with multiple framework support this gets even more complicated.
Currently I'm not aware of a plan to support loading multiple versions of higher-level frameworks (like ASP.NET) side-by-side into the same process (even if they depend on the same underlying NETCore.App). So the resolution logic here will have to consider all of that.
That said, it already does so, so the implementation is probably almost ready, but I think it's worth noting it here, that this is not just about the runtime version, but also about all the frameworks.

AaronRobinsonMSFT: @vitek-karas The concept of different frameworks is something I am still working on fully incorporating in my thought process - thanks for keeping me honest.

Since the framework is also a consideration, it seems that the `deps.json` file is going to be a requirement, where as the `runtimeconfig.json` could be optional? Would that address your concerns here?

vitek-karas: I'm pretty sure frameworks are specified in `runtimeconfig.json`. `deps.json` can only customize the frameworks a little (override assembly at a time), but usually it just does nothing to them. So this would make `runtimeconfig.json` basically mandatory. It might be possible to not have `deps.json`... but I'm not sure.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,138 @@
...
+    <clrClass
+        clsid="{3C58BBC9-3966-4B58-8EE2-398CBBC9FDC4}"
+        name="NetServer.Server"
+        threadingModel="Both"
+        runtimeVersion="v4.0.30319">
+    </clrClass>
+</assembly>
+```
+
+Due to the above issues with traditional RegFree manifests and .NET classes, an alternative system must be employed to enable a low-impact style of class registration for .NET Core.
```
wjk: Wouldn't a bog-standard `<comClass />` SxS manifest entry work here? It wouldn't be able to specify the CLR class information, but it would allow the specification of the server DLL. Furthermore, since COM is by definition Windows only, would it be possible to embed the CLSID-to-CLR-class information as a native resource in the DLL that is listed in the manifest? This is very easy to do using .NET; I use the [ResourceLib](https://github.com/resourcelib/resourcelib) library to do this usually, but you could P/Invoke `UpdateResource()` and friends manually from kernel32 if you really wanted.

I don't think that doing this would affect SCD/FDD any, as the DLL that is listed in the manifest must be present in the same directory hierarchy as the manifest (because Windows is too stupid to look anywhere else). The DLL that is listed in the manifest can then resolve the managed assemblies any way it sees fit.

AaronRobinsonMSFT: > Wouldn't a bog-standard <comClass /> SxS manifest entry work here?

Potentially, but it is a bit more complicated than that. The user shouldn't be expected to author their own server, so it would need to be provided - which is the shim. However, the shim's manifest would then need to declare the `GUID` of the class it intends to provide so SxS would discover the server. Basically this approach is what I am doing and the broad strokes of your suggestion are valid, but like everything in COM the devil is in the details. I will updating this doc in the next few days.

> because Windows is too stupid to look anywhere else
That is actually because of a very complex security scenario and was intentionally designed that way.

Documentation/design-docs/COM-activation.md
```diff
@@ -61,10 +61,9 @@ When `DllGetClassObject()` is called in an activation scenario, the following wi
...
 1) Using the existing `hostfxr` library, attempt to discover the desired CLR and target [framework](https://docs.microsoft.com/en-us/dotnet/core/packages#frameworks).
-    * If a CLR is present, the requested `RuntimeVersion` will be validated against that CLR. If version satisfiability fails, activation will fail.
+    * If a CLR is present, the requested CLR version will be validated against that CLR. If version satisfiability fails, activation will fail.
     * If a CLR is **not** present, an attempt will be made to create a satisfying CLR instance. Failure to create an instance will result in activation failure.
 1) A request to the CLR will be made via a new method for class activation within a COM environment.
     * The ability to load the assembly and create an `IClassFactory` instance will require exposing a new function that can be called from `hostfxr`.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: > I feel this should state that the Coreclr hosting interface has a new method for loading an assembly and creating an instance.

The `hostfxr` implementation wraps entry points into the coreclr hosting interface, so both statements are true, but the real work will probably be done in coreclr. I will clarify this and based on Sven's recent work (dotnet/coreclr#19486) with the start up hooks, this bullet becomes less interesting if needed at all.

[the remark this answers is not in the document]
AaronRobinsonMSFT: The example there is purely an entry point that is easy to call from the native side. The implementation will leverage many of the existing functions hanging off of the `Marshal` class.

Documentation/design-docs/COM-activation.md
```diff
@@ -61,10 +61,9 @@ When `DllGetClassObject()` is called in an activation scenario, the following wi
...
+    * `Codebase` (**required**) - an absolute path to the assembly to load. If a [`runtimeconfig.json`](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md) file exists adjacent to the assembly, that file will be used to describe CLR configuration details. Refer to documentation for the `runtimeconfig.json` format when the file is [optional](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md#what-produces-the-files-and-where-are-they).
 1) Using the existing `hostfxr` library, attempt to discover the desired CLR and target [framework](https://docs.microsoft.com/en-us/dotnet/core/packages#frameworks).
-    * If a CLR is present, the requested `RuntimeVersion` will be validated against that CLR. If version satisfiability fails, activation will fail.
+    * If a CLR is present, the requested CLR version will be validated against that CLR. If version satisfiability fails, activation will fail.
     * If a CLR is **not** present, an attempt will be made to create a satisfying CLR instance. Failure to create an instance will result in activation failure.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: Will clarify that point.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: @wjk This is an example of what it sounds like you are suggesting: https://github.com/AaronRobinsonMSFT/COMInterop/tree/coreclr_example. It is what I am currently investigating as a possible solution. Let me know what you think. Thanks.

wjk: Looks perfect! All we would really need to do then is to hook this code up to the `*.deps.json` mechanism (because I don't think it can parse that as-is). Furthermore, with a bit of expansion, this approach could well implement the out-of-proc COM hosting I was asking about in https://github.com/dotnet/coreclr/issues/18490. Cool beans!

AaronRobinsonMSFT: >All we would really need to do then is to hook this code up to the *.deps.json mechanism (because I don't think it can parse that as-is).

I am unclear at the moment if we would use `runtimeconfig.json` or `deps.json`. There has been much talk about that. @vitek-karas should be able to provide some better insight into those arguments though. The current implementation is basically how I am planning on adding support in the coreclr repo for adding .NET activation testing, as you can imagine actual product support is much more complicated. 

>... this approach could well implement the out-of-proc COM hosting ...

Out of proc COM hosting is definitely up in the air with respect to support in .NET Core. There are some real complex scenarios regarding proxies and I don't know if/how they would fit into the overall .NET Core platform itself. Definitely doable, but would require a significant investment to add support for out-of-proc even if we have in-proc.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,213 @@
...
+
+>[`HRESULT DllCanUnloadNow();`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-dllcanunloadnow)
+
+When `DllGetClassObject()` is called in an activation scenario, the following will occur:
+
+1) Determine additional registration information needed for activation.
+    * **The location and mechanics of accessing this registration information is still TBD**
+    * `Assembly` (**required**) - the [Fully-Qualified Name](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names) of the assembly.
+    * `Class` (**required**) - the full type name (e.g. `MyCompany.MyProduct.MyClass`).
+    * `Codebase` (**required**) - an absolute path to the assembly to load. If a [`runtimeconfig.json`](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md) file exists adjacent to the assembly, that file will be used to describe CLR configuration details. Refer to documentation for the `runtimeconfig.json` format when the file is [optional](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md#what-produces-the-files-and-where-are-they).
```
vitek-karas: > Refer to documentation for the `runtimeconfig.json` format when the file is [optional]
I can't really parse this. I would expect something like
> Refer to documentation for the `runtimeconfig.json` format and what happens if the file is missing.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,213 @@
...
+            public Guid InterfaceId;
+            public int AssemblyCount;
+            public IntPtr AssemblyList;
+            public IntPtr ClassFactoryDest;
+        }
+
+        public static class ComActivator
+        {
+            ...
+            public static int GetClassFactoryForTypeInternal(ref ComActivationContextInternal context);
```
vitek-karas: Above you define the class with name `ComActivationContext` but here it's called `ComActivationContextInternal`

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: I will clarify this.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Oops.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,213 @@
...
+            type="win32"
+            name="COMClientPrimitives"
+            version="1.0.0.0"/>
+
+        <dependency>
+            <dependentAssembly>
+            <!-- RegFree COM - CoreCLR Shim -->
+            <assemblyIdentity
+                type="win32"
+                name="CoreShim.X"
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: >the `CoreShim` would be something which is part of the client app, right?

For the RegFree approach, yes. For the registry scenario, I would suggest the shared framework location. We still have to define the deployment experience for this. Now that I have a PR out for how to test, we can start to delve into the mechanics of the dev and deployment experience.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,213 @@
...
+                version="1.0.0.0"/>
+            </dependentAssembly>
+        </dependency>
+
+        <dependency>
+            <dependentAssembly>
+            <!-- RegFree COM - Managed assembly -->
+            <assemblyIdentity
+                type="win32"
+                name="NETServer"
```
Petermarcu: I'd be careful here on naming, We should be clear this is a COM server, not to be confused with the ASP.NET Web Server.

AaronRobinsonMSFT: That is purely an example for illustrative purposes.

AaronRobinsonMSFT:
@jkotas and @morganbr Would really appreciate you confirming our recent conversation is properly reflected in this updated proposal. Thanks.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,198 @@
...
+
+1) Determine additional registration information needed for activation.
+    * The shim will check for an embedded manifest. If the shim does not contain an embedded manifest, the shim will check if a file with the `<host_library_name>.clsidmap` naming format exists adjacent to it.
+    * The manifest will contain a mapping from [`CLSID`](https://docs.microsoft.com/en-us/windows/desktop/com/com-class-objects-and-clsids) to managed assembly name and the [Fully-Qualified Name](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names) for the type. The exact format of this manifest is an implementation detail, but will be identical whether it is embedded or a loose file.
+    * The manifest will define an exhaustive list of .NET classes it is permitted to provide.
+    * If a [`runtimeconfig.json`](https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md) file exists adjacent to the managed assembly, that file will be used to describe CLR configuration details. The documentation for the `runtimeconfig.json` format defines under what circumstances this file is optional.
```
jkotas: I think `runtimeconfig` should be next to the shim, not per assembly.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,198 @@
...
+>[`HRESULT DllCanUnloadNow();`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-dllcanunloadnow)
+
+>[`HRESULT DllRegisterServer();`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682162(v=vs.85).aspx)
+
+>[`HRESULT DllUnregisterServer();`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms691457(v=vs.85).aspx)
+
+When `DllGetClassObject()` is called in a COM activation scenario, the following will occur:
+
+1) Determine additional registration information needed for activation.
+    * The shim will check for an embedded manifest. If the shim does not contain an embedded manifest, the shim will check if a file with the `<host_library_name>.clsidmap` naming format exists adjacent to it.
```
jkotas: Do we expect the shim going to be renamed to contain your project name by the tooling? 

I think it should get renamed to make it clear it was customized for your project.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: Agreed. That does make things a bit cleaner, but could potentially complicate the packaging logic.

Documentation/design-docs/COM-activation.md
AaronRobinsonMSFT: > Do we expect the shim going to be renamed to contain your project name by the tooling?

In the default case, yes. There should be an option to not rename, but that is definitely not the default. In this particular document I am focusing on CLR mechanics/contracts, not the specific UX around the dev experience. The renaming of the host and how it should be used in a secure manner are being left to the official CLI documentation and/or examples.

Documentation/design-docs/COM-activation.md
```diff
@@ -0,0 +1,199 @@
...
+>[`HRESULT DllCanUnloadNow();`](https://docs.microsoft.com/en-us/windows/desktop/api/combaseapi/nf-combaseapi-dllcanunloadnow)
+
+>[`HRESULT DllRegisterServer();`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682162(v=vs.85).aspx)
+
+>[`HRESULT DllUnregisterServer();`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms691457(v=vs.85).aspx)
+
+When `DllGetClassObject()` is called in a COM activation scenario, the following will occur:
+
+1) Determine additional registration information needed for activation.
+    * The shim will check for an embedded manifest. If the shim does not contain an embedded manifest, the shim will check if a file with the `<shim_name>.clsidmap` naming format exists adjacent to it. Build tooling will be expected to handle shim customization, including renaming the shim to be based on the managed assembly's name (e.g. `NetComServer.dll` could have a custom shim called `NetComServer.shim.dll`).
```
morganbr: Why also allow a sidecar? It wouldn't be possible to ensure a sidecar was signed.

jkotas: > the dev cycle is much more complicated

Which part of the dev cycle is much more complicated? The manifest handling will be hidden behind msbuild target. I assume that this msbuild target will embed it by default.

AaronRobinsonMSFT: The following will be more complicated:

* Incremental build - MSBuild does handle a lot of this but there are many edge cases around auto-generated files that may crop up depending on how we do project references or where in the pipeline we create the manifest.

* Single shim multiple managed assemblies - When multiple managed binaries that proffer COM classes we will need to coordinate with the shim customizing target to ensure all of them can be hosted by the same shim. We are going to want the option for a single shim for multiple managed assemblies for size sensitive systems.

* Diagnostics for the pipeline - Instead of hiding the entire system if we make the pipeline a series of well-known and defined steps/targets we can optimized it by making a single MSBuild target, but we can also help users to help themselves and point at the loose file option when there is an issue. This also enables a reasonable testing of the pipeline.

I do not want to make magic in COM activation for .NET Core. Requiring embedding or hiding the manifest turns into magic - which is fine when it works, but we will have bugs and allowing an escape hatch for customer should be our primary goal - unblocking user scenarios.

> I assume that this msbuild target will embed it by default.

I agree that is a reasonable default.

jkotas: If we support embedding, we need to build and test all what you have listed and it needs to better works well if it is the default. Having second option is just more work, more tests and more documentation on top of it. I think we should be introducing the second options here only if we get feedback that the one embedding option does not work well.

We can drill into details on the incremental build, multiple assemblies and diagnostic concerns once we get to the tooling design and implementation.

danmosemsft:
Kudos for a super clear spec 👏

Issue #4366, opened 2018-07-13
Port fix for custom host to 2.2


steveharter:
Port commit from master https://github.com/dotnet/core-setup/commit/84fa44f892b42ad9235e4ecb756975d8c8f46f1f to the 2.2 branch.

Per issue https://github.com/dotnet/core-setup/issues/4315

@pakrym can you please verify the issue is fixed in master once a new build is available?

pakrym:
The issue doesn't reproduce with `3.0.0-preview1-26810-04` hostpolicy.

steveharter:
@pakrym should I target 2.1.x instead of 2.2? The 2.1.x fix will make it into 2.2 automatically.

pakrym:
The feature we use hosted CLR in only ships in 2.2 so 2.2 is enough for us.

pakrym:
Recently we reenabled tests that we causing AV in hostpolicy and started running them with AppVerifier on. We see ocasional AppVerifier failures in the same area.

Recent dump: `\\pkrymets\exchange\hostpolicy\iisexpress.exe.5292.dmp.cab`

muratg:
@jeffschwMSFT will this be assigned to someone on your side now?

jeffschwMSFT:
@muratg can we open a new issue for this appverifier related issue?  Please assign to @sdmaclea

muratg:
@pakrym could you please do the needful?

pakrym:
After running tests a bit longer we found that crashes still occur even if appverifier is not running.

So it's the same issue as being fixed here it just happens a bit more infrequently.

pakrym:
Opened https://github.com/dotnet/core-setup/issues/4646

PR #4478, opened 2018-08-15
Port fix for custom host

steveharter:
[original issue](https://github.com/dotnet/core-setup/issues/4315)

fixes	https://github.com/dotnet/core-setup/issues/4366 (2.2 port issue)

This is a direct port of https://github.com/dotnet/core-setup/commit/84fa44f892b42ad9235e4ecb756975d8c8f46f1f except for some additional test infrastructure that needed to be pulled in to merge the new test.

#### Description
The issue only applies when the CoreCLR is being hosted by ASP.NET (not the standard dotnet.exe or apphost.exe)

If managed code throws an unhandled exception before the host’s background “breadcrumb writer” thread is done writing breadcrumbs (for potential servicing scenarios) then an AV occurs in the breadcrumb code.

This issue does not need to be fixed in 2.1.x since ASP.NET did not use the custom host until 2.2

#### Customer Impact
Without the fix, the ASP.NET host cannot catch the AV exception and continue execution (or log the exception)

#### Regression?
No

#### Risk
Minimal.  The change occurs in native code in hostpolicy.dll that is executed after managed code has started, and the fix does not affect the user’s managed code.

### DIFF
--- a/src/corehost/cli/breadcrumbs.cpp
+++ b/src/corehost/cli/breadcrumbs.cpp
@@ -7,13 +7,22 @@
 #include "trace.h"
 #include "breadcrumbs.h"
 
-breadcrumb_writer_t::breadcrumb_writer_t(const std::unordered_set<pal::string_t>* files)
+breadcrumb_writer_t::breadcrumb_writer_t(bool enabled, const std::unordered_set<pal::string_t>* files)
     : m_status(false)
+    , m_enabled(enabled)
     , m_files(files)
 {
-    if (!pal::get_default_breadcrumb_store(&m_breadcrumb_store))
+    if (enabled && !pal::get_default_breadcrumb_store(&m_breadcrumb_store))
     {
         m_breadcrumb_store.clear();
+    }
+}
+
+breadcrumb_writer_t::~breadcrumb_writer_t()
+{
+    if (m_enabled)
+    {
+        end_write();
     }
 }
 
@@ -21,22 +30,25 @@ breadcrumb_writer_t::breadcrumb_writer_t(const std::unordered_set<pal::string_t>
 // thread to write breadcrumbs.
 void breadcrumb_writer_t::begin_write()
 {
-    trace::verbose(_X("--- Begin breadcrumb write"));
-    if (m_breadcrumb_store.empty())
+    if (m_enabled)
     {
-        trace::verbose(_X("Breadcrumb store was not obtained... skipping write."));
-        m_status = false;
-        return;
-    }
+        trace::verbose(_X("--- Begin breadcrumb write"));
+        if (m_breadcrumb_store.empty())
+        {
+            trace::verbose(_X("Breadcrumb store was not obtained... skipping write."));
+            m_status = false;
+            return;
+        }
 
-    trace::verbose(_X("Number of breadcrumb files to write is %d"), m_files->size());
-    if (m_files->empty())
-    {
-        m_status = true;
-        return;
+        trace::verbose(_X("Number of breadcrumb files to write is %d"), m_files->size());
+        if (m_files->empty())
+        {
+            m_status = true;
+            return;
+        }
+        m_thread = std::thread(write_worker_callback, this);
+        trace::verbose(_X("Breadcrumbs will be written using a background thread"));
     }
-    m_thread = std::thread(write_worker_callback, this);
-    trace::verbose(_X("Breadcrumbs will be written using a background thread"));
 }
 
 // Write the breadcrumbs. This method should be called
@@ -88,4 +100,3 @@ bool breadcrumb_writer_t::end_write()
     trace::verbose(_X("--- End breadcrumb write %d"), m_status);
     return m_status;
 }
-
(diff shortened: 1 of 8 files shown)


steveharter:
@jeffschwMSFT (or others with permission) I will be OOF until 8\27. If this is ready to go in before I return, please merge (marked servicing approved without any further comments to delay)

danmosemsft:
@jeffschwMSFT per above you can just merge this assuming it's reviewed etc.

jeffschwMSFT:
cc @vitek-karas @swaroop-sridhar @sbomer

sbomer:
@steveharter From the original issue:
> Note that the proposed ~breadcrumb_writer_t will not be called with the normal host (dotnet\apphost) in the app exception case because it requires the host (or a caller) being compiled with SEH exception support (/EHa) with an appropriate try\catch

> Assumption: we only care about the original issue here - the AV when using a custom host - and not guarantying that the breadcrumb thread finishes gracefully with the normal host.

Is this something we should fix for the normal host too? It looks like dotnet/apphost could run into the same issue if using a servicing index.

Otherwise LGTM.

vitek-karas:
Is there some doc or something somewhere which would describe what the bread crumbs are for?

steveharter:
@vitek-karas 

>> Is there some doc or something somewhere which would describe what the bread crumbs are for?

I don't believe so. The original core-setup code lived in the CLI repo. Here's a search to help see the history: https://github.com/dotnet/cli/search?q=breadcrumb&type=Issues

steveharter:
@sbomer 

>> Is this something we should fix for the normal host too? It looks like dotnet/apphost could run into the same issue if using a servicing index.

The AV issue does not occur in the normal host (dotnet.exe or apphost). However, the breadcrumbs will not be written either unless we add the /EHa + try/catch. Is this what you are suggesting? Since it only takes one successful launch (and a minimum amount of time to run before exiting) of any app in order to write the breadcrumb info, I'm not sure if the breadcrumb writing is worth "hardening".

FWIW the /EHa does make the size of the executable, hostpolicy.dll and hostfxr.dll slightly larger (around 40K total), so it doesn't come for free either.

sbomer:
@steveharter It looked to me (at first glance) like the breadcrumbs could be written by the normal host, along the `execute_host_command` code path. If that's not the case, then I'm not concerned about this.

If there's a way to turn on breadcrumbs for the normal host, then based on the description in the original issue it sounds like it could have the same issue.

steveharter:
@sbomer the breadcrumbs are written for *all* hosts - aspnet, dotnet.exe, apphost.exe. The code to write the breadcrumbs is in not in the host executable, it's in hostpolicy.dll which is present for each runtime (Microsoft.NETCore.App) and called by each host executable (host exe -> hostfxr -> hostpolicy).

The original issue with the AV only surfaced with aspnet.  However, the "fix" for the AV applies to all hosts since it the fix is in hostpolicy.dll.

The comment I added about not adding /EHa + try/catch is just asserting that I didn't "harden" or add any logic to the host (or hostpolicy) to write the breadcrumbs in the exception case.

sbomer:
@steveharter Here was my understanding of the issue:
- App throws an exception
- Host exits the run function before calling end_write()
- Breadcrumb writer is still running, but the writer local is gone, causing an AV

With the fix, the destructor guarantees that the run thread waits for writing to finish before destroying the writer. But this comment makes it sound like the destructor isn't guaranteed to run...

> Note that the proposed ~breadcrumb_writer_t will not be called with the normal host (dotnet\apphost) in the app exception case because it requires the host (or a caller) being compiled with SEH exception support (/EHa) with an appropriate try\catch

I think the destructor should run in the normal host too, otherwise we hit the same issue.

> The comment I added about not adding /EHa + try/catch is just asserting that I didn't "harden" or add any logic to the host (or hostpolicy) to write the breadcrumbs in the exception case.

Maybe I misunderstood the first comment, and the destructor will always run. :) In that case we're good.

Anyway, you don't need to block this PR on my questions - just trying to make sure I understand the fix.

steveharter:
@sbomer without /EHa + try/catch I don't believe the stack unwinds during the exception case, so no destructors will be called. The threads are immediately terminated and thus no AV.

sbomer:
If the destructor isn't called, what prevents the breadcrumb thread from trying to write after the writer local has gone out of scope, but before the breadcrumb thread is terminated?

steveharter:
For aspnet the destructor is called, but for the normal dotnet\apphost it is not. For dotnet\apphost, the breadcrumb thread is immediately killed when the unhandled managed exception is thrown (which causing an unhandled C++ SEH exception, killing the process without stack unwinding). 

Compiling with /EHa would allow the code to catch the SEH exception, and the try/catch would be needed to unwind the stacks and call the destructors.

This test explains it a bit: [Breadcrumb_thread_does_not_finish_when_app_has_unhandled_exception](https://github.com/dotnet/core-setup/pull/4478/files#diff-747c7d5469f172ba5df742bb0d8f5079R94)

Issue #4472, opened 2018-08-15
windows.winmd should not be included in netcoreapp distro


jkotas:
## Steps to reproduce

Download https://dotnetcli.blob.core.windows.net/dotnet/Runtime/master/dotnet-runtime-latest-win-x64.zip 

## Expected  behavior

Windows.winmd is not in package

## Actual behavior

Windows.winmd is in the package

jkotas:
cc @luqunl

luqunl:
I am working on

PR #4493, opened 2018-08-20
Exclude Windows.winmd into NetCoreApp distro

luqunl:
Fix https://github.com/dotnet/core-setup/issues/4472

One of my previous change(enable crossgen on S.R.WindowsRuntime) accidentally includes windows.winmd into netcoreapp distro.

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -144,8 +144,8 @@
   <!-- Fetches all the file items from the packages that we want to redist -->
   <Target Name="GetFilesFromPackages" DependsOnTargets="GetPackagePaths">
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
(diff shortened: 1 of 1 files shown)


weshaggard:
@luqunl have you verified that this fix works? I see see signing validation failures due to Windows.winmd in  todays build, see https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=1980335&view=logs.

luqunl:
@weshaggard , I had verified that after this change, the windows.winmd shouldn't exists in NetCoreApp package. Let me do some more investigation.

luqunl:
Windows.winmd doesn't exists in https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.Private.CoreFx.NETCoreApp/4.6.0-preview1-26827-01

luqunl:
Found out it do exists in https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App/3.0.0-preview1-26827-01

luqunl:
@ericstj , yes, you are right. I checked couple packages and found out one package still contains windows.winmd file.

PR #4498, opened 2018-08-27
Lowercase bin dir name to align with corefx and others

ViktorHofer:
Fixes failures in https://github.com/dotnet/core-setup/pull/4489

We are using a property in buildtools that point to the obj directory in the repo (https://github.com/dotnet/buildtools/blob/0fa455c550abb4f49f4830a7f2fa78609f373fd0/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.props#L40). CI fails on Unix because the path is case-sensitive which it isn't on Windows. Lower-casing the bin directory here in core-setup to align with corefx and buildtools.

### DIFF
--- a/RepoDirectories.props
+++ b/RepoDirectories.props
@@ -6,7 +6,7 @@
     <SourceDir>$(RepoRoot)src/</SourceDir>
 
     <!-- Output directories -->
-    <BinDir Condition="'$(BinDir)'==''">$(RepoRoot)Bin/</BinDir>
+    <BinDir Condition="'$(BinDir)'==''">$(RepoRoot)bin/</BinDir>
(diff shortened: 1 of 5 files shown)


ViktorHofer:
CI will not pass as the groovy script changes will only work after being commited.

ViktorHofer:
@dotnet-bot test ci

buildpipeline/Core-Setup-Signing-Validation.json
```diff
@@ -121,7 +121,7 @@
       },
       "inputs": {
         "SourceFolder": "$(PB_SourcesDirectory)",
-        "Contents": "*.log\nBin\\SigningValidation\\Logs\\**",
+        "Contents": "*.log\nbin\\SigningValidation\\Logs\\**",
```
safern: Is this also lower case? This is an official build definition, I wonder if this changed as well.

PR #4500, opened 2018-08-27
Remove Windows.winmd in ref

luqunl:
Also exclude windows.winmd in ref folder

Fix https://github.com/dotnet/core-eng/issues/4047

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -167,8 +167,8 @@
     </ItemGroup>
 
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' == ''">
-      <!-- Not RID-specific: include all reference files -->
-      <FilesToPackage Include="@(Reference)">
+      <!-- Not RID-specific: include all reference files except files in Microsoft.TargetingPack.Private.WinRT -->
+      <FilesToPackage Include="@(Reference)" Condition="'%(Reference.NuGetPackageId)' != 'Microsoft.TargetingPack.Private.WinRT'">
         <TargetPath>ref/$(PackageTargetFramework)</TargetPath>
       </FilesToPackage>
 


luqunl:
cc: @weshaggard

src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -167,8 +167,8 @@
     </ItemGroup>
 
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' == ''">
-      <!-- Not RID-specific: include all reference files -->
-      <FilesToPackage Include="@(Reference)">
+      <!-- Not RID-specific: include all reference files except files in Microsoft.TargetingPack.Private.WinRT -->
+      <FilesToPackage Include="@(Reference)" Condition="'%(Reference.NuGetPackageId)' != 'Microsoft.TargetingPack.Private.WinRT'">
```
weshaggard: @ericstj Would this be better done by adding this package to PackageToExclude list and letting the filtering handle it https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/depProj.targets#L104? 

Or are customizing the build too much to allow the normal filtering to work?

ericstj: We could, but we still need the path to the Windows.winmd.  That depproj filtering was more for indirect dependencies where we couldn't control them.

If we wanted a simpler solution, I'd just put `ExcludeAssets="All"` on the `PackageReference` then rather than find the path to `windows.winmd` in the resolved assets we'd need to hardcode it like `$(PackagesDir)\Microsoft.TargetingPack.Private.WinRT\$(MicrosoftTargetingPackPrivateWinRTPackageVersion)\lib\netstandard1.0\windows.winmd`.

I'm ok with the current approach since we're addressing all the issues.

luqunl: If you guys were ok for this approach, I will check this in  to solve https://github.com/dotnet/core-eng/issues/4047

weshaggard: Yes go ahead and get this in. Although I'd like to think about potentially improving it in the future.

Issue #4512, opened 2018-08-30
[2.1.3]: Apps fail when app-local binary is lower than the version provided by a shared runtime


natemcmaster:
The original issue is described in #4376. This was been fixed already in the master branch, but this is causing problems for 2.1.x servicing. We should backport the fix (#4423) to 2.1.x.

cc @steveharter @eerhardt @joshfree

PR #4521, opened 2018-08-31
Always compare TPA versions (#4423)

steveharter:
Fixes https://github.com/dotnet/core-setup/issues/4512
Original issue: https://github.com/dotnet/core-setup/issues/4376

**Description**
Compare assemblyVersion\fileVersion for patch roll-forward cases to support better serviceability and to prevent run-time FileLoadException from occurring due to assembly-to-assembly versions being too low. Previously the runtime only compared versions for minor- and major roll-forward scenarios. The reason was because it allowed for the app to have complete control over any duplicated assemblies that exist in the app bin location, to support build-from-source and other explicit scenarios.

**Customer Impact**
If an application wishes to always load their "app local" copies of framework assemblies (for example, if they have a build-from-source scenario with new features or bug fixes) then they should prevent Minor\Major roll-forward and patch roll-forward from occurring by modifying their runtimeconfig.json to add applyPatches:false and rollForwardOnNoCandidateFx" : 0 like so:
```
{
  "runtimeOptions": {
    "tfm": "netcoreapp2.1",
    "applyPatches" : false,
    "rollForwardOnNoCandidateFx" : 0,
    "framework": {
      "version": "<yourversion>"
    }
  }
}
```

**Regression?**
No

**Risk**
Some, although should be rare. As noted in the description, this helps with serviceability when we release new versions of framework assemblies in a patch release. However, if the customer always expects the older "app local" version to be loaded then they will need to either:
- Manually apply the runtimeconfig.json changes above to prevent patch versions from being loaded
- Don't install the patch version
- Modify the application to not require a reference to the framework assembly
- Modify the application to not require an older version of the framework assembly


danmosemsft:
Waiting on email.

jamshedd:
Approved for 2.1.5. Test plan required to mitigate risk.

steveharter:
Test plan added at https://github.com/dotnet/core-setup/issues/4541. In general, there are two areas of concern:
1) Having proper manual\automated testing to ensure correct assembly is loaded and the bug is fixed
2) Unexpected\unknown side effects (besides the Customer Impact identified in the description of this PR around the app expecting "app local" to be used during patch roll-forward).

The test plan covers 1 and adds additional tests in second commit in this PR that verify TPA version checks in additional scenarios, ensuring behavior is as exepected between app and framework(s) regarding who will win by TPA version, ties and ensure roll-forward behavior (including patch) does not affect that.

steveharter:
Feedback from separate mail thread indicates this is OK to merge

PR #4549, opened 2018-09-07
Bump versions & Mark packages stable for 1.1.10

wtgodbe:
@weshaggard @safern PTAL

CC @crummel


pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -130,7 +130,10 @@
       <Version>4.0.1</Version>
     </NETCoreApp10Dependency>
     <NETCoreApp10Dependency Include="System.Net.Http">
-      <Version>4.1.3</Version>
+      <Version>4.1.4</Version>
+    </NETCoreApp10Dependency>
+    <NETCoreApp10Dependency Include="System.Net.Http.WinHttpHandler">
```
weshaggard: This one shouldn't be needed as it isn't referenced directly by Microsoft.NETCore.App.

PR #4552, opened 2018-09-10
Add new RIDs for DotNetHostPolicy in stable.packages.props

wtgodbe:
Some builds were failing because I missed adding entries for the new RIDs. We also need to build DotNetHost & DotNetHostResolver since Debian9 was introduced in this release.

@weshaggard @safern @eerhardt PTAL

### DIFF
--- a/pkg/stable.packages.props
+++ b/pkg/stable.packages.props
@@ -577,10 +577,22 @@
     <StablePackage Include="runtime.debian.8-x64.Microsoft.NETCore.DotNetHostPolicy">
       <Version>1.1.10</Version>
     </StablePackage>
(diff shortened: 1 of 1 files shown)


pkg/stable.packages.props
```diff
@@ -568,56 +568,158 @@
     </StablePackage>
 
     <!-- NETCore 1.1.10 core-setup -->
+    <StablePackage Include="Microsoft.DotNet.PlatformAbstractions">
```
weshaggard: Are we planning to ship these?

eerhardt: No - they are now removed.

PR #4570, opened 2018-09-13
Mark PlatformAbstractions & DependencyModel as stable

wtgodbe:
@weshaggard @safern PTAL

CC @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.PlatformAbstractions/project.json
+++ b/src/Microsoft.DotNet.PlatformAbstractions/project.json
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.1.10-servicing-*",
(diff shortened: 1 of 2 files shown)


weshaggard:
@wtgodbe do we need to ship these? IIRC we generally don't stabilize and ship these.

wtgodbe:
We built them for this release because we added Debian9 for the 1.1.10 release - @eerhardt

wtgodbe:
Actually I think I got these mixed up with the host* packages, we may not need this - looking

wtgodbe:
Ah yeah this is the wrong change, what I need to do is revert https://github.com/dotnet/core-setup/commit/51107bf988797510f8221a0a58fb57fac8be4210#diff-dc55be7422231765e25bc0b6326acbd2. Closing.

PR #4577, opened 2018-09-17
COM Activation host

AaronRobinsonMSFT:
Add new comhost library for COM activation
 - Supports reading clsidmap as file or resource
 - Supports registration and unregistration of COM server

See #4476 for design

cc @vitek-karas @jkotas @morganbr @luqunl @jeffschwMSFT

### DIFF
--- a/src/corehost/build.cmd
+++ b/src/corehost/build.cmd
@@ -34,6 +34,7 @@ if /i [%1] == [rid]         ( set __TargetRid=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [toolsetDir]  ( set "__ToolsetDir=%2"&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [hostver]     (set __HostVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [apphostver]  (set __AppHostVersion=%2&&shift&&shift&goto Arg_Loop)
+if /i [%1] == [comhostver]  (set __ComHostVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [fxrver]      (set __HostResolverVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [policyver]   (set __HostPolicyVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [commit]      (set __CommitSha=%2&&shift&&shift&goto Arg_Loop)
@@ -122,9 +123,9 @@ exit /b 1
 :GenVSSolution
 :: Regenerate the VS solution
 
-echo Calling "%__nativeWindowsDir%\gen-buildsys-win.bat %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__HostResolverVersion% %__HostPolicyVersion%"
+echo Calling "%__nativeWindowsDir%\gen-buildsys-win.bat %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__ComHostVersion% %__HostResolverVersion% %__HostPolicyVersion%"
 pushd "%__IntermediatesDir%"
-call "%__nativeWindowsDir%\gen-buildsys-win.bat" %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__HostResolverVersion% %__HostPolicyVersion% %__PortableBuild%
+call "%__nativeWindowsDir%\gen-buildsys-win.bat" %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__ComHostVersion% %__HostResolverVersion% %__HostPolicyVersion% %__PortableBuild%
 popd
 
 :CheckForProj
--- a/src/corehost/build.cmd
+++ b/src/corehost/build.cmd
@@ -34,8 +34,7 @@ if /i [%1] == [rid]         ( set __TargetRid=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [toolsetDir]  ( set "__ToolsetDir=%2"&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [hostver]     (set __HostVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [apphostver]  (set __AppHostVersion=%2&&shift&&shift&goto Arg_Loop)
-if /i [%1] == [comhostver]  (set __ComHostVersion=%2&&shift&&shift&goto Arg_Loop)
-if /i [%1] == [fxrver]      (set __HostResolverVersion=%2&&shift&&shift&goto Arg_Loop)
+if /i [%1] == [fxrver]      (set __HostFxrVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [policyver]   (set __HostPolicyVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [commit]      (set __CommitSha=%2&&shift&&shift&goto Arg_Loop)
 
@@ -123,9 +122,9 @@ exit /b 1
 :GenVSSolution
 :: Regenerate the VS solution
 
-echo Calling "%__nativeWindowsDir%\gen-buildsys-win.bat %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__ComHostVersion% %__HostResolverVersion% %__HostPolicyVersion%"
+echo Calling "%__nativeWindowsDir%\gen-buildsys-win.bat %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__HostFxrVersion% %__HostPolicyVersion% %__PortableBuild%"
 pushd "%__IntermediatesDir%"
-call "%__nativeWindowsDir%\gen-buildsys-win.bat" %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__ComHostVersion% %__HostResolverVersion% %__HostPolicyVersion% %__PortableBuild%
+call "%__nativeWindowsDir%\gen-buildsys-win.bat" %~dp0 "%__VSVersion%" %__BuildArch% %__CommitSha% %__HostVersion% %__AppHostVersion% %__HostFxrVersion% %__HostPolicyVersion% %__PortableBuild%
 popd
 
 :CheckForProj
--- a/src/corehost/cli/comhost/clsidmap.cpp
+++ b/src/corehost/cli/comhost/clsidmap.cpp
@@ -1,0 +1,178 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "comhost.h"
+#include <trace.h>
+
+#include <cpprest/json.h>
+using namespace web;
+
+using comhost::clsid_map_entry;
+using comhost::clsid_map;
+
+#ifdef _WIN32
+bool GetModuleFileNameWrapper(HMODULE hModule, pal::string_t* recv);
+#endif
+
+namespace
+{
+    pal::dll_t get_current_module()
+    {
+#ifdef _WIN32
+        HMODULE hmod = nullptr;
+        BOOL res = ::GetModuleHandleExW(
+            GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+            reinterpret_cast<LPCWSTR>(&get_current_module),
+            &hmod);
+        assert(res != FALSE);
+        return hmod;
+
+#else
+        return nullptr;
+
+#endif
+    }
+
+    pal::string_t get_current_module_name()
+    {
+#ifdef _WIN32
+        HMODULE hmod = (HMODULE)get_current_module();
+        if (hmod != nullptr)
+        {
+            pal::string_t name;
+            if (GetModuleFileNameWrapper(hmod, &name))
+                return name;
+        }
+#endif
+
+        return {};
+    }
+
+    HRESULT string_to_clsid(_In_ const pal::string_t &str, _Out_ CLSID &clsid)
+    {
+#ifdef _WIN32
+        // If the first character of the GUID is not '{' then COM will
+        // attempt to look up the string in the CLSID table.
+        if (str[0] == _X('{'))
+        {
+            if (SUCCEEDED(::CLSIDFromString(str.data(), &clsid)))
+                return S_OK;
+        }
+#endif
+
+        return __HRESULT_FROM_WIN32(ERROR_INVALID_DATA);
+    }
+
+    clsid_map parse_stream(_In_ pal::istream_t &json_map_raw)
+    {
+        // [TODO] handle BOM
+
+        // Parse JSON
+        json::value json_map = json::value::parse(json_map_raw);
+        json::object &json_obj = json_map.as_object();
+
+        // Process JSON and construct a map
+        HRESULT hr;
+        clsid_map mapping;
+        for (std::pair<utility::string_t, json::value> &prop : json_obj)
+        {
+            CLSID clsidMaybe;
+            hr = string_to_clsid(prop.first, clsidMaybe);
+            if (FAILED(hr))
+            {
+                assert(false && "Invalid CLSID");
+                continue;
+            }
+
+            clsid_map_entry e{};
+
+            json::object &val = prop.second.as_object();
+            e.assembly = val.at(_X("assembly")).as_string();
+            e.type = val.at(_X("type")).as_string();
+
+            mapping[clsidMaybe] = std::move(e);
+        }
+
+        return mapping;
+    }
+
+    class memory_buffer : public std::basic_streambuf<pal::istream_t::char_type>
+    {
+    public:
+        memory_buffer(_In_ DWORD dataInBytes, _In_reads_bytes_(dataInBytes) void *data)
+        {
+            auto raw_begin = reinterpret_cast<pal::istream_t::char_type*>(data);
+            setg(raw_begin, raw_begin, raw_begin + (dataInBytes / sizeof(pal::istream_t::char_type)));
+        }
+    };
+
+    clsid_map get_json_map_from_resource()
+    {
+#ifdef _WIN32
+        HMODULE hmod = (HMODULE)get_current_module();
+        if (hmod != nullptr)
+        {
+            // [TODO] Load resource properly
+            HRSRC resHandle = ::FindResourceW(hmod, MAKEINTRESOURCEW(27), MAKEINTRESOURCEW(72));
+            if (resHandle == nullptr)
+                return {};
+
+            DWORD size = ::SizeofResource(hmod, resHandle);
+            HGLOBAL resData = ::LoadResource(hmod, resHandle);
+            if (resData == nullptr || size == 0)
+                return {}; // [TODO] Throw error
+
+            LPVOID data = ::LockResource(resData);
+            if (data == nullptr)
+                return {}; // [TODO] Throw error
+
+            memory_buffer resourceBuffer{ size, data };
+            pal::istream_t stream{ &resourceBuffer };
+            return parse_stream(stream);
+        }
+#endif
+
+        return{};
+    }
+
+    clsid_map get_json_map_from_file()
+    {
+        pal::string_t map_file_name = get_current_module_name();
+        if (map_file_name.size() > 0)
+        {
+            map_file_name += _X(".clsidmap");
+            if (pal::file_exists(map_file_name))
+            {
+                pal::ifstream_t file{ map_file_name };
+                return parse_stream(file);
+            }
+        }
+
+        return{};
+    }
+}
+
+clsid_map comhost::get_clsid_map()
+{
+    // CLSID map format
+    // {
+    //      "<clsid>": {
+    //          "assembly": <assembly_name>,
+    //          "type": <type_name>
+    //      },
+    //      ...
+    // }
+
+    // Find the JSON data that describes the CLSID mapping
+    clsid_map mapping = get_json_map_from_resource();
+    if (mapping.empty())
+    {
+        trace::verbose(_X("JSON map resource stream not found"));
+
+        mapping = get_json_map_from_file();
+        if (mapping.empty())
+            trace::verbose(_X("JSON map .clsidmap file not found"));
+    }
+
+    return mapping;
+}
--- a/src/corehost/cli/comhost/clsidmap.cpp
+++ b/src/corehost/cli/comhost/clsidmap.cpp
@@ -3,6 +3,8 @@
 
 #include "comhost.h"
 #include <trace.h>
+#include <utils.h>
+#include <error_codes.h>
 
 #include <cpprest/json.h>
 using namespace web;
@@ -10,47 +12,22 @@ using namespace web;
 using comhost::clsid_map_entry;
 using comhost::clsid_map;
 
-#ifdef _WIN32
-bool GetModuleFileNameWrapper(HMODULE hModule, pal::string_t* recv);
-#endif
+bool GetModuleHandleFromAddress(void *addr, HMODULE *hModule);
 
 namespace
 {
     pal::dll_t get_current_module()
     {
-#ifdef _WIN32
         HMODULE hmod = nullptr;
-        BOOL res = ::GetModuleHandleExW(
-            GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
-            reinterpret_cast<LPCWSTR>(&get_current_module),
-            &hmod);
-        assert(res != FALSE);
-        return hmod;
+        bool res = GetModuleHandleFromAddress(&get_current_module, &hmod);
+        assert(res);
+        (void)res;
 
-#else
-        return nullptr;
-
-#endif
-    }
-
-    pal::string_t get_current_module_name()
-    {
-#ifdef _WIN32
-        HMODULE hmod = (HMODULE)get_current_module();
-        if (hmod != nullptr)
-        {
-            pal::string_t name;
-            if (GetModuleFileNameWrapper(hmod, &name))
-                return name;
-        }
-#endif
-
-        return {};
+        return (pal::dll_t)hmod;
     }
 
     HRESULT string_to_clsid(_In_ const pal::string_t &str, _Out_ CLSID &clsid)
     {
-#ifdef _WIN32
         // If the first character of the GUID is not '{' then COM will
         // attempt to look up the string in the CLSID table.
         if (str[0] == _X('{'))
@@ -58,17 +35,26 @@ namespace
             if (SUCCEEDED(::CLSIDFromString(str.data(), &clsid)))
                 return S_OK;
         }
-#endif
 
         return __HRESULT_FROM_WIN32(ERROR_INVALID_DATA);
     }
 
-    clsid_map parse_stream(_In_ pal::istream_t &json_map_raw)
+    clsid_map parse_stream(_Inout_ pal::istream_t &json_map_raw)
     {
-        // [TODO] handle BOM
+        skip_utf8_bom(&json_map_raw);
 
         // Parse JSON
-        json::value json_map = json::value::parse(json_map_raw);
+        json::value json_map;
+        try
+        {
+            json_map = json::value::parse(json_map_raw);
+        }
+        catch (const json::json_exception&)
+        {
+            trace::error(_X("Embedded .clsidmap format is invalid"));
+            throw HResultException{ StatusCode::InvalidConfigFile };
+        }
+
         json::object &json_obj = json_map.as_object();
 
         // Process JSON and construct a map
@@ -108,37 +94,32 @@ namespace
 
     clsid_map get_json_map_from_resource()
     {
-#ifdef _WIN32
         HMODULE hmod = (HMODULE)get_current_module();
-        if (hmod != nullptr)
-        {
-            // [TODO] Load resource properly
-            HRSRC resHandle = ::FindResourceW(hmod, MAKEINTRESOURCEW(27), MAKEINTRESOURCEW(72));
-            if (resHandle == nullptr)
-                return {};
+        if (hmod == nullptr)
+            return{};
 
-            DWORD size = ::SizeofResource(hmod, resHandle);
-            HGLOBAL resData = ::LoadResource(hmod, resHandle);
-            if (resData == nullptr || size == 0)
-                return {}; // [TODO] Throw error
+        HRSRC resHandle = ::FindResourceW(hmod, MAKEINTRESOURCEW(RESOURCEID_CLSIDMAP), MAKEINTRESOURCEW(RESOURCETYPE_CLSIDMAP));
+        if (resHandle == nullptr)
+            return {};
 
-            LPVOID data = ::LockResource(resData);
-            if (data == nullptr)
-                return {}; // [TODO] Throw error
+        DWORD size = ::SizeofResource(hmod, resHandle);
+        HGLOBAL resData = ::LoadResource(hmod, resHandle);
+        if (resData == nullptr || size == 0)
+            throw HResultException{ HRESULT_FROM_WIN32(::GetLastError()) };
 
-            memory_buffer resourceBuffer{ size, data };
-            pal::istream_t stream{ &resourceBuffer };
-            return parse_stream(stream);
-        }
-#endif
+        LPVOID data = ::LockResource(resData);
+        if (data == nullptr)
+            throw HResultException{ E_UNEXPECTED }; // This should never happen in Windows 7+
 
-        return{};
+        memory_buffer resourceBuffer{ size, data };
+        pal::istream_t stream{ &resourceBuffer };
+        return parse_stream(stream);
     }
 
     clsid_map get_json_map_from_file()
     {
-        pal::string_t map_file_name = get_current_module_name();
-        if (map_file_name.size() > 0)
+        pal::string_t map_file_name;
+        if (pal::get_own_module_path(&map_file_name))
         {
             map_file_name += _X(".clsidmap");
             if (pal::file_exists(map_file_name))
--- a/src/corehost/cli/comhost/clsidmap.cpp
+++ b/src/corehost/cli/comhost/clsidmap.cpp
@@ -92,8 +92,9 @@ namespace
         }
     };
 
-    clsid_map get_json_map_from_resource()
+    clsid_map get_json_map_from_resource(bool &found_resource)
     {
+        found_resource = false;
         HMODULE hmod = (HMODULE)get_current_module();
         if (hmod == nullptr)
             return{};
@@ -102,6 +103,7 @@ namespace
         if (resHandle == nullptr)
             return {};
 
+        found_resource = true;
         DWORD size = ::SizeofResource(hmod, resHandle);
         HGLOBAL resData = ::LoadResource(hmod, resHandle);
         if (resData == nullptr || size == 0)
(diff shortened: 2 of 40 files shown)


src/corehost/cli/comhost/clsidmap.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+#ifdef _WIN32
+bool GetModuleFileNameWrapper(HMODULE hModule, pal::string_t* recv);
+#endif
+
+namespace
+{
+    pal::dll_t get_current_module()
+    {
+#ifdef _WIN32
```
jkotas: These `#ifdefs` should not be necessary since this is for Windows only.

If it is ever ported to non-Windows, we can add them at that time.

AaronRobinsonMSFT: Sounds good.

AaronRobinsonMSFT:
cc @sdmaclea

AaronRobinsonMSFT:
@dotnet-bot test all please

AaronRobinsonMSFT:
@dotnet-bot test all please

AaronRobinsonMSFT:
@dotnet-bot test Linux-musl x64 Debug Build please

AaronRobinsonMSFT:
Looking to get this committed so I can stop rebasing :) The second part of this is to update hostfxr.

cc @vitek-karas @jkotas @sdmaclea

src/corehost/build.cmd
```diff
@@ -34,6 +34,7 @@ if /i [%1] == [rid]         ( set __TargetRid=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [toolsetDir]  ( set "__ToolsetDir=%2"&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [hostver]     (set __HostVersion=%2&&shift&&shift&goto Arg_Loop)
 if /i [%1] == [apphostver]  (set __AppHostVersion=%2&&shift&&shift&goto Arg_Loop)
+if /i [%1] == [comhostver]  (set __ComHostVersion=%2&&shift&&shift&goto Arg_Loop)
```
jkotas: Does this argument be passed in somewhere?

I am wondering why we have so many different versions - why can't all hosts have same version?

AaronRobinsonMSFT: That is a great question. I would like to do that, but alas the entire hosting design is an amalgam of various designs and considerations that don't seem to have a cohesive design. I could change it _but_ this is plumbed through the build and deployment mechanisms so a bunch would have to change. I am more than willing to do that though.

@vitek-karas Any concerns you can think of? Are you planning on doing this?

vitek-karas: My understanding is that the versions are just a symptom of a different intent. I think we use the versions to basically "mark" the components which we want to update in a servicing release.
(I think, but I'm not sure, that if the version didn't change, we don't ship/update the binary again).

This is super important for `dotnet.exe` - updating this binary requires a reboot and that is something we try to avoid at almost all costs. Basically any servicing release will not update this, and even feature releases may decide to not update it either.

As for the other components I don't really know why we version them separately. I think this goes more to the setup team, as they consume these/version these.
My take:
* `hostfxr` - should simply build the latest version always - and ship it. At runtime we always pick the latest available. The only "tricky" part are preview releases, where in theory preview might have bugs, but any invocation will pick up the latest (which might be preview). But this is a behavior we already shipped with many times.
* `hostpolicy` - this ships with the runtime, should be versioned with the runtime and so on. We do ship it as a separate Nuget package, but that is probably just a historical thing due to it living in a different repo from the runtime. Otherwise this library "belongs" to CoreCLR.
* `apphost` - SDK uses this to publish apps (and now even to build them). It needs to be packaged separately since it's a component which should not be part of the runtime package which is installed on machines. That said it should version with the runtime, no reason not to do that. SDK should pick the matching version for whatever runtime it picked to publish the app for.
* `comhost` - same story as `apphost`. For this change I think we should make the `comhost` and `apphost` versions the same. In fact I would expect we ship them in the same NuGet package - which leads to the question on how are we going to ship it - I assume the same as `apphost` since it's again a component consumed only by SDK.

/cc @rakeshsinghranchi  for the setup related topics

AaronRobinsonMSFT: Will do this now that I am merging the entire COM host work into this PR.

AaronRobinsonMSFT:
@vitek-karas @sdmaclea @jeffschwMSFT @jkoritzinsky,

This is about ready. I am still thinking through some error cases on how to propagate information to users, but the overall review is as intended. Please take a look.

The build/sdk side of this can be found at https://github.com/dotnet/sdk/pull/2814

src/corehost/cli/comhost/clsidmap.cpp
```diff
@@ -0,0 +1,165 @@
...
+bool GetModuleHandleFromAddress(void *addr, HMODULE *hModule);
+
+namespace
+{
+    pal::dll_t get_current_module()
+    {
+        HMODULE hmod = nullptr;
+        bool res = GetModuleHandleFromAddress(&get_current_module, &hmod);
+        assert(res);
+        (void)res;
```
jeffschwMSFT: (void)res; ?

src/corehost/cli/comhost/exports.cpp
```diff
@@ -0,0 +1,250 @@
...
+#endif // _WIN32
+
+COM_API HRESULT STDMETHODCALLTYPE DllGetClassObject(
+    _In_ REFCLSID rclsid,
+    _In_ REFIID riid,
+    _Outptr_ LPVOID FAR* ppv)
+{
+    // Check if the CLSID map contains a mapping
+    clsid_map map;
+    RETURN_HRESULT_IF_EXCEPT(map = comhost::get_clsid_map());
```
jeffschwMSFT: Do we foresee a world where reading the json every time could either 1) produce inconsistent results or 2) be a performance issue?

src/corehost/cli/coreclr.cpp
```diff
@@ -5,95 +5,132 @@
...
+            && coreclr_execute_assembly != nullptr
+            && coreclr_create_delegate != nullptr);
 
-    pal::unload_library(g_coreclr);
+        return true;
+    }
+    
+    void coreclr_unload()
+    {
+        assert(g_coreclr != nullptr && coreclr_initialize != nullptr);
```
jeffschwMSFT: Doesn't coreclr keep a ref count for who started the runtime?  Should we check before unloading?

src/corehost/cli/coreclr.cpp
```diff
@@ -5,95 +5,132 @@
...
+
+        pal::unload_library(g_coreclr);
+    }
 }
 
-pal::hresult_t coreclr::initialize(
+pal::hresult_t coreclr_t::create(
+    const pal::string_t& libcoreclr_path,
```
jeffschwMSFT: NIT: Any reason to change the name from initialize to create?  The delegate remains coreclr_initialize

src/corehost/cli/coreclr.cpp
```diff
@@ -102,10 +139,140 @@ pal::hresult_t coreclr::execute_assembly(
...
+        return StatusCode::Success;
+    }
+
+    return coreclr_shutdown(_host_handle, _domain_id, latchedExitCode);
+}
+
+namespace
+{
+    const char *PropertyNameMapping[] =
+    {
```
jeffschwMSFT: What is this needed for?  In general having this defined in as few places as possible is desirable.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1234,6 +1437,85 @@ int fx_muxer_t::execute(
...
+
+    // Read config
+    fx_definition_vector_t fx_definitions;
+    auto app = new fx_definition_t();
+    fx_definitions.push_back(std::unique_ptr<fx_definition_t>(app));
+
+    pal::string_t runtime_config = _X("");
+    fx_reference_t override_settings;
+    int rc = read_config(*app, host_info.app_path, runtime_config, override_settings);
+    if (rc != StatusCode::Success)
```
jeffschwMSFT: Is there a need to cache this results?  Creating a lot of delegates could make this code path really hot.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1234,6 +1437,85 @@ int fx_muxer_t::execute(
     return result;
 }
 
+int fx_muxer_t::get_coreclr_delegate(
```
jeffschwMSFT: This method is doing some really heavy lifting about resolving hostfx, initializing coreclr, etc.  Where alternative smaller building blocks considered to not have them all in this entry point?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1234,6 +1437,85 @@ int fx_muxer_t::execute(
...
+    fx_reference_t override_settings;
+    int rc = read_config(*app, host_info.app_path, runtime_config, override_settings);
+    if (rc != StatusCode::Success)
+        return rc;
+
+    runtime_config_t app_config = app->get_runtime_config();
+    bool is_framework_dependent = app_config.get_is_framework_dependent();
+    if (is_framework_dependent)
+    {
+        fx_name_to_fx_reference_map_t newest_references;
```
jeffschwMSFT: If possible, I would like to see this 'common' path be resolved using the common path and not here.  My concern is that we have multiple places to keep in sync and drift is likely.

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -566,3 +395,54 @@ SHARED_API hostfxr_error_writer_fn hostfxr_set_error_writer(hostfxr_error_writer
...
+//     delegate
+//          Requested delegate.
+// Return value:
+//     The error code result.
+//
+// A new CoreCLR instance will be created or reused if the existing instance can satisfy the configuration
+// requirements supplied by the runtimeconfig.json file.
+//
+SHARED_API int32_t hostfxr_get_coreclr_delegate(
+    const pal::char_t* libhost_path,
```
jeffschwMSFT: This is an export, correct?  When adding a new export we should always keep in mind 'what future proofing should we do?'  This may be the ultimate set of parameters, but if it is not we are likely stuck with a hostfxr_get_coreclr_delegate2 in the future.

src/corehost/cli/coreclr.cpp
jeffschwMSFT: Answering my own question - this is a refactor.  Follow-up, any reason for this set and the enum to be in separate files?

src/corehost/cli/hostpolicy.cpp
```diff
@@ -443,16 +563,69 @@ SHARED_API int corehost_main_with_output_buffer(const int argc, const pal::char_
...
+    {
+        return StatusCode::LibHostInvalidArgs;
+    }
+
+    args.trace();
+    return StatusCode::Success;
+}
+
+SHARED_API int corehost_get_coreclr_delegate(coreclr_delegate_type type, void **delegate)
+{
```
jeffschwMSFT: Similar future proofing question - is this all the parameters we could imagine needing?

src/corehost/cli/comhost/clsidmap.cpp
AaronRobinsonMSFT: This is done to avoid code analysis issues with "unused variables". In release builds the `assert` macro is defined as `#define assert(expression) ((void)0)` which means `expression` is not used and therefore during code analysis an unused variable is detected. Adding this statement avoids that.

src/corehost/cli/comhost/exports.cpp
AaronRobinsonMSFT: 1) No.
2) Quite possibly. I am more than willing to cache that but until we can start measuring perf in this scenario I am apprehensive to optimize some of these paths beyond obvious paths that are idempotent.

src/corehost/cli/coreclr.cpp
AaronRobinsonMSFT: This is logical equivalent to the previous code paths. The only difference now is that it is clear what is actually happening. I do not know of any internal ref counting.

src/corehost/cli/comhost/clsidmap.cpp
jkoritzinsky: It's too bad we can't use the C++ `[[maybe_unused]]` attribute here. Maybe we should add an `UNUSED(x)` macro to help make this clearer?

src/corehost/cli/coreclr.cpp
AaronRobinsonMSFT: Because this is creating a `coreclr_t` instance. Initialization is an implementation detail of the type and is an implementation detail. This is a static factory method for the new `coreclr_t` type where as the previous implementation was designed around everything being static.

src/corehost/cli/coreclr.cpp
AaronRobinsonMSFT: Yes. There are common values that are always included and there are extra ones. Reconciling them is a goal, but until we have a definitive process for that or what are exposed, I did this. More than willing to put all strings in a single place if that is preferred but operating on enum's is much less error prone than copying around strings. Enums are also much easier to make into stable APIs than strings.

src/corehost/cli/fxr/fx_muxer.cpp
AaronRobinsonMSFT: 100 % agreed. You have touched on a pontentially idempotent optimization, but it may only be idempotent based on how we define previously loaded coreclr instances and how that equivalency is defined.

src/corehost/cli/fxr/fx_muxer.cpp
AaronRobinsonMSFT: You have found the only other entry point that should be needed. Merging the app/exe, which is done, and the new lib method might be beneficial but in this case I don't see how it would benefit much. More than willing to reconsider but it would require yet more refactoring that doesn't give as much of a win as other changes.

src/corehost/cli/fxr/hostfxr.cpp
AaronRobinsonMSFT: Yep. This is something I am looking into for a slight modification for IJW work. There is a small chance one more argument is added and I am working on that now.

src/corehost/cli/hostpolicy.cpp
AaronRobinsonMSFT: See previous comment on `hostfxr_get_coreclr_delegate`.

AaronRobinsonMSFT:
@jeffschwMSFT Seven new exports have been added:

`comhost.dll` (new library) - There are 4 exports for COM and are beyond our control. Their definition and semantics are defined per Microsoft documentation.

`hostfxr` - New export is `hostfxr_get_coreclr_delegate`. This function is largely stable, but an additional argument may be added prior to merge for future scenarios (e.g. IJW).

`hostpolicy` - New exports are `corehost_get_coreclr_delegate` and `corehost_ensure_load`. These functions are stable and there will be no need for them to be altered. The `corehost_ensure_load` is identical to existing `corehost_load`, but has slightly different semantics. This function may not be needed based on further discussion of error cases in review.

Regarding test plan, the plan is the tests - which are rather exhaustive. I am also working locally by stepping through each scenario (`dotnet`, `apphost`, and `comhost`) during debugging and validating existing semantics aren't altered. Additional validation will unfortunately fall to preview feedback.

AaronRobinsonMSFT:
Also please observe the updated document details (`COM-activation.md`) in PR for the design and an exhaustive description of the `comhost` activation steps. Additional details will be added once error cases and CLR configuration equivalence are defined.

jeffschwMSFT:
> hostfxr - New export is hostfxr_get_coreclr_delegate. This function is largely stable, but an additional argument may be added prior to merge for future scenarios (e.g. IJW).

I think this is a reasonable approach, though looking ahead I can imagine as we potentially expand the hosting interface at the coreclr level, that we may want to export that provides an instance of a loaded coreclr and not re-expose all the exports.  

> The corehost_ensure_load is identical to existing corehost_load

I will take a deeper look at the code change around what semantics are slightly different.

AaronRobinsonMSFT:
>we potentially expand the hosting interface at the coreclr level, that we may want to export that provides an instance of a loaded coreclr

I am hesitant to do that. Exposing that kind of export would imply a policy for how coreclr operates. I would think that at some point we will be able to support multiple instances of the same coreclr running in process. Squirreling that logic away in coreclr seems like too much policy in my mind. Once we have the "libhost" scenario defined this will hopefully reduce to a trivial number the users who need to load the `coreclr` library directly.

vitek-karas:
I agree with Jeff here... there are other asks for a way to use the existing hosting to find runtime/frameworks/TPA but then have a way to do something special with the runtime. This is one such example where the desire is to create a delegate rather than execute app's entry point.
ASP.NET has a slightly different requirement of adding a "try/catch" around main via the hosting layer, which might be similar and use a different version of execute_assembly to do that.

In general it might be a nice API (generic) to "load" the runtime and "unload" - leaving the particulars of what is done with the runtime to the caller. The unload is mostly trivial, but I think we should do that for completeness and for some of the special cases (shutting down tracing, releasing memory, unloading dlls, and so on).

src/corehost/cli/hostpolicy.cpp
```diff
@@ -347,19 +431,55 @@ void trace_hostpolicy_entrypoint_invocation(const pal::string_t& entryPointName)
...
+//
+// Loads and initilizes the hostpolicy.
+//
+// The hostpolicy will be reinitialized with the current
+// host_interface_t argument.
+//
+SHARED_API int corehost_load(host_interface_t* init)
+{
+    return corehost_load_internal(init, true /* reinit */);
+}
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: I have no idea. This was undefined but the existing code had this logic and I didn't want to change it. I think that @vitek-karas mentioned something about it, but alas none of us know why this was there other than some ominous comment:

```cpp
// Re-initialize global state in case of re-entry
```

AaronRobinsonMSFT:
>This is one such example where the desire is to create a delegate rather than execute app's entry point.

Agreed, that is the whole point of the `get_coreclr_delegate()` function.

src/corehost/cli/fxr/hostfxr.cpp
morganbr: For IJW, I'd think we'd want something that starts the runtime and maybe loads an assembly, but not a delegate back.

AaronRobinsonMSFT: @morganbr There is no assumption about the specifics of the delegate in this case. I realize whenever someone hears 'delegate' they think managed delegate. But consider the signature of the function, it is returning a function pointer. It could be a function pointer to an internal native function or a managed function, literally anything. Also, the enum should be observed which will indicate what scenario the delegate is supposed to be supporting. For IJW, this could be a function pointer that takes 100 arguments or none. It is up to us and what we need.

AaronRobinsonMSFT:
Spoke offline with @vitek-karas and I am going to remove the enum value and generic export. There are a myriad of reasons this is being done. I will update PR shortly.

src/corehost/cli/hostpolicy.cpp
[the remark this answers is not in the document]
AaronRobinsonMSFT: I have removed this logic since it appears it was written to support a re-entrant test 😖

AaronRobinsonMSFT:
Based on changes in coreclr (https://github.com/dotnet/coreclr/pull/22317), some additional changes are needed here.

AaronRobinsonMSFT:
@jeffschwMSFT @vitek-karas @sdmaclea @jkoritzinsky Any additional feedback here?

src/corehost/cli/fxr/fx_muxer.cpp
jkoritzinsky: As @jeffschwMSFT mentions, it would be nice to refactor this into a separate method to make it easier to reuse as part of the IJW hosting work. I can do that in the IJW PR though after this is merged in.

AaronRobinsonMSFT: I can do that, but honestly, the last 3 lines and signature are all that is special. IJW will be similar.

jkoritzinsky: Sounds good. I'll handle this as part of the IJW work.

AaronRobinsonMSFT:
Documentation should be updated.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -1,279 +1,374 @@
...
+                    set_app_paths = (pal::cstrcasecmp(_hostpolicy_init.cfg_values[i].data(), "true") == 0);
+                }
+
+                properties.add(_hostpolicy_init.cfg_keys[i].data(), _hostpolicy_init.cfg_values[i].data());
+            }
+
+            // App paths and App NI paths
+            if (set_app_paths)
+            {
+                properties.add("APP_PATHS", _app_base_cstr.data());
```
vitek-karas: Why not have these properties defined in the enum as well?

src/corehost/cli/hostpolicy.cpp
```diff
@@ -1,279 +1,374 @@
...
+                properties.add("APP_NI_PATHS", _app_base_cstr.data());
+            }
+
+            // Startup hooks
+            pal::string_t startup_hooks;
+            if (pal::getenv(_X("DOTNET_STARTUP_HOOKS"), &startup_hooks))
+            {
+                pal::pal_clrstring(startup_hooks, &_startup_hooks_cstr);
+                properties.add("STARTUP_HOOKS", _startup_hooks_cstr.data());
```
vitek-karas: And this one as well.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -1,279 +1,374 @@
...
+    std::shared_ptr<coreclr_t> &coreclr)
+{
+    coreclr = g_lib_coreclr.lock();
+    if (coreclr != nullptr)
     {
-        property_keys.push_back("APP_PATHS");
-        property_keys.push_back("APP_NI_PATHS");
-        property_values.push_back(app_base_cstr.data());
-        property_values.push_back(app_base_cstr.data());
+        // [TODO] Validate the current CLR instance is acceptable for this request
```
vitek-karas: Not sure if this is still needed, in any case we need to deal with the TODO somehow.

src/corehost/corehost.cpp
```diff
@@ -134,16 +140,95 @@ bool resolve_fxr_path(const pal::string_t& host_path, const pal::string_t& app_r
...
+bool resolve_fxr_path(const pal::string_t& root_path, pal::string_t* out_dotnet_root, pal::string_t* out_fxr_path)
+{
+    // If a hostfxr exists in root_path, then assume self-contained.
+    if (library_exists_in_dir(root_path, LIBFXR_NAME, out_fxr_path))
+    {
+        trace::info(_X("Resolved fxr [%s]..."), out_fxr_path->c_str());
+        out_dotnet_root->assign(root_path);
+        return true;
+    }
+
```
vitek-karas: What's the reason to not support `DOTNET_ROOT`?
I would expect the `hostfxr` search to be basically identical between apphost and libhost as they are both local to the app/component and both need to support framework dependent.
The only difference I could see us do is the check above - libhost + self-contained - not sure we want to support that right now (I don't think we thought it through...) - there's also the minor fact that SDK can't really produce that.

src/corehost/corehost.cpp
```diff
@@ -161,32 +246,79 @@ bool resolve_fxr_path(const pal::string_t& host_path, const pal::string_t& app_r
...
+    {
+        trace::error(_X("The library %s was found, but loading it from %s failed"), LIBFXR_NAME, fxr_path.c_str());
+        trace::error(_X("  - Installing .NET Core prerequisites might help resolve this problem."));
+        trace::error(_X("     %s"), DOTNET_CORE_INSTALL_PREREQUISITES_URL);
+        return StatusCode::CoreHostLibLoadFailure;
+    }
+
+    // Leak fxr
+
+    auto entry_del = (hostfxr_get_com_activation_delegate_fn)pal::get_symbol(fxr, "hostfxr_get_com_activation_delegate");
```
vitek-karas: Nit: please rename the variable to `hostfxr_get_com_activation_delegate`, it makes the code much easier to read since it looks exactly like if there's a direct call (and no `GetProcAddress` magic).

src/corehost/cli/comhost/clsidmap.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include <utils.h>
+#include <error_codes.h>
+
+#include <cpprest/json.h>
+using namespace web;
+
+using comhost::clsid_map_entry;
+using comhost::clsid_map;
+
+bool GetModuleHandleFromAddress(void *addr, HMODULE *hModule);
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: That is already done. See `pal.unix.cpp`.

src/corehost/cli/comhost/clsidmap.cpp
```diff
@@ -0,0 +1,165 @@
...
+        // Process JSON and construct a map
+        HRESULT hr;
+        clsid_map mapping;
+        for (std::pair<utility::string_t, json::value> &prop : json_obj)
+        {
+            CLSID clsidMaybe;
+            hr = string_to_clsid(prop.first, clsidMaybe);
+            if (FAILED(hr))
+            {
+                assert(false && "Invalid CLSID");
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: This is a core invariant based on tooling writing `CLSID` strings _or_ a nefarious user. This is an example of a classic assert rather than some kind of state we can get into.

src/corehost/cli/hostpolicy.cpp
AaronRobinsonMSFT: No real reason other than we optionally set them rather than always set them. The enum was designed to be a place for unified documentation of what should be considered set. In cases where it is a niche case, I left them out. I can add if that is preferred though.

src/corehost/corehost.cpp
AaronRobinsonMSFT: But it isn't a direct call. It is a call based on some very important logic (i.e. the call is depending on another library). Obscuring that doesn't help readability of code, it loses the context of a loose dependency. I will change it if that is the style, but I think the proposed argument for changing it is flawed.

src/corehost/cli/comhost/clsidmap.cpp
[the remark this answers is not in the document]
vitek-karas: Sorry - I don't see it. All I see in `pal.unix.cpp` is `get_own_module_path`. The comment is about moving `get_current_module` to pal as well.

src/corehost/cli/comhost/clsidmap.cpp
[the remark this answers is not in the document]
vitek-karas: OK - makes sense.

src/corehost/cli/hostpolicy.cpp
vitek-karas: I don't feel strongly about this either... feel free to leave it as is... 
I personally would probably add the enums... but it's just a feeling.

src/corehost/corehost.cpp
vitek-karas: Hmm - OK - I didn't see it that way.
What I found confusing was the last line of this function - call to `entry_del`:
* I didn't find this as a function in the source code (I should have thought about function pointers, but still)
* I was trying to make sense of what it does from the name - and it read as "delete entry" or something along those lines.

Using the name of the export would have made it very clear to me that this is call to hostfxr and what the entry point is.
So I'm OKish to go with a name which doesn't necessarily match the export exactly, but please do rename this to something more descriptive.

src/corehost/cli/comhost/clsidmap.cpp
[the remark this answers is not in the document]
AaronRobinsonMSFT: Ah missed your point. Okay, will add a placeholder for that as well.

src/corehost/cli/comhost/clsidmap.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        // Parse JSON
+        json::value json_map;
+        try
+        {
+            json_map = json::value::parse(json_map_raw);
+        }
+        catch (const json::json_exception&)
+        {
+            trace::error(_X("Embedded .clsidmap format is invalid"));
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: So basically call `hostfxr_set_error_writer(nullptr)`? I would then check if the return is non-nullptr and if it is, reset it?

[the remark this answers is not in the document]
AaronRobinsonMSFT: Oh... bad juju. if it is null, we write to stderr... boo. Okay, will supply a nop writer - sad panda.

src/corehost/cli/hostpolicy.cpp
AaronRobinsonMSFT: That is a post preview TODO and somewhat related to the version roll forward logic.

AaronRobinsonMSFT:
> At least one or two tests which call the new export and verify that it does its thing.

Unsure what I would validate in that scenario. There is no result to confirm really.

> Alongside that we should start thinking what it would take to write more extensive tests

This is related to the complete E2E. As soon as I get my changes in, @jkoritzinsky can submit the SDK updates, at which point we can actually write an E2E test and validate things in the CI. I will create an issue.

AaronRobinsonMSFT:
See https://github.com/dotnet/core-setup/issues/5113 for testing.

src/corehost/corehost.cpp
AaronRobinsonMSFT: I don't think supporting `DOTNET_ROOT` for COM hosting is the right thing to do purely because there is no need at the moment since this only works on Windows and our install locations are well-defined and the fact that it is yet another option that makes the support matrix even more complicated. In line with that, the work being done for version roll forward and how COM will always do major version should make it unnecessary. I can add it, but I don't see a strong need for it. I am in favor of making it relevant in the libhost scenario though.

src/corehost/cli/comhost/clsidmap.cpp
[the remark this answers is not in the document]
vitek-karas: Well - stderr is/was the default... so it made sense to use nullptr for the "default" behavior. We can change that.. but then we would have to somehow define another constant for the "default".

src/corehost/cli/hostpolicy.cpp
vitek-karas: OK - maybe create an issue for it - just for tracking :-)

src/corehost/corehost.cpp
vitek-karas: My view is about predictability. We already have feedback that it's "surprising" that `dotnet.exe` doesn't support `DOTNET_ROOT` while `apphost` does.
The other perspective is testability (even for our own tests) - `DOTNET_ROOT` is the only way to redirect comhost to a private copy of framework. apphost is kind of similar (although it does support self-contained).
IJW will be in the same boat.
We could add a "test-only" env variable to redirect, but given the apphost has it public, I would just use it everywhere.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -1,279 +1,376 @@
...
+            // Runtime options config properties.
+            for (int i = 0; i < _hostpolicy_init.cfg_keys.size(); ++i)
+            {
+                // Provide opt-in compatible behavior by using the switch to set APP_PATHS
+                if (pal::cstrcasecmp(_hostpolicy_init.cfg_keys[i].data(), "Microsoft.NETCore.DotNetHostPolicy.SetAppPaths") == 0)
+                {
+                    set_app_paths = (pal::cstrcasecmp(_hostpolicy_init.cfg_values[i].data(), "true") == 0);
+                }
+
+                properties.add(_hostpolicy_init.cfg_keys[i].data(), _hostpolicy_init.cfg_values[i].data());
```
AaronRobinsonMSFT: @vitek-karas Do you know why we are pushing this into the coreclr property bag? I am unable find any consumption of this property in the coreclr repo? Looks like it is purely for policy, so I don't think we should be propagating it. Unless there is some argument to be made for looking at why we set the `APP*PATHS` properties when debugging coreclr.

src/corehost/corehost.cpp
```diff
@@ -267,6 +267,14 @@ bool get_latest_fxr(pal::string_t fxr_root, pal::string_t* out_fxr_path)
 
 #if defined(CURHOST_LIB)
 
+namespace
+{
+    void libhost_default_error_writer(const pal::char_t*)
```
vitek-karas: I would call this `libhost_null_error_writer` - default is really the stderr one.

src/corehost/corehost.cpp
AaronRobinsonMSFT: Boo. I abhor test-only env variables. Will bring back support.

src/corehost/corehost.cpp
```diff
@@ -295,10 +303,20 @@ int get_com_activation_delegate(pal::string_t *app_path, com_activation_fn *dele
...
         return StatusCode::CoreHostEntryPointFailure;
 
+    auto set_error_writer = (hostfxr_set_error_writer_fn)pal::get_symbol(fxr, "hostfxr_set_error_writer");
+    if (set_error_writer != nullptr)
+    {
+        // Set the error writer during COM activation to the default.
+        // If the previous value is non-null, then set it to the defined value.
+        hostfxr_error_writer_fn prev = set_error_writer(libhost_default_error_writer);
+        if (prev != nullptr)
+            (void)set_error_writer(prev);
```
vitek-karas: I don't understand this - the end result is that if there was a writer, you leave it there.
* There really should not be one - otherwise whoever set it made a mistake
* Regardless - we don't want to write the COM related errors to a random writer - think about it as operation-scoped. COM operation should be isolated from everything else going on, just like you use ALC to isolate its assembly loading.

AaronRobinsonMSFT: I disagree here. Since this is thread local, if the thread was used to activate a CoreCLR instance, then it should be respected. If it wasn't set, then we keep the value, but users should have the final say if they set an error writer no?

src/corehost/corehost.cpp
```diff
@@ -307,7 +325,7 @@ int get_com_activation_delegate(pal::string_t *app_path, com_activation_fn *dele
     app_path_local.replace(app_path_local.begin() + idx, app_path_local.end(), _X(".dll"));
 
     *app_path = std::move(app_path_local);
-    return entry_del(host_path.c_str(), dotnet_root.c_str(), app_path->c_str(), (void**)delegate);
+    return get_com_delegate(host_path.c_str(), dotnet_root.c_str(), app_path->c_str(), (void**)delegate);
```
vitek-karas: You should reset the error writer here to its original value. To make things clean.
There's a `propagate_error_writer` class in utils to make this simple and robust.

src/corehost/cli/hostpolicy.cpp
vitek-karas: I don't know the answer to this. I know that they are exposed through `AppContext.GetData` so there could be use cases in managed code. But I suspect it was basically a "why not" solution. Which is not a great reason. Unfortunately I don't think we can remove it - it would be a breaking change and we don't really have a good enough reason (other than it should not have been done in the first place)

src/corehost/corehost.cpp
vitek-karas: The scenario I'm thinking about is like this:
* Host (for example our apphost in GUI mode would do this) which sets a custom error writer
* It calls into hostfxr to execute an app
* Later on the app does a COM activation on an object which happens to be written in .NET Core as well - it could be that it's trying things out and fully expects the operation to fail.
* The code flow gets into this code (as part of the COM activation) - given the code we would leave the custom error writer in place
* The COM activation fails - we write some errors to the custom writer - so in our case it goes all the way down to the original apphost
* The managed app catches the exception/failure and moves on since it expected the failure
* Much later on it exits - code gets to the apphost which sees that there are some unreported errors - so it pops up a dialog with the error message - which makes no sense in that context.

In general I view the COM activation as a self-contained operation which should not leak anything outside - errors included. Or maybe a different way to put it - the caller should be able to handle the error completely, which this wouldn't allow.

AaronRobinsonMSFT: > * Much later on it exits - code gets to the apphost which sees that there are some unreported errors - so it pops up a dialog with the error message - which makes no sense in that context.

@vitek-karas I'm sorry but this is test nonsense and just back practice. The scenario described is around how poorly this repo does testing _not_ helping a developer understand what is going on in their application. By explicitly setting the activation path to the null stream there is no way for a COM server consumer to know why the activation failed. This is making the situation harder. I agree that having a noisy logs is painful but that is more an artifact of just logging everything instead of logging things are aren't expected (i.e. warnings) or actions that fail (i.e. errors). When an error occurs there should be a full dump, but in the case where nothing fails or defaults are expected then the logs should be generally thin. I would be more inclined to have a conversation about how to reduce existing noise in the logs rather than hide COM activation logic which has been and will be with this suggestion the hardest area to debug.

AaronRobinsonMSFT: Chatted offline with Vitek. I will remove this and instead rely on `IErrorInfo` since that is the COM way.

AaronRobinsonMSFT:
@vitek-karas @jkoritzinsky @jeffschwMSFT 

I have submitted the coreclr repo changes. This PR and the SDK are all that remain for functionality. Once the SDK changes can be consumed here, we will author some additional tests - see #5113.

PR #4581, opened 2018-09-18
Bump version to 1.1.11 & Mark packages non-stable

wtgodbe:
@safern @weshaggard PTAL

### DIFF
--- a/src/Microsoft.DotNet.PlatformAbstractions/project.json
+++ b/src/Microsoft.DotNet.PlatformAbstractions/project.json
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.1.10-servicing-*",
+  "version": "1.1.11-servicing-*",
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
--- a/TestAssets/TestProjects/PortableApp/project.json
+++ b/TestAssets/TestProjects/PortableApp/project.json
@@ -9,7 +9,7 @@
       "dependencies": {
         "Microsoft.NETCore.App": {
           "type": "platform",
-          "version": "1.1.10"
+          "version": "1.1.11-servicing-*"
         },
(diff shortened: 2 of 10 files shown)


src/Microsoft.DotNet.PlatformAbstractions/project.json
```diff
@@ -1,6 +1,6 @@
 {
   "description": "Abstractions for making code that uses file system and environment testable.",
-  "version": "1.1.10-servicing-*",
+  "version": "1.1.11-servicing-*",
```
safern: This never ships stable right? Is there need to bump the version?

wtgodbe: @eerhardt do we not need to bump this/DependencyModel?

eerhardt: To make servicing (and understanding) easier, we've taken the approach that PlatformAbstractions and DependencyModel version with the rest of the branch.

For example: https://www.nuget.org/packages/Microsoft.Extensions.DependencyModel/

image

@safern is correct that we don't need to mark these packages as 'stable' when we aren't actually servicing them. However, I think it is still good practice to keep the version numbers all in sync in the branch. That way when looking at myget or other places of the packages produced by the branch, you can make sense of it all. So moving everything to `1.1.11-servicing-*` now is good. And then when we stablize `1.1.11`, we only mark stable the packages we want to ship. Then everything moves together to `1.1.12-servicing-*`.

safern: Sounds reasonable. I just wanted to understand why they were being bumped since when I was managing the servicing branches I recall I wouldn't bump those unless we were going to service them. Thanks for explaining 😄

weshaggard: Yeah I'm OK with always versioning these as it does make it a little easier to reason about, especially in these servicing branches. However as already stated we shouldn't ship them unless they have changes.

PR #4586, opened 2018-09-19
Add a target for netstandard2.0

### LINKED ISSUE #3680: Microsoft.Extensions.DependencyModel should target netstandard2.0
We should have at least a dependency group for ns2.0 so that we don't bring down unnecessary dependencies.

eerhardt:
Adding netstandard2.0 TFMs to the PlatformAbstractions and DependencyModel libraries so we don't force users to bring down unnecessary dependencies.

Fix #3680

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
+++ b/src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
@@ -3,13 +3,26 @@
 
   <PropertyGroup>
     <Description>Abstractions for reading `.deps` files.</Description>
-    <TargetFrameworks>net451;netstandard1.3;netstandard1.6</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.3;netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>net451;netstandard1.3;netstandard1.6;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.3;netstandard1.6;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
+
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'netstandard2.0'">
+      <PropertyGroup>
+        <NewtonsoftJsonPackageVersion>11.0.1</NewtonsoftJsonPackageVersion>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <NewtonsoftJsonPackageVersion>9.0.1</NewtonsoftJsonPackageVersion>
+      </PropertyGroup>
+    </Otherwise>
+  </Choose>
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'netstandard1.6' ">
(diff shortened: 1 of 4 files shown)


src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -3,13 +3,26 @@
...
-    <TargetFrameworks>net451;netstandard1.3;netstandard1.6</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.3;netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>net451;netstandard1.3;netstandard1.6;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard1.3;netstandard1.6;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
 
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'netstandard2.0'">
+      <PropertyGroup>
+        <NewtonsoftJsonPackageVersion>11.0.1</NewtonsoftJsonPackageVersion>
```
weshaggard: Do we really want to bump the version of this package for NS2.0? That might have an impact on the SDK which I believe consumes this and last I checked only supported version 9.0.1 of this package.

ericstj: I think we need to to in order to accomplish the goal of this PR.  FWIW ASP.NET depends on this lib and already lifts Newtonsoft.Json to 11.0.0.

eerhardt: `11.0.1` is the first version of Newtonsoft.Json that supports/targets NS2.0. My thinking was that it doesn't make sense for DependencyModel to target NS2.0, but then reference a version of another package that still brings in all the `1.x` dependencies that we are trying to avoid with this change.

However, you bring up a good point that upgrading this version will force any user (including the SDK or others that want to run inside VS) to move to Newtonsoft.Json `11.0.1`. And possibily that isn't acceptable. Maybe if other users want to stop bringing in `1.x` dependencies, they can upgrade Newtonsoft in their package/app.

@ericstj - thoughts?

/cc @livarcocc @nguerrera @dsplaisted

ericstj: Referencing a newer version of N.J appears to eclipse the version referenced by Microsoft.Extensions.DependencyModel and avoids downloading the old N.J (and thus the old dependencies).  I know NuGet garuntees the final resolved behavior so this would indeed reduce the packages referenced.  Not sure if NuGet garuntees download behavior, for instance I recall NuGet 2 was greedy+depth-first in traversing the package graph so you'll often download things you don't need, but at least *today* it seems to also avoid the download if the same reference to this package also updates N.J.

I think the deciding factor here is up-stack dependencies.  IIRC SDK is the driver for most functional changes to this library so we need to be certain we are making changes that cater to their needs.

nguerrera: There may be perf consequences to loading a version of N.J newer than VS uses in certain scenarios. We may already be in that situation, though.

nguerrera: I think this is likely fine for 3.0+, which would be Dev16+. We should check that VS will be on 11.0+ there.

nguerrera: Oh, this only bumps the version for netstandard2.0, which our build task desktop target (net46) wouldn't consume. So that's not likely to cause any issue for the SDK unless we retarget.

That said, I think for 3.0 we may well retarget to net472 for 3.0 and I still see N.J 9.0 in dev16 builds.

nguerrera: Following up on whether that will get bumped.

ericstj: Specific TFM always wins over netstandard TFM.

nguerrera: Yeah, just double checking. No reason to hold this change back for SDK.

eerhardt:
test Windows_NT arm Debug Build
test Windows_NT x64 Release Build

eerhardt:
Thanks. Done.

weshaggard:
>@weshaggard @ericstj - any other thoughts? Or is this OK to merge?

I'm OK with it as well. Hopefully we don't have to revisit the decision but we should keep this in the back of our mind as the dependency starts to flow.

PR #4621, opened 2018-10-02
Publish cross-bitness cross-architecture tools on Linux/arm32

echesakovMSFT:
**dotnet/coreclr** now builds and publishes Hostx64/arm32 CrossGen (https://github.com/dotnet/coreclr/pull/19933) as well as *obsolete* Hostx86/arm32 CrossGen during Linux/arm32 official build.

This PR:
1. switches from using Hostx86/arm32 to Hostx64/arm32 CrossGen for producing native images and debugging files (i.e. perfmap files) in core-setup on Linux **and**
2. includes both Hostx86/arm32 and Hostx64/arm32 cross-architecture tools in *runtime.linux-arm64.Microsoft.NETCore.App\*.nupkg* produced by core-setup

These changes would allow up-stack components to use x64 CrossGen during their Linux/arm32 builds without requiring to run under the Docker (this was needed to be able to run a Linux x86 CrossGen application on Linux/x64 build machine).

@eerhardt Please take a look.
/cc @RussKeldorph

skip_ci_please

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
+++ b/src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
@@ -167,6 +167,10 @@
         <TargetPath>runtimes$(_crossDir)/native</TargetPath>
         <IsNative>true</IsNative>
       </FilesToPackage>
+      <FilesToPackage Condition="'$(TargetArchitecture)' == 'arm'" Include="$(_jitPackagePath)/runtimes$(_obsoleteCrossDir)/native/*.*">
+        <TargetPath>runtimes$(_obsoleteCrossDir)/native</TargetPath>
+        <IsNative>true</IsNative>
+      </FilesToPackage>
     </ItemGroup>
 
     <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' == ''">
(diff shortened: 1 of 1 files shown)


src/pkg/projects/Microsoft.NETCore.App/src/Microsoft.NETCore.App.depproj
```diff
@@ -77,6 +78,7 @@
       <_runtimePackagePath>$(PackagesDir)/$(_runtimePackageId.ToLowerInvariant())/$(_runtimePackageVersion)</_runtimePackagePath>
 
       <_crossGenPath>$(_runtimePackagePath)/tools$(_crossDir)/crossgen$(ApplicationFileExtension)</_crossGenPath>
+      <_crossGenPath Condition="'$(TargetArchitecture)' == 'arm' AND '$(OS)' == 'Windows_NT'">$(_runtimePackagePath)/tools$(_obsoleteCrossDir)/crossgen$(ApplicationFileExtension)</_crossGenPath>
```
eerhardt: If this is `obsolete`, then why are we still using it?  If it is valid (i.e. not obsolete) on Windows, then we should call this property something else.

RussKeldorph: This property is temporary to allow us to keep the `x86_arm` directory around long enough to move all the upstack dependents off of it and onto `x64_arm`.  Once there are no more dependents, this will be removed.  I suppose we could call it `deprecated`, but the transition period should be so brief that I'm not sure the terminology is important.

RussKeldorph: Actually, sorry, I missed your emphasis on Windows.  You're right, since this is staying on Windows for now, `obsolete` is not a good choice.

PR #4700, opened 2018-10-25
Add os supported versions to host

### LINKED ISSUE #3638: Host should add supportedOs to application manifest
See https://github.com/dotnet/sdk/issues/1899.

Essentially we should add a application manifest file to dotnet.exe with all the known [supportedOs ](https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx) entries.

vitek-karas:
This change is Windows only.
* Add explicit manifest to `dotnet.exe` which specifies the supported OS versions.
* Remove any manifest from `apphost.exe` as it should get the manifest from the app when it's customized.

Fixes #3638

### DIFF
--- a/src/corehost/cli/dotnet/dotnet.manifest
+++ b/src/corehost/cli/dotnet/dotnet.manifest
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- A list of the Windows versions that this application has been tested on and is
+           is designed to work with. Uncomment the appropriate elements and Windows will 
+           automatically selected the most compatible environment. -->
+
+      <!-- Windows 7 -->
+      <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}" />
+
+      <!-- Windows 8.1 -->
+      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
+
+      <!-- Windows 10 -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
+    </application>
+  </compatibility>
+</assembly>
--- a/src/test/Assets/TestProjects/TestWindowsOsShimsApp/Program.cs
+++ b/src/test/Assets/TestProjects/TestWindowsOsShimsApp/Program.cs
@@ -1,0 +1,149 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace TestWindowsOsShimsApp
+{
+    public static class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
+            Console.WriteLine(string.Join(Environment.NewLine, args));
+            Console.WriteLine($"Framework Version:{GetFrameworkVersionFromAppDomain()}");
+
+        #if WINDOWS
+            Version osVersion = RtlGetVersion();
+            if (osVersion == null)
+            {
+                Console.WriteLine("Failed to get OS version through RtlGetVersion.");
+            }
+            else
+            {
+                Console.WriteLine($"Detected true OS version: {osVersion.Major}.{osVersion.Minor}");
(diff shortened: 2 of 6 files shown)


vitek-karas:
@ericstj Any idea on how to test this? I thought about an app similar to your sample in the issue, but in order to validate it the test would somehow have to be able to determine the correct version of the OS it runs on - without any compatibility magic. Not sure if there's a way to do that.

swaroop-sridhar (approved):

sbomer (approved):
LGTM, thank you for fixing this!

src/corehost/cli/dotnet/dotnet.manifest
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
```
ericstj: You should look at the default manifest provided by the tool chain (and already in dotnet.exe) and merge it with this.  I believe it has things like requested elevation level.

ericstj (commented):
Make sure you test some set of apps with this change.  Probably the built in apps plus the templates.  We don’t want to break any of the basic scenarios and this does have that potential (due to turning off OS compatibility shims).

src/corehost/cli/dotnet/dotnet.manifest
vitek-karas: That is already done by the linker - it merges all manifests - so this one and the one generate by it with the existing settings.

vitek-karas:
by built-in apps you mean all the "dotnet something" tools?
I was thinking about adding some automated tests for this... but I can't figure out how to write the test verification (as it would see the same potentially wrong version of the OS or shims) - any ideas?

src/corehost/cli/dotnet/dotnet.manifest
ericstj: Cool, so long as you've verified that merge behavior that's good.

ericstj:
Yeah, run dotnet <something> tools.  MSBuild is a good one, you could drop this version into a corefx repo and make sure it can still build, that'd be a nice test.  For templates I was primarily thinking about dotnet new mvc, webapi, etc.  Make sure none of those fall on the floor.

WRT to an automated test: you can call RtlGetVersion and it won't lie.  This is what we do in System.Runtime.InteropServices.RuntimeInformation, but in that case we only expose it as a string:
https://github.com/dotnet/corefx/blob/1b16ca1c4bf1ea1a817bc4070362c53341e0ca10/src/Common/src/Interop/Windows/NtDll/Interop.RtlGetVersion.cs#L11-L34

So I imagine an automated test could fetch the current version from RtlGetVersion then ask VerifyVersionInfo to see if we're greater than or equal to that version.  That'd have the nice characteristic of automatically failing when running on a new OS when we need to make a change.

vitek-karas:
I verified this with the new dotnet.exe:
- I can `dotnet new` and then `dotnet run` project types console,winforms,wpf,web,mvc,webapi - everything seems to work just fine
- I can build the CoreFx repo with the new dotnet.exe

I manually exported and double checked the manifest embeded in the new dotnet.exe - it contains the original manifest plus the supported OSes.

ericstj:
test OSX x64 Release Build

src/test/Assets/TestProjects/TestWindowsOsShimsApp/Program.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    public static class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
+            Console.WriteLine(string.Join(Environment.NewLine, args));
+            Console.WriteLine($"Framework Version:{GetFrameworkVersionFromAppDomain()}");
+
+        #if WINDOWS
```
ericstj: Do you need the `#ifdef`?  It looks like you only ever run this on Windows.

vitek-karas: It is not needed, but I personally would prefer to keep it - makes it more explicit. I don't like code which doesn't run, but also doesn't work at all.

vitek-karas:
test OSX x64 Release Build

ericstj:
test OSX x64 Release Build

PR #4739, opened 2018-11-10
disable tests on FreeBSD

wfurt:
The previous change did not work. I was watching tests crawling SLOWLY and timed out at 2H.
They did not hang - there were just super slow. Something we should investigate next week.
On my D4 VM it takes 35 minutes to build & test and it takes ~ 20 minutes on my old desktop. 

I was trying to set BP_SkipTest=false or pass in -SkipTest=True but that conflicted with value passed in down from general section and build did no like same value set twice with different values. 

Here is global run https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2198702&view=logs

Linux with test: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=2198703

BSD without: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=2198709&view=logs

### DIFF
--- a/buildpipeline/Core-Setup-FreeBSD-BT.json
+++ b/buildpipeline/Core-Setup-FreeBSD-BT.json
@@ -200,6 +200,9 @@
     "PB_PackageVersionPropsUrl": {
       "value": ""
     },
+    "PB_SkipTests": {
+      "value": "true"
+    },
     "PACKAGEVERSIONPROPSURL": {
       "value": "$(PB_PackageVersionPropsUrl)"
     }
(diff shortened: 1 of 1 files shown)


buildpipeline/Core-Setup-FreeBSD-BT.json
```diff
@@ -200,6 +200,9 @@
     "PB_PackageVersionPropsUrl": {
       "value": ""
     },
+    "PB_SkipTests": {
+      "value": "true"
```
bartonjs: Why set this and a manual build argument?  Is something anticipating an output if this isn't set to false?

wfurt: not needed. Here is update run. https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/results?buildId=2199610&view=logs

PR #4741, opened 2018-11-12
Skip publishing the freebsd symbols, since the redist from coreclr does not yet work with this step.

bartonjs:


joperezr:
So I guess this means that we won't be able to debug on freebsd? Or do we have an alternative way for people to get the symbols?

bartonjs:
> So I guess this means that we won't be able to debug on freebsd? 

Right now we're still in the process of getting the first official SDK build for FreeBSD built (it's currently using a hand-built version placed directly on the build machines).  Once we get the SDK pipeline flowing we'll be going back and removing the bootstrap buttressing.

PR #4751, opened 2018-11-13
Allow missing assembly dependencies for components.

vitek-karas:
When resolving assembly dependencies for components, don't fail if the assembly doesn't exist on the disk (but it's still present in the .deps.json as a runtime asset).
For dynamically loaded components this behavior makes sense as such components are typically distributed as just the necessary pieces, with everything else provided by the app.

This change is also effectively a workaround for dotnet/sdk#2660, but I think it makes sense on its own for dynamically loaded components. The resolver which is this functionality only returns file paths. It already has a contract that returning null means it could not resolve the required asset. It's up to the caller to decide if this is an issue, or if it's OK to continue.

Right now this is a temporary change mostly intended as a workaround for dotnet/sdk#2660. Issue #4752 tracks the future work to revisit this change and either revert it or keep it (possibly with a knob).

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -411,7 +411,8 @@ bool report_missing_assembly_in_manifest(const deps_entry_t& entry, bool continu
  */
 bool deps_resolver_t::resolve_tpa_list(
         pal::string_t* output,
-        std::unordered_set<pal::string_t>* breadcrumb)
+        std::unordered_set<pal::string_t>* breadcrumb,
+        bool ignore_missing_assemblies)
 {
     const std::vector<deps_entry_t> empty(0);
     name_to_resolved_asset_map_t items;
@@ -444,7 +445,7 @@ bool deps_resolver_t::resolve_tpa_list(
                 return true;
             }
 
-            return report_missing_assembly_in_manifest(entry);
+            return report_missing_assembly_in_manifest(entry, ignore_missing_assemblies);
         }
         else
         {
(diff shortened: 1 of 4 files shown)


swaroop-sridhar:
>> For dynamically loaded components this behavior makes sense ... 
Why would the deps.json of a plugin contain reference to binaries that are actually not found at runtime? 
(I understand it may be because of a bug, to be resolved in dotnet/sdk#2660, but it is not desirable behavior).

It would be better for the plugin to not mention the DLL in its deps.json, but let the app satisfy the dependency somehow. If deps.json actually has a reference to a binary, it likely wants to use that specific DLL instead of the app-version. 

So, I'm concerned that this change may cause silent unintentional behavior. At least, we should have a switch to control the effect.

vitek-karas:
`report_missing_assembly_in_manifest` has a `continueResolving` parameter which basically means to ignore errors (it reports it only as a warning not a fatal error). It's already used for cases where the app is trying to override assembly from framework and also for some special cases in self-contained apps.

vitek-karas:
@swaroop-sridhar I agree that this is partially a workaround for the bug above. I'm not sure if outright failing is the correct behavior though.

I'm somewhat torn on this. On one hand this just makes things work. On the other it does introduce a bit of an imprecision to the system.

What about this:
We will take this for Preview1, so that the system works end to end without big issues (since the sdk issue is unlikely to be fixed for Preview1).
I filed a separate issue #4752 to revisit this decision after the SDK issue is fixed.

swaroop-sridhar:
>>report_missing_assembly_in_manifest has a continueResolving parameter which basically means to ignore errors (it reports it only as a warning not a fatal error). It's already used for cases where the app is trying to override assembly from framework and also for some special cases in self-contained apps. 

Sure, but I'm still not sure whether applying this to every deps.json resolution is good. 
Ex: we ignoring missing dependencies in the TPA/main-app. 
A bug (in the tooling or environment) manifesting like reference number mismatch or incorrectly marked servicing asset may go silently undetected.

swaroop-sridhar:
I just saw your other comment; I agree that there is value and use-cases to this change.
But this is a customer visible behavior. If we think we can change it after Preview 1 (say off by default, on under a flag), I'm OK with that.

vitek-karas:
If we really think that reverting this after Preview1 might cause trouble, then we could add it under a flag... but I think if the end to end works (differently due to changes in SDK), then it should be fine to change it after preview.

PR #4807, opened 2018-11-26
Small SDK lookup fix

vitek-karas:
If the global.json specifies SDK version but we can't find that version in a given search location, the code would end up checking if a folder with name "-1.-1.-1" exist. This is because we're left with empty version structure which is represented as -1.-1.-1.
Not only this is a small perf issue (unnecessarily looking at disk), but if the folder existed by any chance we would have returned that version for use.

Also improves the doc describing the SDK version algorithm by linking to the public docs which have detailed description.

### DIFF
--- a/Documentation/design-docs/multilevel-sharedfx-lookup.md
+++ b/Documentation/design-docs/multilevel-sharedfx-lookup.md
@@ -41,7 +41,10 @@ There are two possibilities for a muxer: it can be a framework-dependent app or 
 
 In the first case the app file path should have been specified as an argument to the dotnet.exe.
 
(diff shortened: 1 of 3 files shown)


vitek-karas:
@johnbeisner - can you please take a look?

johnbeisner:
@vitek-karas 
"but if the folder existed by any chance [-1,-1,-1] we would have returned that version for use."
I do not believe that is the case - the inclution of the -1,-1,-1 versioning was addressed:

Issue:
https://github.com/dotnet/core-setup/issues/1760

PR:
https://github.com/dotnet/core-setup/pull/2793
Look at the first commit for clarity...

Issue #4637, opened 2018-10-05
Runtime dynamic multilevel lookup - muxer


johnbeisner:
Parent issue: https://github.com/dotnet/cli/issues/10100

SDK muxer [the host] and multi-level lookup: https://github.com/dotnet/cli/issues/10153

PR #4857, opened 2018-12-12
Multilevel lookup includes a custom SDK installation directory

ghost:
Reference: https://github.com/dotnet/cli/issues/10100

Addresses:
https://github.com/dotnet/core-setup/issues/4637
https://github.com/dotnet/cli/issues/10153

**Synopsis:**
The method 'get_global_dotnet_dirs' retrieves the default and custom SDK installation directories. We are culling out duplicates as it is entirely possible that the custom and default SDK installation directories are the same.
The method 'get_sdk_self_registered_dir' retrieves the custom installation directory of SDKs if the registy key exists and has a value. Please note that the lookup occurs in the 32-bit registry in all cases.

**Manual Testing:**

```
C:\Users\jbeisner>"C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\dotnet.exe" --list-sdks
2.1.500 [C:\Program Files\dotnet\sdk]
3.0.100-preview-009817 [C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\sdk]
3.0.100-preview-009820 [D:\dotnet\x64\sdk]
```

```
C:\Users\jbeisner>"C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\dotnet.exe" --info
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-preview-009820
 Commit:    799a33c6ba

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17134
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   D:\dotnet\x64\sdk\3.0.100-preview-009820\

Host (useful for support):
  Version: 3.0.0-preview-27211-0
  Commit:  8af9445da6

.NET Core SDKs installed:
  2.1.500 [C:\Program Files\dotnet\sdk]
  3.0.100-preview-009817 [C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\sdk]
  3.0.100-preview-009820 [D:\dotnet\x64\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0-preview-18579-0056 [C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0-preview-18579-0056 [D:\dotnet\x64\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0-preview-27122-01 [C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0-preview-27205-02 [D:\dotnet\x64\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.0.0-alpha-27128-4 [C:\Users\jbeisner\AppData\Local\Microsoft\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 3.0.0-alpha-27128-4 [D:\dotnet\x64\shared\Microsoft.WindowsDesktop.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download
```


nguerrera:
Is this ready for review? Asking because it's marked WIP.

sdmaclea (commented):
LGTM except for the ToDo item

src/corehost/common/pal.windows.cpp
```diff
@@ -199,14 +199,24 @@ bool pal::get_default_servicing_directory(string_t* recv)
 
 bool pal::get_global_dotnet_dirs(std::vector<pal::string_t>* dirs)
 {
-    pal::string_t dir;
-    if (!get_default_installation_dir(&dir))
+    pal::string_t default_dir;
+    pal::string_t custom_dir;
+    bool gdir_found = false;
```
vitek-karas: What is `gdir`? I assume it means "global dir" per the method name. In that case please don't use abbreviations which are not common. Anyway, I think just `dir_found` would work well here.

src/corehost/common/pal.windows.cpp
```diff
@@ -227,6 +237,48 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
...
     append_path(recv, _X("dotnet"));
+
+    return true;
+}
+
+bool pal::get_sdk_self_registered_dir(pal::string_t* recv)
+{
+    DWORD size = 0;
+    const HKEY hkey = HKEY_LOCAL_MACHINE;
+	// The registry search occurs in the 32-bit registry in all cases.
```
vitek-karas: Nit: indentation

src/corehost/common/pal.windows.cpp
```diff
@@ -227,6 +237,48 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
...
+    // Determine the size of the buffer
+    LONG result = ::RegGetValueW(hkey, sub_key, value, flags, nullptr, nullptr, &size);
+    if (result != ERROR_SUCCESS || size == 0)
+    {
+        return false;
+    }
+
+    // Get the key's value
+    std::vector<wchar_t> buffer(size/sizeof(wchar_t));
+    result = ::RegGetValueW( hkey, sub_key, value, flags, nullptr, &buffer[0], &size);
```
vitek-karas: Nit: space before `hkey`

src/corehost/common/pal.windows.cpp
```diff
@@ -227,6 +237,48 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
...
+    if (result != ERROR_SUCCESS)
+    {
+        return false;
+    }
+
+    pal::char_t* text(buffer.data());
+    recv->clear();
+    recv->assign(text);
```
vitek-karas: Can we just call `recv->assign(buffer.data())` ? - I don't see a need for the `text` local variable.
Also - is it necessary to call `clear` - I would think that `assign` overwrites completely previous value.

vitek-karas (changes requested):
We need some testing around this.
I'm not sure how to go about it though - as normally testing this would require admin rights (to modify machine wide registry).
Maybe we'll have to introduce some kind of test-only behavior or something like that... 
Maybe we could ask owners of the `Microsoft.Win32.Registry` APIs as I assume they have to run into similar problems

src/corehost/common/pal.windows.cpp
```diff
@@ -246,9 +246,11 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
 
 bool pal::get_sdk_self_registered_dir(pal::string_t* recv)
 {
+	recv->clear();
```
vitek-karas: Nit: Indentation

src/corehost/common/pal.windows.cpp
```diff
@@ -227,6 +237,44 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
...
+    return true;
+}
+
+bool pal::get_sdk_self_registered_dir(pal::string_t* recv)
+{
+    DWORD size = 0;
+    const HKEY hkey = HKEY_LOCAL_MACHINE;
+    const DWORD flags = RRF_RT_REG_SZ | RRF_SUBKEY_WOW6432KEY;
+
+    //TODO: Use the get_arch() method here; includes arm & arm64
```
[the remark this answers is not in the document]
johnbeisner: Done...

src/corehost/common/pal.windows.cpp
```diff
@@ -227,7 +240,44 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
...
 
     append_path(recv, _X("dotnet"));
+
+    return true;
+}
+
+bool pal::get_sdk_self_registered_dir(pal::string_t* recv)
+{
+#if _TARGET_ARM_
+    //  Self-registered SDK installation directory is not supported for win-arm
```
nguerrera: Is there a fundamental reason why this cannot be supported on ARM, or are we scoping it out?

src/corehost/common/pal.windows.cpp
```diff
@@ -227,7 +240,44 @@ bool pal::get_default_installation_dir(pal::string_t* recv)
...
+
+    // Determine the size of the buffer
+    LONG result = ::RegGetValueW(hkey, sub_key.c_str(), value, flags, nullptr, nullptr, &size);
+    if (result != ERROR_SUCCESS || size == 0)
+    {
+        return false;
+    }
+
+    // Get the key's value
+    std::vector<wchar_t> buffer(size/sizeof(wchar_t));
```
nguerrera: Should we be using `pal::char_t` over `wchar_t` throughout? We're relying on them being the same on Windows in passing sub_key.c_str() to RegGetValueW anyway, correct?

nguerrera (commented):
LGTM overall, just a couple of questions.

As for testing, we just discussed having some test alternative to the reg key, say env var `DOTNET_TEST_SDK_SELF_REGISTERED_DIR_<arch>`. 

This means that we'd still leave coverage of the last mile in actually reading the registry to manual testing. We can add to the end-to-end test plan for that. @vitek-karas Would that work for you?

src/corehost/common/pal.windows.cpp
johnbeisner: @nguerrera 
The 'RegGetValueW' symbol does not resolve in the win-arm compiler - plus the self-registered location is a feature of the native SDK bundle installer.

src/corehost/common/pal.windows.cpp
johnbeisner: Good point - we should be consistent...

vitek-karas:
I'm perfectly fine with some test hook. Do we have a precedent for using env variables as test hooks? (just curious, I'm no opposed to the idea).
The only worry I have is that somebody may start using it, and we'll likely break them eventually - but if we name it right and don't document it (and make it explicit in the code) then I'm fine.
I would go with underscore first - to make it very explicit, that this is not something to be used externally.

src/corehost/common/pal.h
```diff
@@ -221,6 +221,7 @@ namespace pal
     bool get_global_dotnet_dirs(std::vector<pal::string_t>* recv);
 
     bool get_default_installation_dir(pal::string_t* recv);
+    bool get_sdk_self_registered_dir(pal::string_t* recv);
```
vitek-karas: There doesn't seem to be a point in having this in the pal header. The function is not used outside of pal.
Removing it would also let us remove it from the Unix pal (where it's not used at all).

vitek-karas (commented):
Please add tests for some of the other places which are impacted by this change. Searching for usage of pal::get_global_dotnet_dirs shows 5 places where it's used.

johnbeisner:
@vitek-karas 
Adding **true** multi-level SDK and Runtime look-up; moving non-multi-level SDK version resolution tests.

vitek-karas (approved):
LGTM - thanks!

src/corehost/common/pal.windows.cpp
```diff
@@ -12,6 +12,9 @@
 #include <ShlObj.h>
 #include <ctime>
 
+#if !defined(RRF_SUBKEY_WOW6432KEY)
+#define RRF_SUBKEY_WOW6432KEY  0x00020000
```
pakrym: This flag wouldn't work on non-win10 operation systems.

/cc @vitek-karas we may want to improve win8-9-2012 coverage for host tests.

Issue #4291, opened 2018-06-21
Latest Microsoft.DiaSymReader doesn't work with .NET Core 2.1


eerhardt:
## Steps to reproduce
In MSBuild on .NET Core 2.1, write a task that uses the latest version of Microsoft.DiaSymReader to convert PDBs from portable to windows pdbs.

An example of this is [GenFacades Task](https://github.com/dotnet/buildtools/blob/ecfc7cca1b4f9ad9b836dc2c097399b7c8daaa57/src/GenFacades/GenFacades.Core/GenFacades.cs#L239-L261) in dotnet/buildtools.

## Expected  behavior
The task should work fine.

## Actual behavior

```
F:\git\corefx\Tools\partialfacades.task.targets(64,5): error : Specified method is not supported. [F:\git\corefx\src\System.Diagnostics.Debug\src\System.Diagnostics.Debug.csproj]
F:\git\corefx\Tools\partialfacades.task.targets(64,5): error :    at Microsoft.DiaSymReader.SymUnmanagedReaderFactory.CreateReaderWithMetadataImport[TSymUnmanagedReader](Stream pdbStream, Object metadataImport, SymUnmanagedReaderCreationOptions options) in /_/src/Microsoft.DiaSymReader/Reader/SymUnmanagedReaderFactory.cs:line 68 [F:\git\corefx\src\System.Diagnostics.Debug\src\System.Diagnostics.Debug.csproj]
F:\git\corefx\Tools\partialfacades.task.targets(64,5): error :    at Microsoft.DiaSymReader.Tools.SymReaderHelpers.CreateWindowsPdbReader(Stream pdbStream, MetadataReader metadataReader) in F:\git\symreader-converter\src\Microsoft.DiaSymReader.Converter\Utilities\SymReaderHelpers.cs:line 41 [F:\git\corefx\src\System.Diagnostics.Debug\src\System.Diagnostics.Debug.csproj]
F:\git\corefx\Tools\partialfacades.task.targets(64,5): error :    at Microsoft.DiaSymReader.Tools.SymReaderHelpers.CreateWindowsPdbReader(Stream pdbStream, PEReader peReader) in F:\git\symreader-converter\src\Microsoft.DiaSymReader.Converter\Utilities\SymReaderHelpers.cs:line 36 [F:\git\corefx\src\System.Diagnostics.Debug\src\System.Diagnostics.Debug.csproj]
[...]

PR #4860, opened 2018-12-13
Port buildpipeline json to Azure devops Yaml

rakeshsinghranchi:
Tasks accomplished:

- [x] Convert buildpipeline json to Azure devops yaml

- [x] Publish to Build asset registry

- [x] Enable building CI against internal and external repos consuming appropriate machine pool

- [x] Enable building Official builds from internal repo

- [x] New build definition doesn't build FreeBSD for now.

- [x] Removed building Ubuntu 17.10 and Fedora 26 , as are not supported anymore.

- [x] Enabled building Windows leg using dotnet core by replacing Visual Studio MSBuild.

- [ ] Build Product using Arcade SDK . This is out of scope for this Pull request. Goal is to fully consume Arcade before 3.0 release.

- [ ] Remove buildpipeline json files. Clean up work will happen in subsequent PR.

Full test build : https://dnceng.visualstudio.com/internal/_build/results?buildId=57485

Publish changes w.r.t current Product orchestration in Core-Setup: 
1. Finalize-Publish leg doesn't update Versions repo as it B.A.R takes care of managing the assets.


DotnetCLIVersion.txt
```diff
@@ -1 +1 @@
-2.1.301
+2.1.401
```
rakeshsinghranchi: This was required to resolve symbol publish issue. https://github.com/dotnet/core-setup/issues/4291

dependencies.props
```diff
@@ -45,13 +45,13 @@
   <!-- Package versions used as toolsets -->
   <PropertyGroup>
     <FeedTasksPackage>Microsoft.DotNet.Build.Tasks.Feed</FeedTasksPackage>
-    <FeedTasksPackageVersion>3.0.0-preview1-03430-01</FeedTasksPackageVersion>
+    <FeedTasksPackageVersion>2.2.0-beta.18578.9</FeedTasksPackageVersion>
```
rakeshsinghranchi: Consuming Arcade built Build Tasks Feed to generate asset manifest file.

dependencies.props
```diff
@@ -45,13 +45,13 @@
...
   </PropertyGroup>
 
   <!-- Publish symbol build task package -->
   <PropertyGroup>
     <PublishSymbolsPackage>Microsoft.SymbolUploader.Build.Task</PublishSymbolsPackage>
-    <PublishSymbolsPackageVersion>1.0.0-beta-63412-03</PublishSymbolsPackageVersion>
+    <PublishSymbolsPackageVersion>1.0.0-beta-63604-05</PublishSymbolsPackageVersion>
```
rakeshsinghranchi: Required to fix finalize-publish issue : https://github.com/dotnet/arcade/issues/1425

eng/jobs/finalize-publish.yml
```diff
@@ -0,0 +1,94 @@
...
+    - task: MicroBuildSigningPlugin@2
+      displayName: Install MicroBuild plugin for Signing
+      inputs:
+        signType: $(SignType)
+        zipSources: false
+        feedSource: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
+      continueOnError: false
+      condition: and(succeeded(), in(variables['SignType'], 'real', 'test'))
+
+    - task: MSBuild@1
```
rakeshsinghranchi: This step performs signing and sign task (microbuild ) doesn't work with dotnet core yet. Thus, left behind as MSBuild Task. Active issue : https://github.com/dotnet/arcade/issues/1000

publish/publish.proj
```diff
@@ -32,7 +32,7 @@
 
   <Target Name="PublishFinalOutput"
           Condition="'$(Finalize)' == 'true'"
-          DependsOnTargets="PublishCoreHostPackages;SetupPublishSymbols;PublishSymbols;FinalizeBuildInAzure;UpdateVersionsRepo" />
```
rakeshsinghranchi: Versions repo update is not required anymore. @mmitche confirmed this can be removed .

publish/publish.proj
```diff
@@ -234,7 +234,7 @@
   <Target Name="SetupPublishSymbols" Condition="'$(SymbolServerPath)'!=''" >
     <PropertyGroup>
       <SymbolExpirationInDays Condition="'$(SymbolExpirationInDays)'=='' and '$(SymbolExpirationDate)'==''">1</SymbolExpirationInDays>
-      <ConvertPortablePdbsToWindowsPdbs>true</ConvertPortablePdbsToWindowsPdbs>
+      <ConvertPortablePdbsToWindowsPdbs>false</ConvertPortablePdbsToWindowsPdbs>
```
rakeshsinghranchi: Same as in Arcade publish task. https://github.com/dotnet/arcade/blob/ffc7a871add67b229c7fd41ad1cdff41da926312/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj#L138

publish/publish.proj
```diff
@@ -262,39 +262,32 @@
...
     <Error Condition="'$(PackagesUrl)'==''" Text="Missing property PackagesUrl" />
     <Error Condition="'$(TransportFeedAccessToken)'==''" Text="Missing property TransportFeedAccessToken" />
     <PropertyGroup>
       <ExpectedFeedUrl>$(PackagesUrl)</ExpectedFeedUrl>
+      <AssetManifestFilePath>$(BUILD_SOURCESDIRECTORY)/artifacts/log/AssetManifest/Manifest.xml</AssetManifestFilePath>
     </PropertyGroup>
     <ItemGroup>
       <ItemsToPush Remove="@(ItemsToPush)" />
       <ItemsToPush Include="@(_DownloadedStandardPackages)"/>
+       <ItemsToPush Include="@(_DownloadedSymbolsPackages)"/>
```
rakeshsinghranchi: Symbols and Standard nuget packages are published to same feed, thus, simplifying here.

publish/publish.proj
```diff
@@ -262,39 +262,32 @@
...
     </ItemGroup>
     <PushToBlobFeed ExpectedFeedUrl="$(ExpectedFeedUrl)"
                     AccountKey="$(TransportFeedAccessToken)"
```
rakeshsinghranchi: Publishes only nuget packages , what about other assets like native installers, zips, checksums ? Is it required ?

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+# Platform: Linux 
+- template: /eng/jobs/linux-build.yml
+  parameters:
+    Jobs:
+    # Linux glibc X64 leg
+    - job: 
+      displayName: Linux_x64_glibc
+      agentOS: Linux
+      variables:
+        RunArguments: docker run --privileged --rm
```
MichaelSimons: Another thing that seems like it would help simplify and reduce duplication is to have these Run, Build, and Publish args defined in the template and then define args for the values that are different.  For example the RunArgs are the same except for the image/container name and an optional environment variable.  Encapsulating this in a single spot would help avoid unintentional differences.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+  - template: /eng/jobs/validate-signing.yml
+    parameters:
+      pool:
+        ${{ if eq(variables['System.TeamProject'], 'public') }}:
+          name: NetCorePublic-Int-Pool
+          queue: Windows.10.Amd64.Open
+        ${{ if ne(variables['System.TeamProject'], 'public') }}:
+          name: dotnet-internal-temp
+      dependsOn: 
+          - Linux_x64_glibc
```
MichaelSimons: It feels like these names should be more descriptive.  `Linux_x64_glibc` is ambiguous in this context.  Something like `Build_Linux_x64_glibx` would be more useful.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+          - Linux_x64_Alpine36
+          # - Linux_x64_FreeBSD
+          - Linux_x64_Rhel6
+          - Windows_x64
+          - Windows_x86
+          - Windows_Arm
+          - Windows_Arm64
+          - OSX
+  # Finalize Publish 
+  - template: /eng/jobs/finalize-publish.yml
```
MichaelSimons: Why are any parameters needed here?  Isn't finalize-publish only invoke in one place?  Can these just be defined within the template?  This comment applies to any template that is only used once.

eng/jobs/linux-build.yml
```diff
@@ -0,0 +1,178 @@
...
+        /p:ConfigurationGroup=$(_BuildConfiguration)
+        /p:OSGroup=Linux
+        /p:OfficialBuildId=$(OfficialBuildId)
+        /p:StabilizePackageVersion=$(IsStable)
+
+      SourcesDirectory: '$(Build.SourcesDirectory)'
+
+      # Packaging related variables
+      # Debian
+      CommonDebianRunCommand: run --rm --name centos-7-d485f41-20173404063424ubuntu-14.04-debpkg-e5cf912-20175003025046
```
mmitche: Can we not use the native azure devops container support here?  That avoids a whole lot of complexity in managing containers.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
+trigger:
+- master
+
+variables:
+  - name: TeamName
+    value: dotnet-core-acquisition
+  # Skip Running CI tests
+  - name: SkipTests
```
mmitche: Do we run tests in our official build?  It seems odd to skip these otherwise.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+        maxParallel: 2
+        matrix: 
+          debug:
+            _BuildConfiguration: Debug
+          release:
+            _BuildConfiguration: Release
+
+    # Linux Rhel6
+    - job:
+      agentOS: Linux
```
mmitche: what are these agentOS variables used for?

eng/jobs/finalize-publish.yml
```diff
@@ -0,0 +1,94 @@
+parameters:
+  Jobs: []
+jobs:
+- ${{ each job in parameters.Jobs }}: 
+  - job: ${{ job.displayName }}
+    displayName: ${{ job.displayName }}
+    # Run only if all build legs succeeded
+    condition: and(succeeded('Linux_x64_glibc'),
```
mmitche: Is this necessary? Isn't this the default for dependsOn?

mmitche:
@chcosta is added to the review. #Closed

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+          _PublishRidAgnosticPackages: false
+          _TargetArchitecture: x86
+      pool:
+        ${{ if eq(variables['System.TeamProject'], 'public') }}:
+          name: NetCorePublic-Int-Pool
+          queue: Windows.10.Amd64.Open
+        ${{ if ne(variables['System.TeamProject'], 'public') }}:
+          name: dotnet-internal-temp
+      strategy:
+        maxParallel: 2
```
chcosta: maxParallel isn't needed. all it would do would be to constrain the number of parallel jobs you run if your matrix ever grows

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+            _BuildConfiguration: Debug
+          release:
+            _BuildConfiguration: Release
+
+    # Windows Arm64
+    - job:
+      displayName: Windows_Arm64
+      agentOS: Windows
+      variables:
+        _AdditionalMSBuildArguments: "/p:SkipTests=true  /p:StabilizePackageVersion=$(IsStable)"
```
chcosta: nit: there's a double space, and i'm not certain that you need to `"` this

eng/jobs/finalize-publish.yml
```diff
@@ -0,0 +1,94 @@
...
+      displayName: Publish (no PublishType)
+      inputs: 
+        solution: $(Build.SourcesDirectory)\publish\publish.proj
+        platform: $(_TargetArchitecture)
+        configuration: $(_BuildConfiguration)
+        msbuildVersion: 15.0
+        msbuildArchitecture: $(_TargetArchitecture)
+        msbuildArguments: '/p:Finalize=true 
+        /p:OfficialPublish=true 
+        /p:PublishType=$(PublishType)
```
chcosta: Here's an example of where there could be weird behavior because if you removed `/p:PublishType=$(PublishType)` your build would exhibit the same behavior because PublishType would still get set.

eng/jobs/finalize-publish.yml
```diff
@@ -0,0 +1,94 @@
...
+        /p:OfficialPublish=true 
+        /p:PublishType=$(PublishType) 
+        /p:SignType=$(SignType) 
+        /p:StabilizePackageVersion=$(IsStable) 
+        /p:TargetArchitecture=$(_TargetArchitecture) 
+        $(_BlobFeedArgs) 
+        $(_CommonPublishArgs) 
+        $(_NugetFeedArgs) 
+        $(_SymbolServerArgs) 
+        /flp:v=diag;LogFile=$(Build.SourcesDirectory)\\finalizepublish.log'
```
chcosta: Is the `\\` intentional?

You may want to switch this to something like `/bl:$(Build.SourcesDirectory)\finalizepublish.binlog` for performance / readability reasons.

build logs should go into some artifacts folder rather than into the root of the repo

eng/jobs/finalize-publish.yml
```diff
@@ -0,0 +1,94 @@
...
+      inputs:
+        SourceFolder: '$(Build.SourcesDirectory)'
+        Contents: |
+          *.log
+          *.binlog
+        TargetFolder: '$(Build.StagingDirectory)\BuildLogs'
+      continueOnError: true
+      condition: succeededOrFailed()
+
+    - task: CopyFiles@2
```
chcosta: couldn't you use `eng/common/job/job.yml` and simplify a lot of this file?

eng/jobs/linux-build.yml
```diff
@@ -0,0 +1,178 @@
...
+jobs:
+- ${{ each job in parameters.Jobs }}: 
+  - job: ${{ job.displayName }}
+    displayName: ${{ job.displayName }}
+    pool: ${{ job.pool }}
+    strategy: ${{ job.strategy }}
+    variables: 
+      ${{ insert }}: ${{ job.variables }}
+
+      CommonMSBuildArgs: /flp:v=diag /clp:v=detailed
```
chcosta: switch to a binlog

eng/jobs/linux-build.yml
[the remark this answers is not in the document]
chcosta: Is this really the name of the container?  It looks like two container names got merged

eng/jobs/linux-build.yml
```diff
@@ -0,0 +1,178 @@
...
+    # Publish only if internal and not PR 
+    - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+      - script: $(RunArguments)
+          ./Tools/msbuild.sh
+          ./publish/publish.proj
+          $(PublishArguments)
+          $(_CommonPublishArgs)
+        displayName: Publish
+        workingDirectory: '$(SourcesDirectory)'
+        condition: and(succeeded(), eq(variables._BuildConfiguration, 'Release'))
```
chcosta: Other repos are using `_BuildConfig`.  

It seems odd to conditionalize this step in this way via the `_BuildConfiguration` variable.  Couldn't you define "PublishType" on the matrix and set it to "none" or something for the Debug leg and thus let the build step handle not publishing?  ie, move the logic into the build rather than giving your build yaml this control.  That would make adjusting this easier since it would be in one place.

eng/jobs/validate-signing.yml
```diff
@@ -0,0 +1,46 @@
...
+  condition: and(succeeded('Linux_x64_glibc'),
+                succeeded('Linux_Arm64'),
+                succeeded('Linux_Arm'),
+                succeeded('Linux_x64_Alpine36'),
+                succeeded('Linux_x64_Rhel6'),
+                succeeded('Windows_x64'),
+                succeeded('Windows_x86'),
+                succeeded('Windows_Arm'),
+                succeeded('Windows_Arm64'),
+                notIn(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'PullRequest'))
```
chcosta: Can you use this conditional to handle omitting the signing validation template?

https://github.com/dotnet/arcade/blob/master/azure-pipelines.yml#L28

```
${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
```

azure-pipelines.yml
rakeshsinghranchi: Yes,I noticed this too. Will set to 'false' by default.

azure-pipelines.yml
rakeshsinghranchi: This can be removed, will take care of this.

eng/jobs/finalize-publish.yml
rakeshsinghranchi: Private testing confirmed that dependsOn just waits for the execution of the dependent job but doesn't check the run result ( succeeded or failed ) . Thus, added this condition.

azure-pipelines.yml
rakeshsinghranchi: Thanks for the tips,will remove it .

azure-pipelines.yml
rakeshsinghranchi: Is there a tool that you use to check for double space ? :) . Will remove double space and double quotes too.

eng/jobs/osx-build.yml
```diff
@@ -0,0 +1,50 @@
+parameters:
+  pool: {}
+  variables: {}
+jobs:
+- job: OSX
+  pool: ${{ parameters.pool }}
+  strategy: ${{ parameters.strategy }}
+  variables: 
+    SourcesDirectory: '$(Build.SourcesDirectory)'
```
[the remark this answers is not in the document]
rakeshsinghranchi: Makes sense, will get rid of SourcesDirectory.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
+trigger:
+- master
```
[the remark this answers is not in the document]
rakeshsinghranchi: I was thinking about setting  PR trigger as branch policy setting. Will refer arcade yaml on how to set the trigger for PR .

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
+trigger:
```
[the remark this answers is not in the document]
rakeshsinghranchi: Will do for each build leg. Thanks for suggestion.

azure-pipelines.yml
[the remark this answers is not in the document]
mmitche: Shouldn't be set in the UI.  You want:
```
pr:
- master
```

eng/jobs/finalize-publish.yml
chcosta: Hmmm... that seems broken. 

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/multiple-phases?tabs=yaml&view=vsts#conditions

> By default, a job runs if it does not depend on any other job, or if all of the jobs that it depends on have completed and succeeded. You can customize this behavior by forcing a job to run even if a previous job fails or by specifying a custom condition.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+    - group: DotNet-DotNetCli-Storage
+    # Used for publishing to myget
+    - group: DotNet-MyGet-Publish
+    # Used for dotnet/versions update
+    - group: DotNet-Versions-Publish
+
+    # BlobFeed update
+    - name: _PublishBlobFeedUrl
+      value: https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json
+    - name: _BlobFeedArgs
```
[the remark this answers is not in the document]
rakeshsinghranchi: Thanks Chris for your response. Resolving this now.

azure-pipelines.yml
```diff
@@ -0,0 +1,486 @@
...
+    pool:
+      ${{ if eq(variables['System.TeamProject'], 'public') }}:
+        name: Hosted macOS
+      ${{ if ne(variables['System.TeamProject'], 'public') }}:
+        name: Hosted Mac internal
+    strategy:
+      maxParallel: 5
+      matrix: 
+        OSX-x64-debug:
+          _AdditionalBuildArguments: "-PortableBuild=true -strip-symbols -SkipTests=$(SkipTests)"
```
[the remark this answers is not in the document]
rakeshsinghranchi: Agreed, it will simplify things.

eng/jobs/validate-signing.yml
rakeshsinghranchi: It is already taken care of here : https://github.com/dotnet/core-setup/pull/4860/files#diff-fec826feae04e51c0d94076385408bdcR428 . Agreed, it will be more readable if done here too.

eng/jobs/finalize-publish.yml
rakeshsinghranchi: I didn't consider this earlier. Looking at job.yml , consuming it as is would need refactoring on how product builds , artifacts are laid down on the machine. Plan is to consume Arcade in building core-setup before 3.0 ships and this implicitly taken care of then.

eng/jobs/windows-build.yml
```diff
@@ -0,0 +1,180 @@
...
+        feedSource: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
+      continueOnError: false
+      condition: and(succeeded(), in(variables['SignType'], 'real', 'test'))
+
+    # Initialize building
+    - script: $(SourcesDirectory)\build.cmd
+        -- $(CommonMSBuildArgs)
+        /t:BuildTraversalBuildDependencies /flp:v=diag
+      displayName: Build traversal build dependencies
+      condition: succeeded()
```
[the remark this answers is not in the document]
rakeshsinghranchi: Are you referring  to just condition: succeeded() or the Build traversal build dependencies step ?

eng/jobs/linux-build.yml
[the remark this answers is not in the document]
rakeshsinghranchi: Ideally, it should be the image tag name + System.JobId to make it unique. Will take care of this.

azure-pipelines.yml
rakeshsinghranchi: Agreed, will take care this .

eng/jobs/linux-build.yml
rakeshsinghranchi: @chcosta  did share the container support details. Per https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/container-steps.md , this feature is on hold . Also, implementing this latest design change would impact the delivery date. It can be taken care of in next iteration of work pertaining to Arcade.

eng/jobs/linux-build.yml
chcosta: Yes, it does.  But so does a diag log

eng/jobs/finalize-publish.yml
rakeshsinghranchi: For now, will switch to binary logging . Updating the path for logs will be taken care of during next phase of migration to Arcade SDK.

eng/jobs/finalize-publish.yml
```diff
@@ -0,0 +1,94 @@
...
+        signType: $(SignType)
+        zipSources: false
+        feedSource: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
+      continueOnError: false
+      condition: and(succeeded(), in(variables['SignType'], 'real', 'test'))
+
+    - task: MSBuild@1
+      displayName: Publish (no PublishType)
+      inputs: 
+        solution: $(Build.SourcesDirectory)\publish\publish.proj
```
mmitche: @rakeshsinghranchi I noticed the manifest is only tracking the symbol and nuget packages.  What about the msis, exes, zips, etc.

publish/publish.proj
rakeshsinghranchi: @mmitche  , this is the same question I have here . I will look what it takes to includes all these missing assets. BTW, do we need checksums too ?

eng/jobs/finalize-publish.yml
rakeshsinghranchi: Do we need checksums too ?

eng/jobs/linux-build.yml
```diff
@@ -0,0 +1,178 @@
...
+
+    # Only for glibc leg, here we produce RPMs and Debs
+    - ${{ if eq(job.displayName,'Linux_x64_glibc')}}:
+      - task: CopyFiles@2
+        displayName: 'Copy built Portable linux-x64 binaries to staging directory'
+        inputs:
+          SourceFolder: '$(SourcesDirectory)/bin/obj/linux-x64.$(_BuildConfiguration)/sharedFrameworkPublish'
```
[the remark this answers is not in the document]
rakeshsinghranchi: Good suggestion, needs more thoughts on how to achieve this. I will keep this as a separate issue. https://github.com/dotnet/core-setup/issues/4866

publish/publish.proj
jcagme: Yes. I think we'd even create the container if it does not exist

eng/jobs/windows-build.yml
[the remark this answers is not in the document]
dagood: Just the condition--succeeded is the default in my experience when you leave off `condition:` entirely. I checked the docs and it's not clear there yet, but https://github.com/MicrosoftDocs/vsts-docs/issues/2762 tracks that.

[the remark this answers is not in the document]
dagood: On a similar note, `continueOnError: false` is in a few places but `false` [is already the default](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/tasks?view=vsts&tabs=yaml#task-control-options) for that.

azure-pipelines.yml
rakeshsinghranchi: My initial thought was to keep input yml (azure-pipelines.yml) consistent across build legs. This improves readability of the file w.r.t to dependency and pool as it provides overview of job flow for the entire pipeline. Removing parameters and placing the values in place does reduce the overall length of the file. As per suggestion, will update all the jobs that are called only once

eng/jobs/finalize-publish.yml
rakeshsinghranchi: Will take care of this.

eng/jobs/linux-build.yml
rakeshsinghranchi: Agreed, will do. Just note that earlier publishtype value was set only once, now it is set in each build leg.

eng/jobs/osx-build.yml
```diff
@@ -0,0 +1,50 @@
...
+        $(SourcesDirectory)/publish/publish.proj
+        $(_CommonPublishArgs)
+        /p:PublishType=$(PublishType)
+        /p:TargetArchitecture=$(_TargetArchitecture) 
+        /p:ConfigurationGroup=$(_BuildConfiguration) 
+        /p:PortableBuild=$(_PortableBuild) 
+        /p:OSGroup=OSX 
+        $(_AdditionalMSBuildArguments)  
+      displayName: Publish 
+      condition: and(succeeded(), eq(variables._BuildConfiguration, 'Release'))
```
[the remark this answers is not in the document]
rakeshsinghranchi: _BuildConfig variable  (passed via strategy) isn't available during job creation and thus, couldn't move it to step If condition check

mmitche:
@rakeshsinghranchi There should be other queues with VS2017 installed. @MattGal?

rakeshsinghranchi:
@mmitche , I used Hosted VS2017 and it worked fine for the PR build. https://github.com/dotnet/core-setup/pull/4860/checks?check_run_id=42599936 . Waiting for sign-off from reviewers.

MattGal:
Always use hosted if you can, but to answer @mmitche 's question, some other queues with VS 2017 include:

VSTS Agent pools: (Assuming you only want "external" ones here)
dotnet-external-temp
dotnet-external-temp-vs2017
dnceng-windows-spanish-external-temp

Helix queues:
Windows.10.Amd64.ClientRS3.DevEx.Open
Windows.10.Amd64.ClientRS3.VS2017.Open
Windows.10.Amd64.ClientRS4.DevEx.15.8.Open
Windows.10.Amd64.ClientRS4.DevEx.Open
Windows.10.Amd64.ClientRS4.VS2017.Open

PR #4930, opened 2019-01-04
Delete buildpipeline json files as we have moved to azure devops yaml based build definitions

rakeshsinghranchi:


rakeshsinghranchi:
@mmitche  , I still see Jenkins PR builds getting kicked-off.

PR #4936, opened 2019-01-08
Enable FreeBSD leg

rakeshsinghranchi:


eng/jobs/linux-build.yml
```diff
@@ -22,9 +22,11 @@ jobs:
   - job: ${{ parameters.displayName }}
     displayName: ${{ parameters.displayName }}
     pool:
-      ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      ${{ if eq(parameters.displayName,'Build_Linux_x64_FreeBSD') }}:
+        name: dnceng-freebsd-internal
```
dagood: It seems bad that this is available in the public project for PR CI.

eng/jobs/linux-build.yml
```diff
@@ -22,9 +22,11 @@ jobs:
...
+      ${{ if and(eq(variables['System.TeamProject'], 'public'), ne(parameters.displayName,'Build_Linux_x64_FreeBSD')) }}:
         name: dnceng-linux-external-temp
-      ${{ if ne(variables['System.TeamProject'], 'public') }}:
+      ${{ if and(ne(variables['System.TeamProject'], 'public'), ne(parameters.displayName,'Build_Linux_x64_FreeBSD')) }}:
```
dagood: Can this be nested so `ne(parameters.displayName,'Build_Linux_x64_FreeBSD')` is only written out once?

eng/jobs/linux-build.yml
```diff
@@ -224,6 +226,7 @@ jobs:
       condition: succeededOrFailed()
 
     # Force clean up machine in case any docker images are left behind
-    - script: docker system prune -f
-      displayName: Run Docker clean up
-      condition: succeededOrFailed()
\ No newline at end of file
+    - ${{ if ne(parameters.displayName,'Build_Linux_x64_FreeBSD')}}:
```
dagood: \<nit> Missing space after `,`

eng/jobs/linux-build.yml
rakeshsinghranchi: I couldn't find an example of nested if and wasn't sure if it is supported. Will look up and update.

eng/jobs/linux-build.yml
rakeshsinghranchi: yeah, looks like core-sdk too uses the same way. https://github.com/dotnet/core-sdk/commit/7c7f7bae6aa113c2b0bcf3b699dda6ecbf561e77 .

dagood: I'll start a mail thread.

eng/jobs/linux-build.yml
rakeshsinghranchi: How do you find this? :)

dagood: lol, well, I did miss that this is also the case in some of the other conditions. 😉

rakeshsinghranchi: oh, yeah. In many places, almost all of the conditions :( . I wish had vs code auto-format setting on for this type of nit issue.

azure-pipelines.yml
```diff
@@ -130,15 +130,14 @@ jobs:
...
-#     additionalMSBuildArgs: /p:OutputRid=rhel.6-x64
-#     displayName: Build_Linux_x64_FreeBSD
-#     dockerImage: microsoft/dotnet-buildtools-prereqs:centos-6-376e1a3-20174311014331
-#     osGroup: FreeBSD
-#     portableBuild: true
-#     skipTests: true
-#     targetArchitecture: x64
+- template: /eng/jobs/linux-build.yml
+  parameters:
+    displayName: Build_Linux_x64_FreeBSD
```
dagood: Per the mail thread, this leg will have to be internal-only for now: there's no publicly available pool for FreeBSD yet, only the internal one. (I expect FreeBSD legs in CI would fail if we reran it right now.)

wfurt:
BTW we will need to pass in override variables like we did before or update DotnetCLIVersion.txt There is no 2.x version for FreeBSD. core-sdk used 3.,0 preview 1 build.

azure-pipelines.yml
rakeshsinghranchi: >Per the mail thread, this leg will have to be internal-only for now: there's no publicly available pool for FreeBSD yet, only the internal one. (I expect FreeBSD legs in CI would fail if we reran it right now.)

I pushed another change to enable FreeBSD leg for internal builds only.

dagood:
@wfurt Back when we did have the FreeBSD pool, the leg reports it succeeded: https://dnceng.visualstudio.com/public/_build/results?buildId=68484. It does say `Unsupported OS 'FreeBSD' detected. Downloading linux-x64 tools` though. What symptoms would you expect?

wfurt:
that is because old build used environmental variables to point at locally installed SDK. If core-setup uses arcade now, we should either use updated version or use same values as corefx. (may be fastest to talk on IM)

wfurt:
We can do little bit more and we can remove all special handling.

```
[toweinfu@toweinfu-fbsd /mnt/resource/core-setup]$ git diff
diff --git a/DotnetCLIVersion.txt b/DotnetCLIVersion.txt
index 6f1c03c6..2b1526a3 100644
--- a/DotnetCLIVersion.txt
+++ b/DotnetCLIVersion.txt
@@ -1 +1 @@
-2.1.401
\ No newline at end of file
+3.0.100-preview-009812
diff --git a/init-tools.sh b/init-tools.sh
index 5e6abb3c..281c2cdd 100755
--- a/init-tools.sh
+++ b/init-tools.sh
@@ -86,7 +86,10 @@ if [ ! -e "$__DOTNET_PATH" ]; then
                     exit 1
                 fi
                 ;;
-
+            FreeBSD)
+                OS=FreeBSD
+                __PKG_RID=freebsd
+                ;;
             Linux)
                 __PKG_RID=linux
                 OS=Linux
```

>  Total tests: 109. Passed: 109. Failed: 0. Skipped: 0.
>  Test Run Successful.
>  Test execution time: 5.5528 Seconds
>
> Build succeeded.

With this we would 
- not need variable override or local SDK.
- can do crossgen
- can run all tests

The CLI update is needed as there is no 2.1x version. 
I choose version used by core-sdk to decrease variations but we can also use any newer available version.

azure-pipelines.yml
dagood: Ah, yeah, I missed that it changes the basics depending on what's running:

https://github.com/dotnet/core-setup/blob/a8f4ea60d6bb22762777c0097e7dbfeef45a355c/eng/jobs/linux-build.yml#L35-L43

`bash-build.yml` then, I guess?

azure-pipelines.yml
```diff
@@ -130,15 +130,15 @@ jobs:
...
-#     displayName: Build_Linux_x64_FreeBSD
-#     dockerImage: microsoft/dotnet-buildtools-prereqs:centos-6-376e1a3-20174311014331
-#     osGroup: FreeBSD
-#     portableBuild: true
-#     skipTests: true
-#     targetArchitecture: x64
+- ${{ if ne(variables['System.TeamProject'], 'public') }}:
+  - template: /eng/jobs/linux-build.yml
+    parameters:
+      displayName: Build_Linux_x64_FreeBSD
```
bartonjs: Since FreeBSD is its own OS family (not related to Linux), should this just be `Build_FreeBSD_x64`?

dagood:
@wfurt Thanks, filed https://github.com/dotnet/core-setup/issues/4945 to capture this with context. Is it reasonable to start doing FreeBSD builds before making those changes? The way I would view it is: are the env vars causing damage (creating a less useful product), or are they workarounds where maintenance cost is the main concern?

wfurt:
It should be possible to build without it @dagood. I was not sure why the override did not work and avoiding it is one way how to solve it.

dagood:
LGTM. Running an internal build on these changes to see if the leg works when it's enabled would also be nice, if you haven't done that.

As for the thread, IMO we can adjust the legs later when we get a resolution and at least get assets producing for now.

azure-pipelines.yml
dagood: @rakeshsinghranchi Or, actually, I do think this change is worthwhile. (Also brought up in https://github.com/dotnet/core-setup/pull/4936#discussion_r246215809.)

rakeshsinghranchi:
> LGTM. Running an internal build on these changes to see if the leg works when it's enabled would also be nice, if you haven't done that.

I haven't kicked-off an internal build. We can fix any if we hit any hiccup. With my initial PR , I had ensured that leg assets were published to blobstorage.

azure-pipelines.yml
dagood: I'll submit a new PR with these changes.

PR #4946, opened 2019-01-09
Build definition renames for FreeBSD

dagood:
`Build_Linux_x64_FreeBSD` => `Build_FreeBSD_x64`

`linux-build.yml` => `bash-build.yml` (Definitely open to suggestions on this one....)

Followup from https://github.com/dotnet/core-setup/pull/4936#discussion_r246409102.

/cc @wfurt @bartonjs

### DIFF
--- a/eng/jobs/bash-build.yml
+++ b/eng/jobs/bash-build.yml
@@ -22,9 +22,9 @@ jobs:
   - job: ${{ parameters.displayName }}
     displayName: ${{ parameters.displayName }}
     pool:
-      ${{ if and(eq(parameters.displayName, 'Build_Linux_x64_FreeBSD'), ne(variables['System.TeamProject'], 'public')) }}:
+      ${{ if and(eq(parameters.displayName, 'Build_FreeBSD_x64'), ne(variables['System.TeamProject'], 'public')) }}:
         name: dnceng-freebsd-internal
-      ${{ if ne(parameters.displayName, 'Build_Linux_x64_FreeBSD') }}:
+      ${{ if ne(parameters.displayName, 'Build_FreeBSD_x64') }}:
         ${{ if eq(variables['System.TeamProject'], 'public') }}:
           name: dnceng-linux-external-temp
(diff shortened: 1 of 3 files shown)


eng/jobs/finalize-publish.yml
```diff
@@ -8,7 +8,7 @@ jobs:
                   succeeded('Build_Linux_Arm64'),
                   succeeded('Build_Linux_Arm'),
                   succeeded('Build_Linux_x64_Alpine36'),
-                  succeeded('Build_Linux_x64_FreeBSD'),
+                  succeeded('Build_FreeBSD_x64'),
```
bartonjs: I'd fix the sorting on the two changes in this file.

eng/jobs/bash-build.yml
```diff
@@ -22,24 +22,24 @@ jobs:
   - job: ${{ parameters.displayName }}
     displayName: ${{ parameters.displayName }}
     pool:
-      ${{ if and(eq(parameters.displayName, 'Build_Linux_x64_FreeBSD'), ne(variables['System.TeamProject'], 'public')) }}:
+      ${{ if and(eq(parameters.displayName, 'Build_FreeBSD_x64'), ne(variables['System.TeamProject'], 'public')) }}:
```
bartonjs: (Up on line 8)  does `osGroup` need to be changed to FreeBSD for this group?

dagood: OSGroup does need to be FreeBSD, but the value in this file is only the default--it's already overridden here:

https://github.com/dotnet/core-setup/blob/c5f6ce6670ba67cf1e5aaec87c1053aa4fc62bf9/azure-pipelines.yml#L134-L138

eng/jobs/finalize-publish.yml
dagood: Hmm, noticed the non-alpha order was in sync with `azure-pipelines.yml`. I'm going to sort that too.

Issue #4949, opened 2019-01-10
FreeBSD official build leg fails


dagood:
Build: https://dev.azure.com/dnceng/internal/_build/results?buildId=69569

```
/mnt/resource/agent/_work/3/s/src/pkg/projects/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj(62,5):
  error MSB4062: The "GenerateFileVersionProps" task could not be loaded from the assembly /mnt/resource/agent/_work/3/s/Tools/local/netstandard2.0/core-setup.tasks.dll.
  Assembly with same name is already loaded Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.
```

This issue tracks fixing this and reenabling this leg.

dagood:
In source-build, I've experienced this sort of error when MSBuild node reuse is enabled. The `--ci`/`-ci` Arcade script option disables node reuse, and our builds aren't currently passing it--adding it is probably the fix.

dagood:
Forgot that even though we have Arcade scripts in the repo, we don't use them yet. `/nr:false` is the MSBuild arg way, and `MSBUILDDISABLENODEREUSE=1` is an env var way.

dagood:
Fixed, successful internal build: https://dev.azure.com/dnceng/internal/_build/results?buildId=70295&_a=summary

dagood:
This kind of issue happened again when Core-Setup built within 10 minutes of CoreCLR building, although `Microsoft.Build.Tasks.CodeAnalysis.dll` failed this time (for task `Microsoft.CodeAnalysis.BuildTasks.Csc`): https://dev.azure.com/dnceng/internal/_build/results?buildId=77984.

I noticed that Core-Setup has some `Exec` tasks that run `dotnet msbuild` commands, so I think we need to use the env var approach to make sure the setting propagates into that inner command. Submitting a PR to add that.

dagood:
FreeBSD agents are now not accepting jobs. Internal mail thread ongoing: seems to be that there's a new agent version needed.

dagood:
The infra issue should be over, so the leg should work if we enabled it. Filed https://github.com/dotnet/core-setup/issues/5083 to track that.

(Note: the leg is still disabled right now.)

PR #4950, opened 2019-01-10
Disable node reuse in AzDO builds on persistent machines, add FreeBSD

dagood:
See https://github.com/dotnet/core-setup/issues/4949 for info about the issue/fix.

I wasn't able to repro the original issue--I'd bet it's because it had been too long since a build happened on the FreeBSD agent and the old MSBuild node disappeared.

I also applied the fix to Windows because we use persistent machines there, too. In the future we'll be on Arcade and use the `-ci`/`--ci` flag to apply generally known tweaks we need for CI.

Working on running a testing build with this fix.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,15 +71,15 @@ jobs:
   ################################################################################
   # Build Bash legs (Linux and FreeBSD)
   ################################################################################
-# - ${{ if ne(variables['System.TeamProject'], 'public') }}:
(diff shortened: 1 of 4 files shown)


dagood:
My test internal build got through all the legs, only publishing is in progress now: https://dev.azure.com/dnceng/internal/_build/results?buildId=70240. I'm not concerned about publish and it takes a long time, so I'm going ahead with this merge to get an official build going with FreeBSD.

Issue #4944, opened 2019-01-09
Remove LTTng dependency from runtime-deps package


MichaelSimons:
Liblttng (and it's dependency userspacercu) are not actually required to build or run .NET Core applications. They are only required when building the CoreCLR product.

Liblttng is an optional component that can be used alongside lttng-tools to collect tracing data from running .NET Core applications. Removing liblttng saves some disk space. Given that it can't be used on its own without lttng-tools, there isn't a reason to have it included in the runtime-deps packages. We should remove it to help those who are striving for super-small dotnet installs.

Related to https://github.com/dotnet/dotnet-docker/pull/877

cc @rakeshsinghranchi, @brianrob

MichaelSimons:
cc @dagood

Issue #4789, opened 2018-11-19
Implement a build system for Snaps


dleeapho:
.NET Core Snaps are currently shipping but the process is manual. We need to implement a build system (based on Arcade)

PR #4991, opened 2019-01-16
Remove LTTng dependency from runtime-deps packages

dagood:
Removes the `liblttng-ust0` dependency from runtime-deps packages to reduce size. I tried out a dev build's Debian 9 deb package in a container, and a self-contained app started working after installing the updated package.

My test was a simple console app, so it doesn't rule out more complex dependencies. However, https://github.com/dotnet/core-setup/issues/4944 goes into more detail about how this dependency isn't needed. The change has also already been rolled out to the runtime deps and sdk Docker images.

/cc @brianrob @rakeshsinghranchi

### DIFF
--- a/src/pkg/packaging/deb/dotnet-runtime-deps-debian_config_debian.8-x64.json
+++ b/src/pkg/packaging/deb/dotnet-runtime-deps-debian_config_debian.8-x64.json
@@ -32,7 +32,6 @@
         "libc6":{},
         "libgcc1":{},
         "libgssapi-krb5-2":{},
-        "liblttng-ust0":{},
         "libstdc++6":{},
         "zlib1g":{},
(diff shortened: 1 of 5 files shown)


rakeshsinghranchi:
Looks good to me.

dagood:
Thinking about the snaps again, @leecow, do you need to make changes to your snap generation to remove LTTng and match this change?

@MichaelSimons Yeah, I saw them, but the snap files in this repo aren't currently used and not kept up to date. (See https://github.com/dotnet/core-setup/pull/3660 for why they're here.) I'll be following up on that when we actually add a snaps build system with https://github.com/dotnet/core-setup/issues/4789.

leecow:
Yes, lttng is included as a staged package for snap builds. Will remove the next time I refresh.

Issue #3311, opened 2017-10-14
Remove dependency from JSON.NET


bradwilson:
Today, you have a dependency on JSON.NET in some fundamental components (for example, [Microsoft.Extensions.DependencyMode](https://www.nuget.org/packages/Microsoft.Extensions.DependencyModel/1.0.0)). This has the unfortunate side-effect that it becomes either difficult or impossible for end users to choose something newer without either running up against potential incompatibilities, or being outright unable to do so at all (for example, in a plugin model application which leverages M.E.D to load dependencies, but does so after the older JSON.NET has already been loaded).

An example of this dependency which becomes a hard-blocker is unit testing w/ xUnit.net, which uses M.E.D to load the dependencies of the unit test libraries. See: https://github.com/xunit/xunit/issues/1060

I've previously discussed this with a few team members, but notably @DamianEdwards indicated that removing this in a point release (f.e. the upcoming 2.1) would be considered to be a breaking change. I'm not 100% sure I agree with that assessment, but either way, this is not a great situation for end users to be in, to have dependencies on things your team doesn't ship and doesn't control the revisions of.

Petermarcu:
outside of dependency model, I believe there are other parts of the ASP.NET stack that also have this dependency. Is it the case that if we don't remove all of them, then we still have the problem. I would like to see us free JSON.net from the platform although, we'd have to work through the details. We can probably move some of the platform components to use the low allocating Json parsers in corefx lab over time.

bradwilson:
There are two versions of this problem:

## The "indirect" model

This is caused by applications that are _indirectly_ consuming a library which has a competing version of a dependency (for example, the unit testing system in xUnit.net, or a plug-in system for an application). In this case the version linked by the originating application is loaded before you're even aware that a conflict may come up, so whatever version the application has already loaded will be the one that's used. The dynamically loaded code (f.e., the plug-in or the unit tests) get silently stuck with the application's preference.

In desktop CLR days, we had both app domains and cross-process RPC options that could be used to reduce or eliminate these problems, which xUnit.net does use (as probably do some applications that use plug-in models, if for no other reason than isolation). In .NET Core land, there is really nothing similar, because the platform has eliminated app domains, and AFAIK does not include any simple cross-process RPC, either (like old-school remoting).

## The "direct" model

[...]

PR #5009, opened 2019-01-19
Update M.E.DependencyModel to use S.T.Json source for netstandard2.0

ahsonkhan:
Partially addresses https://github.com/dotnet/core-setup/issues/3311

Initial attempt to port Microsoft.Extensions.DependencyModel to use the System.Text.Json source package for TFM netstandard2.0. **Note:** ~There are some API changes going through so this is still WIP.~ We have the JsonReader and JsonWriter available atm so only using the components that are ready.

The System.Text.Json binary is in-box on .NET Core 3.0.

We can't remove the reference to Newtonsoft since M.E.DependencyModel has multiply TFMs: `net451;netstandard1.3;netstandard1.6;netstandard2.0`

**Is it possible to drop support for all but netstandard2.0?** Otherwise, we have to conditionally include the source package.

See https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/source_package/README.md

Here are the diffs to try to highlight what changed during the port:
- DependencyContextJsonReader.cs: https://www.diffchecker.com/KEX0UhJd
- DependencyContextWriter.cs: https://www.diffchecker.com/GByfr9Gi
- JsonTextReaderExtensions.cs: https://www.diffchecker.com/7NfdaYUn

Related PR: https://github.com/dotnet/core-setup/pull/5010

cc @joshfree, @bartonjs, @steveharter, @eerhardt, @KrzysztofCwalina, @JamesNK, @glennc


src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Buffers;
+using System.Diagnostics;
+using System.IO;
+
+namespace Microsoft.Extensions.DependencyModel
+{
+    internal struct StreamBufferWriter : IBufferWriter<byte>, IDisposable
```
ahsonkhan: We should consider adding such an adapter between IBufferWriter and Stream, publicly.

cc @stephentoub, @davidfowl

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -6,29 +6,41 @@
     <TargetFrameworks>net451;netstandard1.3;netstandard1.6;netstandard2.0</TargetFrameworks>
```
ahsonkhan: What should we do about the other TFMs?

src/managed/Microsoft.Extensions.DependencyModel/JsonTextReaderExtensions.Netstandard2_0.cs
```diff
@@ -0,0 +1,139 @@
...
+                    }
+                }
+
+                return true;
+            }
+
+            return false;
+        }
+
+        internal static void Skip(this ref Utf8JsonReader jsonReader)
```
ahsonkhan: There is an issue to add this API: https://github.com/dotnet/corefx/issues/33295

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Netstandard2_0.cs
```diff
@@ -0,0 +1,442 @@
...
+            if (runtimeLibrary != null)
+            {
+                // Add runtime-agnostic assets
+                AddAssets(DependencyContextStrings.RuntimeAssembliesKey, runtimeLibrary.RuntimeAssemblyGroups.GetDefaultGroup(), ref jsonWriter);
+                AddAssets(DependencyContextStrings.NativeLibrariesKey, runtimeLibrary.NativeLibraryGroups.GetDefaultGroup(), ref jsonWriter);
+                AddResourceAssemblies(runtimeLibrary.ResourceAssemblies, ref jsonWriter);
+
+                // Add runtime-specific assets
+                bool wroteObjectStart = false;
+                wroteObjectStart = AddRuntimeSpecificAssetGroups(DependencyContextStrings.RuntimeAssetType, runtimeLibrary.RuntimeAssemblyGroups, wroteObjectStart, ref jsonWriter);
```
ahsonkhan: This was necessary since the writer is forward only (while building a JObject and writing at the end did not care about the order).

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,837 @@
...
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        private const int UnseekableStreamInitialRentSize = 4096;
+
+        private static ArraySegment<byte> ReadToEnd(Stream stream)
```
ahsonkhan: Borrowed from https://github.com/dotnet/corefx/blob/d1da417dbf38079f06b397bdf3c8f51b94e62238/src/System.Text.Json/src/System/Text/Json/JsonDocument.Parse.cs#L182. I am working on providing better stream support with the reader, but it still results in the calling code being overly complex.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+                Runtimes = runtimes,
+                Natives = natives,
+                Compilations = compilations,
+                RuntimeTargets = runtimeTargets,
+                Resources = resources,
+                CompileOnly = compileOnly
+            };
+        }
+
+        private IEnumerable<Dependency> ReadTargetLibraryDependencies(ref Utf8JsonReader reader)
```
ahsonkhan: I am not sure why this API was public. I have marked it as private here.
```C#
public IEnumerable<Dependency> ReadTargetLibraryDependencies(JsonTextReader reader)
```

src/managed/Microsoft.Extensions.DependencyModel/JsonTextReaderExtensions.Netstandard2_0.cs
```diff
@@ -0,0 +1,139 @@
...
+                }
+                throw CreateUnexpectedException(ref reader, "a JSON true or false literal token");
+            }
+            return reader.GetBooleanValue();
+        }
+
+        internal static Exception CreateUnexpectedException(ref Utf8JsonReader reader, string expected)
+        {
+            return new FormatException($"Unexpected character encountered, excepted '{expected}' " +
+                                       $"at line {reader.CurrentState._lineNumber} position {reader.CurrentState._bytePositionInLine}");
```
ahsonkhan: The `Utf8JsonReader` doesn't track/create a path.

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
davidfowl: Yes please. We have something similar in SignalR https://github.com/aspnet/AspNetCore/blob/958e9049894137bb08c0d239dca0cda7c850a691/src/SignalR/common/Shared/MemoryBufferWriter.cs#L14 (ignore the fact that it's also a Stream, we just did that for other reasons).

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+            {
+                compilationOptions = CompilationOptions.Default;
+            }
+
+            Target runtimeTarget = SelectRuntimeTarget(targets, runtimeTargetName);
+            runtimeTargetName = runtimeTarget?.Name;
+
+            if (runtimeTargetName != null)
+            {
+                var seperatorIndex = runtimeTargetName.IndexOf(DependencyContextStrings.VersionSeperator);
```
bartonjs: Typo: sepArator

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
davidfowl: Also the PipeWriter implemenation over a Stream is 90% similar https://github.com/aspnet/AspNetCore/blob/958e9049894137bb08c0d239dca0cda7c850a691/src/SignalR/common/Shared/MemoryBufferWriter.cs#L14.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+                if (runtimeTarget == null)
+                {
+                    runtimeTarget = compileTarget;
+                    framework = ridlessTarget.Name;
+                }
+            }
+
+            if (runtimeTarget == null)
+            {
+                throw new FormatException("No runtime target found");
```
bartonjs: I'm rather surprised that core-setup isn't using resource strings

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+
+            return new DependencyContext(
+                new TargetInfo(framework, runtime, runtimeSignature, isPortable),
+                compilationOptions,
+                CreateLibraries(compileTarget?.Libraries, false, libraryStubs).Cast<CompilationLibrary>().ToArray(),
+                CreateLibraries(runtimeTarget.Libraries, true, libraryStubs).Cast<RuntimeLibrary>().ToArray(),
+                runtimeFallbacks ?? Enumerable.Empty<RuntimeFallbacks>());
+        }
+
+        private Target SelectRuntimeTarget(List<Target> targets, string runtimeTargetName)
```
bartonjs: If the two different readers need to be maintained the code that's just copied between them should be in a shared file (essentially something like DependencyContextJsonReader.cs, DependencyContextJsonReader.SystemTextJson.cs, DependencyContextJsonReader.NewtonSoftJson.cs; where only the parts that are touching the JSON are in the satellite partials)

bartonjs: Though, since it looks like you added extension methods to make for a fewer-lines-of-code-change, I wonder if the better answer for this project isn't to make project local reader type which just shims across the two other readers (it could be that there's a tricky thing somewhere, but flipping tabs shows that these are remarkably similar files)

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+
+            if (!_stringPool.TryGetValue(s, out string result))
+            {
+                _stringPool[s] = s;
+                result = s;
+            }
+            return result;
+        }
+
+        private class Target
```
eerhardt: Let's not duplicate all these classes/structs.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.DependencyModel
+{
+    public class DependencyContextJsonReader : IDependencyContextReader
```
eerhardt: In general, it feels like there is too much duplication in this file. Is there a way to factor this out so the entire DependencyContextJsonReader isn't duplicated? Similar to the way we have "pal" layers in corefx - have a common DependencyContextJsonReader.cs file that contains all the common code, and then partial classes for all the "TFM-specific" code.

src/managed/Microsoft.Extensions.DependencyModel/JsonTextReaderExtensions.Netstandard2_0.cs
```diff
@@ -0,0 +1,139 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.DependencyModel
+{
+    internal static class JsonTextReaderExtensions
```
eerhardt: It feels like all these methods are things that most consumers would want... and it is obviously a non-trivial amount of code.

bartonjs: @eerhardt A lot of the code in these extension methods is just about changing the exception type,  and combining the move and get operations.  If InvalidOperationException were acceptable then `ReadAsString()` is just

```C#
reader.Next();
return reader.GetStringValue();
```

The return value of Next is important when using the reader in async/chunked mode; and an exception model is what it is...

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -6,29 +6,41 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
+    <Compile Remove="DependencyContextJsonReader.Netstandard2_0.cs" />
+    <Compile Remove="DependencyContextWriter.Netstandard2_0.cs" />
+    <Compile Remove="JsonTextReaderExtensions.Netstandard2_0.cs" />
+    <Compile Remove="StreamBufferWriter.Netstandard2_0.cs" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
bartonjs: The conditions around ns20 seem odd...

ns20:
* Pull in the source package for JSON
* Use the old .cs files

!ns20:
* Pull in NewtonSoft.Json
* Use the new code

I'm sorta surprised that anything builds.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Netstandard2_0.cs
joshfree: probably worth a comment in the code then until this moves to the JsonDocument (writer)

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
joshfree: This is something that @JeremyKuhne might have looked into.
/cc @danmosemsft

src/managed/Microsoft.Extensions.DependencyModel/JsonTextReaderExtensions.Netstandard2_0.cs
ahsonkhan: > IMO you should only be using public APIs from the reader.

Agreed.

However, I don't really understand the need for this method. Why not rely on the exceptions thrown by the JsonReader itself? But, you are right that we may want to expose these publicly.

cc @eerhardt

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Netstandard2_0.cs
ahsonkhan: I don't think we would move this to JsonDocument (as it is written today) especially since this is writing (not reading). I was only suggesting that we needed this `bool wroteObjectStart` due to the ordering of the writes being important (while the ordering of adding to `JObject` is not really important).

src/managed/Microsoft.Extensions.DependencyModel/JsonTextReaderExtensions.Netstandard2_0.cs
eerhardt: Looks like the code was that way since it was migrated from using `JObject` to simply the `JsonTextReader`.

https://github.com/dotnet/cli/commit/a02d002352517994a53b5e232c57db1c9724599a

I assume it is so we can give "good" exception messages about exactly where we expect a new object/array/etc. What is the exception message you get when you rely on the exceptions thrown by the JsonReader itself?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,841 @@
...
+        {
+            var jsonReader = new Utf8JsonReader(jsonData, isFinalBlock: true, state: default);
+
+            jsonReader.ReadStartObject();
+
+            var runtime = string.Empty;
+            var framework = string.Empty;
+            var isPortable = true;
+            string runtimeTargetName = null;
+            string runtimeSignature = null;
```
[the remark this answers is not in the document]
ahsonkhan: Fair, though I wanted to avoid too many unnecessary changes to make the diff cleaner to review. I can change it in both places if necessary.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: Didn't introduce it, so would have to change it in both places. Will fix.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
ahsonkhan: > I'm sorta surprised that anything builds.

The source files are auto-included, so the conditions are reversed (which is why it builds fine). Note, we are removing not including.

ns20:
* Pull in the source package for JSON
* **Remove** ~Use~ the old .cs files

!ns20:
* Pull in NewtonSoft.Json
* **Remove** ~Use~ the new code

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: Here is the diff between the two implementations: https://www.diffchecker.com/TBfQTDQ0
Essentially, we have the following key differences:
- pass by ref explicitly (with the ref keyword) vs. pass by ref implicitly (passing a class)
- `(string)reader.Value` vs. `reader.GetString()`

We could consider doing if/defs around the function signatures OR pass the `JsonTextReader` class by ref explicitly (even though that is redundant). And then we could have a GetStringValue helper (similar to `IsTokenTypeProperty`). That would get rid of most of the diffs. Thoughts on this approach?

How could we shim across the readers?

src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonReaderTest.cs
```diff
@@ -30,7 +30,7 @@ public void ReadsRuntimeTargetInfo()
         ""signature"":""target-signature""
     },
     ""targets"": {
-        "".NETCoreApp,Version=v1.0/osx.10.10-x64"": {},
+        "".NETCoreApp,Version=v1.0/osx.10.10-x64"": {}
```
ahsonkhan: **Question:** Do we expect the json payloads to contain comments? If so, we need to use a more forgiving option within the reader that allows comments.

src/managed/Microsoft.Extensions.DependencyModel/JsonTextReaderExtensions.Netstandard2_0.cs
ahsonkhan: > If position info isn't already public then this is a good argument for providing it. If you need it here then other users of the reader will also likely use it.

https://github.com/dotnet/corefx/issues/34768

ahsonkhan:
Any other PR feedback or questions/concerns around the API?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            ArraySegment<byte> drained = ReadToEnd(stream);
+
+            DependencyContext dependencyContext;
+            try
+            {
+                dependencyContext = ReadCore(drained.AsSpan(0, drained.Count));
```
stephentoub: Won't the parameterless version of AsSpan() use the array segment's offset and count implicitly?  That seems better, not only because it's simpler, but also because this current code is assuming that the returned array segment begins at offset 0, which if ReadToEnd changed, it might not.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+            }
+            catch
+            {
+                // Holds document content, clear it before returning it.
+                drained.AsSpan(0, drained.Count).Clear();
```
stephentoub: Ditto

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+                ArrayPool<byte>.Shared.Return(drained.Array);
+                throw;
+            }
+            finally
+            {
+                // Holds document content, clear it before returning it.
+                drained.AsSpan(0, drained.Count).Clear();
```
stephentoub: Ditto; will stop commenting on that.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+                ArrayPool<byte>.Shared.Return(drained.Array);
```
stephentoub: I don't understand the catch and finally block here... won't this end up clearing, but much worse, returning the same array twice in the case of an exception?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+            }
+            return dependencyContext;
```
stephentoub: Why the complication of defining the dependencyContext outside of the try/finally?  Why not just:
``C#
ArraySegment<byte> drained = ReadToEnd(stream);
try
{
    return ReadCore(drained.AsSpan());
}
finally
{
    drained.AsSpan().Clear();
    ArrayPool<byte>.Shared.Return(drained.Array);
}
```
?

File: src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs [hunk 22]
```diff
@@ -0,0 +1,628 @@
...
+        }
+
+        private static bool IsTokenTypeProperty(JsonTokenType tokenType)
+            => tokenType == JsonTokenType.PropertyName;
+
+        private DependencyContext ReadCore(ReadOnlySpan<byte> jsonData)
```
stephentoub: Can this be static?  It's hard for me to see whether this is using any instance state.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+            string runtimeSignature = null;
+
+            CompilationOptions compilationOptions = null;
+            List<Target> targets = null;
+            Dictionary<string, LibraryStub> libraryStubs = null;
+            List<RuntimeFallbacks> runtimeFallbacks = null;
+
+            while (reader.Read() && IsTokenTypeProperty(reader.TokenType))
+            {
+                switch (reader.GetString())
```
stephentoub: We need to materialize these strings?  We can't get back the span and do a comparison based on spans?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+            {
+                compilationOptions = CompilationOptions.Default;
+            }
+
+            Target runtimeTarget = SelectRuntimeTarget(targets, runtimeTargetName);
+            runtimeTargetName = runtimeTarget?.Name;
+
+            if (runtimeTargetName != null)
+            {
+                var separatorIndex = runtimeTargetName.IndexOf(DependencyContextStrings.VersionSeparator);
```
stephentoub: Nit: var => int

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+                }
+                else
+                {
+                    framework = runtimeTargetName;
+                }
+            }
+
+            Target compileTarget = null;
+
+            var ridlessTarget = targets.FirstOrDefault(t => !IsRuntimeTarget(t.Name));
```
stephentoub: Nit: var => Target (I don't know what core-setup's coding guidelines are, but if it's the same as corefx, then please review the use of such var's, and I'll stop commenting on these)

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
bartonjs: Yep, I totally missed that it was Remove instead of Include.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+        }
+
+        private void ReadRuntimeTarget(ref Utf8JsonReader reader, out string runtimeTargetName, out string runtimeSignature)
+        {
+            runtimeTargetName = null;
+            runtimeSignature = null;
+
+            reader.ReadStartObject();
+
+            while (reader.TryReadStringProperty(out string propertyName, out string propertyValue))
```
stephentoub: Same question about extracting strings and whether spans could be used first, then materialized into strings only when needed.  Of course, maybe perf isn't a big factor here, in which case clarity wins.  I'll stop asking about this.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+
+                do
+                {
+                    if (expectedLength == 0 && rented.Length == written)
+                    {
+                        byte[] toReturn = rented;
+                        rented = ArrayPool<byte>.Shared.Rent(checked(toReturn.Length * 2));
+                        Buffer.BlockCopy(toReturn, 0, rented, 0, toReturn.Length);
+                        // Holds document content, clear it.
+                        ArrayPool<byte>.Shared.Return(toReturn, clearArray: true);
```
stephentoub: What happens if the stream grows after its Length was initially checked in the CanSeek section above?  Won't we start calling stream.Read with invalid arguments?  For robustness, you might want to just treat the initial sizing as a best guess rather than guaranteed.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+                    }
+
+                    lastRead = stream.Read(rented, written, rented.Length - written);
+                    written += lastRead;
+                } while (lastRead > 0);
+
+                return new ArraySegment<byte>(rented, 0, written);
```
stephentoub: Why are we reading the whole stream contents into an array and then parsing that, rather than reading and parsing as we go?  Is it because we expect these inputs to be rather small?

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            _stream = stream ?? throw new ArgumentNullException(nameof(stream));
+
+            _rentedBuffer = ArrayPool<byte>.Shared.Rent(bufferSize);
+        }
+
+        public void Advance(int count)
+        {
+            _stream.Write(_rentedBuffer, 0, count);
+            _rentedBuffer.AsSpan(0, count).Clear();
```
stephentoub: Why is this clear needed?

Also, I'd have expected it to be an error condition if you advanced further than you'd requested memory for.  That doesn't need to happen / be tracked?

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            if (sizeHint < 0)
+                throw new ArgumentException(nameof(sizeHint));
+
+            if (sizeHint == 0)
+            {
+                int newSize = _rentedBuffer.Length == 0 ? MinimumBufferSize : checked(_rentedBuffer.Length * 2);
+                byte[] temp = _rentedBuffer;
+                _rentedBuffer = ArrayPool<byte>.Shared.Rent(newSize);
+                temp.AsSpan().Clear();
```
stephentoub: What's the intended behavior if you call GetMemory without having called Advance?  This will end up discarding any such data, rather than copying it over from temp to _rentedBuffer.  Is that intentional?

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            else if (sizeHint > _rentedBuffer.Length)
+            {
+                var temp = _rentedBuffer;
+                _rentedBuffer = ArrayPool<byte>.Shared.Rent(sizeHint);
+                temp.AsSpan().Clear();
+                ArrayPool<byte>.Shared.Return(temp);
+            }
+            Debug.Assert(_rentedBuffer.Length > 0);
+            return _rentedBuffer;
```
stephentoub: Rather than duplicating the whole method, can't this just be:
```C#
public Span<byte> GetSpan(int sizeHint = 0) => GetMemory(sizeHint).Span;
```
?  Is the extra overhead of going through the `Memory<T>` prohibitive?  And if it is, could you just refactor all of this into a private helper that returns the array, and then wrap GetMemory and GetSpan around that?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
```diff
@@ -0,0 +1,628 @@
...
+                switch (reader.GetString())
+                {
+                    case DependencyContextStrings.RuntimeTargetPropertyName:
+                        ReadRuntimeTarget(ref reader, out runtimeTargetName, out runtimeSignature);
+                        break;
+                    case DependencyContextStrings.CompilationOptionsPropertName:
+                        compilationOptions = ReadCompilationOptions(ref reader);
+                        break;
+                    case DependencyContextStrings.TargetsPropertyName:
+                        targets = ReadTargets(ref reader);
```
bartonjs: I'm still a bit concerned about the amount of code that would need to be updated in two places if we need to keep supporting two different readers.

If it were me, I'd revert DependencyContextJsonReader, then add a new `internal ref struct UnifiedJsonReader` which provides the interface that this type wants (`ReadStartObject()`, et cetera) and has one implementation using Utf8JsonReader and another using the NewtonSoft one.  (And then update all the methods to take the `ref UnifiedJsonReader`, of course)

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: I realize that this code can be optimized for performance/allocations but the intent was to keep the diff as minimal as possible as part of this port to highlight what is absolutely necessary.

I will file an issue to make optimizations and avoid materializing strings where possible, and defer performance or cosmetic changes to the future.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: > Of course, maybe perf isn't a big factor here, in which case clarity wins. I'll stop asking about this.

Based on the Json.NET usage within this library, I got the sense that perf isn't a big factor. We can still optimize the implementation for netstandard2.0 given the new library though.

https://github.com/dotnet/core-setup/issues/5034

Resolving as out of scope for this PR.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: I am borrowing this helper from corefx. Should it be fixed there as well?
https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L176

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: I am not sure how much we should invest in this `UnifiedJsonReader` approach if we can drop the other TFMs and the dependency on Newtonsoft.Json all together.

Do you think I should revert and unify even if we will remove one of the impls soon? In other words, it is acceptable to support two readers in the short term while we evaluate the TFMs required and then consolidate only if we decide to keep the `JsonTextReader` impl as well?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: > For robustness, you might want to just treat the initial sizing as a best guess rather than guaranteed.

I don't follow what you mean here. Can you elaborate what you recommend we do with the initial size (by which I am assuming you mean `expectedLength`)?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: > Why are we reading the whole stream contents into an array and then parsing that, rather than reading and parsing as we go?

Mainly because reading partially from the stream and passing that reader along the various methods is quite complex since the `Utf8JsonReader` doesn't have built-in support for streams. I will see if I can make it work but we need to provide APIs to support the stream scenario.

> Is it because we expect these inputs to be rather small?

I am not sure how big the input can get. Generally deps.json files are not that large (i.e. < 1MB). @eerhardt, what is the typical (and largest) size you expect as input to `DependencyContextJsonReader`?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
stephentoub: I mean right now this code will blow up if the stream is seekable but its length changes over the course of the method.  I'm suggesting instead that you still presize the array based on the initially read Length, but then don't do anything differently in the rest of the method than for the !CanSeek case.

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
ahsonkhan: You are right, the clear isn't needed, and yes, I need to do the input validation here.

I spent some time discussing with @davidfowl around a stream<->IBufferWriter impl, and the conclusion was that we shouldn't have such an adapter (and definitely not one that writes to the stream synchronously on calls to Advance).

Instead, we should buffer the whole output in memory (in some re-sizable array-backed IBufferWriter) and then copy the data to the stream asynchronously (using a `CopyToAsync(Stream stream)` method).

Something like the `ArrayFormatter`: https://github.com/dotnet/corefxlab/blob/87990483caf421669efb1f8f63f333b065d553b7/src/System.Text.Formatting/System/Text/Formatting/Formatters/ArrayFormatter.cs#L43

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
ahsonkhan: > This will end up discarding any such data, rather than copying it over from temp to _rentedBuffer. Is that intentional?

If I understand the contract correctly, then yes, this is intentional. The expectation is that the caller can start writing at index 0 of the returned `Memory<byte>`. So, if you didn't Advance past the previously written data, then that data will be lost.

cc @pakrym

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
stephentoub: > You are suggesting I remove the resize logic all together.

No, the opposite, that you always use the resize logic even if CanSeek == true... basically remove the "expectedLength == 0 &&" from the if check.  In the CanSeek case, you'd also want to ask for an array that was Length+1 rather than Length, so that you'd never hit this resize in the common case.

stephentoub: > Can the stream length shrink as well?

Yes:
```C#
using System;
using System.IO;

class Program
{
    static void Main()
    {
        using (Stream theStream = new FileStream("tmp.dat", FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite))
        {
            Console.WriteLine(theStream.Length);
            ChangeLength("tmp.dat", 10);
            Console.WriteLine(theStream.Length);
            ChangeLength("tmp.dat", 5);
            Console.WriteLine(theStream.Length);
        }
    }

    static void ChangeLength(string path, int length)
    {
        using (Stream tmp = new FileStream(path, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite))
            tmp.SetLength(length);
    }
}
```
outputs:
```
0
10
5
```

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: No, it can't be static. Many of the methods it calls reference `Pool(string s)` which uses the `_stringPool`.
```C#
private readonly IDictionary<string, string> _stringPool = new Dictionary<string, string>();
```

A few of the other methods can be made static since they don't call the Pool instance method. Should I update those? For example:
```C#
private static List<string> ReadPropertyNames(ref Utf8JsonReader reader);
private static List<RuntimeFallbacks> ReadRuntimes(ref Utf8JsonReader reader);
```

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
ahsonkhan: Note that this is a breaking change. Previously we exposed Json.NET type via this API:
```C#
public IEnumerable<Dependency> ReadTargetLibraryDependencies(JsonTextReader reader);
```

It doesn't look like anyone is calling this API though:
- https://github.com/search?l=C%23&q=ReadTargetLibraryDependencies&type=Code
- https://apisof.net/catalog/Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.ReadTargetLibraryDependencies(JsonTextReader)

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
bartonjs: The check against zero was to avoid the seekable stream resizing when it didn't change.  Doing `+ 1` and letting the resize run if it hits seems fine.

(The hint could also subtract off Position if it wanted more accuracy :smile:)

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
pakrym: Yeah, without advancing the data is lost.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
bartonjs: > Do you think I should revert and unify even if we will remove one of the impls soon?

Yep.  If the older refs get dropped you just stop building the Newtonsoft-using version.  The shim type performs the same role as the current extension methods.

And if we stay in limbo a long time (or decide we need to keep building the older TFMs), it keeps the limbo state cleaner.

src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonReaderTest.cs
ahsonkhan: The existing implementation does not support comments inside the payload.

If comments exist, the `DependencyContextJsonReader` throws `FormatException` (since `JsonTextReader` allows comments by default). If the intention was to allow & ignore comments, the `DependencyContextJsonReader` needs to be more robust against the comment token type and ignore it, explicitly.

If we use the new `Utf8JsonReader`, we could either disallow, skip, or allow comments.

1. **[partially breaking change - different exception type]** If we disallow comments (default behavior of the reader), then we throw `JsonReaderException` whenever we see a comment (since it is considered invalid JSON as per spec).
2. **[partially breaking change - what was invalid before is supported now]** If we skip comments, then the behavior is now different (since comments become supported, and ignored).
3. **[non-breaking change]** If we allow comments, then we continue to throw `FormatException`, just like the current implementation.

If we want to match the current behavior (i.e. #3), we have to allow comments to pass through even though that gets detected by `DependencyContextJsonReader` as an error and a `FormatException` is thrown. I much rather have option 1 or 2 where we either support skipping comments (and successfully read the deps file) OR we throw `JsonReaderException` explicitly.

@eerhardt what option would you prefer? I have gone with option 3 in the meantime and added a test encapsulating the current behavior.

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Netstandard2_0.cs
eerhardt: It appears this was an oversight when it was [first written](https://github.com/dotnet/cli/pull/3972/files#diff-ceddfbfdc43e20bec4e1c154cdddd637R362). Making it `private` is a good change IMO.

src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonReaderTest.cs
eerhardt: In this case, I think I'd be fine with changing the exception type, as long as the exception still conveys a decent error message (what is the problem, which line/column, etc). I'd find it hard to believe someone is relying on this behavior, and if they are that they'd have a hard time moving to the new behavior.

Honestly, I'd also be fine with ignoring the comments as well.

src/managed/Microsoft.Extensions.DependencyModel/ArrayBufferWriter.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (stream == null)
+                throw new ArgumentNullException(nameof(stream));
+
+            await stream.WriteAsync(_rentedBuffer, 0, _written, cancellationToken).ConfigureAwait(false);
+
+            _rentedBuffer.AsSpan(0, _written).Clear();
+            _written = 0;
+        }
+
+        internal void CopyTo(Stream stream)
```
eerhardt: (nit) why have some methods `internal` and some `public` on an `internal` class?

ahsonkhan: This particular API was made internal mainly because we likely won't have this synchronous API public on ArrayBufferWriter if we decide to ship it from corefx (only the async one would be public). All others should be public or private. I don't mind making it public if there is a need for it.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -5,30 +5,42 @@
...
-      </PropertyGroup>
-    </Otherwise>
-  </Choose>
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
-    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
eerhardt: IMO following the existing `Choose-When` pattern cleans this up. That way we aren't spaghetti'ing between `!= netstandard2.0` and `== netstandard2.0`, and we can easily see what happens for `ns20` and otherwise.

src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonReaderTest.cs
```diff
@@ -30,7 +28,7 @@ public void ReadsRuntimeTargetInfo()
...
```
eerhardt: What's really weird is that using DependencyModel `v2.1`, I tried adding trailing commas and I was getting exceptions being thrown from the current implementation. I'm surprised that these tests are passing.

src/managed/Microsoft.Extensions.DependencyModel/ArrayBufferWriter.cs
```diff
@@ -0,0 +1,133 @@
...
+        public ArrayBufferWriter(int initialCapacity = MinimumBufferSize)
+        {
+            if (initialCapacity <= 0)
+                throw new ArgumentException(nameof(initialCapacity));
+
+            _rentedBuffer = ArrayPool<byte>.Shared.Rent(initialCapacity);
+            _written = 0;
+        }
+
+        public async Task CopyToAsync(Stream stream, CancellationToken cancellationToken = default)
```
eerhardt: Would it make sense to delete the methods that aren't being called?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Utf8JsonReader.cs
```diff
@@ -0,0 +1,94 @@
...
+    public partial class DependencyContextJsonReader : IDependencyContextReader
+    {
+        public DependencyContext Read(Stream stream)
+        {
+            if (stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            ArraySegment<byte> drained = ReadToEnd(stream);
```
eerhardt: I find `drained` an odd name for this variable. Would `buffer` be better?

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
```diff
@@ -0,0 +1,166 @@
...
+                return defaultValue;
+            }
+            if (_reader.TokenType != JsonTokenType.True && _reader.TokenType != JsonTokenType.False)
+            {
+                throw CreateUnexpectedException(ref _reader, "a JSON true or false literal token");
+            }
+            return _reader.GetBoolean();
+        }
+
+        private static Exception CreateUnexpectedException(ref Utf8JsonReader reader, string expected)
```
eerhardt: (minor) It'd be nice to have to not have this method duplicated.

ahsonkhan: To keep the current custom exception message, it has to be. JsonTextReader returns the path.
I can extract out the creation of the exception message, but that's about it.

src/managed/Microsoft.Extensions.DependencyModel/StreamBufferWriter.Netstandard2_0.cs
ahsonkhan: https://github.com/dotnet/corefx/issues/34894

ahsonkhan:
Build failures for Arm/Arm64 on Windows are unrelated: https://dev.azure.com/dnceng/public/_build/results?buildId=81361
```text
2019-01-28T10:44:06.7388911Z   Calling "D:\a\1\s\src\corehost\Windows\gen-buildsys-win.bat D:\a\1\s\ "15 2017" arm b41cae7e8c4dee8b3e2bda9fec2f17b5d29c5fbf 3.0.0-preview-27328-2 3.0.0-preview-27328-2 3.0.0-preview-27328-2 3.0.0-preview-27328-2"
2019-01-28T10:44:06.7389019Z   "Computed RID for native build is win-arm"
2019-01-28T10:44:06.7389139Z   "C:\Program Files\CMake\bin\cmake.exe" D:\a\1\s\src\corehost\Windows\.. "-DCMAKE_SYSTEM_VERSION=10.0" "-DCLI_CMAKE_RUNTIME_ID:STRING=win-arm" "-DCLI_CMAKE_HOST_VER:STRING=3.0.0-preview-27328-2" "-DCLI_CMAKE_APPHOST_VER:STRING=3.0.0-preview-27328-2" "-DCLI_CMAKE_HOST_FXR_VER:STRING=3.0.0-preview-27328-2" "-DCLI_CMAKE_HOST_POLICY_VER:STRING=3.0.0-preview-27328-2" "-DCLI_CMAKE_PKG_RID:STRING=win-arm" "-DCLI_CMAKE_COMMIT_HASH:STRING=b41cae7e8c4dee8b3e2bda9fec2f17b5d29c5fbf" "-DCLI_CMAKE_PLATFORM_ARCH_ARM=1" "-DCMAKE_INSTALL_PREFIX=D:/a/1/s/src/corehost/../../bin/win-arm.Debug/corehost" "-DCLI_CMAKE_RESOURCE_DIR:STRING=D:\a\1\s\src\corehost\..\..\bin\obj\win-arm.Debug\hostResourceFiles" -G "Visual Studio 15 2017 ARM" 
2019-01-28T10:44:11.3988591Z   -- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.
2019-01-28T10:44:25.0285854Z   -- The C compiler identification is unknown
2019-01-28T10:44:29.2688030Z   -- The CXX compiler identification is unknown
2019-01-28T10:44:29.3789095Z   CMake Error in CMakeLists.txt:
2019-01-28T10:44:29.3789556Z     No CMAKE_C_COMPILER could be found.
2019-01-28T10:44:29.3789731Z   
2019-01-28T10:44:29.3789913Z   
2019-01-28T10:44:29.3790220Z   
2019-01-28T10:44:29.3790372Z   CMake Error in CMakeLists.txt:
2019-01-28T10:44:29.3790550Z     No CMAKE_CXX_COMPILER could be found.
2019-01-28T10:44:29.3790697Z   
2019-01-28T10:44:29.3790843Z   
2019-01-28T10:44:29.3791011Z   
2019-01-28T10:44:29.3900213Z   -- Configuring incomplete, errors occurred!
2019-01-28T10:44:29.5281405Z   See also "D:/a/1/s/bin/obj/win-arm.Debug/corehost/CMakeFiles/CMakeOutput.log".
2019-01-28T10:44:29.5290897Z   See also "D:/a/1/s/bin/obj/win-arm.Debug/corehost/CMakeFiles/CMakeError.log".
2019-01-28T10:44:29.5359326Z   Failed to generate native component build project!
2019-01-28T10:44:29.5359586Z D:\a\1\s\src\corehost\build.proj(79,5): error MSB3073: The command "D:\a\1\s\src\corehost\build.cmd Debug arm apphostver 3.0.0-preview-27328-2 hostver 3.0.0-preview-27328-2 fxrver 3.0.0-preview-27328-2 policyver 3.0.0-preview-27328-2 commit b41cae7e8c4dee8b3e2bda9fec2f17b5d29c5fbf rid win-arm portable" exited with code 1.
2019-01-28T10:44:29.5507299Z 
2019-01-28T10:44:29.5614879Z Build FAILED.
2019-01-28T10:44:29.5628283Z 
2019-01-28T10:44:29.5633185Z D:\a\1\s\src\corehost\build.proj(79,5): error MSB3073: The command "D:\a\1\s\src\corehost\build.cmd Debug arm apphostver 3.0.0-preview-27328-2 hostver 3.0.0-preview-27328-2 fxrver 3.0.0-preview-27328-2 policyver 3.0.0-preview-27328-2 commit b41cae7e8c4dee8b3e2bda9fec2f17b5d29c5fbf rid win-arm portable" exited with code 1.
2019-01-28T10:44:29.5638856Z     0 Warning(s)
2019-01-28T10:44:29.5642108Z     1 Error(s)
```

cc @dotnet/dnceng

MattGal:
@ahsonkhan I looked into this in https://github.com/dotnet/core-setup/issues/5052, see notes there.

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
eerhardt: OK, this makes sense. Thanks!

---
In reply to: [251344200](https://github.com/dotnet/core-setup/pull/5009#discussion_r251344200) [](ancestors = 251344200)

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -5,30 +5,52 @@
...
-        <NewtonsoftJsonPackageVersion>11.0.1</NewtonsoftJsonPackageVersion>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+        <!-- Suppress warnings for unnecessary CLSCompliance attributes -->
+        <NoWarn>3021</NoWarn>
       </PropertyGroup>
+      <ItemGroup>
+        <PackageReference Include="Microsoft.Bcl.Json.Sources" Version="4.6.0-preview.19072.2">
+          <PrivateAssets>All</PrivateAssets>
+        </PackageReference>
+        <PackageReference Include="System.Memory" Version="4.5.2" />
```
eerhardt: Do we really need `4.5.2`?

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -5,30 +5,52 @@
...
     <When Condition="'$(TargetFramework)' == 'netstandard2.0'">
+      <!--These properties are necessary to build the source package-->
       <PropertyGroup>
-        <NewtonsoftJsonPackageVersion>11.0.1</NewtonsoftJsonPackageVersion>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+        <!-- Suppress warnings for unnecessary CLSCompliance attributes -->
+        <NoWarn>3021</NoWarn>
       </PropertyGroup>
+      <ItemGroup>
+        <PackageReference Include="Microsoft.Bcl.Json.Sources" Version="4.6.0-preview.19072.2">
```
eerhardt: We should be putting the version numbers in [https://github.com/dotnet/core-setup/blob/master/eng/Versions.props](https://github.com/dotnet/core-setup/blob/master/eng/Versions.props)

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
ahsonkhan: As opposed to what? 4.5.2 is the latest version available.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
ahsonkhan: I followed the format for existing package references within this csproj which included version numbers here. Should they move to Versions.props too?

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
eerhardt: Right, but is the latest version the right thing to choose? Typically libraries don't update their dependencies just because there is a new version out. Doing so causes anyone who references you to use the latest version, which may not be the right thing.

Is there a specific reason you choose `4.5.2`?  or was it just because it was the latest version?

---
In reply to: [251573648](https://github.com/dotnet/core-setup/pull/5009#discussion_r251573648) [](ancestors = 251573648)

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
eerhardt: No, the existing package reference version (Newtonsoft) should remain here, since we never want it updated or shared with other libraries.

Putting this version in Versions.props will get it automatically updated through normal dependency flow PRs.

---
In reply to: [251574076](https://github.com/dotnet/core-setup/pull/5009#discussion_r251574076) [](ancestors = 251574076)

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
ahsonkhan: > Typically libraries don't update their dependencies just because there is a new version out. Doing so causes anyone who references you to use the latest version, which may not be the right thing.

Good point. I didn't have a specific need for the latest. I believe I can safely downgrade to `4.5.0`.

src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonReaderTest.cs
```diff
@@ -39,6 +38,27 @@ public void ReadsRuntimeTargetInfo()
             context.Target.RuntimeSignature.Should().Be("target-signature");
         }
 
+        [Fact]
+        public void ReadsRuntimeTargetInfoWithCommentsIsInvalid()
+        {
+            var exception = Assert.Throws<JsonReaderException>(() => Read(
```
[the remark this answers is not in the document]
ahsonkhan: The tests target `netcoreapp2.0` so they don't run on non-netstandard2.0. We can add older TFMs to test the Newtonsoft code-paths.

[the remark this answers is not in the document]
ahsonkhan: Based on offline discussion and https://github.com/dotnet/core-setup/issues/5034, it's not worth adding TFMs for this since we will remove the non-netstandard2.0 code paths.

PR #5010, opened 2019-01-19
[WIP] Update core-setup local tools to use S.T.Json for netstandard2.0 instead

ahsonkhan:
Initial attempt to port tools-local to use the `System.Text.Json` source package for TFM netstandard2.0. **Note:** There are some API changes going through so this is still WIP. We have the JsonReader and JsonWriter available atm so only using the components that are ready.

The System.Text.Json binary is in-box on .NET Core 3.0.

I added TODOs to use the `JsonDocument` and `Deserializer` once they are available. Using the `Utf8Reader` in these scenarios is less than ideal and requires a lot of code change.

**This PR is mainly to experiment to gather feedback and showcase the capabilities. Hence, marked as no merge.**

We can't remove the reference to Newtonsoft since the tools are targeting net46 on Unix.

**Is it possible to drop support for all but netstandard2.0?** Otherwise, we have to conditionally include the source package.

See https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/source_package/README.md

Related PR: https://github.com/dotnet/core-setup/pull/5009

cc @joshfree, @bartonjs, @steveharter, @eerhardt, @KrzysztofCwalina, @JamesNK, @glennc, @ericstj


tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+                }
+
+                // Open the Config Json and read the values into the model
+                TextReader projectFileReader = File.OpenText(ConfigJsonFile);
+                if (projectFileReader != null)
+                {
+                    string jsonFileText = projectFileReader.ReadToEnd();
+
+                    // TODO: Replace the use of the Utf8JsonReader with the Deserializer once it is available.
+                    ConfigJson configJson = GetConfigJson(jsonFileText);
```
ahsonkhan: This isn't a great way to deserialize a POCO object. I used the JsonReader just as an experiment.

tools-local/tasks/ChangeEntryPointLibraryName.Netstandard2_0.cs
```diff
@@ -0,0 +1,245 @@
...
+                    case JsonTokenType.EndObject:
+                        writer.WriteEndObject();
+                        break;
+                }
+            }
+        }
+
+        private void ReadWritePropertyName(string propertyName, ref Utf8JsonReader reader, ref Utf8JsonWriter writer)
+        {
+            if (propertyName == "targets")
```
ahsonkhan: We need to use the `JsonDocument` here. The order in which the "targets" property and the "libraries" property show up could affect the end results here if we use the Reader and Writer. This relies on "targets" coming up before "libraries".

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+        public string FPMParameters { get; set; }
+
+        public override bool Execute()
+        {
+            try
+            {
+                if (!File.Exists(ConfigJsonFile))
+                {
+                    throw new FileNotFoundException($"Expected file {ConfigJsonFile} was not found.");
+                }
```
stephentoub: Won't File.OpenText throw the same exception, and without the inherent race condition?  You're just trying to provide a better exception message in some cases?

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+
+                // Open the Config Json and read the values into the model
+                TextReader projectFileReader = File.OpenText(ConfigJsonFile);
+                if (projectFileReader != null)
```
stephentoub: This won't ever be null.

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+        {
+            try
+            {
+                if (!File.Exists(ConfigJsonFile))
+                {
+                    throw new FileNotFoundException($"Expected file {ConfigJsonFile} was not found.");
+                }
+
+                // Open the Config Json and read the values into the model
+                TextReader projectFileReader = File.OpenText(ConfigJsonFile);
```
stephentoub: This should be disposed of when you're done with it; otherwise you're leaving the file open until it's eventually closed by a finalizer.

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
stephentoub: So we're opening the file, reading all of its contents into a string (which will involve converting from UTF8 to UTF16 if it's saved as UTF8), and then turning around and encoding that string into a UTF8 byte[]?  Isn't the point of having a UTF8-based JSON reader to avoid that?

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+        {
+            byte[] utf8Json = Encoding.UTF8.GetBytes(jsonFileText);
+            var configJson = new ConfigJson();
+
+            var reader = new Utf8JsonReader(utf8Json, isFinalBlock: true, state: default);
+
+            while (reader.Read() && reader.TokenType == JsonTokenType.PropertyName)
+            {
+                string propertyName = reader.GetStringValue().ToLower();
+                if (propertyName == Maintainer_Name_String)
```
stephentoub: Does perf not matter here?  Or could we instead compare span to span rather than materializing the string?  And if we stick with strings, could the comparison be done as ordinal ignore case rather than lower casing the property name?

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+                string str = File.ReadAllText(changelogFile);
+                str = str.Replace("{PACKAGE_NAME}", configJson.Package_Name);
+                str = str.Replace("{PACKAGE_VERSION}", package_version);
+                str = str.Replace("{PACKAGE_REVISION}", configJson.Release.Package_Revision);
+                str = str.Replace("{URGENCY}", configJson.Release.Urgency);
+                str = str.Replace("{CHANGELOG_MESSAGE}", configJson.Release.Changelog_Message);
+                str = str.Replace("{MAINTAINER_NAME}", configJson.Maintainer_Name);
+                str = str.Replace("{MAINTAINER_EMAIL}", configJson.Maintainer_Email);
+                // The date format needs to be like Wed May 17 2017
+                str = str.Replace("{DATE}", DateTime.UtcNow.ToString("ddd MMM dd yyyy"));
```
stephentoub: Ok, we obviously don't care about perf in this code :)

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
bartonjs: I was going to ask the same question, but then I realized that the reader does encoding-sniffing.  So if someone saves the file as UTF-16 this would still successfully read it.

If we're willing to say "the file is UTF-8" (or, more likely, "the file is 7-bit ASCII") then this can be simplified a lot.

tools-local/tasks/ChangeEntryPointLibraryName.Netstandard2_0.cs
bartonjs: https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.Bcl.Json.Sources says JsonDocument is available now (4.6.0-preview.19073.1 is bigger than the previous version)

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
```diff
@@ -0,0 +1,581 @@
...
+        private static readonly string Full_Text_String = nameof(License.Full_Text).ToLower();
+
+        private static readonly string RpmDependency_Package_Name_String = nameof(RpmDependency.Package_Name).ToLower();
+        private static readonly string RpmDependency_Package_Version_String = nameof(RpmDependency.Package_Version).ToLower();
+    }
+
+    /// <summary>
+    /// Model classes for reading and storing the JSON. 
+    /// </summary>
+    public class ConfigJson
```
bartonjs: As with the other PR; there's a lot of duplication between this file and the alternate for this file.

eerhardt:
> This PR is mainly to experiment to gather feedback and showcase the capabilities. Hence, marked as no merge.

Is the intention to ever put this code in? I don't see a valid reason for complicating this code.

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
ahsonkhan: Yep, doesn't look like it from what I gather.

tools-local/tasks/BuildFPMToolPreReqs.Netstandard2_0.cs
ahsonkhan: Originally written this way: https://github.com/dotnet/core-setup/pull/2462/files#diff-0ebce9e49dc50f19e82e0d3ceada8568R43

dleeapho:
Closing as stale.

ahsonkhan:
Sounds good. Will re-visit later.

PR #5033, opened 2019-01-23
Disable UpdateVersionsRepo: unblock official build

dagood:
It's unknown at this point what's causing dotnet/versions publish to fail, and it only fails in the official build definition. Removing the target for now to get a build through. I will likely add this target back later with additional clean steps depending on the results of https://github.com/dotnet/core-setup/pull/5032.


dagood:
Skipping CI: file is only used in official builds, and this has been flipped back and forth before successfully.

PR #5042, opened 2019-01-24
Fix registry access on legacy windows

### LINKED ISSUE #5028: RRF_SUBKEY_WOW6432KEY used in SDK lookup implementation only supported on Windows 10
https://github.com/dotnet/core-setup/blob/2f528e5f158aa41789f7967cf8e8b868a8173d95/src/corehost/common/pal.windows.cpp#L243-L247

@vitek-karas @johnbeisner

Added in https://github.com/dotnet/core-setup/pull/4857

vitek-karas:
Fix registry access on legacy Windows

The bug was that `RRF_SUBKEY_WOW6432KEY` is only supported on Win10. On legacy Windows it has no effect and 64bit registry is accessed.

The fix is to use `RegOpenKeyEx` which takes the access right parameter which can specify `KEY_WOW64_32KEY`. This makes 64bit app access the 32bit registry and this one is supported on anything above Windows 2000 (excluded).

Fixes #5028

### DIFF
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -12,10 +12,6 @@
 #include <ShlObj.h>
 #include <ctime>
 
-#if !defined(RRF_SUBKEY_WOW6432KEY)
-#define RRF_SUBKEY_WOW6432KEY  0x00020000
-#endif
-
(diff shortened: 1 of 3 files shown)


pakrym:
LGTM

PR #5049, opened 2019-01-28
Fix the timestamp timezone in tracing in Windows

### LINKED ISSUE #5048: host tracing reports incorrect timezine in timestamps on Windows
Enabling tracing and running `dotnet.exe` on Windows prints out:
```
Tracing enabled @ Mon Jan 28 12:09:07 2019 Pacific Standard Time
```

Doing the same on Linux prints out:
```
Tracing enabled @ Mon Jan 28 12:09:13 2019 GMT
```

This is on the same machine with both Windows and Linux set to PST and showing the same local time of 4:09 AM.
Ignoring the format differences (abbreviated timezone versus verbose one), the timezone reported on Windows is wrong since it correctly shows the UTC time, but then shows PST instead of GMT/UTC timezone.

As far as I can tell this is a bug/difference in behavior in time/gmtime C++ functions used by the tracing.

vitek-karas:
It seems that printing out timezone information for a tm structure on Windows always shows the local timezone, even if the tm is returned by gmtime.
The simplest fix seems to be to simply print out GMT always, after all gmtime always returns GMT time.
This also makes the output match Linux formating as Windows otherwise uses verbose zone names, while Linux uses abbreviations.

Fixes #5048

### DIFF
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -50,7 +50,7 @@ pal::string_t pal::get_timestamp()
     std::time_t t = std::time(0);
     const std::size_t elems = 100;
     char_t buf[elems];
-    std::wcsftime(buf, elems, _X("%c %Z"), std::gmtime(&t));
+    std::wcsftime(buf, elems, _X("%c GMT"), std::gmtime(&t));
 
     return pal::string_t(buf);
 }


vitek-karas:
CI failures is #5052 everything else passes.

symphony2014:
@vitek-karas But, how can I disable this tracing ?

vitek-karas:
@symphony2014 the tracing should be off by default. You can enable it by setting COREHOST_TRACE=1 environment variable, but when it's set to 0 (or not set), the tracing should be disabled.

PR #5050, opened 2019-01-28
Fix lookup of coreclr.dll in .deps.json to not rely on relative paths

### LINKED ISSUE #4978: Hostpolicy can't find coreclr.dll if it's relative path has no directory separator in it
With the usage of runtime packs in dotnet/cli#10007 the relative path of some native components in these packs may be in the "current dir". It means that the relative path of `coreclr.dll` specified in the `.deps.json` will be `coreclr.dll` without any directory separator.

This breaks the logic which finds `coreclr.dll` in [this code](https://github.com/dotnet/core-setup/blob/ca431d0cefee116b5c4ca292961ab60ddd81611f/src/corehost/cli/deps_resolver.cpp#L576).

The code looks for `/coreclr.dll` to make sure that it doesn't match for example `mycoreclr.dll`. So we need to keep the piece with includes the directory separator in the check. The simplest way to fix this is to not look at the relative path from the asset but look in the full resolved path which is guaranteed to be absolute and thus must have a directory separator in front of the library name.

vitek-karas:
The existing code expects the .deps.json record for coreclr.dll (and clrjit.dll) to have at least one directory separator, since that was always the case up until now.
With recent changes to targeting packs, these records are just file names, and the above logic breaks.

Fix the logic by lookup at the resolved absolute path, which is guaranteed to have a directory separator before the library name (we need the separator to match whole file names).

Added a test which validates the new scenario.

Fixes #4978 .

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -573,13 +573,15 @@ void deps_resolver_t::init_known_entry_path(const deps_entry_t& entry, const pal
     {
         return;
     }
-    if (m_coreclr_path.empty() && ends_with(entry.asset.relative_path, _X("/") + pal::string_t(LIBCORECLR_NAME), false))
+
+    assert(pal::is_path_rooted(path));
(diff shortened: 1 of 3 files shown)


vitek-karas:
CI failure is #5052, everything else passes.

Issue #4987, opened 2019-01-16
Tizen Core-Setup nuget package version for master in dotnet/version


gbalykov:
## Steps to reproduce

We build nuget packages for CoreCLR/CoreFX/Core-Setup based on the updates of `dotnet/version` repo (specifically, https://github.com/dotnet/versions/tree/master/build-info/dotnet/core-setup/master for core-setup master). 

For Core-Setup the last update of `master` was on 29th of December with version `preview-27218-01`, however, both CoreCLR and CoreFX version have been updated regularly since then. 

Now, CoreFX `preview.19065.1` requires Core-Setup `preview-27218-01`, but we have not built it, as there were no changes in `dotnet/versions`.

Has the version update process changed?

cc @alpencolt

gbalykov:
cc @jkotas

dleeapho:
@gbalykov is this for Tizen build triggering?

gbalykov:
@dleeapho yes, forgot to add info

dleeapho:
@dagood per our discussion we can either reinstate the dotnet/version info or look at how maestro++/BAR infrastructure can provide this info. /cc @markwilkie

dagood:
Filed https://github.com/dotnet/arcade/issues/1818 to see how this will work with BAR. I'm taking this issue to turn on dotnet/versions publishing again (if there isn't a quick and easy answer to how to use BAR).

gbalykov:
@dagood `Latest.txt` doesn't seem to be updated now for core-setup master

dagood:
I worked around the issue I was seeing with PR https://github.com/dotnet/core-setup/pull/5069, and we're starting to get dotnet/versions commits again: https://github.com/dotnet/versions/commits/master/build-info/dotnet/core-setup/master 🎉 

@gbalykov Can you check if those commits are good and will let you get your infra running again?

gbalykov:
@dagood thank you! I see that versions for core-setup are updated now. Same issue seems to be with CoreCLR, as `https://github.com/dotnet/versions/tree/master/build-info/dotnet/coreclr/master` are not updated. Is the reason the same as for core-setup?

dagood:
Sweet, thanks for checking.

Yeah, CoreCLR also hasn't been publishing to dotnet/versions in their yaml builds. Filed https://github.com/dotnet/coreclr/issues/22398 to implement that.

PR #5069, opened 2019-02-01
Add dotnet/versions publish as hosted job

dagood:
If dotnet/versions publish is enabled in the main finalize/publish job, it fails because VersionTools is an old version. The reason for the old VersionTools is unknown. To narrow it down or work around it, use a hosted agent to ensure machine state isn't a factor, such as previous builds' tooling carrying over somehow.

For https://github.com/dotnet/core-setup/issues/4987. This will likely work around the issue. If it still repros in the official build, it will show that agent environment isn't a factor which would be extremely strange.


dagood:
CI failures are known, unrelated: https://github.com/dotnet/core-setup/issues/5052.

Issue #5108, opened 2019-02-07
Managed API is missing documentation (xmldocs)


natemcmaster:
Most of the API in [src/managed/](https://github.com/dotnet/core-setup/tree/master/src/managed) ships in a NuGet package, but is missing API documentation. It would be useful to provide API docs for the public API in

* Microsoft.DotNet.PlatformAbstractions
* Microsoft.Extensions.DependencyModel

eerhardt:
Appears to be a duplicate of https://github.com/dotnet/core-setup/issues/4372. Closing in favor of that issue instead.

PR #5109, opened 2019-02-07
Generate xml docs for managed API

natemcmaster:
Even if the file is mostly empty, I'd still like to have it generated. These files will eventually end up in the Microsoft.AspNetCore.App.Ref targeting pack.


natemcmaster:
This doesn't resolve https://github.com/dotnet/core-setup/issues/5108 -- the API is still undocumented. This is about making sure, at the very least, a doc .xml file exists. (cref https://github.com/aspnet/AspNetCore/pull/7355).

Issue #4772, opened 2018-11-19
Targeting Pack: netcoreapp MSI, installers


dleeapho:

dagood:
MSI is complete and building in official builds. I filed:
* https://github.com/dotnet/core-setup/issues/5174 to track a standalone Windows installer (the MSI is not friendly enough for this)
* https://github.com/dotnet/core-setup/issues/5175 for linux packages
* https://github.com/dotnet/core-setup/issues/5176 for MacOS installer

PR #5114, opened 2019-02-08
Create targeting pack MSI installers

dagood:
This change has the targeting pack pkgproj create an install layout based on the nupkg's data/ dir. That's picked up by the targeting pack installer project the same way it is for hostfxr and shared framework. In general, the targeting pack installer project is a copy of existing installers with very little changed.

In official builds, build the full platform manifest on win-x86, not just win-x64. Both platforms' installers need the full manifest.

---

A wrinkle about building on x64 and x86: the `Microsoft.NETCore.App.Ref` nupkg is built once on each, so there's a chance of differences. I diffed the installed bits, and everything's identical except ordering of the lines in the platform manifest.

Ultimately it would be safest to feed one targeting pack nupkg to all the legs to package up. For now I could sort the platform manifest if order affects how the file's used at all.

For https://github.com/dotnet/core-setup/issues/4772

---

The three installer projects in core-setup were very copy-paste heavy, and this PR adds a fourth. I plan to refactor this to remove a lot of the duplication, and I'm waiting to do that to get the MSI validated in the SDK ASAP. I mentioned this in a mail thread but missed it in the original PR description.

List of things I plan to do before I make another installer:

* Remove `generatemsi.ps1` powershell scripts, use msbuild projects with shared targets.
  * Peek at wixproj, but likely stick with `Exec`s for now.
* Consider unifying targeting pack installer creation with targeting pack pkgproj. This could make the targeting pack creation entry point cleaner.
  * I also want to look into merging `Microsoft.NETCore.App.Ref.pkgproj` into `Microsoft.NETCore.App.pkgproj`, and using it to drive other packs (Runtime, AppHost) as well. Same goal: cleaner entry point.
* Share WiX project files to avoid boilerplate.

### DIFF
--- a/src/pkg/packaging/dir.props
+++ b/src/pkg/packaging/dir.props
@@ -8,10 +8,12 @@
     <!-- Package prebuilt binaries -->
     <UsePrebuiltPortableBinariesForInstallers>false</UsePrebuiltPortableBinariesForInstallers>
     
+    <!-- Use trailing '/'. Heat.exe doesn't accept a trailing '\'.-->
     <PackagesIntermediateDir>$(IntermediateOutputRootPath)packages/</PackagesIntermediateDir>
     <SharedHostPublishRoot>$(IntermediateOutputRootPath)sharedHost/</SharedHostPublishRoot>
     <HostFxrPublishRoot>$(IntermediateOutputRootPath)hostFxr/</HostFxrPublishRoot>
     <SharedFrameworkPublishRoot>$(IntermediateOutputRootPath)sharedFx/</SharedFrameworkPublishRoot>
+    <TargetingPackPublishRoot>$(IntermediateOutputRootPath)Microsoft.NETCore.App.Ref/layout/</TargetingPackPublishRoot>
     <CombinedPublishRoot>$(IntermediateOutputRootPath)combined-framework-host/</CombinedPublishRoot>
   </PropertyGroup>
 
@@ -28,6 +30,7 @@
     <ProductBrandSuffix Condition="'$(ReleaseBrandSuffix)'!=''">$(ProductionVersion) $(ReleaseBrandSuffix)</ProductBrandSuffix>
     <SharedHostBrandName>$(ProductBrandPrefix) Host - $(ProductBrandSuffix)</SharedHostBrandName>
     <HostFxrBrandName>$(ProductBrandPrefix) Host FX Resolver - $(ProductBrandSuffix)</HostFxrBrandName>
+    <TargetingPackBrandName>$(ProductBrandPrefix) App Targeting Pack - $(ProductBrandSuffix)</TargetingPackBrandName>
     <SharedFrameworkBrandName>$(ProductBrandPrefix) Runtime - $(ProductBrandSuffix)</SharedFrameworkBrandName>
   </PropertyGroup>
 </Project>
--- a/src/pkg/packaging/dir.props
+++ b/src/pkg/packaging/dir.props
@@ -30,7 +30,7 @@
     <ProductBrandSuffix Condition="'$(ReleaseBrandSuffix)'!=''">$(ProductionVersion) $(ReleaseBrandSuffix)</ProductBrandSuffix>
     <SharedHostBrandName>$(ProductBrandPrefix) Host - $(ProductBrandSuffix)</SharedHostBrandName>
     <HostFxrBrandName>$(ProductBrandPrefix) Host FX Resolver - $(ProductBrandSuffix)</HostFxrBrandName>
-    <TargetingPackBrandName>$(ProductBrandPrefix) App Targeting Pack - $(ProductBrandSuffix)</TargetingPackBrandName>
+    <TargetingPackBrandName>$(ProductBrandPrefix) Targeting Pack - $(ProductBrandSuffix)</TargetingPackBrandName>
     <SharedFrameworkBrandName>$(ProductBrandPrefix) Runtime - $(ProductBrandSuffix)</SharedFrameworkBrandName>
   </PropertyGroup>
 </Project>
--- a/src/pkg/packaging/windows/package.targets
+++ b/src/pkg/packaging/windows/package.targets
@@ -47,6 +47,10 @@
         <Output TaskParameter="GeneratedGui" PropertyName="HostFxrUpgradeCode" />
       </GenerateGuidFromName>
 
+      <GenerateGuidFromName Name="$(TargetingPackInstallerFile)">
+        <Output TaskParameter="GeneratedGui" PropertyName="TargetingPackUpgradeCode" />
+      </GenerateGuidFromName>
+
       <ItemGroup>
         <WixOutputs Include="$(WixObjRoot)host">
             <InputDir>$(SharedHostPublishRoot)</InputDir>
@@ -59,6 +63,12 @@
             <BrandName>$(HostFxrBrandName)</BrandName>
             <InstallerName>$(HostFxrInstallerFile)</InstallerName>
             <UpgradeCode>$(HostFxrUpgradeCode)</UpgradeCode>
+        </WixOutputs>
+        <WixOutputs Include="$(WixObjRoot)targetingpack">
+            <InputDir>$(TargetingPackPublishRoot)</InputDir>
+            <BrandName>$(TargetingPackBrandName)</BrandName>
+            <InstallerName>$(TargetingPackInstallerFile)</InstallerName>
+            <UpgradeCode>$(TargetingPackUpgradeCode)</UpgradeCode>
         </WixOutputs>
         <!-- shared framework has extra arguments so it needs to be in a separate group -->
         <WixOutputs2 Include="$(WixObjRoot)sharedframework">
--- a/src/pkg/packaging/windows/package.targets
+++ b/src/pkg/packaging/windows/package.targets
@@ -40,15 +40,15 @@
                                  OverwriteDestination="false" />
 
       <GenerateGuidFromName Name="$(SharedHostInstallerFile)">
-        <Output TaskParameter="GeneratedGui" PropertyName="SharedHostUpgradeCode" />
+        <Output TaskParameter="GeneratedGuid" PropertyName="SharedHostUpgradeCode" />
       </GenerateGuidFromName>
 
       <GenerateGuidFromName Name="$(HostFxrInstallerFile)">
-        <Output TaskParameter="GeneratedGui" PropertyName="HostFxrUpgradeCode" />
+        <Output TaskParameter="GeneratedGuid" PropertyName="HostFxrUpgradeCode" />
       </GenerateGuidFromName>
 
       <GenerateGuidFromName Name="$(TargetingPackInstallerFile)">
-        <Output TaskParameter="GeneratedGui" PropertyName="TargetingPackUpgradeCode" />
+        <Output TaskParameter="GeneratedGuid" PropertyName="TargetingPackUpgradeCode" />
       </GenerateGuidFromName>
 
       <ItemGroup>
@@ -82,7 +82,7 @@
       <MakeDir Directories="$(WixObjRoot);@(WixOutputs);@(WixOutputs2)" />
 
       <GenerateGuidFromName Name="$(SharedFrameworkInstallerFile)">
-        <Output TaskParameter="GeneratedGui" PropertyName="SharedFxUpgradeCode" />
+        <Output TaskParameter="GeneratedGuid" PropertyName="SharedFxUpgradeCode" />
       </GenerateGuidFromName>
 
       <PropertyGroup>
@@ -110,7 +110,7 @@
      </PropertyGroup>
 
      <GenerateGuidFromName Name="$(SharedBundle)">
-       <Output TaskParameter="GeneratedGui" PropertyName="SharedBundleCode" />
+       <Output TaskParameter="GeneratedGuid" PropertyName="SharedBundleCode" />
      </GenerateGuidFromName>
 
      <PropertyGroup>
--- a/src/pkg/packaging/windows/targetingpack/generatemsi.ps1
+++ b/src/pkg/packaging/windows/targetingpack/generatemsi.ps1
@@ -30,35 +30,35 @@ $InstallFilesWixobj = "$WixObjRoot\install-files.wixobj"
 
 function RunHeat
 {
-    $result = $true
+    $Result = $true
     pushd "$WixRoot"
 
     Write-Host Running heat.. to $InstallFileswsx
-    write-host "Root $TargetingPackPublishRoot"
+    Write-Host "Root $TargetingPackPublishRoot"
 
     .\heat.exe dir `"$TargetingPackPublishRoot`" `
-    -nologo `
-    -template fragment `
-    -sreg -gg `
-    -var var.TargetingPackSrc `
-    -cg InstallFiles `
-    -srd `
-    -dr DOTNETHOME `
-    -out $InstallFileswsx | Out-Host
+        -nologo `
+        -template fragment `
+        -sreg -gg `
+        -var var.TargetingPackSrc `
+        -cg InstallFiles `
+        -srd `
+        -dr DOTNETHOME `
+        -out $InstallFileswsx | Out-Host
 
     if($LastExitCode -ne 0)
     {
-        $result = $false
+        $Result = $false
         Write-Host "Heat failed with exit code $LastExitCode."
     }
 
     popd
-    return $result
+    return $Result
 }
 
 function RunCandle
 {
-    $result = $true
+    $Result = $true
     pushd "$WixRoot"
 
     Write-Host Running candle..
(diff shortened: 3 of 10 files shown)


src/pkg/packaging/dir.props
```diff
@@ -28,6 +30,7 @@
     <ProductBrandSuffix Condition="'$(ReleaseBrandSuffix)'!=''">$(ProductionVersion) $(ReleaseBrandSuffix)</ProductBrandSuffix>
     <SharedHostBrandName>$(ProductBrandPrefix) Host - $(ProductBrandSuffix)</SharedHostBrandName>
     <HostFxrBrandName>$(ProductBrandPrefix) Host FX Resolver - $(ProductBrandSuffix)</HostFxrBrandName>
+    <TargetingPackBrandName>$(ProductBrandPrefix) App Targeting Pack - $(ProductBrandSuffix)</TargetingPackBrandName>
```
nguerrera: "App Targeting Pack" is odd branding. I suggest removing App.

dagood: This expands to `Microsoft .NET Core App Targeting Pack`, is `Microsoft .NET Core Targeting Pack` actually specific enough?

src/pkg/projects/Microsoft.NETCore.App.Ref/Microsoft.NETCore.App.Ref.pkgproj
```diff
@@ -39,4 +42,35 @@
...
+
+    <ItemGroup>
+      <TargetingPackDataEntries
+        Include="@(TargetingPackNupkgEntries)"
+        Condition="$([System.String]::new('%(TargetingPackNupkgEntries.Identity)').StartsWith('data/'))" />
+    </ItemGroup>
+
+    <ZipFileExtractToDirectory
+      SourceArchive="$(TargetingPackNupkgFile)"
+      DestinationDirectory="$(DataLayoutDir)ref/$(Id)/$(Version)/"
```
nguerrera: I count 3 to 1 on votes to keep packs/ over ref/

https://github.com/dotnet/designs/pull/50/files/b78838c38284d45f25798922dd6c01ca526df7dc#diff-ebf265d0476041dd8cab82fa24ab5b0e

Counting @dsplaisted on the 1 side for his email that said to use ref/.  I otherwise don't recall anyone having a strong opinion to change packs/

I think ref/ is inaccurate and also reads redundantly with .Ref in path.

src/pkg/packaging/dir.props
nguerrera: I think that's good branding for the base targeting pack (without app). The others will be ASP.NET Core and Windows Desktop to distinguish form .NETCore. The App name in the packages is unfortunate in my opinion. I think it goes back to .NETCore being taken in TFMs before what we now call .NET Core being .NETCoreApp -> NETCore.App. It's a weird name because you can use it in libraries IMHO.

I'm haven't seen ".NET Core App" before in branding.

dagood: I see, my view of the product has previously always been in nuget packages, so .NET Core App looks ordinary to me. Thanks for the details, will change.

johnbeisner:
@dagood 
Is this MSI a stand-alone MSI or is/will it included in a bundle installation of some sort?

nguerrera:
This will be included in the sdk installer. We would also like for it to be installable on its own so that you can have say 3.1 sdk + 3.0 runtime + 3.0 targeting pack as a possible deployment. (I.e. only one set of tools but bits needed to target old tfm with new tools.

See dotnet/designs#50

dagood:
My understanding is that this will be similar to the runtime installers: Core-Setup builds runtime MSIs that Core-SDK chains, but Core-Setup also produces a runtime bundle installer.

However, a targeting pack bundle would only chain a single MSI (I think). The runtime bundle chains 3 MSIs. Maybe we can theme the MSI and use that as the user-facing targeting pack installer? I'm not sure, though, if there's something special the bundle does I don't know about yet, if theming somehow only works on bundle installers, or if it's as simple as wanting an EXE rather than MSI.

src/pkg/projects/Microsoft.NETCore.App.Ref/Microsoft.NETCore.App.Ref.pkgproj
dsplaisted: I'm happy with `packs/` too.

dsplaisted:
Do we need separate x64 and x86 MSIs, or can we bundle the same MSI into the different architecture .NET Core SDKs, and just pass in a different value for DOTNET_ROOT?

dagood:
In the mail thread we're kind of using `DOTNETHOME` and `DOTNET_ROOT` interchangeably. Going with that, using the x64 installer as it is now, we can't install to two locations at once just using `DOTNETHOME`. If I do this:

```
.\dotnet-targeting-pack-3.0.0-preview-27408-0-win-x64.msi DOTNETHOME=C:\install-1
.\dotnet-targeting-pack-3.0.0-preview-27408-0-win-x64.msi DOTNETHOME=C:\install-2
```

The first installs into `C:\install-1`, which works. The second command gives me the change/repair/remove options. This seems like reasonable behavior--it's the same product, we're only telling it where to install.

Poking around, instance transforms seem to address this. ([Doc](https://docs.microsoft.com/en-us/windows/desktop/Msi/installing-multiple-instances-with-instance-transforms), [tutorial-ish post](http://windows-installer-xml-wix-toolset.687559.n2.nabble.com/Multiple-Instance-Transforms-Walkthrough-Proposed-Simple-Addition-to-WiX-to-Make-Them-Easier-td708828.html).) Adding this:

```xml
<Property Id="PACKINSTANCE" Value="DefaultValue" />
<InstanceTransforms Property="PACKINSTANCE">
    <Instance ProductCode="ec60557f-0793-4b6c-b181-56017b9d8d3e" ProductName="$(var.ProductName) (x64)" UpgradeCode="57f5e45d-4ab9-43b5-a0f4-4d94577c4ee7" Id="x64"/>
    <Instance ProductCode="2c7c073c-cf5d-4a90-8f53-f204029590d9" ProductName="$(var.ProductName) (x86)" UpgradeCode="c7ea6c78-442f-44fb-87ec-07b0399c5954" Id="x86"/>
</InstanceTransforms>
```

lets me install two instances with:

```
.\dotnet-targeting-pack-3.0.0-preview-27408-0-win.msi TRANSFORMS=:x86 DOTNETHOME=C:\x86 MSINEWINSTANCE=1
.\dotnet-targeting-pack-3.0.0-preview-27408-0-win.msi TRANSFORMS=:x64 DOTNETHOME=C:\x64 MSINEWINSTANCE=1
```

The SDK bundle would select one when chaining, and I suppose the user-facing targeting pack installer would have to let the user pick.

---

Stepping back, I'm not convinced there's a business need to do this. The only scenario I can think of is letting a developer skip downloading a targeting pack twice if they want to install it for x86 *and* x64 SDKs on the same machine. It even seems limited to manual targeting pack installs, because the x86 and x64 SDK bundles would already each have a copy of the targeting pack MSI.

Also, will it be confusing to devs to have x64/x86-specific downloads for everything else, but have an x64/x86 option in the targeting pack installer itself?

src/pkg/packaging/windows/package.targets
```diff
@@ -47,6 +47,10 @@
         <Output TaskParameter="GeneratedGui" PropertyName="HostFxrUpgradeCode" />
       </GenerateGuidFromName>
 
+      <GenerateGuidFromName Name="$(TargetingPackInstallerFile)">
+        <Output TaskParameter="GeneratedGui" PropertyName="TargetingPackUpgradeCode" />
```
ericstj: Gui?

src/pkg/projects/Microsoft.NETCore.App.Ref/Microsoft.NETCore.App.Ref.pkgproj
```diff
@@ -39,4 +42,35 @@
...
+      <Output TaskParameter="Entries" ItemName="TargetingPackNupkgEntries" />
+    </ZipFileGetEntries>
+
+    <ItemGroup>
+      <TargetingPackDataEntries
+        Include="@(TargetingPackNupkgEntries)"
+        Condition="$([System.String]::new('%(TargetingPackNupkgEntries.Identity)').StartsWith('data/'))" />
+    </ItemGroup>
+
+    <ZipFileExtractToDirectory
```
ericstj: Why do you pull them from the package?  Just a couple targets before this you have the entire contents in items (`@(File)` if I remember correctly).

src/pkg/packaging/windows/targetingpack/targetingpack.wxs
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
```
ericstj: All of this appears to be boilerplate yet it is still copy/pasted in every installer?

src/pkg/packaging/windows/targetingpack/variables.wxi
```diff
@@ -0,0 +1,26 @@
...
+  <?define Program_Files="ProgramFilesFolder"?>
+  <?define Win64AttributeValue=no?>
+  <?elseif $(var.Platform)=x64?>
+  <?define Program_Files="ProgramFiles64Folder"?>
+  <?define Win64AttributeValue=yes?>
+  <?else?>
+  <?error Invalid Platform ($(var.Platform))?>
+  <?endif?>
+
+  <?define DependencyKey   = "Dotnet_CLI_TargetingPack_$(var.BuildVersion)_$(var.Platform)"?>
```
ericstj: Why CLI?

src/pkg/packaging/windows/targetingpack/generatemsi.ps1
```diff
@@ -0,0 +1,162 @@
...
+function RunHeat
+{
+    $result = $true
+    pushd "$WixRoot"
+
+    Write-Host Running heat.. to $InstallFileswsx
+    write-host "Root $TargetingPackPublishRoot"
+
+    .\heat.exe dir `"$TargetingPackPublishRoot`" `
+    -nologo `
```
MichaelSimons: Indenting the arg lines would help the readability.

src/pkg/packaging/windows/targetingpack/generatemsi.ps1
```diff
@@ -0,0 +1,162 @@
...
+
+if(-Not (RunLight))
+{
+    Exit -1
+}
+
+if(!(Test-Path $TargetingPackMSIOutput))
+{
+    throw "Unable to create the Targeting Pack MSI."
+    Exit -1
```
MichaelSimons: Does this get executed?  Doesn't throw prevent Exit from getting executed?

Why is there a mixture of using Exit -1 vs throwing?  The script doesn't seem consistent.

src/pkg/packaging/windows/targetingpack/generatemsi.ps1
```diff
@@ -0,0 +1,162 @@
...
+$InstallFileswsx = "$WixObjRoot\install-files.wxs"
+$InstallFilesWixobj = "$WixObjRoot\install-files.wixobj"
+
+function RunHeat
+{
+    $result = $true
+    pushd "$WixRoot"
+
+    Write-Host Running heat.. to $InstallFileswsx
+    write-host "Root $TargetingPackPublishRoot"
```
MichaelSimons: `Write-Host` vs `write-host` inconsistency

src/pkg/packaging/windows/targetingpack/generatemsi.ps1
```diff
@@ -0,0 +1,162 @@
...
+    return $result
+}
+
+function RunCandle
+{
+    $result = $true
+    pushd "$WixRoot"
+
+    Write-Host Running candle..
+    $AuthWsxRoot =  Join-Path $PackagingRoot "windows\targetingpack"
```
MichaelSimons: Inconsistency in casing of local variables.  PascalCased vs camelCased - $AuthWsxRoot vs $result

src/pkg/packaging/windows/package.targets
dagood: I might as well fix it in this PR. Nothing points to this being anything but a typo.

src/pkg/projects/Microsoft.NETCore.App.Ref/Microsoft.NETCore.App.Ref.pkgproj
dagood: There's a bit of reasoning in the target comment:

> This extracts from the nupkg (zip) directly. An alternative would be using the PackageFile    items, but there's some risk of handling TargetPath metadata differently than NuGet does. Using the nupkg directly ensures results are identical.

It is slower to decompress it, I'm just coding a little defensively here.

src/pkg/packaging/windows/targetingpack/targetingpack.wxs
dagood: Yeah... I'm definitely planning to look into using some common WiX files as part of refactoring.

src/pkg/packaging/windows/targetingpack/variables.wxi
dagood: Looks like this name in general came from when the CLI repo originally split into Core-Setup. We may need to keep this in the other installers for continuity (although I haven't verified this) but I should definitely change it for the new installer. Thanks for spotting this.

joeloff:
Is the plan to apply MSTs to a baseline MSI? Bundles don't support this scenario AFAIK. Properties like the ProductCode/UpgradeCode are usually extracted into the metadata of the bundle. From what I've been able to find, the only way to support transforms in an MSI inside a bundle is to implement a custom bootstrapper.

dagood:
I hadn't heard of MSTs ("MSI transform files" for my own reference) before. It looks like I created a [embedded transform](https://docs.microsoft.com/en-us/windows/desktop/msi/embedded-transforms) based on [the command I ran it with](https://docs.microsoft.com/en-us/windows/desktop/msi/applying-transforms). I was expecting that the bundle could use the properties I used (`TRANSFORMS=:x86 DOTNETHOME=C:\x86 MSINEWINSTANCE=1`) but maybe this is special territory?

I stopped short of prototyping something when I realized there might not be a point in doing it. But it does seems worth understanding this a little more to eliminate the option on a technical note rather than just lack of need.

> Bundles don't support this scenario AFAIK. Properties like the ProductCode/UpgradeCode are usually extracted into the metadata of the bundle.

I see... I've got to wonder why they're extracted, but I do see that in the bundle now. Definitely raises the difficulty.

> From what I've been able to find, the only way to support transforms in an MSI inside a bundle is to implement a custom bootstrapper.

I saw hints of this searching around online, but couldn't tell if they were in the same situation. I'm guessing we strongly want to avoid making a custom bootstrapper if possible?

dagood:
Merging to get these in official builds.

Now we know more problems to work through to get a single targeting pack installer for x86/x64, but still no need. Still interested to know more, but moving on for now.

dagood:
@dsplaisted, https://dev.azure.com/dnceng/internal/_build/results?buildId=95378 produced these MSIs:
* https://dotnetcli.blob.core.windows.net/dotnet/Runtime/3.0.0-preview4-27413-11/dotnet-targeting-pack-3.0.0-preview4-27413-11-win-x64.msi
* https://dotnetcli.blob.core.windows.net/dotnet/Runtime/3.0.0-preview4-27413-11/dotnet-targeting-pack-3.0.0-preview4-27413-11-win-x86.msi

joeloff:
@dagood if we have two separate MSIs there shouldn't be a problem adding them to the bundles. The problem with transforms using only a single MSI is that the bundle would only be aware of the one MSI's metadata. So if you install it using the transform properties, the bundle won't be aware that it installed a different product and won't be able to track it.

dagood:
Yeah, I'd say two separate MSIs is a clear default plan. Simple and in line with what we've already been doing everywhere else, just a *tiny* bit *potentially* wasteful in terms of file size.

> The problem with transforms using only a single MSI is that the bundle would only be aware of the one MSI's metadata. So if you install it using the transform properties, the bundle won't be aware that it installed a different product and won't be able to track it.

Hmm, so basically the bundle must use those extracted properties, and since we'd only want to bundle one copy of the targeting pack MSI (the entire point!), it can only extract one set of properties to track with? Or maybe bundling can't extract properties at all for the transforms? (Unless we make a custom engine, which I'm still assuming we strongly don't want to do?)

src/pkg/projects/Microsoft.NETCore.App.Ref/Microsoft.NETCore.App.Ref.pkgproj
ericstj: I don't find that defensive, it's just adding complexity.  We define TargetPath and consume it, you can rely on it, of if you can't that's a bug that we fix.

Another thing to consider is inserting a layout phase between collecting assets and pack.  That way the content of the pack can be defined by the layout instead of vice-versa.

dagood: That's true. I did add a file in https://github.com/dotnet/core-setup/pull/5168 that has a non-folder TargetPath, which is the kind of thing I wanted to avoid handling--but it looks like that's actually the only one, and I don't need to do it that way. I'll get rid of this.

A layout-first approach sounds much better. I'll shoot for that.

PR #5134, opened 2019-02-11
Add CI for release/3.0

wtgodbe:
@dagood @safern PTAL

CC @danmosemsft @leecow

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,6 @@
 trigger:
 - master
+- release/3.0
 
 variables:
   - name: TeamName
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,11 @@ trigger:
 - master
 - release/3.0
 
+pr:
+- master
+- release/3.0
+
+
 variables:
   - name: TeamName
     value: dotnet-core-acquisition


safern:
Would it make sense to also define the PR trigger in the yaml and stop overriding it from the build def itself?

Also, we should batch CI builds.

See: https://github.com/dotnet/corefx/blob/master/.azure-ci.yml#L5-L13

wtgodbe:
I wasn't aware we triggered branches in CI from the build def, seems to me it'd be better to do it in the YAML file. Does that entail anything other than removing the override from the build def?

As for batching, I agree we should move to that, but I'll leave it for a separate PR

safern:
You need to remove the override and add the pr section to the yaml to specify PR triggers as well.
```
- pr:
....
```

wtgodbe:
@dagood @safern from what I'm seeing, the Core-Setup builds don't overwrite the YAML triggers - am I missing something?

dagood:
Interesting, you're right. My guess would be the def is currently trying to run PR validation on any branch (if it has valid yaml?), and the `pr:` change will limit it. Seems like a good idea in case a dev branch comes up that uses a bunch of build resources but doesn't actually need validation.

PR #5149, opened 2019-02-13
Update branding for preview4

wtgodbe:
CC @dagood @danmosemsft @mmitche @leecow

### DIFF
--- a/BranchInfo.props
+++ b/BranchInfo.props
@@ -12,7 +12,7 @@
     <IncludeBuildNumberInPackageVersion Condition="'$(StabilizePackageVersion)' != 'true'">true</IncludeBuildNumberInPackageVersion>
 
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
-    <ReleaseBrandSuffix>Preview</ReleaseBrandSuffix>
+    <ReleaseBrandSuffix>Preview4</ReleaseBrandSuffix>
     <Channel>master</Channel>
     <ContainerName>dotnet</ContainerName>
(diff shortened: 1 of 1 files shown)


dagood:
So `master` will produce `3.0.0-preview4`, `3.0.0-preview5`, etc., and `release/3.0` will produce `3.0.0-preview`? Is it ok that nuget versions will have a weird sort? (`3.0.0-preview4` will be later than `3.0.0-preview`, even when `release/3.0` is building for preview 5?)

wtgodbe:
@dagood Currently release/3.0 is building `preview3`

BranchInfo.props
```diff
@@ -12,7 +12,7 @@
     <IncludeBuildNumberInPackageVersion Condition="'$(StabilizePackageVersion)' != 'true'">true</IncludeBuildNumberInPackageVersion>
```
dagood: Same issue with `PreReleaseLabel` a few lines up.

BranchInfo.props
```diff
@@ -12,7 +12,7 @@
...
 
     <ReleaseSuffix>$(PreReleaseLabel)</ReleaseSuffix>
-    <ReleaseBrandSuffix>Preview</ReleaseBrandSuffix>
+    <ReleaseBrandSuffix>Preview4</ReleaseBrandSuffix>
```
dagood: This is also missing a space.

dagood: I wouldn't be opposed to using MSBuild property functions to lowercase and remove the space to keep brand and prerelease label in sync.

wtgodbe:
My bad, it should be fixed now. Will do something similar in Master

PR #5170, opened 2019-02-14
[release/2.1] Set createLogFile to false for msbuild steps, fix official build

dagood:
createLogFile "true" sets flp and expects the log file it specifies to be present when the step completes. Our arguments include another flp parameter that overrides the task's built-in flp, causing the file not to exist, failing the build. Fix this by setting createLogFile "false".


dagood:
I'd expect the Tizen failure due to difficulty building a matching nupkg automatically. Merging.

PR #5185, opened 2019-02-15
IJW Host

jkoritzinsky:
Add a hosting shim for IJW assemblies on Windows. 

When an IJW assembly is loaded, the shim will swap out the tokens in the vtfixup table with delayed-activation thunks that activate the runtime when called (to avoid the Windows Loader Lock). The shim also provides an entry-point to resolve these delayed-activation thunks back to tokens for the case where the IJW assembly is being loaded from managed code.

IJW executables will be built into DLLs (same as the rest of the .NET Core ecosystem), so we don't need any additional entrypoints.

cc: @jkotas @jeffschwMSFT @sdmaclea 

See dotnet/coreclr#22636 for the runtime component.


src/corehost/cli/ijwhost/pedecoder.h
```diff
@@ -0,0 +1,130 @@
...
+private:
+
+    bool HasReadyToRunHeader() const
+    {
+        return FindReadyToRunHeader() != nullptr;
+    }
+
+    READYTORUN_HEADER * FindReadyToRunHeader() const;
+
+    bool CheckDirectory(IMAGE_DATA_DIRECTORY *pDir) const
```
jkotas: I do not think you need this. It is ok to assume that the IJW image is valid.

src/corehost/cli/ijwhost/pedecoder.h
```diff
@@ -0,0 +1,123 @@
...
+private:
+
+    bool HasReadyToRunHeader() const
+    {
+        return FindReadyToRunHeader() != nullptr;
+    }
+
+    READYTORUN_HEADER * FindReadyToRunHeader() const;
+
+    IMAGE_SECTION_HEADER* RvaToSection(std::uint32_t rva) const;
```
jkotas: RvaToSection and FindFirstSection can be delete too.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -161,9 +161,6 @@ static int execute_host_command(
     if (code != StatusCode::Success)
         return code;
 
-    // Previous hostfxr trace messages must be printed before calling trace::setup in hostpolicy
-    trace::flush();
-
```
vitek-karas: I don't understand why we need to remove the flush here?
All the other cases have it for a good reason - this one as well...

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1412,13 +1381,47 @@ int fx_muxer_t::execute(
...
+static int get_delegate_from_runtime(
+    const pal::string_t& impl_dll_dir,
+    corehost_init_t* init,
+    const char* corehost_entrypoint,
+    void** delegate)
 {
-    assert(host_info.is_valid());
+    pal::dll_t corehost;
+    hostpolicy_contract host_contract{};
+    corehost_get_delegate_fn host_entry = nullptr;
```
vitek-karas: Can we call this with a more descriptive name? Maybe `corehost_entrypoint_function` or something like that - basically something which makes it easy to correlate to the `corehost_entrypoint` argument.

src/corehost/cli/fxr_resolver.h
```diff
@@ -0,0 +1,34 @@
...
+    void **delegate);
+using hostfxr_main_fn = int(*)(const int argc, const pal::char_t* argv[]);
+using hostfxr_main_startupinfo_fn = int(*)(
+    const int argc,
+    const pal::char_t* argv[],
+    const pal::char_t* host_path,
+    const pal::char_t* dotnet_root,
+    const pal::char_t* app_path);
+using hostfxr_error_writer_fn = void(*)(const pal::char_t* message);
+using hostfxr_set_error_writer_fn = hostfxr_error_writer_fn(*)(hostfxr_error_writer_fn error_writer);
```
vitek-karas: I would prefer if we would separate the function pointers into their own header. They don't really belong to the resolver.
Maybe `hostfxr.h`? (Might even ship it for people to easily include one day)

src/corehost/cli/hostpolicy.cpp
```diff
@@ -615,6 +614,54 @@ SHARED_API int corehost_get_com_activation_delegate(void **delegate)
         delegate);
 }
 
+SHARED_API int corehost_get_load_and_execute_in_memory_assembly_delegate(void** delegate)
```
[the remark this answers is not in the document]
vitek-karas: I think I agree now. Basically what we want is a way to ask for a special "runtime" entry point. And we have 3 of them now (will have 4 soon). I agree it would be nice to have just one export for all of them - it would also make it a bit more explicit ... "these are special things, please don't use them directly".

src/corehost/cli/ijwhost/exports.cpp
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
vitek-karas: I know that @AaronRobinsonMSFT  started this in the comhost, but I think I don't like the generic `exports.cpp` name of this file. It makes it hard to recognzie what I'm looking at in the editor, if nothing else. (It also means we have at least two files with the exact same name in the repo close to each other).

I would prefer to use the approach which `hostfxr` and `hostpolicy` already uses, that is `hostfxr.cpp` and `hostpolicy.cpp`. So in this case `ijwhost.cpp`. Can you please also rename the one in `comhost` to `comhost.cpp`?

src/corehost/cli/ijwhost/ijwhost.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
vitek-karas: I don't really understand why we have `exports.cpp` and then this file... why not merge them into one?

src/corehost/cli/fxr/fx_muxer.cpp
jkoritzinsky: I think that one accidentally got removed at some point when I was trying to refactor some common code out that I couldn’t do the way I wanted. I’ll revert it

src/corehost/cli/ijwhost/exports.cpp
AaronRobinsonMSFT: @vitek-karas As a historical note, when creating a C++ component I normally create a file called `exports.cpp` for a library and `main.cpp` for a console. These files represent contain the library/console contract for consumption. In the case of the console it will contain `main`/`wmain` per the OS contract and in the case of the library it will contain the exported functions for consumption. It makes it very clear to consumers of the source where to started when tracing code since the `exports.cpp` compilation unit will contain the only "public" functions.

vitek-karas: I understand that point of view... it's just that for me personally - I don't like having multiple files with the same names in one solution. But I can learn to live with that. This was more of a nit... so if I'm the only one... feel free to ignore.

jkotas: We should follow consistent style for all hosts in the repo; or ideally across .NET Core C++ components.

FWIW, the `exports.cpp` convention does not scale. It works only if the library exports small number of methods.

AaronRobinsonMSFT: @jkotas I disagree with it not scaling. Like many things "does not scale" is predicated on a specific design orthodoxy. I prefer libraries have small concise number of exports rather than a large number which makes unit testing and mental modelling simpler. In fact, if I recall the .NET Core system is/was trying to do just this by making many smaller libraries rather than large ones. We can argue over what large or small means, but it can work in practice - especially if you are creating a third party library. Creating complete generic platforms tends to fail due to wanting a single library, but like the `coreclr` or `hostfxr` libraries the number of exports should be kept small and concise because they are _not_ a generic platform API but simple library contracts that are intended to be configured and then activated.

@vitek-karas I am not pushing back, just offering why I generally do that for this kind of project. I absolutely understand your perspective and if that is your preference let's do it.

src/corehost/cli/ijwhost/IJWBootstrapThunkCPU.h
```diff
@@ -0,0 +1,74 @@
...
+#include "pal.h"
+
+extern "C" void VTableBootstrapThunkInitHelperStub();
+
+class VTableBootstrapThunkCode;
+
+class VTableBootstrapThunk
+{
+private:
+    VTableBootstrapThunkCode *GetCode();
```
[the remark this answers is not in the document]
jkoritzinsky: This class effectively acts as an opaque class over each of the arch-specific `VTableBootstrapThunkCode` classes.

src/corehost/cli/ijwhost/arm/IJWBootstrapThunkCPU.cpp
```diff
@@ -0,0 +1,231 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "pedecoder.h"
+#include "IJWBootstrapThunkCPU.h"
```
jkotas: Nit: I would drop the CPU suffix from the name. This suffix was a workaround for build system issue 15 years ago.

src/corehost/cli/ijwhost/IJWBootstrapThunkCPU.h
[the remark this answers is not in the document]
jkoritzinsky: I'm working on refactoring this to just directly expose the `VTableBootstrapThunkCode` class from the respective arch.

src/corehost/cli/ijwhost/exports.cpp
jkotas: I agree with you that creating small unit-testable components is a good thing. I disagree that there should be one shipping .dll per the small unit-testable component. We have tried to do that with .NET Core and it did not work well for us. It is hard to manage dependencies between the components when they are small .dlls and it costs performance to load many small .dlls.

> what large or small means

For me, small .dll is less than 1MB.

src/corehost/cli/comhost/comhost.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
jkotas: We have stopped using this license header in 2016. Looks like `core-setup` was missed during the license header cleanup.

src/corehost/cli/hostpolicy.cpp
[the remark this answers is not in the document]
jkoritzinsky: I've updated the code to use an enum in the layer between hostfxr and hostpolicy. Do we want to use an enum as well on the public API surface of hostfxr? I think we can do that as well if we want to.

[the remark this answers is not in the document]
vitek-karas: Yes - I think we should do the same change in hostfxr as well.

vitek-karas:
Can you please add some reasoning why we removed the ability to execute IJW assembly (.exe)?

jeffschwMSFT:
@vitek-karas 
> Can you please add some reasoning why we removed the ability to execute IJW assembly (.exe)?

We discussed with the Microsoft Visual C++ team and the proposed project system will only support dll creation.  Only having dlls also fits the .NET Core appmodel.

@jkoritzinsky and I discussed the challenges of enabling IJW exe's and due to how they work for .NET Framework, there are technical challenges which I feel we can scope out for now.

src/corehost/cli/hostfxr.h
```diff
@@ -7,20 +7,35 @@
...
+};
+
+using hostfxr_get_delegate_fn = int32_t(*)(
+    const pal::char_t* host_path,
+    const pal::char_t* dotnet_root,
+    const pal::char_t* app_path,
+    hostfxr_delegate_type type,
+    void** delegate);
+
+using hostfxr_get_delegate_fn_old = int(*)(
```
AaronRobinsonMSFT: Why is this one needed?

src/corehost/cli/comhost/comhost.cpp
```diff
@@ -90,7 +90,7 @@ namespace
 
         // Leak fxr
 
-        auto get_com_delegate = (hostfxr_get_delegate_fn)pal::get_symbol(fxr, "hostfxr_get_com_activation_delegate");
+        auto get_com_delegate = (hostfxr_get_delegate_fn)pal::get_symbol(fxr, "hostfxr_get_runtime_delegate");
```
AaronRobinsonMSFT: ~~`get_com_delegate`~~ -> `get_runtime_delegate`

src/corehost/cli/hostfxr.h
jkoritzinsky: I thought I removed that before committing...

AaronRobinsonMSFT:
cc @elinor-fung

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -1406,13 +1378,47 @@ int fx_muxer_t::execute(
...
+    coreclr_delegate_type type,
+    void** delegate)
 {
-    assert(host_info.is_valid());
+    pal::dll_t corehost;
+    hostpolicy_contract host_contract{};
+    corehost_get_delegate_fn coreclr_delegate = nullptr;
+
+    int code = load_hostpolicy(impl_dll_dir, &corehost, host_contract, "corehost_get_coreclr_delegate", &coreclr_delegate);
+    if (code != StatusCode::Success)
```
vitek-karas: We should write out a good error message in this case. This can happen if somebody is trying to run COM or IJW on .NET Core 2.* since the `hostpolicy` of 2.* won't have the necessary export. We should be able to detect that specific case and report an error in lines of "The COM component must target netcoreapp3.0 or higher version" or something like that.

src/corehost/cli/ijwhost/ijwhost.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+        if (pe.IsILOnly() || (!pe.HasManagedEntryPoint() && !pe.HasNativeEntryPoint()))
+        {
+            // If there is no user entry point, then we don't want the
+            // thread start/stop events going through because it slows down
+            // thread creation operations
+            DisableThreadLibraryCalls(hInst);
+        }
+
+        // Install the bootstrap thunks
```
vitek-karas: The host must call `trace::setup()` the first time it gets called - the tracing will otherwise not work - by default it's disabled and it doesn't do lazy init on every `trace::` call. 
Please note that comhost has the same bug - can you please also look at that and fix it there?

src/corehost/cli/ijwhost/ijwhost.cpp
```diff
@@ -0,0 +1,152 @@
...
+        if (!patch_vtable_entries(pe))
```
vitek-karas: Another problem is `trace::error`. By default this writes to `stderr` even if tracing is disabled - so it does that in 99.9% of cases.

I don't know enough about the IJW scenarios to be able to tell what should be the right behavior. That said, by default the process will get failures into its `stderr` output, which might not be desirable.

In case of COM we redirect error writing to a buffer and report it through COM error reporting. The main scenario there is if something tries to load a COM component which fails, it should be able to handle the failure "Silently".

I don't know if IJW has such scenario where failure to load IJW assembly should also be possible to handle "Silently". But I think it's probably safer to do that. The one scenario which comes to mind is a simple `Assembly.LoadFile` on IJW assembly - which may fail and the caller would expect to be able to simply `try/catch-ignore` around it without any consequences. If we write errors to stderr that would not be the case.

Note that even if we redirect error writing, if tracing is enabled all error output goes to the trace as well - so that would be a way to debug such failures (enable host tracing to get full information).

src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -0,0 +1,189 @@
...
+        // If we ignore the failure to patch bootstrap thunks we will come to this same
+        // function again, causing an infinite loop of "Failed to start the .NET Core runtime" errors.
+        // As we were taken here via an entry point with arbitrary signature,
+        // there's no way of returning the error code so we just throw it.
+
+        trace::error(_X("Failed to start the .NET Core runtime. Error code %d"), status);
+
+#pragma warning (push)
+#pragma warning (disable: 4297)
+        throw status;
```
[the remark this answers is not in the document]
jkoritzinsky: This is a pure C++ exception. It's the same behavior as .NET Framework in this case.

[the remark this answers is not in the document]
jkoritzinsky: This call also doesn't happen during load. This is in the delayed-activation thunk, so the code here is called by whatever method first calls a managed method in the IJW host. As a result, we have absolutely no idea what code called us, so there's no good way to report the errors to the caller since the caller can be literally anything.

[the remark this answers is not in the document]
vitek-karas: I understand... 
I guess we picked C++ since this is most likely called by the native parts of the C++/CLI and thus something which at least resembles C++ and probably has catch handlers for it... but I'm just guessing.

Anyway - it is a conscious choice with history behind it - good enough for me :-)

src/corehost/cli/ijwhost/ijwhost.cpp
jkoritzinsky: So there's two areas in the IJW host that trace errors/warnings:

1. During load, if we cannot change the page permissions on the memory page that the vtfixup table is located in, we log an error when it's unrecoverable and a warning when it's recoverable. In the .NETFX implementation, these logs are just debug-only auto-fail asserts instead, so that is another option for those.

2. When starting the runtime upon the first call to the native function. Since we're in an arbitrary call stack (like I mentioned above with the exception throw) we have no knowledge whatsoever of who is calling us or what scenario we are in. Additionally, this code-path is not executed at all if IJW is first loaded as a managed assembly since the .NET assembly loader patches the vtable with the JIT stubs before anything can call these thunks. I can't think of any good options for alternative error logging here nor can I think of any scenarios where we would want this to be handled "Silently".

Documentation/design-docs/IJW-activation.md
```diff
@@ -0,0 +1,103 @@
...
+
+Additionally, .NET Framework has support for a legacy code-path to forward calls to `_CorDllMain` to a user-provided `DllMain`. These code-paths are prone to locking under the loader lock if they call into any managed code or if the user-provided `DllMain` is a managed method, so support for a managed `DllMain` implementation is deprecated. Additionally, the Visual C++ compiler no longer generates a `DllMain` implementation for library initialization. Instead, the compiler generates a static module constructor to initialize any global state.
+
+### .NET Core IJW Activation
+
+Like with COM activation, our intent is to avoid a system-wide shim for IJW activation, especially since the host DLL needs to be linked to any C++/CLI assemblies. This new library (henceforth called the 'shim') will export functions to fulfill the requirements that the Visual C++ compiler needs to compile C++/CLI assemblies. Since we do not need to support backward compatibility with previously compiled mixed-mode assemblies, we are free to rename the exported functions while finalizing the design.
```
AaronRobinsonMSFT: ~~any C++/CLI assemblies~~ -> "all C++/CLI assemblies"

Documentation/design-docs/IJW-activation.md
```diff
@@ -0,0 +1,103 @@
...
+       public static int LoadAndExecuteInMemoryAssembly(IntPtr handle, int argc, [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.LPWStr, SizeParamIndex = 1)] string[] argv); /* argc is required for marshalling to know how large to make the argv array */
+   }
+   ```
+
+   Note this API would not be exposed outside of `System.Private.CoreLib` unless we decide to do so.
+   * The loading of the assembly will take place in the default `AssemblyLoadContext`.
+
+#### IJW DLLs and Delayed-Activation Thunks
+
+When `_CorDllMain()` is called when the DLL, the following will occur:
```
AaronRobinsonMSFT: "When `_CorDllMain()` is called when the DLL"?

Documentation/design-docs/IJW-activation.md
```diff
@@ -0,0 +1,103 @@
...
+2) Call the native `DllMain` if the user provided one.
+3) If `_CorDllMain` was called because the DLL is being unloaded from the process:
+   1) Deallocate the thunks allocated for the calling DLL.
+
+The .NET Framework implementation uses the runtime's `PEDecoder` class to validate that the PE is a .NET assembly, bail out early if it is IL only, get the native entry point if it exists, get the vtfixup table, and read RVAs from the module. Our options here are to either port over the `PEDecoder` class to core-setup, move the shim to coreclr and copy a chunk of code from core-setup to coreclr along with it to enable the shim to activate the runtime through hostfxr, or some other code-sharing model. The `PEDecoder` class heavily uses the CoreCLR `CONTRACT` and `CHECK` APIs, so porting it over to core-setup is not a simple task. A subset has been ported to be able to test behavior. Since the OS will validate the image when loading it, we can avoid having to port over all of the validation code and just add a check for the NT header that .NET uses to specify that it is a .NET assembly.
```
AaronRobinsonMSFT: Is this paragraph still needed?

Documentation/design-docs/IJW-activation.md
```diff
@@ -0,0 +1,103 @@
...
+    ```
+
+  Note this API would not be exposed outside of `System.Private.CoreLib` unless we decide to do so.
+  * The loading of this assembly will take place in an isolated `AssemblyLoadContext`.
+
+The naming of these APIs is designed to be useful for non-IJW scenarios as well, such as possibly Single-Exe.
+
+When the runtime loads the assembly, it needs to know if each element in the vtfixup table is a token or a stub. In .NET Framework, this check is implemented by the runtime querying `mscoree.dll` by looking up callbacks. When the runtime is traversing the vtfixup table and updating the entries to point to JIT stubs, it queries `mscoree.dll` if the module has stubs. If the module has stubs, it calls back into `mscoree.dll` to query the stub data structures for the metadata token. Otherwise, it grabs the token from the slot.
+
+We will implement it similarly, by having CoreCLR call back into `ijwhost.dll` if it is present.
```
AaronRobinsonMSFT: When will it not be present? This is unusual phrasing, can you clarify?

src/corehost/cli/fxr_resolver.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+        fx_ver_t max_ver;
+        for (const auto& dir : list)
+        {
+            trace::info(_X("Considering fxr version=[%s]..."), dir.c_str());
+
+            pal::string_t ver = get_filename(dir);
+
+            fx_ver_t fx_ver;
+            if (fx_ver_t::parse(ver, &fx_ver, false))
```
AaronRobinsonMSFT: Can you comment what `false` is in the call to `fx_ver_t::parse()`.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -607,11 +606,24 @@ SHARED_API int corehost_get_com_activation_delegate(void **delegate)
...
+    case coreclr_delegate_type::load_in_memory_assembly:
+        return coreclr->create_delegate(
+            "System.Private.CoreLib",
+            "Internal.Runtime.InteropServices.InMemoryAssemblyLoader",
+            "LoadInMemoryAssembly",
+            delegate);
+    default:
+        return StatusCode::LibHostInvalidArgs;
+    }
+
```
AaronRobinsonMSFT: Remove line.

src/corehost/cli/ijwhost/AMD64/bootstrap_thunk.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <pshpack1.h>
+class bootstrap_thunk
+{
+private:
+    // 49 BA 78 56 34 12 78 56 34 12 mov         r10,1234567812345678h
+    // 49 BB 34 12 34 12 34 12 34 12 mov         r11,1234123412341234h
+    // 41 FF E3                      jmp         r11
+
+    static BYTE             s_mov_r10[2];
```
AaronRobinsonMSFT: I would remove these static's from the class. Also, they should be marked `const`.

src/corehost/cli/ijwhost/AMD64/bootstrap_thunk.h
```diff
@@ -0,0 +1,72 @@
...
+    std::uint32_t           m_token;                        // 4 bytes
+    enum {
+        e_TOKEN_IS_DEF = 0x1
+    };
+    UINT32                  m_flags;                        // 4 bytes
+
+    pal::dll_t               m_dll;            // pal::dll_t of this module
+                                                    // 8 bytes
+    std::uintptr_t               *m_slot;             // VTable slot for this thunk
+                                                    // 8 bytes
```
AaronRobinsonMSFT: There are some sizes defined here in comments. and I don't see a lot of value and in some cases could be misleading. Can these be removed? Is there value?

src/corehost/cli/ijwhost/CMakeLists.txt
```diff
@@ -0,0 +1,95 @@
...
+    ../fxr/fx_ver.cpp
+    pedecoder.cpp
+    bootstrap_thunk_chunk.cpp
+    ${ARCH_SPECIFIC_FOLDER_NAME}/bootstrap_thunk.cpp
+    Exports.def
+)
+
+if(WIN32)
+    list(APPEND SOURCES
+        Exports.def)
```
AaronRobinsonMSFT: Can we use `__declspec(dllexport)` instead of the DEF file?

src/corehost/cli/ijwhost/corhdr.h
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+#ifndef CORHDR_H
```
AaronRobinsonMSFT: Can I assume you trimmed as much as possible?

src/corehost/cli/ijwhost/corhdr.h
```diff
@@ -0,0 +1,64 @@
...
+    mdtString               = 0x70000000,       //
+    mdtName                 = 0x71000000,       //
+    mdtBaseType             = 0x72000000,       // Leave this on the high end value. This does not correspond to metadata table
+} CorTokenType;
+
+#define mdTokenNil                  ((mdToken)0)
+
+typedef enum ReplacesGeneralNumericDefines
+{
+    IMAGE_DIRECTORY_ENTRY_COMHEADER     =14,
```
AaronRobinsonMSFT: Please add a reference to the PE spec.

src/corehost/cli/ijwhost/i386/bootstrap_thunk.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+//=================================================================================
+// Returns the token of this thunk
+std::uint32_t bootstrap_thunk::get_token()
+{
+    std::uint32_t ulTok = 0;
+    BYTE *pbTok = (BYTE *)&ulTok;
+
+    memcpy(pbTok, &m_tok[0], 3);
```
AaronRobinsonMSFT: `memcpy(pbTok, &m_tok[0], sizeof(m_tok));`

src/corehost/cli/ijwhost/i386/bootstrap_thunk.cpp
```diff
@@ -0,0 +1,94 @@
...
+//
+void bootstrap_thunk::initialize(std::uintptr_t pThunkInitFcn,
+                                          pal::dll_t dll,
+                                          std::uint32_t token,
+                                          std::uintptr_t *pSlot)
+{
+    
+    // First fill in the token portion of the struct.
+    BYTE *pbTok = (BYTE *)(&token);
+    memcpy(&m_tok[0], pbTok, 3);
```
AaronRobinsonMSFT: `memcpy(&m_tok[0], pbTok, sizeof(m_tok));`

src/corehost/cli/ijwhost/i386/bootstrap_thunk.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // Now set up the thunk code
+    std::uintptr_t pFrom;
+    std::uintptr_t pTo;
+    
+    // This is the call to the thunk bootstrapper function
+    pFrom = (std::uintptr_t)&m_code.m_thunkFcn + sizeof(m_code.m_thunkFcn);
+    pTo = pThunkInitFcn;
+    m_code.m_call            = 0xe8;
```
AaronRobinsonMSFT: Document `0xe8`.

src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -0,0 +1,192 @@
...
+#define COR_VTABLE_NOT_PTRSIZED COR_VTABLE_32BIT
+#else
+#define COR_VTABLE_PTRSIZED     COR_VTABLE_32BIT
+#define COR_VTABLE_NOT_PTRSIZED COR_VTABLE_64BIT
+#endif
+
+namespace
+{
+    std::mutex g_thunkChunkLock{};
+    bootstrap_thunk_chunk* g_pVtableBootstrapThunkChunkList = nullptr;
```
AaronRobinsonMSFT: By definition this will be `nullptr`, can be removed.

src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -0,0 +1,192 @@
...
+    for (size_t i = 0; i < numFixupRecords; ++i)
+    {
+        numThunks += pFixupTable[i].Count;
+    }
+
+    size_t chunkSize = sizeof(bootstrap_thunk_chunk) + sizeof(bootstrap_thunk) * numThunks;
+
+    void* pbChunk = HeapAlloc(g_heapHandle, 0, chunkSize);
+
+    if (pbChunk == nullptr)
```
AaronRobinsonMSFT: I want to talk about this. This will occur if the process runs out memory, so we return `false` and now what? Do we have a recourse here? My concern here is if we can't allocate the vtable for fixups, the function will return and then eventually the CLR will try to do something. I am more of the opinion that at IJW startup, we should A/V instead of attempting to handle an OOM.

src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -0,0 +1,192 @@
...
+        g_pVtableBootstrapThunkChunkList = chunk;
+    }
+
+    trace::setup();
+
+    size_t currentThunk = 0;
+    for(size_t i = 0; i < numFixupRecords; ++i)
+    {
+#ifndef _WIN64
+        assert((pFixupTable[i].Type & (COR_VTABLE_FROM_UNMANAGED | COR_VTABLE_FROM_UNMANAGED_RETAIN_APPDOMAIN)) && "managed->managed vtablefixup slots are not supported!");
```
AaronRobinsonMSFT: Why is this related to x86?

src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    std::uintptr_t thunkAddress = *(pThunk->get_slot_address());
+
+    return thunkAddress;
+}
+
+void release_bootstrap_thunks(PEDecoder& pe)
+{
+    // Is this an IJW module
+    if (!pe.IsILOnly())
```
AaronRobinsonMSFT: Flip and return early.

src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -0,0 +1,192 @@
...
+                break;
+            }
+        }
+    }
+}
+
+
+bool are_thunks_installed_for_module(pal::dll_t instance)
+{
+    std::lock_guard<std::mutex> _{g_thunkChunkLock};
```
AaronRobinsonMSFT: Instead of `_` use `lock`.

src/settings.cmake
```diff
@@ -247,3 +251,33 @@ if (WIN32 AND (CLI_CMAKE_PLATFORM_ARCH_ARM OR CLI_CMAKE_PLATFORM_ARCH_ARM64))
 	      message("Using Windows SDK version ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
       endif()
 endif ()
+
+if (WIN32)
+    if(CLI_CMAKE_PLATFORM_ARCH_ARM)
+      # Explicitly specify the assembler to be used for Arm32 compile
+      if($ENV{__VSVersion} STREQUAL "14 2015")
```
AaronRobinsonMSFT: Does "14 2015" need to be updated?

src/corehost/cli/comhost/comhost.cpp
vitek-karas: I filed #5224 for that.

src/corehost/cli/ijwhost/ijwhost.cpp
vitek-karas: In both cases the errors seem recoverable from the outside caller perspective. Basically the logic is as follows. Let's say I have code like this:
``` C#
try
{
  Assembly wpfAssembly = Assembly.LoadFile("wpf_ijw_Assembly.dll");
}
catch (Exception ex)
{
  Console.WriteLine("Something failed, but we can live without WPF for now.");
}
```

If we want/need to support the above scenario then I would also expect the above code to not write anything to `stderr`. If we don't "swallow" error writing in the host, it will pollute the `stderr`. 

I understand that there's no good way to report detailed error information and I think it's OKish in this case. Debugging can be done by turning on host tracing (`COREHOST_TRACE=1`) which would include the errors as well any other tracing we do in the host.

But we simply should not write to `stderr` unless we're explicitly asked to in "component" scenarios. This is unfortunate consequence of the fact that host was original written as the owner of the process, in which case writing to `stderr` made sense. Since we've added lot more scenarios where the host doesn't own the process (comhost, ijwhost, nethost) or where `stderr` is a bad choice (GUI apps on Windows).

src/settings.cmake
vitek-karas: We now officially support only VS 2017 and VS 2019.

src/corehost/cli/ijwhost/ijwthunk.cpp
jkoritzinsky: This function is called as part of the native module load, so if we fail to allocate, then we'll return false from `_CorDllMain` and tell Windows that we failed to load the module successfully. As a result, this module won't be successfully loaded into the process, let alone the CLR if we fail to allocate here.

src/corehost/cli/ijwhost/CMakeLists.txt
jkoritzinsky: We can't use `__declspec(dllexport)` because the MSVC compiler (for `_CorDllMain`) and CoreCLR (for `GetTokenForVTableEntry`) are relying on the entry-point being the same name on all platforms instead of accounting for the stdcall mangling. We can make an active decision to change it, but at least for now I've decided to keep it simpler by just using a DEF file to expose an unmangled name on all platforms.

src/corehost/cli/ijwhost/corhdr.h
jkoritzinsky: Yep. I only pulled in what I needed. The only extra pieces are extra members of some of the enums since I didn't think we would want to have partial enum definitions that didn't match coreclr in here.

src/corehost/cli/ijwhost/AMD64/bootstrap_thunk.h
```diff
@@ -0,0 +1,72 @@
...
+    BYTE                    m_mov_r11[2];
+    BYTE                    m_val_r11[8];
+    BYTE                    m_jmp_r11[3];   // total 23 bytes
+    BYTE                    m_padding[1];   // 1 byte to pad to 24
+    // Data for the thunk
+    std::uint32_t           m_token;                        // 4 bytes
+    enum {
+        e_TOKEN_IS_DEF = 0x1
+    };
+    UINT32                  m_flags;                        // 4 bytes
```
jkotas: Where are these flags used?

src/corehost/cli/ijwhost/corhdr.h
jkotas: You can use `IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR` from `winnt.h` from standard Windows SDK instead.

src/corehost/cli/ijwhost/corhdr.h
```diff
@@ -0,0 +1,64 @@
...
+} IMAGE_COR_VTABLEFIXUP;
+
+#define RidFromToken(tk) ((ULONG32) ((tk) & 0x00ffffff))
+#define TypeFromToken(tk) ((ULONG32)((tk) & 0xff000000))
+#define IsNilToken(tk) ((RidFromToken(tk)) == 0)
+
+
+using mdToken = std::uint32_t;
+
+typedef enum CorTokenType
```
jkotas: Are these token definitions really only used for asserts?

src/corehost/cli/ijwhost/ijwthunk.cpp
AaronRobinsonMSFT: >As a result, this module won't be successfully loaded into the process, let alone the CLR if we fail to allocate here.

That is a good thing for sure. However, the question remains regarding how we or the user can proceed? Does this manifest itself as "module failed to load?" What is the developer experience here?

src/corehost/cli/ijwhost/corhdr.h
jkoritzinsky: On x86 it looks like the stub encodes it in whether or not it is a mdtMethodDef or mdtMemberRef. I don't see this encoding being used anywhere though so I can remove it, and also remove the asserts if you would like.

src/corehost/cli/ijwhost/ijwhost.cpp
```diff
@@ -0,0 +1,152 @@
...
+    // Thus, we do DLL_PROCESS_ATTACH work first, and DLL_PROCESS_DETACH work last.
+    if (dwReason == DLL_PROCESS_ATTACH)
+    {
+        // If this is not a .NET module (has a CorHeader), shouldn't be calling _CorDllMain
+        if (!pe.HasCorHeader())
+        {
+            return FALSE;
+        }
+
+        if (pe.IsILOnly() || (!pe.HasManagedEntryPoint() && !pe.HasNativeEntryPoint()))
```
jkotas: I am not sure how valuable these `IsILOnly` checks are. The IL-only check pulls in definitions of structures like R2R header that this should not really depend on.

If we get here for IL-only image, it means something is really wrong.

I would just check for the entrypoint here, and remove the IsILOnly checks from here and other places.

src/corehost/cli/ijwhost/ijwthunk.cpp
jkoritzinsky: I pulled this assert from the netfx implementation. My thought would be that it probably has to deal with something related to calling conventions. However, I don't see anything x86-specific in the runtime's fixup code, so I don't see why this assert is needed. I'll remove it.

src/corehost/cli/ijwhost/ijwhost.cpp
jkoritzinsky: I'll go remove them.

src/corehost/cli/ijwhost/ijwhost.cpp
jkoritzinsky: In your example the second error cannot occur, in which case I agree with you. I'm totally fine "swallowing" the trace in the initial loader code.

Here's an example of the second case I came up with:

```c
void DoStuff(int(*callback)(int) );

int main()
{
     HMODULE mod = LoadLibrary(W("my-ijw-assembly.dll")); // In the first case, this is the call that will fail.
     int(*myFunc)(int) = GetProcAddress(mod, "MyFunc");
     DoStuff(myFunc);
}

void DoStuff(int(*callback)(int))
{
      // Lots of interesting code here
      // More interesting code

      callback(5); // In the second case the exception would be thrown here if this is the first call into managed code in the IJW assembly.
}
```

Since we're more likely than not going to take down the process with an unhandled exception in this case and also the fact that we're basically inserting all of the runtime startup code into an arbitrary stack somewhere in the user's process, I think silently failing here would not be good since unlike the COM case, there's no good way for us to notify the user that the runtime initialization failed in this case without logging.

src/corehost/cli/ijwhost/corhdr.h
jkotas: Yes, I think it would be nice to make this code as simple as possible, depending on as few things as possible.

src/corehost/cli/ijwhost/ijwthunk.cpp
jkoritzinsky: Yep this manifests as a load failure, and if loading via `Assembly.Load` and family, it throws an exception just like if `Assembly.Load` failed on an IL-only assembly load.

AaronRobinsonMSFT: Sounds good from the managed side, but how about the native side? If I am in native and I make a call into a managed function, how does this come back to native? Can I assume the exception is marshalled as a structured exception or a C++ exception? I am concerned about crashes customers observe in their products and how well it can be diagnosed from a dump or log. I will say if an exception is thrown more than likely the process is coming down - so that is enough for me as long as that is true.

Issue #5181, opened 2019-02-15
WindowsDesktop Targeting Pack: create standalone installers


dagood:
Catch-all for WindowsDesktop targeting pack installers not already split into other issues. This includes standalone bundle installers. (Windows and macOS.)

dagood:
Moving to future to match #5174. WindowsDesktop MSI, macOS component pkg, and Linux packages are already implemented.

dagood:
This work is obviated by the new 3.1 installer. Installing downlevel targeting packs in the new 3.1 installer may be an interesting scenario.

/cc @johnbeisner @KathleenDollard

Issue #5179, opened 2019-02-15
WindowsDesktop Targeting Pack: create nupkg


dagood:

dagood:
Completed by https://github.com/dotnet/core-setup/pull/5214.

PR #5214, opened 2019-02-20
Add WindowsDesktop targeting pack nupkg, refactor pkg projects

dagood:
Adds WindowsDesktop targeting pack nupkg: https://github.com/dotnet/core-setup/issues/5179. This is done by taking bits from the existing Microsoft.WindowsDesktop.App. (This PR doesn't add installers, that's tracked by https://github.com/dotnet/core-setup/issues/5180 / https://github.com/dotnet/core-setup/issues/5181.)

To reduce the number of files and duplication needed to do that, I refactored the pkgproj arrangement and pulled common props/targets together. There may be more that needs to be pulled out later when runtime packs are added in.

Also includes a few fixes that would be merge conflicts with the refactor:
* Stop moving all targeting pack files to `data/`. Let them sit in the same place they would normally.
* Change the Pack nupkg data type from Data to DotnetPlatform.

I'd suggest looking at each commit separately. 🙂 

Example WindowsDesktop targeting pack nupkg: [Microsoft.WindowsDesktop.App.Ref](https://github.com/dotnet/core-setup/files/2885824/Microsoft.WindowsDesktop.App.Ref.3.0.0-preview4-27420-0.nupkg.zip). Diffing all the other nupkgs against a build of the commit I based this on shows no unintended changes.

/cc @nguerrera @dsplaisted @merriemcgaw


src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+
+  <PropertyGroup>
+    <FrameworkPackageName>Microsoft.WindowsDesktop.App</FrameworkPackageName>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.WindowsDesktop.App">
```
ericstj: Shouldn't you instead be picking up the packages that make this thing up and replacing it?  That way we can delete dotnet-trusted.

dagood: I was going for an incremental approach for now, but I'll try moving that over here in case it's not much worse to get it all done at the same time.

src/pkg/projects/depproj.targets
```diff
@@ -0,0 +1,109 @@
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
ericstj: Can you rename this file to something else?

src/pkg/projects/depproj.targets
```diff
@@ -0,0 +1,109 @@
...
+  <!--
+    Shared targets specific to dependency projects (depproj).
+  -->
+
+  <!-- Fetches all the file items from the packages that we want to redist -->
+  <Target Name="GetFilesFromPackages">
+    <ItemGroup Condition="'$(NuGetRuntimeIdentifier)' != ''">
+      <!-- RID-specific: include all runtime files except files in Microsoft.TargetingPack.Private.WinRT -->
+      <FilesToPackage Include="@(ReferenceCopyLocalPaths)" Condition="'%(ReferenceCopyLocalPaths.NuGetPackageId)' != 'Microsoft.TargetingPack.Private.WinRT'">
```
ericstj: Can you delete this condition?  We should fetch Windows.winmd in a different way so that this isn't needed.

src/pkg/projects/depproj.targets
```diff
@@ -0,0 +1,109 @@
...
+
+      <!-- include all doc files -->
+      <_docFilesToPackage Include="%(FilesToPackage.RootDir)%(FilesToPackage.Directory)**\%(FilesToPackage.FileName).xml" />
+
+      <FilesToPackage Include="@(_docFilesToPackage)">
+        <TargetPath>ref/$(PackageTargetFramework)/%(RecursiveDir)</TargetPath>
+      </FilesToPackage>
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(IncludeSourceFilesInPackage)' == 'true'">
```
ericstj: I think we can delete this now.

src/pkg/projects/depproj.targets
```diff
@@ -0,0 +1,109 @@
...
+    </ItemGroup>
+
+    <!-- Add versions file with the hashes of the repos we consume -->
+    <ItemGroup>
+      <FilesToPackage Include="$(IntermediateOutputPath)\$(FrameworkPackageName).versions.txt">
+        <TargetPath></TargetPath>
+      </FilesToPackage>
+    </ItemGroup>
+
+    <!-- on windows workaround max-path -->
```
ericstj: Remove all this since we remove source flowing.

src/pkg/projects/depproj.targets
```diff
@@ -0,0 +1,109 @@
...
+    <WriteLinesToFile Lines="@(_VersionsFileLines)"
+                       File="$(IntermediateOutputPath)\$(FrameworkPackageName).versions.txt"
+                       Overwrite="true"/>
+  </Target>
+
+  <!-- Target overrides (can't be shared with pkgproj) -->
+
+  <Target Name="Build" DependsOnTargets="GenerateHashVersionsFile;GetFilesToPackage" />
+
+  <!-- remove when fixing https://github.com/dotnet/buildtools/issues/1273 -->
```
ericstj: Still applicable?

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
ericstj: I see.  That's fair, but make sure you understand what's going on over there so you can see the path forward and what needs to be shared vs not.  You'll find that a lot of stuff happening in dotnet-trusted overlaps with stuff that's going on here.

dagood: It turns out there's DotNet-Trusted changes I need first to make all the bits available--filed https://github.com/dotnet/core-setup/issues/5219 to track this. I'm going to look a bit more at this in case there's something I'm missing but I don't plan to block this PR on that.

src/pkg/projects/depproj.targets
dagood: Sure, but what's bad about the current name so I don't just give it another bad one? 😄

src/pkg/projects/depproj.targets
dagood: I can make the package reference conditional so it's enabled in `src.builds` for crossgen but not `pkg.builds`. `SkipDesignTimePackageReferences` seems like a reasonable thing to call this.

Smells hacky to me, (specifically, restoring the same project multiple times with different references) but it's simple, removes this condition, and at least it isn't unclear in the source what's going on.

Another idea is to fit some project earlier in the build that specifically restores this package and plops the file in an intermediate path for the depproj to use. I suppose this would be similar to the `external/winrt` approach in CoreFX. This would take some more time to implement--I don't see precedent in Core-Setup.

I'll put the condition approach in this PR for now. Moving it out seems like something that can be done later.

src/pkg/projects/depproj.targets
dagood: Not specific enough? Maybe `framework.dependency.targets`, `framework.packaging.targets`? Shared prefix also seems nice to tell what they're for at a glance.

src/pkg/projects/netcoreapp/src/netcoreapp.depproj
```diff
@@ -110,51 +93,22 @@
...
-    <Error Condition="!Exists('$(CoreFxVersionFile)')" Text="Failed to find the corefx version.txt file at $(CoreFxVersionFile)" />
-
+  <Target Name="GetDependencyVersionFiles" DependsOnTargets="GetPackagePaths;ResolveNuGetPackages">
     <ItemGroup>
-      <_VersionsFileLines Include="core-setup $(LatestCommit)" />
-      <_VersionsFileLines Include="corefx $(CoreFxHash)" />
-      <_VersionsFileLines Include="dotnet/versions/corefx $(CoreFxCurrentRef)" />
-      <_VersionsFileLines Include="coreclr $(CoreClrHash)" Condition="'$(CoreClrHash)'!=''" />
-      <_VersionsFileLines Include="dotnet/versions/coreclr $(CoreClrCurrentRef)" Condition="'$(CoreClrHash)'!=''" />
+      <DependencyVersionFile
```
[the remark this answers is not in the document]
dagood: Fixing. I'm also making it a little more strict than the original by requiring the CoreCLR one.

src/pkg/projects/depproj.targets
ericstj: depproj.targets overlaps with a target named the same thing elsewhere that implements the basic depproj functionality.

Both of your suggestions seem reasonable.

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
ericstj: Sounds good.  Thanks for looking into it and identifying the work.

src/pkg/projects/depproj.targets
ericstj: Cheaper fix would be to make the PackageReference ExcludeAssets=All and then just grab the path from the PackagesDir.  I'm ok with not using nuget resolution for this file.

dagood: I gave `PrivateAssets=All` a try earlier which didn't seem to affect `PrereleaseResolveNuGetPackageAssets`. I'll try `ExcludeAssets`. Agreed it'd be cheaper.

src/pkg/projects/depproj.targets
dagood: It looks like it actually isn't an override, but it's still used. Removing it makes the pkg build fail during `GenerateRuntimeDependencies` because `DetermineRuntimeDependencies` isn't defined in the depproj. (Since only pkgproj gets buildtools `Packaging.targets`.)

I don't follow (1) and (2) in the buildtools issue, it looks like (3) hasn't happened, though: "We call DetermineRuntimeDependencies on all references in the lineup package but should only call it on pkgproj references."

src/pkg/projects/depproj.targets
dagood: Worked! Thanks.

src/pkg/projects/depproj.targets
ericstj: I see, bummer.  Sorry for that.

dagood:
Thanks a bunch for the review, Eric!

@dsplaisted @nguerrera the build is https://dev.azure.com/dnceng/internal/_build/results?buildId=103216. Once it's done and you have time, let me know if you have any feedback about the output `Microsoft.WindowsDesktop.App.Ref`.

(Note that this is not in preview3.)

Issue #5213, opened 2019-02-20
Remove or neuter Microsoft.DotNet.PlatformAbstractions


danmosemsft:
As agreed more broadly (with @dleeapho et al) we are committed to remove (in some sense) Microsoft.DotNet.PlatformAbstractions from core-setup in 3.0. 

After https://github.com/dotnet/corefx/issues/31002 is completed, all its functionality will have equivalents in CoreFX proper - albeit in different shapes.

(Note that Microsoft.DotNet.DependencyModel is not included here - it must run on older runtimes, so it remains in core-setup.)

Options for M.DN.PA:

1. Remove it, and document alternative functionality. Because the existing package remains on NuGet, we avoid breaking existing code.
1. Retain it as is, but replace its implementation by calls into the CoreFX functionality, if semantically possible.
1. As above, but also move into CoreFX.

@eerhardt @dseefeld is the above correct? thoughts?

dagood:
3 seems like the ideal plan.

We can also start with 1, then do 3 later if we end up needing to fix a breaking change. Waiting until a break adds uncertainty to future plans, though, so it seems safer to do it sooner.

eerhardt:
I don't think we should retain it, unless we have a strong reason for it.

[Here](https://github.com/dotnet/corefx/issues/31002#issuecomment-404331527) is my analysis of all the public functionality in this API. It can be boiled down to 3 public APIs that are necessary. Those 3 APIs are being proposed in https://github.com/dotnet/corefx/issues/31002. So if we get those 3 APIs in `corefx`, then I think we can delete this library completely.

The only issue I see with that plan is that `PlatformAbstractions` supports `netstandard1.3` today. It will be a breaking/functionality loss to only support those 3 APIs on `.NET Core 3.0` going forward. I believe it will be better for us to delete this library, but it may negatively impact our customers.

danmosemsft:
I also prefer 1. Given that we will not innovate in it, preserving it seems like all cost and no value.  We still have the option of servicing it if we have to.

If I understand right, deleting it need not block on the new API - any code already working will continue working. We can just delete it?

eerhardt:
We can't just delete it.

I think we need replacement APIs first, especially for `RuntimeEnvironment.GetRuntimeIdentifier()`. What are we going to tell things like the CLI to use? The old `2.1` PlatformAbstractions?

We will also need `Microsoft.Extensions.DependencyModel` to stop using it.

[...]

PR #5218, opened 2019-02-21
Remove DependencyModel's dependency on PlatformAbstractions

eerhardt:
We are planning on removing PlatformAbstractions. The first step is to remove the dependency on it in DependencyModel. The dependency isn't necessary.

Working towards #5213

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/ApplicationEnvironment.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/ApplicationEnvironment.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.Extensions.DependencyModel
+{
+    internal static class ApplicationEnvironment
+    {
+        public static string ApplicationBasePath { get; } = GetApplicationBasePath();
+
+        private static string GetApplicationBasePath()
+        {
+            var basePath =
+#if NET451
+                (string)AppDomain.CurrentDomain.GetData("APP_CONTEXT_BASE_DIRECTORY") ??
+                AppDomain.CurrentDomain.BaseDirectory;
+#else
+                AppContext.BaseDirectory;
+#endif
+            return Path.GetFullPath(basePath);
+        }
+    }
+}
--- a/src/managed/Microsoft.Extensions.DependencyModel/ApplicationEnvironment.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/ApplicationEnvironment.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.DependencyModel
 
         private static string GetApplicationBasePath()
         {
-            var basePath =
+            string basePath =
 #if NET451
                 (string)AppDomain.CurrentDomain.GetData("APP_CONTEXT_BASE_DIRECTORY") ??
                 AppDomain.CurrentDomain.BaseDirectory;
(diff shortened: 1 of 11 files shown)


pranavkm:
@bricelam - looks like this https://github.com/aspnet/EntityFrameworkCore/blob/f38686a5974ab94463ccd68484e9133a13e7e13f/src/EFCore.Sqlite.Core/Infrastructure/SpatialiteLoader.cs#L15 code might be relying on the transitive reference to get to `RuntimeEnvironment`.

src/managed/Microsoft.Extensions.DependencyModel/ApplicationEnvironment.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
steveharter: License doesn't follow the new header (I realize this is a copy of the old code)... Applies to other files as well.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -39,6 +43,7 @@
     <Otherwise>
       <ItemGroup>
         <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+        <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.0.0" />
```
steveharter: Why was this version selected (2.5 years old) -- because it supports older TFMs?

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
```diff
@@ -50,16 +55,11 @@
     </Otherwise>
   </Choose>
 
-  <ItemGroup>
-    <ProjectReference Include="..\Microsoft.DotNet.PlatformAbstractions\Microsoft.DotNet.PlatformAbstractions.csproj" />
-  </ItemGroup>
-
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' Or '$(TargetFramework)' == 'netstandard1.6' ">
+    <PackageReference Include="System.AppContext" Version="4.1.0" />
```
steveharter: Again I assume these package versions were selected based on best fit.

src/managed/Microsoft.Extensions.DependencyModel/ApplicationEnvironment.cs
danmosemsft: This is tracked by https://github.com/dotnet/core-setup/issues/5224. I see there are 100+ files in this repo that need the update.

danmosemsft:
@eerhardt can you comment on @steveharter question about package version? Other stuff is resolved.

src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj
eerhardt: Both of these versions are what is used in `PlatformAbstractions` today, which we were getting through transitive dependencies:

https://www.nuget.org/packages/Microsoft.DotNet.PlatformAbstractions/3.0.0-preview-27324-5

```
System.AppContext (>= 4.1.0)
System.Collections (>= 4.0.11)
System.IO (>= 4.1.0)
System.IO.FileSystem (>= 4.0.1)
System.Reflection.TypeExtensions (>= 4.1.0)
System.Runtime.Extensions (>= 4.1.0)
System.Runtime.InteropServices (>= 4.1.0)
System.Runtime.InteropServices.RuntimeInformation (>= 4.0.0)
```

I am keeping these versions the same as they were before my change.

PR #5228, opened 2019-02-21
Cleanup native build

vitek-karas:
Refactor cmake files to avoid code duplication.
Fix comhost to have native resources on Windows - that is version and other properties.
Turn on parallel build on Windows - on my machine the native build took 125 second before and 40 after this change.


AaronRobinsonMSFT:
@vitek-karas Oops. Looks like during this work, we all missed the inadvertent loss of the `Exports.def` file for `comhost.dll` - alas the managed COM hosting scenario is now broken because the host no longer has the required exports.

PR #5248, opened 2019-02-25
[WIP] Enabling the upgrade of the .NET Core Runtime bundle in a product-band

ghost:
Enabling the 'upgrade' of the .NET Core Runtime bundle in a 'product-band'.

Addresses: https://github.com/dotnet/core-setup/issues/4188
Related: https://github.com/dotnet/core-sdk/pull/530

**BranchInfo.props
src/pkg/packaging/dir.props**
The upgrade code is based on: "an arbitrary string + the .Net Core product-band [3.0] + the RID"
In this way, the upgrade codes, x64 & x86, will correctly change when the SDK product-band changes.

**src/pkg/packaging/windows/package.targets**
Send the stable upgrade code to the bundle generation script.

@dotnet/dotnet-cli

### DIFF
--- a/BranchInfo.props
+++ b/BranchInfo.props
@@ -16,6 +16,7 @@
     <Channel>master</Channel>
     <ContainerName>dotnet</ContainerName>
     <ChecksumContainerName>$(ContainerName)</ChecksumContainerName>
+    <ProductBandVersion>$(MajorVersion).$(MinorVersion)</ProductBandVersion>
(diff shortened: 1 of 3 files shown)


dagood (approved):
We'll need to have the preview1, 2, and 3 upgrade codes as `RelatedBundle`s, right? I'm ok waiting for us to know the preview3 code to implement it, just need to make sure it stays on the radar for preview4 release.

PR #5252, opened 2019-02-26
Use partial crossgen for select assemblies on Linux

MichalStrehovsky:
Measurements show that doing partial crossgen on these assemblies captures about two thirds of the size savings we could achieve by partially crossgenning everything.

Only doing this for Linux, since Linux containers are more size-conscious.


sergiy-k:
@MichalStrehovsky, thank you! @brianrob, do you have any objections against merging this change?

Issue #5254, opened 2019-02-26
Add signing validation infrastructure to master, release/3.0


dagood:
Motivated by https://github.com/dotnet/core-setup/issues/5244, "This is like the 3rd or 4th time this has happened recently."

In release/2.1 and release/2.2, we had signing validation. The port to YAML didn't include this.

We don't use Arcade SDK yet, so this is likely throwaway work to port the build pipeline implementation to yaml. Uptaking Arcade SDK and its validation should obviate this if done first.

AaronRobinsonMSFT:
@dagood This is only part of the issue. If the AspNet repo has other invariants that must be satisfied it too needs to have a CI for updating to a new package.

dagood:
Noted, feel free to link when there's an issue filed for those, but that doesn't impact the work in Core-Setup, right?

AaronRobinsonMSFT:
@dagood Agreed.

dagood:
For short-term continuous validation we could have the build simply download https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/microsoft.dotnet.signcheck/1.0.0-beta.19251.2/microsoft.dotnet.signcheck.1.0.0-beta.19251.2.nupkg and run it on our (Windows) zips. It's self-contained (runs on full framework) so it's easy without Arcade. This is what I did locally.

/cc @natemcmaster

dagood:
We also need to add anti-signing validation for `apphost.exe` and `comhost.dll`. (Not just skip them.)

PR #5255, opened 2019-02-26
Remove comhost from the shared framework generation.

### LINKED ISSUE #5253: Microsoft.NETCore.App shared runtime contains comhost.dll
Version:
3.0.0-preview4-27426-2

Considering https://github.com/dotnet/sdk/pull/2814 is there a reason comhost.dll should be bundled with the runtime?

@vitek-karas @AaronRobinsonMSFT

jkoritzinsky:

Ensure that the comhost is not included in the shared framework. There's some work that may need to be done in the SDK as well to remove the need for the `<ToDelete />` element.

Fixes #5253

### DIFF
--- a/src/sharedFramework/sharedFramework.proj
+++ b/src/sharedFramework/sharedFramework.proj
@@ -58,6 +58,7 @@
       <ToDelete Include="$(SharedFrameworkNameAndVersionRoot)\framework.runtimeconfig.json" />
       <ToDelete Include="$(SharedFrameworkNameAndVersionRoot)\apphost$(ExeSuffix)" />
       <ToDelete Include="$(SharedFrameworkNameAndVersionRoot)\$(LibPrefix)hostfxr$(LibSuffix)" />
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT:
@jkoritzinsky This also needs to be put in the release/3.0 branch I think?

pakrym:
Is there a way to add a test that verifies the new behaviour?

dougbu:
We need this fix in release/preview3 don't we? /cc @mmitche

AaronRobinsonMSFT:
@pakrym I don't see how. The issue here is one of signing validation that was lost in our Arcade transition. Issue https://github.com/dotnet/core-setup/issues/5254 will track the support.

jkoritzinsky:
PR #5256 is a port of this to the release/3.0 branch.

jkoritzinsky:
Sorry about the merge commit instead of squash. I thought I had already set that setting in this repo.

PR #5294, opened 2019-03-01
Remove usage of atomic and rely on lock instead.

AaronRobinsonMSFT:
- Please add a description for changes you are making.

Remove usage of atomic and replace with mutex.

- If there is an issue related to this PR, please add a reference to it.

See https://github.com/dotnet/coreclr/issues/22653

cc @vitek-karas @RussKeldorph

PR #5309, opened 2019-03-03
Remove uses of the shared and weak ptrs 

AaronRobinsonMSFT:

Remove all uses of `std::shared_ptr` and `std::weak_ptr` types since they don't seem to be supported on our ARM32 images.

https://github.com/dotnet/coreclr/issues/22653

### DIFF
--- a/src/corehost/cli/hostpolicy.cpp
+++ b/src/corehost/cli/hostpolicy.cpp
@@ -22,10 +22,10 @@ namespace
     bool g_init_done;
     hostpolicy_init_t g_init;
 
-    std::shared_ptr<coreclr_t> g_coreclr;
+    std::unique_ptr<coreclr_t> g_coreclr;
 
     std::mutex g_lib_lock;
-    std::weak_ptr<coreclr_t> g_lib_coreclr;
+    coreclr_t *g_lib_coreclr;
 
     class prepare_to_run_t
     {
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT:
First attempt at this was https://github.com/dotnet/core-setup/pull/5294. The previous PR addressed part of the issue, but the symbol in question was still being used.

AaronRobinsonMSFT:
cc @MichaelSimons

jeffschwMSFT:
@AaronRobinsonMSFT have we done a symbol dump on these binaries to ensure there are no additional use of atomic*?  Also, is it possible that this may be an onion peel of missing symbols?  Is it possible that once we removal all uses of atomic (even implicitly) that there may be another missing symbol?  I know that we are trying to track down an arm device to try these on, but is there anything additional that we can do to try to find any other possible missing symbols?

AaronRobinsonMSFT:
have we done a symbol dump on these binaries to ensure there are no additional use of atomic*?

@jeffschwMSFT Good question. I confirmed we are in the same place with respect to atomic* symbols as we were before. That may sound equivocating and it is. The symbol in question already existed in `libhostpolicy.so` in the breadcrumbs compilation unit. It was in a function that was never called or referenced, but it is in there. I wasn't able to fully get why it existed, but my guess is, it has to do with using `std:thread`. Either way, the atomic* symbol are in the same state - so yay.

>Is it possible that once we removal all uses of atomic (even implicitly) that there may be another missing symbol?

Possible, but unlikely. The symbol issue was failing immediately prior to entering `libcoreclr.so` at which point there isn't much going on in the host. There is the re-entrant logic in `AssemblyDependencyResolver`, but that code is almost identical to what is called to start up the CLR so I am relatively confident that is okay.

>is there anything additional that we can do to try to find any other possible missing symbols?

Sort of, but it could be costly in the short term. We could force link of all symbols and and fail using the compiler flag `--no-undefined`. I am more than a decade out on gcc flags and tricks, but I do remember using that - I am sure clang has an equivalent. The flag was helpful, but did require maintenance to ensure all libs were properly listed. Also, I was rather naive when I was using it so it may have other side-effects of which I am unaware.

The absolute best thing we can do is enable at least smoke testing on all platforms and/or get partners/stakeholders to consume bits early and often so we avoid the last minute urgency.

PR #5334, opened 2019-03-05
Update to an SDK version with COM host support

AaronRobinsonMSFT:
cc @jkoritzinsky @jeffschwMSFT @vitek-karas @dagood


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "2.1.300"
+    "dotnet": "3.0.100-preview4-010597"
```
dagood: Does the preview3 build `3.0.100-preview3-010431` have this support?

To be sure that we don't break the source-build N-1 chain ([description of N-1](https://github.com/dotnet/source-build/blob/master/Documentation/EngineeringPlanLinuxDistroRepos.md#n-1-toolset)), we should stick to releases.

/cc @dseefeld @crummel @dleeapho

AaronRobinsonMSFT: That is a good question. I don't know, let me see what I can find out.

dagood: Oh, and this is where the actual value is, since Core-Setup isn't on Arcade SDK yet:

https://github.com/dotnet/core-setup/blob/e9692c95240cf23da279fb3f69942591674e6d44/DotnetCLIVersion.txt#L1

I think the global.json is only here to prevent Arcade dependency flow from failing.

AaronRobinsonMSFT:
Depends on #5337 - specifically at least a preview 4 version of the SDK-3.0

dagood:
> preview 4 version

😞 We can't upgrade to that until we release a preview 4 SDK that Red Hat can build from source. Sounds like it ended up that preview 3 doesn't have the support you need?

jkoritzinsky:
Yeah the PR adding the SDK support for creating COM hosts didn't make the Preview 3 cut.

AaronRobinsonMSFT:
>We can't upgrade to that until we release a preview 4 SDK that Red Hat can build from source.

@dagood That is absolutely okay. This was something I was doing because I had cycles and was going to try and make some progress on. We can wait to do the work.

Issue #5341, opened 2019-03-05
ARM CI flaky in auto-update PRs: failure to get Microsoft.NETCore.Platforms


dagood:
https://github.com/dotnet/core-setup/pull/4881#issuecomment-448456974
https://github.com/dotnet/core-setup/pull/5111#issuecomment-462003939
https://github.com/dotnet/core-setup/pull/5247#issuecomment-467222014
https://github.com/dotnet/core-setup/pull/5328#issuecomment-469907199

This kind of thing used to happen with MyGet because it takes some time for them to index the packages after we pushed them. Not specifically to the ARM legs, though. I wouldn't expect this with the blob feed, because Maestro++ shouldn't be creating the upgrade PR until after blob feed publish is definitely completely done.

dagood:
I bet NuGet is using an out of date cache. According to https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders, the http cache is kept for 30 minutes. The ARM builds are (unlike other legs!) on static machines rather than recycled VMs, where some build also using the dotnet-core feed could have very easily run within 30 minutes of our CI starting and refreshed the cache.

Deleting the cache (`%LocalAppData%\NuGet\v3-cache`) as a CI step seems like the solution.

dagood:
After https://github.com/dotnet/core-setup/pull/5343 was merged, this hasn't happened for 4 update CI runs. Closing until it repros again, if it does.

PR #5343, opened 2019-03-06
Try to fix ARM CI flakiness by removing NuGet http cache for Windows builds

dagood:
> Windows builds (specifically ARM builds in CI) use static machines, which may have an out of date cache that NuGet considers alive. Ensure the cache is not used by deleting it.

See https://github.com/dotnet/core-setup/issues/5341. This isn't backed by any info about what's on the machines, just going to give it a shot because it sounds pretty likely to me.

/cc @wtgodbe @safern @danmosemsft, maybe this is a problem in other repos too?

### DIFF
--- a/eng/jobs/windows-build.yml
+++ b/eng/jobs/windows-build.yml
@@ -52,6 +52,11 @@ jobs:
         continueOnError: false
         condition: and(succeeded(), in(variables['SignType'], 'real', 'test'))
 
+    # NuGet's http cache lasts 30 minutes. If we're on a static machine, this may interfere with
+    # auto-update PRs by preventing the CI build from fetching the new version. Delete the cache.
+    - powershell: Remove-Item -Recurse -ErrorAction Ignore "$env:LocalAppData\NuGet\v3-cache"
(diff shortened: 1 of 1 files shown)


safern:
> /cc @wtgodbe @safern @danmosemsft, maybe this is a problem in other repos too?

I haven't seen it but we don't run Windows ARM in CI, only official builds

dagood:
The thing that I think matters is the machine being static (e.g. `dotnet-external-temp`), rather than recycled or done in a Docker container.

safern:
>The thing that I think matters is the machine being static (e.g. dotnet-external-temp), rather than recycled or done in a Docker container.

Ah I see. We use `Hosted VS2017` for CI and dotnet-internal-temp for official builds. Might be worth for you to switch to `Hosted VS2017` since those are recycled.

dagood:
Would certainly like to, it doesn't have dependencies we need though: https://github.com/dotnet/core-setup/issues/4997. 😄 Haven't had time to revisit. Agreed it makes sense for this not to affect you.

Issue #5112, opened 2019-02-08
Add linux-musl-arm64 runtime (Alpine arm64) official build leg


dagood:
Once CoreCLR and CoreFX produce linux-musl-arm64 artifacts, we can add the leg to Core-Setup.

A Docker image for the builds was created by https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/73.

dagood:
Completed by https://github.com/dotnet/core-setup/pull/5345.

PR #5345, opened 2019-03-06
Add linux-musl-arm64 (Alpine) build leg

dagood:
https://github.com/dotnet/core-setup/issues/5112

CoreFX PR: https://github.com/dotnet/corefx/pull/35778
CoreFX build was brought in by https://github.com/dotnet/core-setup/pull/5328

I'm planning to do a mock build to validate framework package content and publish behavior.

/cc @safern @richlander @MichaelSimons

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -122,6 +122,17 @@ jobs:
 
 - template: /eng/jobs/bash-build.yml
   parameters:
+    additionalMSBuildArgs: /p:OutputRid=linux-musl-arm64
+    crossBuild: true
+    displayName: Build_Linux_Arm64_Alpine37
+    dockerImage: microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-cross-arm64-alpine10fcdcf-20190208200917
+    additionalRunArgs: -e ROOTFS_DIR=/crossrootfs/arm64
+    portableBuild: true
+    skipTests: true
+    targetArchitecture: arm64
+
+- template: /eng/jobs/bash-build.yml
+  parameters:
     additionalMSBuildArgs: /p:OutputRid=linux-musl-x64
     displayName: Build_Linux_x64_Alpine36
     dockerImage: microsoft/dotnet-buildtools-prereqs:alpine-3.6-3148f11-20171119021156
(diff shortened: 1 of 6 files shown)


azure-pipelines.yml
```diff
@@ -120,6 +120,17 @@ jobs:
 #       skipTests: true
 #       targetArchitecture: armel
 
+- template: /eng/jobs/bash-build.yml
+  parameters:
+    additionalMSBuildArgs: /p:OutputRid=linux-musl-arm64
+    crossBuild: true
+    displayName: Build_Linux_Arm64_Alpine37
+    dockerImage: microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-cross-arm64-alpine10fcdcf-20190208200917
```
MichaelSimons: Consider pulling this from mcr.microsoft.com/dotnet-buildtools/prereqs

eng/jobs/finalize-publish.yml
```diff
@@ -7,6 +7,7 @@ jobs:
     condition: and(
                   succeeded('Build_Linux_Arm'),
```
MichaelSimons: Do these conditions need to be specified at all?  From https://docs.microsoft.com/en-us/azure/devops/pipelines/process/multiple-phases?view=azure-devops&tabs=yaml#conditions

> By default, a job runs if it does not depend on any other job, or if all of the jobs that it depends on have completed and succeeded.

dagood: Heh, I actually just put in a change in my mock branch to remove this. I think it was necessary at the time because the default is all-or-nothing, and the condition needed to include `ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')`. However, this clause looks like it's handled by a template condition, so I think it can be completely deleted.

I'm also moving the dependsOn into the main file for better visibility.

dagood:
Poking at the nupkgs from the mock build, it looks fine. Identity package points to the runtime package, which exists. I didn't try to run it. Merging to get an official build out. Thanks for the reviews!

dagood:
Successful build:
https://dev.azure.com/dnceng/internal/_build/results?buildId=115788
`3.0.0-preview4-27506-5`

PR #5391, opened 2019-03-08
Split libhost.h into multiple headers and cleanup

vitek-karas:
`libhost.h` contains 3 classes, the `host_interface` which is the interface between `hostfxr` and `hostpolicy` and then the `hostfxr` "copy" of it and the `hostpolicy` "copy" of it. As such it introduces lot of coupling between source files and includes headers all over the place.

This change splits the header into 3 where only the `host_interface.h` is now included by both `hostfxr` and `hostpolicy` and it clearly defines the interface between the two. The rest is specific to either library.

Deleted some functions which were not used by anything.
Moved the rest of the functions in `libhost.cpp` into the places where they were used.

Some of the small cleanup.

This change is pure refactoring (moving code around), no functional changes to anything.


AaronRobinsonMSFT:
@jkoritzinsky I think this is going to play hell with your IJW changes.

src/corehost/cli/fxr/corehost_init.h
```diff
@@ -0,0 +1,173 @@
...
+        hi.host_info_host_path = m_host_info_host_path.c_str();
+        hi.host_info_dotnet_root = m_host_info_dotnet_root.c_str();
+        hi.host_info_app_path = m_host_info_app_path.c_str();
+
+        return hi;
+    }
+
+private:
+
+    static void make_cstr_arr(const std::vector<pal::string_t>& arr, std::vector<const pal::char_t*>* out)
```
AaronRobinsonMSFT: I know this is a refactor, but can we please get away from private static constructs. They are completely unnecessary and break incremental build semantics when one is simply updating implementation in a compilation unit rather than mucking with the interface. Just move the `make_cstr_arr` function into the compilation unit and it will work.

jkoritzinsky:
Yeah I'm mentally preparing for a day of resolving merge conflicts. I don't think it'll be too bad though.

src/corehost/cli/fxr/corehost_init.h
vitek-karas: I'll move all the bodies into a cpp… in a different change to keep things organized.

vitek-karas:
@jkoritzinsky I'll help as much as I can with the merge, just ping me.

PR #5397, opened 2019-03-08
Use the CMake install target to place binaries instead of MSBuild post-processing

jkoritzinsky:
We generate an install target to place our binaries and symbols in the right location via CMake, but we never use it. Additionally, we effectively replicate that functionality in `core-host/build.proj`.

This PR switches the native build to place the native output via the CMake install target and removes the MSBuild scripting that was duplicating it.

Also fixes a few build bugs in our RID construction in `src/corehost/build.sh` with cross-compilation.

### DIFF
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -236,6 +236,7 @@ fi
 
 __cmake_defines="${__cmake_defines} -DVERSION_FILE_PATH:STRING=${__versionSourceFile}"
 
+mkdir -p $__intermediateOutputPath
 pushd $__intermediateOutputPath
 
 echo "Building Corehost from $DIR to $(pwd)"
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -11,7 +11,7 @@ init_rid_plat()
             if [ -e $ROOTFS_DIR/etc/os-release ]; then
                 source $ROOTFS_DIR/etc/os-release
                 __rid_plat="$ID.$VERSION_ID"
-                if [[ "$ID" == "alpine" ]]
+                if [[ "$ID" == "alpine" ]]; then
                     __rid_plat="linux-musl"
                 fi
             fi
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -266,4 +266,4 @@ fi
 popd
 
 set +x # turn off trace
-cmake --build $__intermediateOutputPath --target install
+cmake --build $__intermediateOutputPath --target install --config $__configuration
--- a/src/corehost/build.sh
+++ b/src/corehost/build.sh
@@ -202,7 +202,6 @@ if [ -z $__commit_hash ]; then
     exit -1
 fi
 
-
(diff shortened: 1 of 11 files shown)


src/corehost/build.sh
```diff
@@ -196,9 +202,12 @@ if [ -z $__commit_hash ]; then
     exit -1
 fi
 
+
```
AaronRobinsonMSFT: Remove line.

src/settings.cmake
```diff
@@ -111,23 +111,12 @@ function(strip_symbols targetName outputFilename)
...
-    endif()
-
-    install(PROGRAMS ${install_source_file} DESTINATION .)
+function(install_symbols targetName destination_path)
     if(WIN32)
-        install(FILES ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>/${targetName}.pdb DESTINATION PDB)
+        install(FILES ${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>/${targetName}.pdb DESTINATION ${destination_path})
     else()
-        install(FILES ${strip_destination_file} DESTINATION .)
+        strip_symbols(${targetName} strip_destination_file)
```
AaronRobinsonMSFT: is `strip_symbols` a no-op for Windows from CMake?

src/corehost/cli/comhost/CMakeLists.txt
```diff
@@ -40,4 +40,5 @@ if (WIN32 AND (CLI_CMAKE_PLATFORM_ARCH_ARM OR CLI_CMAKE_PLATFORM_ARCH_ARM64))
     target_link_libraries(comhost Advapi32.lib Ole32.lib OleAut32.lib)
 endif()
 
-install_library_and_symbols (comhost)
\ No newline at end of file
+install(TARGETS comhost DESTINATION corehost)
+install_symbols(comhost corehost)
```
vitek-karas: Can we move this to `lib.cmake` similarly to how it's in `exe.cmake` ?

src/settings.cmake
jkoritzinsky: `strip_symbols` is one of our own functions. It's a no-op on Windows.

src/corehost/cli/comhost/CMakeLists.txt
jkoritzinsky: We could, but because we will have to add an additional install command in `ijwhost` anyway to install the `ijwhost.lib` file to the output directory, I prefer to keep the installs in the same file so they won't be split (and the main ones hidden in lib.cmake).

jkoritzinsky:
Yes, the output is produced to the exact same directories as before this change.

src/corehost/cli/comhost/CMakeLists.txt
vitek-karas: OK - makes sense.

src/settings.cmake
AaronRobinsonMSFT: Sounds good. Looking at the function it appears to already check for Windows. Since that is the case, we should always call it and reduce having redundant branches.

PR #5398, opened 2019-03-08
[release/2.1] Remove host RPM postinstall, install the symlink

dagood:
Ports https://github.com/dotnet/core-setup/pull/5380 and https://github.com/dotnet/core-setup/pull/5451 (fixup PR) to `release/2.1`. (No conflicts: change is identical.) After merging this fix, it'll merge into `release/2.2` as well.

See discussion in https://github.com/dotnet/core-setup/issues/5221.

/cc @leecow 

#### Description

In the dotnet-host RPM package, rather than using postinstall and postuninstall scripts to create and maintain a `/usr/bin/dotnet` symlink, have the RPM package install a symlink file directly. This reduces the complexity of the host package.

#### Customer Impact

Reason for the change: when the scripts don't set up `/usr/bin/dotnet` properly, you get `dotnet: command not found` when you try to run any `dotnet` commands. Removing the postinstall and postuninstall scripts eliminates a potential source of this problem.

Reason to port it: we merged this change into `master`, for 3.0+ RPM packages. If it isn't also ported into `release/2.1` and `release/2.2`, the symlink **will** be removed when downgrading the host (more info at https://github.com/dotnet/core-setup/pull/5380#issuecomment-471001755).

#### Regression?

No.

#### Risk

Minimal. This change is localized to the host RPM package and moves to a more reliable implementation. I've manually tested the `master` PR and this one to make sure the symlink is still created as expected.


vivmishra:
Approved for 2.1.10 and 2.2.4.

Issue #5454, opened 2019-03-12
Update readme.md download link table (3.0, 2.2, 2.0?)


dagood:
The readme is out of date:
* There isn't a 3.0 column.
* There isn't a 2.2 column.
* There is a 2.0 column, but that release is EOL.

There’s a readme table generation shell script I made for source-build's table that would make this table a **lot** easier to maintain. I’m going to bring that over.

@leecow

dagood:
> There’s a readme table generation shell script I made for source-build's table [...] I’m going to bring that over.

That script was too prescriptive and hard to maintain to reapply here. Ended up with a C# task and only generated the table, not the URLs: https://github.com/dotnet/core-setup/pull/5468.

PR #5468, opened 2019-03-13
Maintain readme table: add 3.0 + 2.2 columns, remove 2.0 column, add table tooling

dagood:
Fixes https://github.com/dotnet/core-setup/issues/5454

Adds 3.0 and 2.2 columns and removes the EOL 2.0 column. Link to how the readme will look on the repo home page: https://github.com/dagood/core-setup/tree/maintain-readme#daily-builds.

The wrapping in the cells got worse because of the extra column, but it's not as bad as I thought it would be. I could try some things pretty easily now that the table is generated, though. (Decreasing font size, sub/superscript, ....)

Note the 3.0 links and badges don't exist as of creating this PR--I fixed the channel just recently in https://github.com/dotnet/core-setup/pull/5459 and a build is going through publish right now.

The table tool doesn't go all the way to generating the links automatically. There's enough legacy and weirdness going on it doesn't seem worthwhile to spend more time to try to find the patterns in the URLs. There's also hope that Darc will generate readme-type stuff for us for 3.0+, making the current table maintenance-mode.

/cc @MichaelSimons


dagood:
Good idea, done. The 3.0 build is complete now, so I checked the table and fixed up a link. Should be good to go.

PR #5479, opened 2019-03-14
Restore support for UTF-8 BOM/Preamble on loading JSON.

bartonjs:
This brings in the changes from https://github.com/dotnet/corefx/pull/35896 to the copy in DependencyContextJsonReader.Utf8JsonReader.cs.

Since DependencyContextJsonReader.JsonTextReader.cs uses `new StreamReader(Stream)` this was a behavior that already worked previously that because broken by the move to Utf8JsonReader.

Now it's restored, and there's a test for it.

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Utf8JsonReader.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Utf8JsonReader.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Extensions.DependencyModel
             return ReadCore(reader);
         }
 
-        // Borrowed from https://github.com/dotnet/corefx/blob/ef23e3317ca6e83f1e959ab265a8e59fb8a6dcd9/src/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L176-L225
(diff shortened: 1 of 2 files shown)


src/managed/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.Utf8JsonReader.cs
```diff
@@ -3,13 +3,16 @@
...
 using System.Buffers;
+using System.Diagnostics;
 using System.IO;
 using System.Text.Json;
 
 namespace Microsoft.Extensions.DependencyModel
 {
     public partial class DependencyContextJsonReader : IDependencyContextReader
     {
+        private static ReadOnlySpan<byte> Utf8Bom => new byte[] { 0xEF, 0xBB, 0xBF };
```
eerhardt: Would this be better as a method? That way someone wouldn't make the mistake of calling it more than once in a method.

bartonjs: Perhaps.  The only pattern I've seen for a ReadOnlySpan over the program `.text` segment used a property; but the JIT or crossgen should see a property get just as a normal method... so it should work?

But for now I'm going with "this looks like it does in corefx" and there is "it looks like everywhere else".  But I'll definitely think on it going forward.

PR #5484, opened 2019-03-14
Remove MyGet feed restore sources

dagood:
Depend on blob feeds and the NuGet gallery only.

`NugetProjectModelPackageVersion` is used for build tasks, I just changed it to the nearest version on the gallery, 4.3.0.


wtgodbe:
Does this mean we plan to continue not publishing to myget as well?

dagood:
No, I think that's an independent issue. Microsoft builds should never restore from MyGet feeds in general, but MyGet is still the most reasonable way for someone to browse what we produce.

dagood:
This seems to have broken the official build publish step:
https://dev.azure.com/dnceng/internal/_build/results?buildId=125189

> /root/coresetup/publish/publish.proj(11,3): error MSB4019: The imported project "/root/coresetup/packages/microsoft.symboluploader.build.task/1.0.0-beta-63604-05/build/PublishSymbols.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.

Reverting for now.

Issue #5490, opened 2019-03-15
Recent Targeting Pack MSIs install to wrong location


dagood:
The MSI from our latest daily build installs targeting pack assets to the wrong location:

`C:\Program Files\`**`dotnet\targeting\packs`**`\Microsoft.NETCore.App.Ref\3.0.0-preview4-27515-03`
vs. the correct:
`C:\Program Files\`**`dotnet\packs`**`\Microsoft.NETCore.App.Ref\3.0.0-preview4-27515-03`

It looks like in https://github.com/dotnet/core-setup/pull/5369 I changed how the layout dir is generated to include the type of pack (`targeting`) for Deb/RPM generation but I missed updating the MSI code to account for it.

I found this while working on the WindowsDesktop targeting pack MSI (https://github.com/dotnet/core-setup/issues/5180), when I tried diffing the tables of the old MSI vs. the new MSI. This problem will be fixed with the PR for https://github.com/dotnet/core-setup/issues/5180.

FYI @dsplaisted @nguerrera in case you've hit this.

Issue #40, opened 2016-05-28
Registry entries left behind after uninstalling DotNetCore.1.0.0.RC2-SDK.Preview1-x64


leecow:
Don't know if we're leaving these behind of if they are Windows Installer artifacts

HKEY_CLASSES_ROOT\Installer\Dependencies\Dotnet_CLI_1.0.0.002702_x64
HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows NT\CurrentVersion\AppCompatFlags\Compatibility Assistant\Store
- C:\ProgramData\Package Cache\{2adee9e7-1f53-47dc-8e30-49ff0dec528e}\dotnet-dev-win-x64.1.0.0-preview1-002702.exe
- C:\Users\[username]\AppData\Local\Temp\{827CEE61-99A0-4091-974F-915706DF8B20}\.be\dotnet-dev-win-x64.1.0.0-preview1-002702.exe
- C:\Users\[username]\Downloads\DotNetCore.1.0.0.RC2-SDK.Preview1-x64.exe
HKEY_LOCAL_MACHINE\SOFTWARE\Classes\Installer\Dependencies\Dotnet_CLI_1.0.0.002702_x64

brthor:
@leecow What installer did you use, and how did you find that these files were being left behind?

piotrpMSFT:
@leecow can you provide @brthor with repro steps to get the results you see above? I assume you installed the preview1 bundle, then uninstalled something via ARP. How did you identify the diff specified above? What ARP entry did you uninstall?

leecow:
@brthor @piotrpMSFT This would have been installed via the SDK and/or Runtime MSI and uninstalled through ARP. No diff involved, discovered doing manual spelunking through the registry.

brthor:
First time digging into the MSIs, but with help from @jhendrixMSFT and @ericstj , tracked down the issue to the static component id here:
https://github.com/dotnet/cli/blob/rel/1.0.0/packaging/windows/clisdk/provider.wxs#L5

causing these components not to get removed properly on uninstall. I changed it to be auto generated and confirmed that sXs installation of cli still works, and that the keys are individually removed when the appropriate versions are removed via ARP. 

My understanding is that by sharing the same componentId, these components will be ref-counted and not removed properly when any version of the cli is still installed. However even if they're all uninstalled they may still not be removed because of this issue mentioned by @robmen (I did not test this scenario however).

[...]

PR #5491, opened 2019-03-15
Create targeting pack MSIs through pkgproj, including WindowsDesktop

dagood:
Resolves https://github.com/dotnet/core-setup/issues/5180.

Similar to the Targeting Pack nupkgs, Debian packages, and RPM packages, use `pkgproj` infrastructure to drive Targeting Pack MSI installer builds. This builds the installers for .NET Core and Windows Desktop.

I created a `src\pkg\projects\packaging-tools` dir:
* Gives the shared WiX files a clear place to go.
* Rough staging area for infra that seems shareable (eventually) through Arcade. I'm not planning to do much polishing for that in this PR, though, the main point is the WindowsDesktop MSI.

WiX toolset download and tool execution are also changed from Powershell scripts into MSBuild targets.

This PR also fixes https://github.com/dotnet/core-setup/issues/5490: the .NET Core Targeting Pack MSI had an unintended `targeting` dir in the path it installs to.

/cc @dsplaisted @nguerrera @johnbeisner


src/pkg/projects/packaging-tools/windows/provider.wxs
```diff
@@ -2,7 +2,7 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:dep="http://schemas.microsoft.com/wix/DependencyExtension">
   <?include "Variables.wxi" ?>
   <Fragment>
-    <Component Id="$(var.DependencyKeyId)" Directory="TARGETDIR" Win64="no" Guid="117B1755-5027-4257-A1A9-7C37D81C3D5F">
+    <Component Id="$(var.DependencyKeyId)" Directory="TARGETDIR" Win64="no" Guid="$(var.DependencyKeyGuid)">
```
johnbeisner: What is the functional purpose of the "var.DependencyKeyId" GUID?

dagood: (Took a journey for this one... turns out there's been a bug here that I can fix! 😄)

In the scope of this PR, I made this change so that both the .NET Core and Windows Desktop targeting packs can use the same `wxs` file but have unique GUIDs here.

From the ground up, I don't know why this `Component` has a project-specific GUID. I saw that each existing `provider.wxs` in Core-Setup has a unique GUID in the `Guid` attribute, and I've been keeping to that as the status quo.

Now... if I remove `Guid="$(var.DependencyKeyGuid)"` (to autogenerate), the build succeeds, and I see that the provider component's GUID is some generated value. I think this is ok, because this component puts down a creates a version-specific registry key (`Software\Classes\Installer\Dependencies\dotnet_targeting_pack_24.0.27515_x64`) that won't be shared with future builds. In fact, reading general MSI guidance, using a static component GUID is really bad when the artifact is being renamed: this messes up ref counting.

Looking at [Core-SDK's `provider.wxs`](https://github.com/dotnet/core-sdk/blob/master/src/redist/targets/packaging/windows/clisdk/provider.wxs), it explicitly uses the auto-GUID. In fact, tracing it back, the static GUID was removed in CLI's `provider.wxs` to fix a bug where registry entries are left behind: https://github.com/dotnet/core-setup/issues/40#issuecomment-225323170.

It turns out that that bug affects the Core-Setup MSIs: the reg key stays behind when I install and uninstall (for example) the host MSI.

I'm going to add a change that makes all `provider.wxs`s in Core-Setup auto-generate these GUIDs to fix that.

@johnbeisner let me know if that train of thought doesn't make sense, thanks for pointing this out!

PR #5515, opened 2019-03-20
Disable using clang 3.9 for linux ARM cross build

janvorli:
The clang 3.9 was found to be generating broken code when cross
building for arm/armel long time ago for coreclr repo builds.
It was generating e.g. incorrect code for interlocked operations.
Now with a recent change in core-setup (adding linking of libatomic.a),
it started to generate broken code in core-setup too (leading to
SIGILL in libhostpolicy.so).
This change disables usage of clang 3.9 for ARM cross building, like
we do in the coreclr repo.

Close https://github.com/dotnet/core-sdk/issues/1063

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,7 +91,7 @@ jobs:
   parameters:
     crossBuild: true
     displayName: Build_Linux_Arm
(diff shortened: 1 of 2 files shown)


janvorli:
Hmm, it seems our build docker image doesn't have clang 5.0 installed. That needs to be fixed first.

janvorli:
I've updated the docker image for arm linux build to be the same as the one in coreclr and it is ok now.

PR #5516, opened 2019-03-20
Improve roll forward spec and tests

vitek-karas:
Doc
* Fix description of existing roll forward behavior - release will roll forward to pre-release if Minor and Major is used.
* Added tables with sample roll forward behaviors describing existing code.
* Minor tweaks and clarifications

Tests
* Add a new test validating roll forward to latest patch and pre-release behavior
* Added comments to clarify what the tests do and why they expect certain outcomes


swaroop-sridhar:
I think it is a good idea to add a test case for mismatched shared-framework versions (#4947). I'm OK if this is done as a separate checkin.  Thanks.

vitek-karas:
@swaroop-sridhar The tests I added in the previous PR (this is just improving on that) contain quite a few cases which hit that bug. I had to comment all those out to make the tests pass. For example https://github.com/dotnet/core-setup/blob/e2ead1e8128bfc31af2b86688f86b014ecde1a75/src/test/HostActivationTests/FrameworkResolution/MultipleFrameworks.cs#L62

PR #5537, opened 2019-03-22
Publish Microsoft.Dotnet.Build.Bundle package

swaroop-sridhar:
This change publishes the Bundle tool as a package for consumption in the toolset/sdk repos.

The changes are:
* Enable package publishing for the Bundler
* Add a property in sdk.props so help find the tool easily from the MSBuild tooling
  (similar to https://github.com/mono/linker/pull/487/files)

### DIFF
--- a/src/managed/CommonManaged.props
+++ b/src/managed/CommonManaged.props
@@ -24,7 +24,7 @@
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' And '$(TargetFramework)' != 'netcoreapp2.0' ">
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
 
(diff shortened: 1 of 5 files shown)


src/managed/CommonManaged.props
```diff
@@ -24,7 +24,7 @@
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' != 'netstandard2.0' And '$(TargetFramework)' != 'netcoreapp2.0' ">
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
sbomer: Out of curiosity, do you know why this project or the other managed projects need to disable implicit framework references?

swaroop-sridhar: This project only targets netcoreapp2.0, so it doesn't disable implicit framework references.
The  `PlatformAbstraction` and `DependencyModel` projects disable it for netstandard1.3 and [include specific versions of packages](https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.DotNet.PlatformAbstractions/Microsoft.DotNet.PlatformAbstractions.csproj), but I don't know why.

PR #5574, opened 2019-03-27
Bundler Test: Run using AppHost

swaroop-sridhar:
The bundler test runs a self-contained-published test app twice:
once before and once after bundling.

The initial run (to sanity check the project) was done using
the dotnet host obtained from the SDK. However, the SDK dotnet seems
to be a X64 binary, which failed to run the test-app published for x86.
https://dev.azure.com/dnceng/public/_build/results?buildId=134281&view=logs

This change fixes the problem by invoking both runs using the published AppHost.


swaroop-sridhar:
CC: @dagood

PR #5589, opened 2019-03-27
Depend on M.NC.Platforms 2.1.3 instead of 2.2.0

wtgodbe:
The latest version of M.NC.Platforms from CoreFx release/2.2 is from November, whereas the Platforms package in 2.1 has been updated a couple of times since then (at will be updated again for the next 2.1 release). We should depend on the latest version of that package in Core-Setup 2.x, which means depending on the 2.1 version in both release/2.x branches. This should fix the update step so that we depend on the right version of the package.

We still need to decide whether or not we need to sim-ship the package out of both CoreFx 2.x branches, as many external folks depend on M.NC.Platforms 2.2.0 today.

@ericstj @dagood PTAL

https://github.com/dotnet/core-setup/issues/5562

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -16,7 +16,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <MicrosoftNETCorePlatformsPackageVersion>2.2.0</MicrosoftNETCorePlatformsPackageVersion>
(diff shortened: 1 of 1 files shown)


wtgodbe:
Looks like both CoreFx 2.2 & CoreClr 2.2 also depend on M.NC.Platforms 2.2.0, so maybe we should think more about sim-shipping

wtgodbe:
Closing in favor of https://github.com/dotnet/corefx/pull/36464

PR #5594, opened 2019-03-27
Bundler: allow relative path for source directory

AustinWise:
Also fix a typo.

Without this change, the bundler has an internal error when a relative path is used as the source.

### DIFF
--- a/src/managed/Microsoft.NET.Build.Bundle/Bundler.cs
+++ b/src/managed/Microsoft.NET.Build.Bundle/Bundler.cs
@@ -54,6 +54,10 @@ namespace Microsoft.NET.Build.Bundle
                 throw new BundleException("Dirctory not found: " + OutputDir);
             }
 
+            // Convert relative paths to absolute paths.
+            SourceDir = Path.GetFullPath(SourceDir);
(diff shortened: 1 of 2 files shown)


AustinWise:
It seems like the shared framework is not building for some reason:

2019-03-28T00:01:02.2743087Z /root/coresetup/src/sharedFramework/lockedhost/lockedhost.csproj : error NU1102: Unable to find package Microsoft.NETCore.DotNetHostResolver with version (>= 3.0.0-preview4-27527-22) [/root/coresetup/src/sharedFramework/sharedFramework.proj]

swaroop-sridhar:
@dotnet-bot  test this please

swaroop-sridhar:
@AustinWise can you please add a test case for this? Here: https://github.com/dotnet/core-setup/blob/master/src/test/Microsoft.NET.Build.Bundle.Tests/BundleAndExtract.cs 

Thanks.

AustinWise:
@swaroop-sridhar That's a good point about the test case, I'll try to make a follow up PR when I get some time.

swaroop-sridhar:
Thanks @AustinWise

PR #5650, opened 2019-04-02
Add ijwhost.lib to host package.

jkoritzinsky:
I forgot to include the `ijwhost.lib` exports lib in the Microsoft.NETCore.DotNetAppHost package. C++/CLI will need this in the package for aquisition via NuGet.

### DIFF
--- a/src/corehost/cli/ijwhost/Exports.def
+++ b/src/corehost/cli/ijwhost/Exports.def
@@ -1,3 +1,3 @@
 EXPORTS
-    _CorDllMain            PRIVATE
+    _CorDllMain
     GetTokenForVTableEntry PRIVATE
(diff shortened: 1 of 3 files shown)


jkoritzinsky:
The build passed on this PR, but seems that the update from Azure DevOps got lost before reaching GitHub.

AaronRobinsonMSFT:
Is there any reason not to deploy PDBs for this lib? Having an incomplete codeview section in binaries can cause heck with some diagnostic tools. Not saying it should be deployed, but since this is an open source project, what is the harm?

jkoritzinsky:
The problem with including the pdb is that we were accidentally grabbing the same `ijwhost.pdb` twice, once for `ijwhost.dll` and once for `ijwhost.lib`. The two options I was thinking of were either removing duplicates, which I thought could lead to silent replacement of expected PDBs, or not including the PDB for the lib.

PR #5653, opened 2019-04-03
Bundler: Add test case for handling relative-path input

swaroop-sridhar:
Run the bundle-extract test where inputs are specified as:
 - Absolute paths
 - Relative paths terminated by directory separator
 - Relative paths not terminated by directory separator


swaroop-sridhar:
CC: AustinWise

swaroop-sridhar:
Add a test case for the fix in https://github.com/dotnet/core-setup/pull/5594

AustinWise:
Thanks for adding these test cases.

Issue #5665, opened 2019-04-03
Platform manifest incorrectly lists version '0.0.0.0' for some assemblies


dagood:
This causes downstream repos (ASP.NET Core in particular) to fail in some cases.

This appears to be caused by crossgen stripping out versions, then the manifest generation using those versions.

A workaround is to use the original version from the packages being crossgen'd + redistributed.

(More info in an email thread, this issue is mainly for tracking purposes.)

dougbu:
@Eilon what labels would you like to see on this issue?

dagood:
Adding `blocking-partner`, I've it seen used for issues blocking ASP.NET Core uptake before.

dougbu:
@dagood should this be closed now that @davidwrighton got dotnet/coreclr@283cab341daacc in?

Or, should this become a test issue i.e. cover avoiding similar problems in the future?

dagood:
This isn't fixed until Core-Setup uptakes a new CoreCLR build. The auto-update PR https://github.com/dotnet/core-setup/pull/5657 doesn't have one yet, but I think that's subscribed daily. I'll check if there's a new CoreCLR build and get an update through if so.

dagood:
There hasn't been a build yet. The newest CoreCLR build is hitting https://github.com/dotnet/core-eng/issues/5745: https://dev.azure.com/dnceng/internal/_build/results?buildId=142802.

dagood:
New builds ongoing with the crossgen fix:
https://dev.azure.com/dnceng/internal/_build/results?buildId=143031
https://dev.azure.com/dnceng/internal/_build/results?buildId=143167

dougbu:
Doesn't https://dev.azure.com/dnceng/internal/_build/results?buildId=143167 look like it will complete successfully?

To double-check: the "Update dependencies from https://github.com/dotnet/coreclr build 20190404.73" commit includes both of the fixes ASP.NET Core needs?
- dotnet/coreclr@283cab341daacc
- dotnet/corefx@53c0bae174088c3

dougbu:
Oh, I see https://dev.azure.com/dnceng/internal/_build/results?buildId=143167 won't get agents to do the final push

dougbu:
My second question relates to https://dev.azure.com/dnceng/internal/_build/results?buildId=143256

dagood:
[...]

PR #5667, opened 2019-04-04
Get pre-crossgen files for the platform manifest

dagood:
Workaround for crossgen stripping versions from binaries.

For https://github.com/dotnet/core-setup/issues/5665.

When generating a netcoreapp platform manifest, the `GenerateFileVersionProps` calls into the pkgproj to fetch the pre-crossgen files rather than post-crossgen. The MSBuild and targets chain is based on `GetPackageFiles`.

I tried it on win-x64 and it fixes the 0.0.0.0 versions. I'm doing some before/after package diffs next to make sure it doesn't affect anything else.

/cc @dougbu @JunTaoLuo

### DIFF
--- a/src/pkg/projects/dir.targets
+++ b/src/pkg/projects/dir.targets
@@ -146,13 +146,31 @@
                    '$(BuildFullPlatformManifest)' == 'true'">true</IncludeAllRuntimePackagesInPlatformManifest>
     </PropertyGroup>
 
-    <MSBuild Projects="@(ProjectReference)"
-             Condition="'%(ProjectReference.PackageTargetRuntime)' == '$(PackageRID)' OR
-                        '%(ProjectReference.BuidOnRID)' == '$(PackageRID)' OR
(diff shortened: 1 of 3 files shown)


dagood:
I diffed `Microsoft.NETCore.App` and `runtime.win-x64.Microsoft.NETCore.App` vs. pre and post-break, and the changes I expect are there with no files missing.

dagood:
This workaround isn't needed, crossgen was fixed by https://github.com/dotnet/coreclr/pull/23714. (Still waiting for a build, see https://github.com/dotnet/core-setup/issues/5665.)

ericstj:
I wonder if we should add a validation step to the task that generates the platform manfiest.  If after looking at all files we don't have a fileversion for a .dll or .exe raise an error (that could be suppressed).

dagood:
Heh, that came up in the email thread. I'm planning on adding exactly that. 😄

Issue #5297, opened 2019-03-02
source-build: Remove Microsoft.DiaSymReader.Native prebuilt


dagood:
Part of prebuilt removal: https://github.com/dotnet/source-build/issues/885

I looked at the Core-Setup `Microsoft.DiaSymReader.Native` `1.7.0` prebuilt dependency. I don't believe the source for this package is public, so we can't remove it by adding the repo to source-build.

https://github.com/dotnet/core-setup/blob/855cd7611a68f849e041e13a7836f10eaa6e1ed4/src/pkg/projects/netcoreapp/src/netcoreapp.depproj#L16

Core-Setup redists the package's native DLLs in the Windows runtime packages. Not a problem for build from source: Windows is not a priority.

More interesting, the native DLL version info is included in the platform manifest file (in `Microsoft.NETCore.App`):

```
Microsoft.DiaSymReader.Native.x86.dll|Microsoft.NETCore.App||14.12.25830.2
Microsoft.DiaSymReader.Native.amd64.dll|Microsoft.NETCore.App||14.12.25830.2
Microsoft.DiaSymReader.Native.arm.dll|Microsoft.NETCore.App||14.12.25830.2
```

Removing this dependency would make source-build unable to produce the same platform manifest as the Microsoft build without workarounds. We could hard-code the values, or put these native DLLs through something like the ref-only prebuilt managed DLL infrastructure.

However, the source-built platform manifest file is *already missing* these entries, and more. This is because source-build only produces the manifest for the current platform, where the Microsoft build produces the manifest for all platforms. If this difference is ok, we can remove the dependency during builds from source without any concerns.

dagood:
I started a thread about this with @ericstj, @dsplaisted, and @nguerrera a little over a week ago but I haven't had time to follow up on it. To open it up to more input, here's info from the thread:

> [@dagood] I’m wondering what the consequences are for source-build’s [platform manifest] being different from the Microsoft build’s. Maybe self-contained publish for win-x64 by a Linux source-built SDK is a little broken, or something like that?

> [@ericstj] PlatformManifest never impacts self-contained.  With Self-contained you always have the “real” assets to resolve with.
> 
> The PlatformManifest is used for framework-dependent.  It’s listing out all the things that are in the shared framework but not visible at build time so that conflict resolution can “know” to remove them.
> 
[...]

PR #5682, opened 2019-04-05
Enforce that DLL/EXEs in platform manifest have a FileVersion

dagood:
Prevent https://github.com/dotnet/core-setup/issues/5665 from occurring again by making sure no DLL/EXE FileVersions are `0.0.0.0`.

Not enabled for WindowsDesktop. There are some files to investigate there. We might need to have a baseline to enable the check there if there are valid reasons for the files not to have versions.

I rebased it back before the crossgen issue was fixed and it was able to detect the issues ([full list](https://gist.github.com/dagood/fe57f26c8d7bd9211b2732bed8924b00)):

```
C:\git\core-setup\src\pkg\projects\dir.targets(159,5): error : Missing FileVersion in 121 shared framework files: [C:\git\core-setup\src\pkg\projects\netcoreapp\pkg\Microsoft.NETCore.App.pkgproj]
C:\git\core-setup\src\pkg\projects\dir.targets(159,5): error : C:\git\core-setup\packages/transport.runtime.linux-x64.Microsoft.NETCore.Runtime.CoreCLR\3.0.0-preview4-27602-72\runtimes\linux-x64\native\System.Private.CoreLib.dll [C:\git\core-setup\src\pkg\projects\netcoreapp\pkg\Microsoft.NETCore.App.pkgproj]
C:\git\core-setup\src\pkg\projects\dir.targets(159,5): error : C:\git\core-setup\bin/win-x64.Debug\\crossgen/netcoreapp/win-x64/runtimes/win-x64/lib/netcoreapp3.0/Microsoft.CSharp.dll [C:\git\core-setup\src\pkg\projects\netcoreapp\pkg\Microsoft.NETCore.App.pkgproj]
```

### DIFF
--- a/src/pkg/projects/dir.targets
+++ b/src/pkg/projects/dir.targets
@@ -163,7 +163,8 @@
                               PackageVersion="$(Version)"
                               PlatformManifestFile="$(PlatformManifestFile)"
                               PropsFile="$(PropsFile)"
-                              PreferredPackages="$(Id);@(RuntimeDependency)" />
+                              PreferredPackages="$(Id);@(RuntimeDependency)"
+                              PermitDllAndExeFilesLackingFileVersion="$(PermitDllAndExeFilesLackingFileVersion)" />
   </Target>
 
   <Target Name="IncludeFileVersionPropsFile"
--- a/src/pkg/projects/dir.targets
+++ b/src/pkg/projects/dir.targets
@@ -158,6 +158,14 @@
       <Output TaskParameter="TargetOutputs" ItemName="SharedFrameworkRuntimeFiles" />
     </MSBuild>
 
+    <!--
+      Workaround: zero-versioned Microsoft.VisualBasic.dll in non-Windows CoreFX transport package.
+      See https://github.com/dotnet/corefx/issues/36630
+    -->
+    <PropertyGroup Condition="'$(OSGroup)' != 'Windows_NT'">
+      <PermitDllAndExeFilesLackingFileVersion>true</PermitDllAndExeFilesLackingFileVersion>
+    </PropertyGroup>
+
     <GenerateFileVersionProps Files="@(SharedFrameworkRuntimeFiles)"
                               PackageId="$(Id)"
                               PackageVersion="$(Version)"
--- a/src/pkg/projects/windowsdesktop/dir.props
+++ b/src/pkg/projects/windowsdesktop/dir.props
@@ -4,5 +4,16 @@
 
   <PropertyGroup>
     <FrameworkPackageName>Microsoft.WindowsDesktop.App</FrameworkPackageName>
+
+    <!--
+      Disable FileVersion check for WindowsDesktop. Files with no FileVersion, as of writing:
+
+      runtimes\win-x86\native\PenImc_cor3.dll
+      runtimes\win-x86\native\PresentationNative_cor3.dll
+      runtimes\win-x86\native\wpfgfx_cor3.dll
+      runtimes\win-x86\lib\netcoreapp3.0\DirectWriteForwarder.dll
+      runtimes\win-x86\lib\netcoreapp3.0\System.Printing.dll
+    -->
+    <PermitDllAndExeFilesLackingFileVersion>true</PermitDllAndExeFilesLackingFileVersion>
   </PropertyGroup>
 </Project>
--- a/tools-local/tasks/GenerateFileVersionProps.cs
+++ b/tools-local/tasks/GenerateFileVersionProps.cs
@@ -2,18 +2,21 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
-using Microsoft.Build.Framework;
-using System.IO;
-using System.Collections.Generic;
 using Microsoft.Build.Construction;
+using Microsoft.Build.Framework;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 
 namespace Microsoft.DotNet.Build.Tasks
 {
     public partial class GenerateFileVersionProps : BuildTask
     {
-        const string PlatformManifestsItem = "PackageConflictPlatformManifests";
-        const string PreferredPackagesProperty = "PackageConflictPreferredPackages";
+        private const string PlatformManifestsItem = "PackageConflictPlatformManifests";
+        private const string PreferredPackagesProperty = "PackageConflictPreferredPackages";
+        private static readonly Version ZeroVersion = new Version(0, 0, 0, 0);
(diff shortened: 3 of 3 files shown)


dagood:
Ah... CI failing because crossgen still strips the version on non-Windows as it always has.

I'm thinking we should actually take https://github.com/dotnet/core-setup/pull/5667 too for a little more source-build correctness. In source-build, we can't hop over to Windows to build the platform manifest. It already isn't accurate for other reasons (https://github.com/dotnet/core-setup/issues/5297), but might as well fix this one.

dagood:
Holding off on adding #5667, crossgen shouldn't be stripping version on non-Windows anymore. @davidwrighton is looking at fixing it--I'll wait for that. Once an auto-PR comes through with the change, this PR should go green.

dagood:
I misdiagnosed that! It's just one DLL that CoreFX is building with 0.0.0.0 version on non-Windows platforms.

| File in package cache | FileVersion |
| --- | --- |
| microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\ref\netcoreapp3.0\Microsoft.VisualBasic.dll | 14.7.3060.0 |
| runtime.linux-arm.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\linux-arm\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 0.0.0.0 |
| runtime.linux-arm64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\linux-arm64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 0.0.0.0 |
| runtime.linux-musl-x64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\linux-musl-x64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 0.0.0.0 |
| runtime.linux-x64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\linux-x64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 0.0.0.0 |
| runtime.osx-x64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\osx-x64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 0.0.0.0 |
| runtime.rhel.6-x64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\rhel.6-x64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 0.0.0.0 |
| runtime.win-arm.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\win-arm\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 14.7.3060.0 |
| runtime.win-arm64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\win-arm64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 14.7.3060.0 |
| runtime.win-x64.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\win-x64\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 14.7.3060.0 |
| runtime.win-x86.microsoft.private.corefx.netcoreapp\4.6.0-preview4.19204.3\runtimes\win-x86\lib\netcoreapp3.0\Microsoft.VisualBasic.dll | 14.7.3060.0 |

It appears to only be this particular DLL:

```
Missing FileVersion in 1 shared framework files: [/root/coresetup/src/pkg/projects/netcoreapp/pkg/Microsoft.NETCore.App.Ref.pkgproj]
/root/coresetup/bin/linux-musl-arm64.Release/crossgen/netcoreapp/linux-musl-arm64/runtimes/linux-musl-arm64/lib/netcoreapp3.0/Microsoft.VisualBasic.dll [/root/coresetup/src/pkg/projects/netcoreapp/pkg/Microsoft.NETCore.App.Ref.pkgproj]
```

ericstj:
That's interesting.  Suppress it for now and let's file an issue on CoreFx.

dagood:
Sounds good, filed https://github.com/dotnet/corefx/issues/36630.

ericstj:
Interesting that it isn't missing the version, but the version is set and it is 0.0.0.0

src/pkg/projects/dir.targets
```diff
@@ -158,6 +158,14 @@
       <Output TaskParameter="TargetOutputs" ItemName="SharedFrameworkRuntimeFiles" />
     </MSBuild>
 
+    <!--
+      Workaround: zero-versioned Microsoft.VisualBasic.dll in non-Windows CoreFX transport package.
```
ericstj: I was actually thinking you would constrain to a single filename.

dagood: That would be better, but I'm not sure if the risk of another non-Windows file changing to 0.0.0.0/none before https://github.com/dotnet/corefx/issues/36630 is fixed is high enough to warrant effort here. This is also a backstop for a brand new source-build-only fix that we stumbled into (for a bug nobody has mentioned before) rather than actually preventing a regression.

Not hard to do, but this was a bit quicker and I wanted to get this out of the way.

ericstj: I see.  One thing I've thought about in the past is constructing the platform manfiest from some form of authoring rather than inspecting binaries.  That way we just inspect the binaries in the individual legs to validate that they are represented correctly in the manifest, but the manifest represents a superset of all legs.  When maestro picks up an update it runs a target that commits an updated manifest.

ericstj: Also, it might make sense to do more "layout correctness" checks in a central place.  Things like this and verify closure are relevant for more things than just the places where we run them now.

dagood: Yeah, all this stuff happening recently has been making me think about how core-setup could really use general artifact testing, rather than trying to add little checks to various individual parts of the build.

> When maestro picks up an update it runs a target that commits an updated manifest.

Seems like this would make the build somewhat faster, too, since we wouldn't have to download all platforms' runtime packages on all legs to generate the platform manifest.

Unfortunately not supported in Maestro++ unless it's coded into Darc directly. https://github.com/dotnet/arcade/issues/774. I think it'd have to be pretty general and robust because that limits our opportunities to change/maintain it quite a bit. I think there's also expectations of "fast" and not requiring a real repo clone and build tooling. I can only imagine it being reasonable if everything needed is in BAR metadata.

I wondered a bit about the style of baseline + update baseline target + PR validation build letting you download an updated baseline (we use this in source-build). Problem is the versions change for every single update--so I don't think that would work at all.

Issue #5580, opened 2019-03-27
Crossgen the WindowsDesktop shared framework


dagood:
After gathering the WindowsDesktop shared framework in Core-Setup (https://github.com/dotnet/core-setup/issues/5219), it can share the crossgen infra being used on the .NET Core shared framework.

Issue #5219, opened 2019-02-21
Gather Microsoft.WindowsDesktop.App (WPF, WinForms) in Core-Setup, not DotNet-Trusted


dagood:
DotNet-Trusted currently gathers WPF and WinForms bits and produces Microsoft.WindowsDesktop.App and its runtime packages. We should produce the WindowsDesktop packages in Core-Setup to centralize gathering and take advantage of the shared infrastructure.

Core-Setup has a lot of existing infrastructure similar to DotNet-Trusted's for Microsoft.NETCore.App. We will save infra work by sharing this infra in one repo: Core-Setup already runs crossgen to make Microsoft.NETCore.App runtime packages, and Targeting and Runtime pack creation is being added.

Also, when we do this, the DotNet-Trusted repo (with some WPF caveats) would no longer be in the build graph and could be deleted.

I started looking at this for https://github.com/dotnet/core-setup/pull/5214, but there are some WPF bits that are only built in DotNet-Trusted that we need to work out how to transport first.

/cc @zsd4yr @ericstj

zsd4yr:
@vatsan-madhavan for wpf visability

zsd4yr:
related to https://github.com/dotnet/winforms/issues/471

Lee02:
but this sdk don't windowsDesktop .don't run Linux

dagood:
@Lee02 That's correct, WPF (https://github.com/dotnet/wpf) and WinForms (https://github.com/dotnet/winforms) only run on Windows.

dagood:
To gather status here, I merged [PR 168307](https://devdiv.visualstudio.com/DevDiv/_git/DotNet-Trusted/pullrequest/168307) today to add Microsoft.Private.WindowsDesktop.App transport packages to DotNet-Trusted. [The official build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2485956) produced `Microsoft.Private.WindowsDesktop.App` `3.0.0-preview4-27511-40` (and runtime packages).

Next step is get Core-Setup able to consume `Microsoft.Private.WindowsDesktop.App` to produce `Microsoft.WindowsDesktop.App` packages, with crossgen. We'll synchronize turning that package on in Core-Setup and off in DotNet-Trusted once the Core-Setup changes are ready.

vatsan-madhavan:
[...]

Issue #5663, opened 2019-04-03
Create "empty" Microsoft.NETCore.App package to support a 3.0 workaround


dagood:
To work around not getting some NuGet features that we won't get in time for 3.0, we need to provide a mostly empty Microsoft.NETCore.App package for the SDK to implement some fixes.

Proposal (thanks for this @dsplaisted!):

* Rename the existing `Microsoft.NETCore.App` package to something like `Microsoft.NETCore.App.Internal` or `Microsoft.NETCore.App.Insertion`
* Create a new `Microsoft.NETCore.App` package which has no assets, and only has a dependency on `Microsoft.NETCore.Platforms`

Issues this will help with:

* Package downgrades when version 1.x of `Microsoft.NETCore.App` is in the dependency graph: https://github.com/dotnet/sdk/issues/3044
* Package downgrades for `Microsoft.NETCore.Platforms`
* Platforms package showing up with warning icon in VS solution

Once Runtime Pack work is complete and those are used instead of `Microsoft.NETCore.App`, we can stop producing `Microsoft.NETCore.App.Internal`.

/cc @nguerrera

nguerrera:
You might need to do runtime packs first so that the current old targeting packs don't become .Internal and break things

dagood:
I don't think that will be a problem based on how the infra is set up, but I'll keep the order flexible if something pops up.

dagood:
I normally run with `-skiptests=true` because they've been irrelevant to my work, but I tried running my changes for this workaround through CI and the tests are trashed. The projects being restored are like this:

https://github.com/dotnet/core-setup/blob/a458284434f18ad8acbbf867d7338d66e021ad3a/src/test/Assets/TestProjects/PortableApp/PortableApp.csproj#L3-L7

There are generic "the world is ending" compilation errors like:

> Program.cs(9,13): error CS0103: The name 'Console' does not exist in the current context

but more specifically the runtime packages are missing:

>C:\git\core-setup\bin\tests\win-x64.Release\1\StandaloneAppWithSubDirs\StandaloneAppWithSubDirs.csproj : error : NETSDK1056: **Project is targeting runtime 'win-x64' but did not resolve any runtime-specific packages**. This runtime may not be supported by the target framework.

I'm not really familiar with the tests but I think this is to be expected because Core-Setup is on SDK `2.1.401` so the Microsoft.NETCore.App packages still matter quite a bit.

The first thing that comes to mind is creating:
[...]

PR #5741, opened 2019-04-09
Create WindowsDesktop shared framework nupkgs and MSIs

dagood:
Produces Microsoft.WindowsDesktop.App and its runtime packages and MSIs in Core-Setup. At the moment, these changes would start us producing the same artifacts in two places--we need to coordinate switchover once the PR is ready to merge.

Resolves https://github.com/dotnet/core-setup/issues/5675

To validate I diffed the MSI tables and contents of the shared framework packages and layouts in DOTNET_HOME vs. DotNet-Trusted outputs. Differences all look good or fine to me:

* nupkg
  * The identity package no longer has `Microsoft.NETCore.Platforms` and `Microsoft.NETCore.Targets` dependencies. I believe these are obsolete because this package only exists to be turned into packs.
  * Instead of `version.txt`, has `Microsoft.WindowsDesktop.App.versions.txt`. (Standard convention in core-setup.)
* MSI
  * `Microsoft.WindowsDesktop.App.deps.json` shows a dependency on `Microsoft.WindowsDesktop.App/3.0.0-preview4-27606-0`, not the runtime package directly. This is because in Core-Setup I use the `runtime.json` to resolve the runtime package to be more consistent.
  * DN-T is missing `version.txt`, this PR's MSI includes it.
  * Resource DLLs are carried through fine. (Not a difference, but wanted to call out that I didn't miss it.)

This PR adds the empty `useSharedDesignerContext.txt` file to all Core-Setup packages. It looked like Core-Setup was simply missing this. (Seems obsolete for these packages, but figured might as well to remove the difference.)

I moved the `packaging-tools` files in a separate commit. If the diff across both commits doesn't detect that, you can look at the second commit only.

This isn't crossgen'd yet (https://github.com/dotnet/core-setup/issues/5580) and it still depends on the DotNet-Trusted transport packages (https://github.com/dotnet/core-setup/issues/5219). (Those issues are incremental work on top of this.)

/cc @zsd4yr @vatsan-madhavan @johnbeisner @MichaelSimons


src/pkg/packaging-tools/sharedFramework/Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<Project>
+
+  <!--
+    This Directory.Build.props prevents the SDK from skipping down to the root
+    Directory.Build.props, following the dir.props chain instead.
```
ericstj: Any reason not to just rename?

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
```diff
@@ -0,0 +1,153 @@
+<Project Sdk="Microsoft.NET.Sdk" InitialTargets="GetPackageReference" DefaultTargets="Restore;Publish">
+  <!--
+    This project lays out a shared framework using the SDK targets.
+  -->
+
+  <UsingTask TaskName="ProcessSharedFrameworkDeps" AssemblyFile="$(LocalBuildToolsTaskDir)core-setup.tasks.dll" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
ericstj: I don't suppose we have a central place where we define this?

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
```diff
@@ -0,0 +1,153 @@
...
+    <NETCoreAppMaximumVersion>3.0</NETCoreAppMaximumVersion>
+    <RuntimeIdentifier>$(PackageRID)</RuntimeIdentifier>
+    <PackageId>unused</PackageId>
+
+    <ProjectDepsFileName>$(SharedFrameworkName).deps.json</ProjectDepsFileName>
+
+    <MicrosoftNETPlatformLibrary>do-not-consider-existing-platform</MicrosoftNETPlatformLibrary>
```
ericstj: Can you add a comment?

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
```diff
@@ -0,0 +1,153 @@
...
+
+    <!-- Don't reference NETCore.App -->
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+    <!-- Only Restore from our built packages -->
+    <RestoreSources>$(PackageOutputPath)</RestoreSources>
+    <RestoreConfigFile>$(ProjectDir)NuGet.config</RestoreConfigFile>
```
ericstj: Why are you setting config file?  In dotnet-trusted this was a config file that cleared everything.

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
```diff
@@ -0,0 +1,153 @@
...
+      <SharedFrameworkAssetsFile>$(IntermediateOutputPath)/project.assets.json</SharedFrameworkAssetsFile>
+      <SharedFrameworkDepsFile>$(PublishDir)$(ProjectDepsFileName)</SharedFrameworkDepsFile>
+      <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='Windows_NT'">win</RuntimeGraphGeneratorRuntime>
+      <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='OSX'">osx</RuntimeGraphGeneratorRuntime>
+      <RuntimeGraphGeneratorRuntime Condition="'$(OSGroup)'=='FreeBSD'">freebsd</RuntimeGraphGeneratorRuntime>
+      <RuntimeGraphGeneratorRuntime Condition="'$(RuntimeGraphGeneratorRuntime)'==''">linux</RuntimeGraphGeneratorRuntime>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <TrimPkgsFromDeps Include="runtime.$(PackageTargetRid).microsoft.netcore.dotnetapphost" />
```
ericstj: Do we still have these pacakges in the deps?  I thought we flattened it.

src/pkg/projects/dir.props
```diff
@@ -95,6 +95,11 @@
           Include="$(PackageThirdPartyNoticesFile)" >
       <SkipPackageFileCheck>true</SkipPackageFileCheck>
     </File>
+    <!-- Add a marker to help the designer optimize & share .NET Core packages -->
+    <File Condition="'$(IncludeDesignerMarker)' != 'false'"
+          Include="$(MSBuildThisFileDirectory)useSharedDesignerContext.txt">
```
ericstj: Be careful about this.  @	lutzroeder is this something you expect to see in the shared framework packages?  I suspect the designer might see these when designing for a self-contained app (since it runs against the package and not the shared framework).

nguerrera: Where does this file get deployed exactly, who reads it, and how do they locate it?

Fixing tag, @lutzroeder, cc @chabiss

nguerrera: Judging by https://github.com/dotnet/corefx/commit/8d7870ad127d5b25968835ceb22a6c87eb7e73f8 message, I don't think this should be needed for .NET Core 3.

vatsan-madhavan: //sidebar

That's the only place I found any "documentation" ;-) about _useSharedDesignerContext.txt_ when I was moving `Microsoft.NET.Sdk.WindowsDesktop` to [dotnet/wpf](https://github.com/dotnet/wpf) (thanks @ericstj!). 

Is it documented officially anywhere else - or who should perhaps own the documentation for that file? Would a developr ever need or want to incorporate that into their own package? 

/cc @chabiss

src/pkg/packaging-tools/sharedFramework/Directory.Build.props
dagood: Rename which thing?

Since the import is in the (new) SDK props, it seems most practical to me to keep this around until the entire project moves from `dir.props` to `Directory.Build.props` during the move to Arcade, then this particular one can be removed.

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
dagood: There's `NETCoreAppFramework`, will change. Didn't spot this after the move from dn-t.

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
dagood: Will do. This was for producing Microsoft.NETCore.App on the new infra which I didn't end up finishing for this PR.

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
dagood: I'll revisit this, doesn't seem necessary with `RestoreSources`.

ericstj: I think when I was playing with this before I added stuff until I saw what I expected in the `project` section of the project.assets.json.  This lists all the places NuGet considered when doing the restore and we need to make sure they are all locked down.

src/pkg/projects/dir.props
dagood: I'll get rid of this unless I get confirmation otherwise, only motivation was package content parity. Copying what I mentioned in the PR description here for context:

> This PR adds the empty `useSharedDesignerContext.txt` file to all Core-Setup packages. It looked like Core-Setup was simply missing this. (Seems obsolete for these packages, but figured might as well to remove the difference.)

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
dagood: `Microsoft.WindowsDesktop.App` and its runtime package don't have these dependencies, so currently this doesn't do anything.

`Microsoft.NETCore.App` still has them... I also remember this coming up before but forget what happened. Maybe there was just no particular reason to trim them ahead of time and instead we'd just wait for the package to disappear. (https://github.com/dotnet/core-setup/issues/5663 is a step towards that.)

I'll remove the trimming from this PR. When this infra is used for netcoreapp, the solution will be to remove the dependencies from the package itself.

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
dagood: Removing this line does this (and only this) in the `project.assets.json`, `project` section, and nothing to the sfx contents:

```diff
       "configFilePaths": [
-        "C:\\git\\core-setup\\NuGet.config"
+        "C:\\git\\core-setup\\NuGet.Config",
+        "C:\\Users\\dagood\\AppData\\Roaming\\NuGet\\NuGet.Config",
+        "C:\\Program Files (x86)\\NuGet\\Config\\Microsoft.VisualStudio.Offline.config"
       ],
```

I figure we might as well keep it in case there's some weird stuff the chain of NuGet.Configs might be able to do. I'll add a comment that it isn't doing anything to `Only Restore from our built packages` though.

src/pkg/packaging-tools/sharedFramework/Directory.Build.props
dagood: If you mean rename the root `Directory.Build.props` so it isn't detected, I believe that one's actively used to build the managed packages. Might actually cause source-build problems too, come to think of it, since source-build has a `Directory.Build.props` and the SDK has no search boundaries.

Leaving it as-is for now, will get better with Arcade.

dagood:
I think merging this could break Core-SDK updates, so holding off. Posted my thoughts on how to handle the switchover at https://github.com/dotnet/core-setup/issues/5675#issuecomment-481901179.

PR #5742, opened 2019-04-09
AppHost: Support bundles (stage 1)

swaroop-sridhar:
This changes implements the [app-host support](https://github.com/dotnet/designs/blob/master/accepted/single-file/design.md#the-host ) for executing .net core apps published as a single file.

This change implements [stage 1](https://github.com/dotnet/designs/blob/master/accepted/single-file/staging.md#1-self-extractor), which [extracts out](https://github.com/dotnet/designs/blob/master/accepted/single-file/extract.md ) the embedded files to disk.

On startup, the AppHost detects whether it's own binary is a .net core bundle.
If so, on the first run, the host extracts embedded files to:
  * $DOTNET_BUNDLE_EXTRACT_BASE_DIR/app/id/... if DOTNET_BUNDLE_EXTRACT_BASE_DIR is set
  * $TMPDIR/.net/app/id/... otherwise.

On subsequent runs, the files extracted above are reused.

AppHost (X64) size increase:
Windows 5KB, Linux 11KB, Mac 6KB

Testing:
Tested the single-file extraction locally with several kinds of dotnet apps
(wpf, winforms, web, mvc, console, etc.)
Added a test case that runs a bundled apps with sub directories

### DIFF
--- a/src/corehost/cli/apphost/CMakeLists.txt
+++ b/src/corehost/cli/apphost/CMakeLists.txt
@@ -20,15 +20,17 @@ set(SKIP_VERSIONING 1)
 
 set(SOURCES
     ../fxr/fx_ver.cpp
-    ./bundle/bundle_runner.cpp
-    ./bundle/manifest.cpp
     ./bundle/file_entry.cpp
+    ./bundle/manifest.cpp
+    ./bundle/bundle_runner.cpp
 )
-
-include_directories(./bundle)
 
 set(HEADERS
     ../fxr/fx_ver.h
+    ./bundle/file_type.h
+    ./bundle/file_entry.h
+    ./bundle/manifest.h
+    ./bundle/bundle_runner.h
 )
 
 include(../exe.cmake)
--- a/src/corehost/cli/apphost/CMakeLists.txt
+++ b/src/corehost/cli/apphost/CMakeLists.txt
@@ -3,7 +3,7 @@
 # See the LICENSE file in the project root for more information.
 
 cmake_minimum_required (VERSION 2.6)
-project(apphost)
+project(apphost)	
 set(DOTNET_PROJECT_NAME "apphost")
 
 # Add RPATH to the apphost binary that allows using local copies of shared libraries
@@ -20,7 +20,12 @@ set(SKIP_VERSIONING 1)
 
 set(SOURCES
     ../fxr/fx_ver.cpp
+    ./bundle/bundle_runner.cpp
+    ./bundle/manifest.cpp
+    ./bundle/file_entry.cpp
 )
+
+include_directories(./bundle)
 
 set(HEADERS
     ../fxr/fx_ver.h
--- a/src/corehost/cli/bdl_file_entry.cpp
+++ b/src/corehost/cli/bdl_file_entry.cpp
@@ -2,49 +2,50 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#ifdef FEATURE_APPHOST
-
 #include "bdl_processor.h"
 #include "pal.h"
 #include "error_codes.h"
 #include "trace.h"
 #include "utils.h"
 
+using namespace bundle;
+
 bool file_entry_t::is_valid()
 {
-	return data.offset > 0 && data.size > 0 && 
-		   (file_type_t)data.type < file_type_t::END &&
-		   data.path_length > 0 && data.path_length <= PATH_MAX;
+    return data.offset > 0 && data.size > 0 &&
+        (file_type_t)data.type < file_type_t::__last &&
+        data.path_length > 0 && data.path_length <= PATH_MAX;
 }
 
-file_entry_t* file_entry_t::read(FILE* bundle)
+file_entry_t* file_entry_t::read(FILE* stream)
 {
-	file_entry_t* entry = new file_entry_t();
+    file_entry_t* entry = new file_entry_t();
 
-	// First read the fixed-sized portion of file-entry
-	bdl_processor_t::read(&entry->data, sizeof(entry->data), bundle);
-	if (!entry->is_valid())
-	{
-		trace::error(_X("Invalid FileEntry"));
-		throw StatusCode::BundleExtractionFailure;
-	}
+    // First read the fixed-sized portion of file-entry
+    bundle_runner_t::read(&entry->data, sizeof(entry->data), stream);
+    if (!entry->is_valid())
+    {
+        trace::error(_X("Failure processing application bundle; possible file corruption."));
+        trace::error(_X("Invalid FileEntry detected."));
+        throw StatusCode::BundleExtractionFailure;
+    }
 
-	// Read the relative-path, given its length 
-	pal::string_t& path = entry->relative_path;
-	bdl_processor_t::read_string(path, entry->data.path_length, bundle);
-	
-	// Fixup the relative-path to have current platform's directory separator.
-	if (bundle_dir_separator != DIR_SEPARATOR)
-	{
-		for(size_t pos = path.find(bundle_dir_separator); 
-			pos != pal::string_t::npos; 
-			pos = path.find(bundle_dir_separator, pos))
-		{
-			path[pos] = DIR_SEPARATOR;
-		}
-	}
+    // Read the relative-path, given its length 
+    pal::string_t& path = entry->relative_path;
+    bundle_runner_t::read_string(path, entry->data.path_length, stream);
 
-	return entry;
+    // Fixup the relative-path to have current platform's directory separator.
+    if (bundle_dir_separator != DIR_SEPARATOR)
+    {
+        for (size_t pos = path.find(bundle_dir_separator);
+            pos != pal::string_t::npos;
+            pos = path.find(bundle_dir_separator, pos))
+        {
+            path[pos] = DIR_SEPARATOR;
+        }
+    }
+
+    return entry;
 }
 
-#endif // FEATURE_APPHOST
+
--- a/src/corehost/cli/bdl_file_entry.cpp
+++ b/src/corehost/cli/bdl_file_entry.cpp
@@ -1,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#ifdef FEATURE_APPHOST
+
+#include "bdl_processor.h"
+#include "pal.h"
+#include "error_codes.h"
+#include "trace.h"
+#include "utils.h"
+
+bool file_entry_t::is_valid()
+{
+	return data.offset > 0 && data.size > 0 && 
+		   (file_type_t)data.type < file_type_t::END &&
+		   data.path_length > 0 && data.path_length <= PATH_MAX;
+}
+
+file_entry_t* file_entry_t::read(FILE* bundle)
+{
+	file_entry_t* entry = new file_entry_t();
+
+	// First read the fixed-sized portion of file-entry
+	bdl_processor_t::read(&entry->data, sizeof(entry->data), bundle);
+	if (!entry->is_valid())
+	{
+		trace::error(_X("Invalid FileEntry"));
+		throw StatusCode::BundleExtractionFailure;
+	}
+
+	// Read the relative-path, given its length 
+	pal::string_t& path = entry->relative_path;
+	bdl_processor_t::read_string(path, entry->data.path_length, bundle);
+	
+	// Fixup the relative-path to have current platform's directory separator.
+	if (bundle_dir_separator != DIR_SEPARATOR)
+	{
+		for(size_t pos = path.find(bundle_dir_separator); 
+			pos != pal::string_t::npos; 
+			pos = path.find(bundle_dir_separator, pos))
+		{
+			path[pos] = DIR_SEPARATOR;
+		}
+	}
+
+	return entry;
+}
+
+#endif // FEATURE_APPHOST
--- a/src/corehost/cli/bdl_file_entry.h
+++ b/src/corehost/cli/bdl_file_entry.h
@@ -1,0 +1,59 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#ifndef __BDL_FILE_ENTRY_H__
+#define __BDL_FILE_ENTRY_H__
+
+#if FEATURE_APPHOST
+
+#include <cstdint>
+#include "bdl_file_type.h"
+#include "pal.h"
+
+#pragma pack(push, 1)
+
+// FileEntry: Records information about embedded files.
+// 
+// The bundle manifest records the following meta-data for each 
+// file embedded in the bundle:
+// Fixed size portion (represented by file_entry_inner_t)
+//   - Offset     
+//   - Size       
+//   - File Entry Type       
+//   - path-length  (7-bit extension encoding, 1 Byte due to MAX_PATH)
+// Variable Size portion
+//   - relative path  ("path-length" Bytes)
+
+class file_entry_t
+{
+public:
+    
+    // The inner structure represents the fields that can be 
+    // read contiguously for every file_entry. 
+    struct file_entry_inner_t
+    {
+        int64_t offset;
+        int64_t size;
+        file_type_t type;
+        int8_t path_length;
+    } data;
+    pal::string_t relative_path; // Path of an embedded file, relative to the extraction directory.
+
+	file_entry_t()
+		:data(), relative_path()
+	{
+	}
+
+    static file_entry_t* read(FILE* bundle);
+
+private:
+	static const pal::char_t bundle_dir_separator = '/';
+    bool is_valid();
+};
+
+#pragma pack(pop)
+
+#endif // FEATURE_APPHOST
+
+#endif // __BDL_FILE_ENTRY_H__
--- a/src/corehost/cli/bdl_file_entry.h
+++ b/src/corehost/cli/bdl_file_entry.h
@@ -5,55 +5,52 @@
 #ifndef __BDL_FILE_ENTRY_H__
 #define __BDL_FILE_ENTRY_H__
 
-#if FEATURE_APPHOST
-
 #include <cstdint>
 #include "bdl_file_type.h"
 #include "pal.h"
 
-#pragma pack(push, 1)
-
-// FileEntry: Records information about embedded files.
-// 
-// The bundle manifest records the following meta-data for each 
-// file embedded in the bundle:
-// Fixed size portion (represented by file_entry_inner_t)
-//   - Offset     
-//   - Size       
-//   - File Entry Type       
-//   - path-length  (7-bit extension encoding, 1 Byte due to MAX_PATH)
-// Variable Size portion
-//   - relative path  ("path-length" Bytes)
-
-class file_entry_t
+namespace bundle
 {
-public:
-    
-    // The inner structure represents the fields that can be 
-    // read contiguously for every file_entry. 
-    struct file_entry_inner_t
+
+    // FileEntry: Records information about embedded files.
+    // 
+    // The bundle manifest records the following meta-data for each 
+    // file embedded in the bundle:
+    // Fixed size portion (represented by file_entry_inner_t)
+    //   - Offset     
+    //   - Size       
+    //   - File Entry Type       
+    //   - path-length  (7-bit extension encoding, 1 Byte due to MAX_PATH)
+    // Variable Size portion
+    //   - relative path  ("path-length" Bytes)
+
+    class file_entry_t
     {
-        int64_t offset;
-        int64_t size;
-        file_type_t type;
-        int8_t path_length;
-    } data;
-    pal::string_t relative_path; // Path of an embedded file, relative to the extraction directory.
+    public:
 
-	file_entry_t()
-		:data(), relative_path()
-	{
-	}
-
-    static file_entry_t* read(FILE* bundle);
-
-private:
-	static const pal::char_t bundle_dir_separator = '/';
-    bool is_valid();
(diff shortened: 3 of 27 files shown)


src/corehost/cli/bdl_file_entry.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "utils.h"
+
+bool file_entry_t::is_valid()
+{
+	return data.offset > 0 && data.size > 0 && 
+		   (file_type_t)data.type < file_type_t::END &&
+		   data.path_length > 0 && data.path_length <= PATH_MAX;
+}
+
+file_entry_t* file_entry_t::read(FILE* bundle)
```
vitek-karas: `bundle` is not a good name here - too overloaded. `bundleFile` at least, ideally something without `bundle` completely (though I don't have a good idea right now).

src/corehost/cli/bdl_file_entry.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+	file_entry_t* entry = new file_entry_t();
+
+	// First read the fixed-sized portion of file-entry
+	bdl_processor_t::read(&entry->data, sizeof(entry->data), bundle);
+	if (!entry->is_valid())
+	{
+		trace::error(_X("Invalid FileEntry"));
```
vitek-karas: Please make the errors more descriptive. Consider that they are visible be end users. Not only the developers, but the actual users of the apps, which may not know anything about bundles. The error should start with some kind of humanly readable part and then we can include the technical details.

Maybe something like:
> Failed to open the application bundle. It's possibly corrupted.
> Invalid file entry found.

Please ask some native speaker to polish this...

src/corehost/cli/bdl_file_entry.h
```diff
@@ -0,0 +1,59 @@
...
+    struct file_entry_inner_t
+    {
+        int64_t offset;
+        int64_t size;
+        file_type_t type;
+        int8_t path_length;
+    } data;
+    pal::string_t relative_path; // Path of an embedded file, relative to the extraction directory.
+
+	file_entry_t()
```
vitek-karas: Weird indentation

src/corehost/cli/bdl_file_entry.h
```diff
@@ -0,0 +1,59 @@
...
+#ifndef __BDL_FILE_ENTRY_H__
+#define __BDL_FILE_ENTRY_H__
+
+#if FEATURE_APPHOST
+
+#include <cstdint>
+#include "bdl_file_type.h"
+#include "pal.h"
+
+#pragma pack(push, 1)
```
vitek-karas: We really only need this for the inner structure, right?

src/corehost/cli/bdl_file_entry.h
```diff
@@ -0,0 +1,59 @@
...
+// Variable Size portion
+//   - relative path  ("path-length" Bytes)
+
+class file_entry_t
+{
+public:
+    
+    // The inner structure represents the fields that can be 
+    // read contiguously for every file_entry. 
+    struct file_entry_inner_t
```
vitek-karas: Make it anonymous - the type name is not used anywhere.

src/corehost/cli/bdl_file_type.h
```diff
@@ -0,0 +1,32 @@
...
+// implementations will process certain file_types directly from the bundle.
+
+enum file_type_t : uint8_t
+{
+    assembly,
+    ready2run,
+    deps_json,
+    runtime_config_json,
+    extract,
+    END
```
vitek-karas: Nit: `__last` or something like that would be better. All caps doesn't feel right with the rest of the code.

src/corehost/cli/bdl_manifest.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+manifest_footer_t* manifest_footer_t::read(FILE* bundle)
+{
+	manifest_footer_t* footer = new manifest_footer_t();
+
+	bdl_processor_t::read(footer, num_bytes_read(), bundle);
+
+	if (!footer->is_valid())
+	{
+		trace::info(_X("Manifest footer Invalid"));
```
vitek-karas: Better error message.

src/corehost/cli/bdl_manifest.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+manifest_header_t* manifest_header_t::read(FILE* bundle)
+{
+	manifest_header_t* header = new manifest_header_t();
+
+	// First read the fixed size portion of the header
+	bdl_processor_t::read(&header->data, sizeof(header->data), bundle);
+	if (!header->is_valid())
+	{
+		trace::error(_X("Manifest header incompatible"));
```
vitek-karas: Better error message. Future proofing this might be better to be specific about version mismatches. The rest of the validation can probably remain a black box.

src/corehost/cli/bdl_manifest.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include "bdl_processor.h"
+#include "pal.h"
+#include "error_codes.h"
+#include "trace.h"
+#include "utils.h"
+
+const char *manifest_footer_t::m_expected_signature = ".NetCoreBundle";
+
+bool manifest_footer_t::is_valid()
```
vitek-karas: Nit: ordering. In the header the header goes first, in the cpp the footer goes first.

src/corehost/cli/bdl_processor.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include "pal.h"
+#include "trace.h"
+#include "utils.h"
+
+void bdl_processor_t::seek(long offset, int origin)
+{
+	if (fseek(m_bundle, offset, origin) != 0)
+	{
+		trace::error(_X("Bundle extraction: Seek failure."));
+		throw StatusCode::BundleExtractionIOError;
```
vitek-karas: Please do not use exceptions.

src/corehost/cli/bdl_processor.cpp
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	return file;
+}
+
+// Extract one file from the bundle to disk.
+void bdl_processor_t::extract_file(file_entry_t *entry)
+{
+	FILE* file = create_extraction_file(entry->relative_path);
+	uint8_t* buffer = new uint8_t[entry->data.size];
```
vitek-karas: The `data.size` is not validated in any way, so this may force us to allocate unreasonable amounts of memory.
Also, what if the embedded file is large (500MB, like a graphics, video and so on) - this will allocate the memory for the entire file, potentially causing swapping and so on.
We should stream the file chunk by chunk.

src/corehost/cli/bdl_processor.cpp
```diff
@@ -0,0 +1,316 @@
...
+	seek(entry->data.offset, SEEK_SET);
+	read(buffer, entry->data.size, m_bundle);
+	write(buffer, entry->data.size, file);
+
+	fclose(file);
+}
+
+bool bdl_processor_t::can_reuse_extraction()
+{
+	// Should the pre-extracted files be checked for validity?
```
vitek-karas: This discussion should not be part of the code, but the design doc.

src/corehost/cli/bdl_file_entry.h
```diff
@@ -0,0 +1,59 @@
...
+    {
+        int64_t offset;
+        int64_t size;
+        file_type_t type;
+        int8_t path_length;
+    } data;
+    pal::string_t relative_path; // Path of an embedded file, relative to the extraction directory.
```
vitek-karas: Naming: most classes in the host use the `m_` prefix for member fields. Even in this change the processor uses this, but the file entry here doesn't.

src/corehost/common/pal.windows.cpp
```diff
@@ -457,6 +462,12 @@ bool pal::get_module_path(dll_t mod, string_t* recv)
     return GetModuleFileNameWrapper(mod, recv);
 }
 
+bool pal::get_temp_directory(pal::string_t& tmp_dir)
+{
+	pal::getenv(_X("TEMP"), &tmp_dir);
```
vitek-karas: Probably a better way would be to call `GetTempPath`.

src/corehost/corehost.cpp
```diff
@@ -33,9 +34,18 @@
 #define EMBED_HASH_HI_PART_UTF8 "c3ab8ff13720e8ad9047dd39466b3c89" // SHA-256 of "foobar" in UTF-8
 #define EMBED_HASH_LO_PART_UTF8 "74e592c2fa383d4a3960714caef0c4f2"
 #define EMBED_HASH_FULL_UTF8    (EMBED_HASH_HI_PART_UTF8 EMBED_HASH_LO_PART_UTF8) // NUL terminated
+
+bool is_exe_enabled_for_execution_hack(pal::string_t* app_dll)
```
vitek-karas: I assume this is going away before merge... right?

src/corehost/corehost.cpp
```diff
@@ -109,15 +119,34 @@ int exe_start(const int argc, const pal::char_t* argv[])
         requires_v2_hostfxr_interface = true;
     }
 
-    app_path.assign(get_directory(host_path));
+	bdl_processor_t extractor(host_path);
```
vitek-karas: Naming consistency - extractor/processor. I think extractor is better, and thus we should rename the class to extractor as well.

src/corehost/corehost.cpp
```diff
@@ -89,10 +99,10 @@ int exe_start(const int argc, const pal::char_t* argv[])
     pal::string_t app_path;
     pal::string_t app_root;
     bool requires_v2_hostfxr_interface = false;
-
+	
 #if FEATURE_APPHOST
     pal::string_t embedded_app_name;
-    if (!is_exe_enabled_for_execution(&embedded_app_name))
+    if (!is_exe_enabled_for_execution_hack(&embedded_app_name))
```
vitek-karas: This is obviously broken... as is the apphost won't run in the existing (non-single-file) scenarios.

src/corehost/corehost.cpp
```diff
@@ -33,9 +34,18 @@
 #define EMBED_HASH_HI_PART_UTF8 "c3ab8ff13720e8ad9047dd39466b3c89" // SHA-256 of "foobar" in UTF-8
 #define EMBED_HASH_LO_PART_UTF8 "74e592c2fa383d4a3960714caef0c4f2"
 #define EMBED_HASH_FULL_UTF8    (EMBED_HASH_HI_PART_UTF8 EMBED_HASH_LO_PART_UTF8) // NUL terminated
+
+bool is_exe_enabled_for_execution_hack(pal::string_t* app_dll)
+{
+	pal::string_t host_path;
+	pal::get_own_executable_path(&host_path);
+	*app_dll = strip_executable_ext(get_filename(host_path)).append(_X(".dll"));
```
vitek-karas: We should consider if it's worth to support renaming the bundle. As is, it cannot ever be renamed, otherwise it won't run. That is a relatively unexpected property for an executable.

src/corehost/corehost.cpp
```diff
@@ -102,7 +94,7 @@ int exe_start(const int argc, const pal::char_t* argv[])
 	
 #if FEATURE_APPHOST
     pal::string_t embedded_app_name;
-    if (!is_exe_enabled_for_execution_hack(&embedded_app_name))
+    if (!is_exe_enabled_for_execution(&embedded_app_name))
```
vitek-karas: So the apphost will look just like it does without bundle (including the special stamped path to the app.dll to run), right?

swaroop-sridhar: Yes, no change to this mechanism.

src/corehost/cli/bdl_file_entry.cpp
```diff
@@ -0,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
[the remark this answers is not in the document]
swaroop-sridhar: I was trying to follow the convention similar to `fx_`... abbreviation for frameworks. 
I'll remove the `bdl` prefix and make a new namespace.

src/corehost/cli/bdl_file_entry.h
swaroop-sridhar: My understanding is that only *private* data members are prefixed by `m_` and *private* static members are prefixed by `s_` ([example](https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/fx_definition.h#L36)). Public data members don't have a prefix, and they read like properties ([example](https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/host_startup_info.h#L28)).

The convention used in other header files in this repo is not consistent, but I haven't seen any use of `m_` prefix for public members in the code-base. If you think its better, I can make the fields private and expose accessor functions.

src/corehost/corehost.cpp
swaroop-sridhar: I started with the name `extractor`, but as we progress to subsequent [stages](https://github.com/dotnet/designs/blob/master/accepted/single-file/staging.md) of single-file support, this class will extract fewer things. Instead it will setup data-structures for other parts of the host/runtime to use. So, I'm not sure what would be a good name for it.

src/corehost/cli/bdl_processor.cpp
swaroop-sridhar: @vitek-karas  Can you please expand on why you're suggesting not to use exceptions? 
There are a couple of exception-throws in [existing code](https://github.com/dotnet/core-setup/blob/18780678da576d8c629066f50af30159a8859b2f/src/corehost/cli/ijwhost/ijwthunk.cpp#L136-L145). And exceptions are used by json-parser as well.

src/corehost/cli/bdl_file_entry.cpp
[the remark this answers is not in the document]
vitek-karas: I see - I guess it makes sense. On the other hand FX is a very common abbreviation, "bdl" would be new. My personal preference is to not use abbreviations unless they are very obvious. So if I would start over I would probably even use `framework` instead of `fx`.
But in this particular case the namespace feels like the best solution - it's a very separate piece of functionality...

src/corehost/cli/bdl_processor.cpp
vitek-karas: hmm - I missed the exceptions in the IJW... 😞 
The json parser is basically a necessary evil - we took that code from elsewhere.

The reason I don't like exceptions in this kind of code is that it's very easy to forget and let the exception escape. Which will that wreak havoc on the caller - specifically since it's a C++ exception, so a very weird thing for any language but C++.

The return codes are a bit more messy to work with, but they make error paths very explicit.

It was also my impression that the code base doesn't use them - so for consistency.

I'm open to discussion, but I think we should have a good reason to start using them in general.

src/corehost/cli/bdl_file_entry.h
vitek-karas: I missed the fact that these are public...
OK - I guess I'm fine with it... still not a fan, but it's OK.

src/corehost/corehost.cpp
vitek-karas: "runner", "activator" ... either is pretty bad.
I guess I hate "runner" the least... it's still not nice, but at least it's kind of obvious.

For me the most important bit is consistency, I can learn to live with a "weird" name, as long as it's used everywhere 😄

src/corehost/cli/bdl_processor.cpp
swaroop-sridhar: >> The return codes are a bit more messy to work with, but they make error paths very explicit.

Definitely. I started out with propagating return codes, but the source code was filled with error handling `if-else` checks -- which (subjectively) seemed to obscure the overall code flow. 
The bundle-extraction logic is IO-intensive, and has a higher density of error checks.

>> The reason I don't like exceptions in this kind of code is that it's very easy to forget and let the 
exception escape. 

Conversely, I think it is even more easy to forget checking or propagating an error code leading to failure/unexpected behavior in an unrelated part of the code. 

>> I'm open to discussion, ... 

How about we take a middle ground, where we say that:
* We only throw a Status code exceptions 
* We only throw to indicate critical failure.
* Each subsystem (ex: bundle-handler, deps-resolver) catches its own exceptions. 

I can certainly change the code back to using if-else checks, if we're still not convinced.

vitek-karas: OK - I guess I'll warm up to it eventually. I'm fine if there's a clear entry point to the subsystem where we can easily try/catch and isolate the rest of the code from the exceptions that way.

jkoritzinsky: The IJW host exceptions are also a necessary evil because the function that can throw executes in an arbitrary call stack by code that may or may not know that it is starting up the runtime and as a result has no other way to propagate an error code.

swaroop-sridhar:
@vitek-karas I've committed the changes except file renames/moves in this change https://github.com/dotnet/core-setup/pull/5742/commits/d1813710eddb1f5a1d260726ae15e48c1cae31c0
I'll commit another change that does the file moves, because that obscures the diffing. Thanks.

swaroop-sridhar:
@vitek-karas with the latest commit, I believe I've addressed all of your comments. Thanks.

src/corehost/common/pal.windows.cpp
```diff
@@ -464,8 +459,10 @@ bool pal::get_module_path(dll_t mod, string_t* recv)
 
 bool pal::get_temp_directory(pal::string_t& tmp_dir)
 {
-	pal::getenv(_X("TEMP"), &tmp_dir);
-	return pal::realpath(&tmp_dir);
+    pal::char_t temp_path[MAX_PATH + 1];
+    GetTempPathW(MAX_PATH + 1, temp_path);
```
vitek-karas: Check the result and return false - it's better than to rely on the realpath below - the `temp_path` is uninitialized, so if the `GetTempPath` fails, it will likely leave it uninitialized as well.

src/corehost/corehost.cpp
```diff
@@ -119,7 +111,7 @@ int exe_start(const int argc, const pal::char_t* argv[])
         requires_v2_hostfxr_interface = true;
     }
 
-	bdl_processor_t extractor(host_path);
+	bundle::bundle_runner_t extractor(host_path);
 	StatusCode bundle_status = extractor.extract();
 
 	switch (bundle_status)
```
vitek-karas: Nit: Indentation inside the switch is somewhat inconsistent.

src/corehost/cli/bdl_processor.cpp
```diff
@@ -2,244 +2,328 @@
...
+    if (fwrite(buf, 1, size, stream) != size)
+    {
+        trace::error(_X("Failure extracting contents of the application bundle."));
+        trace::error(_X("I/O failure when writing extracted files."));
+        throw StatusCode::BundleExtractionIOError;
+    }
 }
 
-void bdl_processor_t::read(void* buf, size_t size, FILE* stream)
+void bundle_runner_t::read(void* buf, size_t size, FILE* stream)
```
vitek-karas: Do we ever use the read on anything else but the bundle?
There's a bit of an inconsistency here, `seek` doesn't take the stream to operate on (always uses the bundle), `read` does take it, but it feels like it will always be bundle. `write` takes it as well and that one is on the other hand never the bundle.
Maybe make all of these static methods which always take the stream for symmetry. No strong feeling though.
But I think it would make sense to have at least `seek` and `read` consistent.

src/corehost/cli/apphost/CMakeLists.txt
```diff
@@ -20,8 +20,13 @@ set(SKIP_VERSIONING 1)
...
     ../fxr/fx_ver.cpp
+    ./bundle/bundle_runner.cpp
+    ./bundle/manifest.cpp
+    ./bundle/file_entry.cpp
 )
 
+include_directories(./bundle)
+
 set(HEADERS
     ../fxr/fx_ver.h
```
vitek-karas: Please add the header files as well. It's not needed for the build itself, but it helps make the generated VS solution much easier to use.

src/corehost/cli/apphost/CMakeLists.txt
```diff
@@ -20,8 +20,13 @@ set(SKIP_VERSIONING 1)
 
 set(SOURCES
     ../fxr/fx_ver.cpp
+    ./bundle/bundle_runner.cpp
+    ./bundle/manifest.cpp
+    ./bundle/file_entry.cpp
 )
 
+include_directories(./bundle)
```
vitek-karas: Do we need this? Can't we use relative paths to include from the corehost (and everything else is in the same directory)?

swaroop-sridhar:
Thanks @vitek-karas, I've addressed your  comments in this checkin: https://github.com/dotnet/core-setup/pull/5742/commits/2f379ec292b862bc729fbcaefddcdc2f2bd00dd5

src/corehost/cli/bdl_processor.cpp
swaroop-sridhar: We only `seek()` and `read()` from the bundle file. However, the difference is that while `seek()` is only used within `bundle_runner`, `read()` is used from other types like `manifest` and `file_entry`.  
So, `seek()` was a private member function, while `read()` was a static public function. 

Anyway, I've changed `seek()` now to be a static function taking a stream argument, similar to `read()` and `write().

swaroop-sridhar:
@vitek-karas  I had to make another small change: https://github.com/dotnet/core-setup/pull/5742/commits/3de5b31eaed4ea649a1713667edd9124b2c8df3e to get tmp-directory on Ubuntu.

swaroop-sridhar:
@vitek-karas, I merged the change, in order to facilitate its flow into SDK repo, where I have other changes to be tested with this AppHost. If you have any concerns with https://github.com/dotnet/core-setup/pull/5742/commits/3de5b31eaed4ea649a1713667edd9124b2c8df3e, I'll address it separately. I'm also planning to put out another PR with more test cases.

vitek-karas:
@swaroop-sridhar Thanks! LGTM

PR #5744, opened 2019-04-09
Fix algorithm description with the order problem solution

vitek-karas:
Describe the ordering problem and how it is fixed in the new version.

### DIFF
--- a/Documentation/design-docs/framework-version-resolution.md
+++ b/Documentation/design-docs/framework-version-resolution.md
@@ -267,36 +267,37 @@ To maintain backward compatibility, each framework reference will also have to c
 
 ### Algorithm
 Terminology
-- `config fx references`: framework references for a single `.runtimeconfig.json`, where each reference consists of `name`, `version`, optional `rollForward` and optional `applyPatches`.
-- `newest fx references`: framework references keyed off of framework name that contain the highest framework version requested. It is used to perform "soft" roll-forwards to compatible references of the same framework name without reading the disk or performing excessive re-try (Step 7).
+- `framework reference`: consists of framework `name`, `version`, `rollForward` and optionally `applyPatches`.
+- `config fx references`: `framework references` for a single `.runtimeconfig.json`.
+- `newest fx references`: dictionary of `framework references` keyed off of framework `name` that contain the highest `version` requested and most constrained `rollForward` and `applyPatches`. It is used to perform "soft" roll-forwards to compatible references of the same framework name without reading the disk or performing excessive re-try.
 - `resolved frameworks`: a list of frameworks that have been resolved, meaning a compatible framework was found on disk.
 
 Steps
 1. Determine the `config fx references`:
    * Parse the application's `.runtimeconfig.json` `runtimeOptions.frameworks` section.
-   * Insert each framework reference into the `config fx references`.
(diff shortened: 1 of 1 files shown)


vitek-karas:
I tried to make the algorithm clearer and fix the terminology.

@sdmaclea  do you have an example of a situation where the algorithm fails?

sdmaclea:
> @sdmaclea do you have an example of a situation where the algorithm fails?

The hypothetical situation I was thinking about was.

Developers app references some frameworks, 
* say `ASP.NET` & `JSON_UBER` to simplify discussion.  
* Developper allows aggressive roll forward on both `ASP.NET` & `JSON_UBER`
* `ASP.NET` & `JSON_UBER` both include JSON references
* New release of ASP.NET reference to JSON, which is incompatible with `JSON_UBER`'s reference
* Old version of ASP.NET is present, but app will not use because we don't search all compatible versions of each framework in case of failure.

Documentation/design-docs/framework-version-resolution.md
```diff
@@ -322,6 +329,65 @@ By following these best practices we have optimal run-time performance (less pro
...
+
+Since the reference in standard framework (`ASP.NET` in this sample) has no settings (defaults) this can be very common. Specifically for COM or other components where we will recommend usage of `LatestMinor` or `LatestMajor` for best compatibility.
+Note that with `LatestMajor` the problem is even worse because depending on the order and available framework the references may actually fail to resolve with the old algorithm.
+
+The fixed algorithm doesn't consider the actual hard resolved framework version when computing the effective framework reference. See the algorithm description above. The outcome is that it will effectively always compute the full effective framework reference before hard resolving it and thus is not affected by ordering. There's a small downside to it, that it may need to retry more often. To avoid unnecessary retries the best practices should be followed, specifically the one about not specifying unnecessary framework references. In the above sample, the app doesn't need to specify a framework reference for `Microsoft.NETCore.App` and so it should not do that.
```
elinor-fung: > There's a small downside to it, that it may need to retry more often.

The downside is that it may need to retry more often.

sdmaclea:
> (Personally, I kind of want a flow chart, but I don't think there's an easy way to do that)

It is relatively easy to inline a png sitting next to a doc.  There are UML tools which can generate png files.  I think there are some readily available online.  

I used to use plantuml because I despised WYSIWYG time wasted fiddling with layout.  

Here someones blog on plantuml https://blog.anoff.io/2018-07-31-diagrams-with-plantuml/
Here is a link that embeds a sequence diagram (my preference over flow charts) (flow charts should also be supported).
cached image

vitek-karas:
@sdmaclea It's true that the algorithm is such that selection of a lower-level framework doesn't influence selection of a higher-level framework. In general this should not happen too much unless something in the system uses `LatestMajor`, mostly because:
* Any other setting will not automatically just to a different major version unless necessary (that is nothing else is available)
* Frameworks in general should use default settings and thus allow roll over minor.

Also right now this should not really happen as we will probably ship all frameworks in sync. This might become an issue if we allow 3rd party frameworks with very different dependencies on MS frameworks for example.

I'll add a section with "known issues" to the doc describing this situation and why it's unlikely to become an issue for 3.0 release.

PR #5798, opened 2019-04-11
Build the Bundler as a library

swaroop-sridhar:
Currently, the single-file bundler is
* Built as an app instead of a library
* Invoked from the SDK out-of-proc in order to enforce strict API boundaries

However, since the SDK implementation progressed, we've had to revisit the above decisions.

First, the bundler currently is implemented to take a folder vs a list of files to bundle.
* This causes unnecessary extra copies to the publish directory.
    * The files to be bundled need not be first copied to the publish directory.
    * Instead they can be written directly to the bundle.
* Stale items remaining in the publish directory may be written into the bundle unnecessarily.

Therefore, we decided that the bundler should take tuples of (source file-path, relative-path within bundle) instead of the publish folder as input.
* While it is possible to provide this input on the command-line/response file, it is convenient and more efficient to do it in-proc.
* Further, the SDK team agrees that the bundler is simple enough to run in-proc with the build system.

Therefore, the changes proposed are:
* Bundler will be built into a library Microsoft.NET.HostModel in core-setup repo (similar to DependencyModel)
    * It is not a task directly because we don't have to worry about multi-targeting
    * Other apphost related services like stamping the app.dll name into the apphost will be moded into this library
* A task in the SDK repo will consume this library through package reference (from myget)
    * (WIP) SDK changes: https://github.com/dotnet/sdk/compare/master...swaroop-sridhar:bundle-lib?expand=1


swaroop-sridhar:
CC: @dagood

swaroop-sridhar:
CC: @fadimounir

src/managed/Microsoft.NET.HostModel/Bundle/Bundler.cs
```diff
@@ -161,63 +124,114 @@ FileType InferType(string fileRelativePath, Stream file)
...
+        ///   SourcePath: path to the file to be bundled at compile time
+        ///   RelativePath: path where the file is expected at run time, relative to the app DLL.
+        /// </param>
+        /// <returns>
+        /// The full path the the generated bundle file
+        /// </returns>
+        /// <exceptions>
+        /// ArgumentException if input is invalid
+        /// IOExceptions and ArgumentExceptions from callees flow to the caller.
+        /// </exceptions>
```
nguerrera: I suggest IEnumerable or IReadOnlyList of tuple here? It can be inefficient to have to make a perfectly sized array and arrays also make the input vs output contract harder to spot since they're mutable

nguerrera: I also suggest a custom struct over tuple so that we can add features that require more metadata than relative path with minimal API churn.

swaroop-sridhar: Thanks @nguerrera. I've changed this now.

src/managed/Microsoft.NET.HostModel/Bundle/Bundler.cs
```diff
@@ -224,14 +223,13 @@ public string GenerateBundle(string sourceDir)
             // Sort the file names to keep the bundle construction deterministic.
             Array.Sort(sources, StringComparer.Ordinal);
 
-            var fileSpec = new (string SourcePath, string RelativePath)[sources.Length];
-            for(int i=0; i < sources.Length; i++)
+            List<FileSpec> fileSpecs = new List<FileSpec>();
```
nguerrera: Pass sources.Length to constructor here to avoid unnecessary resizing.

swaroop-sridhar: @nguerrera: done, thanks.

PR #5893, opened 2019-04-17
 Produce mostly empty Microsoft.NETCore.App

dagood:
Resolves https://github.com/dotnet/core-setup/issues/5663

See commit messages for more details on why this ended up being pretty large.

Manual testing was mostly comparing MSI tables, but also installing them and comparing on-disk (and in-registry) results. I also spot-checked the macOS runtime pkg and Debian/RPM packages.

@johnbeisner let me know when you're ready for this, a hard break is expected.

/cc @nguerrera @dsplaisted


dsplaisted:
Does this include continuing to produce a package with the existing contents of the Microsoft.NETCore.App package, but with a different package name?

dagood:
Yep, that's `Microsoft.NETCore.App.Insertion`. Thinking about it again I'm going to switch to the `.Internal` suggestion because of how I ended up using it in Core-Setup too.

dagood:
@MichaelSimons @ericstj can you PTAL?

@johnbeisner do you have time to handle the break core-sdk-side today? I imagine it's mainly changing references to `Microsoft.NETCore.App` to `Microsoft.NETCore.App.Internal` for runtimes. Or maybe @dsplaisted has more context to make these changes?

dir.props
```diff
@@ -276,7 +276,8 @@
     <SymbolPackageOutputPath Condition="'$(SymbolPackageOutputPath)'==''">$(PackageOutputPath)</SymbolPackageOutputPath>
     <PackageSymbolsBinDir Condition="'$(PackageSymbolsBinDir)' == ''">$(PackagesBasePath)symbols/</PackageSymbolsBinDir>
 
-    <SharedFrameworkPublishDir>$(IntermediateOutputRootPath)sharedFrameworkPublish\</SharedFrameworkPublishDir>
+    <!-- Use / instead of \: used directly by WiX which doesn't accept trailing \. -->
```
MichaelSimons: This comment is a little confusing IMO.  Is the ending period punctuation or the trailing character sequence?  I assume the former.  If this is true consider dropping the punctuation and/or quoting the characters.

```suggestion
    <!-- Use '/' instead of '\': used directly by WiX which doesn't accept trailing '\'. -->
```

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
```diff
@@ -113,18 +122,69 @@
...
+  </Target>
+
   <!-- Clean up artifacts that dotnet-publish generates which we don't need -->
   <Target Name="DeleteUnnecessaryPublishOutputs" AfterTargets="Publish">
     <ItemGroup>
       <ToDelete Include="$(PublishDir)apphost$(ExeSuffix)" />
       <ToDelete Include="$(PublishDir)$(LibPrefix)comhost$(LibSuffix)" />
       <ToDelete Include="$(PublishDir)$(LibPrefix)hostfxr$(LibSuffix)" />
       <ToDelete Include="$(PublishDir)$(LibPrefix)ijwhost$(LibSuffix)" />
+      <ToDelete Include="$(PublishDir)$(LibPrefix)ijwhost.lib" />
```
MichaelSimons: Can you explain the difference between these?

```
      <ToDelete Include="$(PublishDir)$(LibPrefix)ijwhost$(LibSuffix)" />
      <ToDelete Include="$(PublishDir)$(LibPrefix)ijwhost.lib" />
```

dir.props
dagood: Good call, your assumption is correct and I prefer the quoting.

src/pkg/packaging-tools/sharedFramework/sharedFramework.csproj
dagood: This is because `LibSuffix` only handles dynamic libraries (`.dll`/`.so`), but we also build ijwhost as a static library (`.lib`/`.a`/?). It's only built on Windows so I don't have to worry about handling other extensions for now.

Adding a comment to explain.

dagood:
This missed the Preview 5 snap.

@dsplaisted @nguerrera let me know if you think this qualifies as a critical bugfix for Preview 5. From the description of what it's solving, it seems like a QoL change that isn't critical.

dsplaisted:
No, we did not want to take this for preview 5.  From our team's POV, you can merge it as soon as the branch is open for preview 6.

Issue #5940, opened 2019-04-19
Fix WindowsDesktop crossgen warnings for DirectWriteForwarder and System.Printing


dagood:
When the build attempts to crossgen these files, we get warnings:

```
Warning: Could not load type 'System.Runtime.CompilerServices.SuppressMergeCheckAttribute' from assembly 'System.Runtime, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. while resolving 0x1000002 - System.Runtime.CompilerServices.SuppressMergeCheckAttribute.
Warning: Could not load type 'System.Runtime.CompilerServices.DecoratedNameAttribute' from assembly 'System.Runtime, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. while resolving 0x1000044 - System.Runtime.CompilerServices.DecoratedNameAttribute.
Warning: Could not load type 'System.Runtime.CompilerServices.AssemblyAttributesGoHereM' from assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. while resolving 0x100006d - System.Runtime.CompilerServices.AssemblyAttributesGoHereM.
Warning: Could not load type 'System.Runtime.CompilerServices.AssemblyAttributesGoHere' from assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. while resolving 0x1000071 - System.Runtime.CompilerServices.AssemblyAttributesGoHere.
Warning: Could not load type 'System.Runtime.InteropServices.Marshal' from assembly 'System.Runtime, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. while resolving 0x1000090 - System.Runtime.InteropServices.Marshal.
Warning: Could not load type 'System.Runtime.CompilerServices.SuppressMergeCheckAttribute' from assembly 'System.Runtime, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. while resolving 0xa000001 - System.Runtime.CompilerServices.SuppressMergeCheckAttribute..ctor.
[...]

PR #5942, opened 2019-04-19
Crossgen the WindowsDesktop shared framework using standardized crossgen infra

dagood:
This moves the crossgen infra into the new shared packaging targets, and uses it to crossgen both netcoreapp and windowsdesktop. I split it into a move commit and a change commit to highlight what I actually had to change to make it apply to both frameworks.

Resolves https://github.com/dotnet/core-setup/issues/5580.

Two outstanding issues:
* https://github.com/dotnet/core-setup/issues/5940 There are warnings due to a missing dependency not being given to crossgen. This is fine until we figure out what the dependency is.
* https://github.com/dotnet/core-setup/issues/5941 We should use `/r` to pass in the specific `WindowsBase.dll` we want, not rely on the order of directories passed to `platform_assemblies_paths`.

@vatsan-madhavan @zsd4yr @merriemcgaw


src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -90,6 +90,243 @@
...
+            GetCrossgenPackagePaths;
+            GetFilesFromPackages">
+    <PropertyGroup>
+      <_crossGenIntermediatePath>$(IntermediateOutputPath)/crossgen</_crossGenIntermediatePath>
+    </PropertyGroup>
+    <!-- Ensure crossgen is executable.  See https://github.com/NuGet/Home/issues/4424 -->
+    <Exec Command="chmod u+x $(_crossGenPath)"
+          Condition="'$(OSGroup)' != 'Windows_NT'" />
+    <ItemGroup>
+      <!-- Exclude run crossgen on System.Runtime.WindowsRuntime for non-Windows due to System.Runtime.WindowsRuntime.dll for non-Windows only throw NotSupported Exception  -->
```
ericstj: IIRC it's actually due to lack of Windows.winmd cross-plat, the reason it's OK is because it's a not supported assembly.

vatsan-madhavan:
lgtm. 

/cc @ryalanms

src/pkg/packaging-tools/framework.dependency.targets
dagood: I didn't have to touch this beyond the move for this PR, will update the comments though.

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -90,6 +90,243 @@
...
+      <!-- Also crossgen on non-windows doesn't support -Platform_Winmd_Paths -->
+      <!-- Exclude crossgen for resource DLLs in 'cs/', 'de/', ... subdirectories. -->
```
dagood: I'm actually not sure why I assume resource DLLs shouldn't be crossgenned. Let me know if these should be, or if there are concrete reasons either way. I'll merge this PR without crossgenning them if I don't hear otherwise.

zsd4yr: isn't crossgening resource dll's sort of a no-op? Because they contain no logical code?

dagood: That would make sense, but crossgenning reduces the size of e.g. `de/PresentationCore.resources.dll` from 103 KB to 94 KB. 😕

vatsan-madhavan: /cc @fadimounir

dagood: GitHub won't let me attach the input/output to help compare, but this RSP will do it in the repro dir I shared via email:

```
-readytorun
-in wd\de\PresentationCore.resources.dll
-out wd-out/PresentationCore.resources.dll
-platform_assemblies_paths wd\;corefx\;coreclr\
-Platform_Winmd_Paths winmd
-JITPath jit\clrjit.dll
```

dagood: Actually, filed https://github.com/dotnet/core-setup/issues/5975 to track this better rather than a PR comment thread.

Issue #5337, opened 2019-03-05
Build with a 3.0 dotnet SDK toolset in master


dagood:
This tracks implementation work to use a 3.0 toolset in `master`. See https://github.com/dotnet/core-setup/pull/5334.

dagood:
Done in arcade-migration dev branch, marking rollout awaiting merge into master.

dagood:
Done: arcade migration merged.

PR #5954, opened 2019-04-20
SingleFile: Add test cases

swaroop-sridhar:
Add test cases to test:
  * Run of self-contained and portable apps.
  * Extraction to custom directories.
  * Reuse of extracted files on subsequent runs.


src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -242,6 +242,7 @@ private void ValidateRequiredDirectories(RepoDirectoriesProvider repoDirectories
             DotNetCli dotnet = null,
             string runtime = null,
             string framework = null,
+            string selfContained=null,
```
sbomer:
```suggestion
            string selfContained = null,
```

src/test/HostActivationTests/BundledAppWithSubDirs.cs
```diff
@@ -18,59 +18,100 @@ public BundledAppWithSubDirs(SharedTestState fixture)
...
-            var bundler = new Microsoft.NET.HostModel.Bundle.Bundler(hostName, singleFileDir);
-            string singleFile = bundler.GenerateBundle(fixture.TestProject.OutputDirectory);
+        private static string GetPublishPath(TestProjectFixture fixture)
+        {
+            return Path.Combine(fixture.TestProject.ProjectDirectory, "publish");
+        }
 
-            // Run the bundled app (extract files)
-            Command.Create(singleFile)
+        private static DirectoryInfo GetBundleDir(TestProjectFixture fixture)
```
sbomer: Maybe move these static functions into a helper class instead of duplicating them in each test source file.

src/test/Microsoft.NET.HostModel.Tests/BundlerConsistencyTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        [Fact]
+        public void TestWithoutSpecifyingHostFails()
+        {
+            var fixture = sharedTestState.TestFixture
+                .Copy();
+
+            var hostName = GetHostName(fixture);
+            var appName = Path.GetFileNameWithoutExtension(hostName);
```
sbomer: Should these tests also be using a `BundleApp` helper?

swaroop-sridhar: The `BundleApp()` helper is a temporary step until Core-setup is updated to use 3.0 SDK for its testing. https://github.com/dotnet/core-setup/issues/5337 
Once the new SDK is available, `BundleApp` helper should be removed in favor of passing the `/p:PublishSingleFile=true` property to publish.

However this test should continue to invoke the bundler directly, since it is a test for the `HostModel` library.

src/test/HostActivationTests/BundledAppWithSubDirs.cs
swaroop-sridhar: I thought about that -- but these are auxiliary one-line helpers. Plus there was no good place to put the helpers because: 
* They depend on the specific way in which files are published, bundled, or extracted (directory names and such). So, it is not a good idea to add them to TestUtilities generically.
* They are used both by HostActivationTests and HostModel test.

So I let them be in the individual tests. I'm happy to extract the code out to a helper if we feel strongly about it.

vitek-karas: If you move the tests into a sub-namespace, then adding a bundle only helpers should be trivial.
Alternatively you can have a base class for the bundle tests and share that way.

I agree that moving these to TestUtils is probably an overkill, but currently there's already duplication in HostActivationTests alone.

src/test/TestUtils/TestProjectFixture.cs
```diff
@@ -268,6 +269,12 @@ private void ValidateRequiredDirectories(RepoDirectoriesProvider repoDirectories
                 publishArgs.Add($"/p:NETCoreAppFramework={framework}");
             }
 
+            if(selfContained != null)
```
vitek-karas: Nit: 
```suggestion
            if (selfContained != null)
```

swaroop-sridhar:
Thanks for the feedback @vitek-karas. I've addressed it in https://github.com/dotnet/core-setup/pull/5954/commits/f832be505f38b251558619e4b9e6ecd509246aee

src/test/HostActivationTests/BundledAppWithSubDirs.cs
swaroop-sridhar: I Created a BundleTest test directory grouping which contains
- Microsoft.NET.HostModel.Bundle tests
- AppHost Bundler tests
- BundleHelpers -- containing common helper methods for the above two tests.

vitek-karas:
I must admit that when I suggested to move the tests into a subfolder I only thought about the HostActivation part of the tests. But what you did also makes sense...

LGTM!

Issue #5969, opened 2019-04-21
Core-Setup => Core-SDK uptake failing in master


wli3:
```
/home/vsts/work/1/s/src/redist/targets/Crossgen.targets(105,5): error : Unable to load Jit Compiler: libclrjit.so, hr=0x8007007e [/home/vsts/work/1/s/src/redist/redist.csproj]
/home/vsts/work/1/s/src/redist/targets/Crossgen.targets(105,5): error : Error: compilation failed for "/home/vsts/work/1/s/artifacts/bin/redist/Debug/dotnet/sdk/3.0.100-preview5-011521/Roslyn/bincore/csc.dll" (0x8007007e) [/home/vsts/work/1/s/src/redist/redist.csproj]

```

https://github.com/dotnet/core-sdk/pull/1627

dagood:
The first update commit in that PR appears to have failed in the same way, and has a single-commit diff:

https://github.com/dotnet/core-setup/compare/04641968d5ab5bda06d046737ef1685b5a9184ec...f78b553ee37fbe18d4d3ba5972b74cfcc560d1aa

Core-SDK PR validation shows that this error happens on ARM and ARM64. I would guess my changes to enable crossgen on WindowsDesktop made the netcoreapp on those archs carry the wrong crossgen or something like that. Looking into it.

dagood:
A diff that definitely seems responsible is `runtime.linux-arm.Microsoft.NETCore.App` now missing `runtimes/x64_arm/native/libclrjit.so`.

PR #5986, opened 2019-04-22
Revert GetRuntimeFilesFromPackages target order

dagood:
Potential fix for https://github.com/dotnet/core-setup/issues/5969.


dagood:
No effect in a local build I ran to try it out. This target was for the cross-targeted `crossgen`, which was working fine.

Now I have a local build that I can use to test further fixes though which runs pretty fast, so I'll validate before next submission.

PR #5987, opened 2019-04-22
Fix _jitPackageDir ref to include cross build jit

dagood:
Fixes https://github.com/dotnet/core-setup/issues/5969.

Refactor went wrong, causing `x64_arm64` (for example) to get missed because `_jitPackagePath` was replaced by `_jitPackageDir`. Verified fix with a local `/p:TargetArchitecture=arm64` build.


dagood:
FYI @dleeapho, this issue was a missed reference to a variable that went away when I did a bit of generalization on the crossgen targets.

PR #6003, opened 2019-04-22
Fix FixLayoutPermissions ordering for AppHost +x permission in installers

dagood:
Make sure `FixLayoutPermissions` runs while generating macOS pkg, and run it before copying to the secondary layout for Deb/RPM.

This is a fix for https://github.com/dotnet/cli/issues/11231. If `x` isn't set in Core-Setup, it needs to be set by the CLI tooling on the published app, which seems like a riskier fix.

### DIFF
--- a/src/pkg/packaging-tools/framework.packaging.targets
+++ b/src/pkg/packaging-tools/framework.packaging.targets
@@ -167,7 +167,9 @@
     Create macOS pkg installer.
   -->
   <Target Name="CreatePkg"
-          DependsOnTargets="GetInstallerProperties">
+          DependsOnTargets="
+            GetInstallerProperties;
+            FixLayoutPermissions">
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


dagood:
Confirmed the fix with a local build using Suspicious Package to check the contents before/after.

wli3:
It also includes Linux? I think Linux should have the same bug

dagood:
This PR should fix it there too. Agreed I'd expect that to have the same issue. Running a build now to validate the bit's set in Deb+RPM.

dagood:
Confirmed, `x` is set for e.g. `/usr/share/dotnet/packs/Microsoft.NETCore.App.Host.linux-x64/3.0.0-preview6-27622-0/runtimes/linux-x64/native/apphost` when installing an RPM/Deb with this change.

PR #6096, opened 2019-04-25
Remove ".Internal" from runtime packages

dagood:
Only the identity package should be ".Internal", not the runtime packages.

The desired set of netcoreapp packages until Core-Setup produces proper runtime packs is:

* `Microsoft.NETCore.App`: workaround package (https://github.com/dotnet/core-setup/issues/5663) that has no runtime.json and only one dependency on Microsoft.NETCore.Platforms.
* `Microsoft.NETCore.App.Internal`: identity package that brings in runtime packages.
* `runtime.<rid>.Microsoft.NETCore.App`: the runtime package, used as a runtime pack by the current SDK. Customers see this so it's not ideal for it to have `.Internal` at the end.

This is done by setting up one `Microsoft.NETCore.App.pkgproj` that creates the workaround package, and another that produces runtime packages. Then, a `Microsoft.NETCore.App.Internal.pkgproj` project references the one that produces runtime packages.

This is probably doable with a single `Microsoft.NETCore.App.pkgproj` but a multi-project approach seems more straightforward than trying to get all the conditions right for identity vs. runtime builds.

@dsplaisted @MichaelSimons @natemcmaster


ericstj:
Seems reasonable, the packaging targets don't really allow for much extensibility here so using a single project would require changes to them: https://github.com/dotnet/arcade/blob/d37270268a65592cae630f1b979b70f74d4614dd/src/Microsoft.DotNet.Build.Tasks.Packaging/src/build/Packaging.targets#L42-L43

dagood:
Thanks! Yeah, I'm hoping to be able to avoid making Tasks.Packaging changes until Core-Setup uses the Arcade one.

PR #6128, opened 2019-04-27
Skip System.Windows.Forms crossgen

dagood:
[Taken from mail thread] Winforms assembly is not able to find the **“Win32Manifest”** resource ( that was embedded into the assembly) and thus native call **toCreateActCtx(..)** fail with  resource not found. This is the case with cross-gen’d assemblies but not with raw assemblies that we build from our repo.

### DIFF
--- a/src/pkg/packaging-tools/framework.dependency.targets
+++ b/src/pkg/packaging-tools/framework.dependency.targets
@@ -192,6 +192,8 @@
         System.Runtime.WindowsRuntime is only supported on Windows.
 
         Resource DLLs in 'cs/', 'de/', ... subdirectories.
+
+        System.Windows.Forms. If crossgen'd, the Win32Manifest resource can't be found.
       -->
       <_filesToCrossGen
         Include="@(FilesToPackage)"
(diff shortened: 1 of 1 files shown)


src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -222,6 +227,8 @@
         so in case of conflicts, DLLs are found in the framework being crossgenned.
       -->
       <_crossgenPlatformDirectories Include="%(_filesToCrossGen.RootDir)%(_filesToCrossGen.Directory)" />
+      <!-- We need to include System.Windows.Forms as a platform assembly even if we don't crossgen it.-->
+      <_crossgenPlatformDirectories Include="@(WinFormsPlatformDirectory)" />
```
dagood: I don't think this line is actually necessary here, only in `CreateCrossGenSymbols`. I'm running a new build now to see if I can remove it.

dagood: Yeah, new build worked (and still fixed the issue).

dagood:
@ericstj, @brianrob (maybe? 😄), @dleeapho, would appreciate if you can take a look. If not, the plan is to merge tomorrow to start work on getting a build through the weekend. This seems to me like a pretty safe change.
/cc @mmitche

sergiy-k:
@dagood, https://github.com/dotnet/coreclr/issues/24301 has been fixed in CoreCLR master. This change can be reverted for Preview 6.

dagood:
@sergiy-k Thanks for the ping, submitted https://github.com/dotnet/core-setup/pull/6379 to revert.

PR #6168, opened 2019-04-29
Increase GUI test timeout and add more diagnostics.

### LINKED ISSUE #6117: Flaky test: Running_AppHost_with_GUI_Reports_Errors_In_Window
https://dev.azure.com/dnceng/internal/_build/results?buildId=170609

```
  [xUnit.net 00:01:47.5846677]     Microsoft.DotNet.CoreSetup.Test.HostActivation.StandaloneAppActivation.Running_AppHost_with_GUI_Reports_Errors_In_Window [FAIL]
  Failed   Microsoft.DotNet.CoreSetup.Test.HostActivation.StandaloneAppActivation.Running_AppHost_with_GUI_Reports_Errors_In_Window
  Error Message:
   Assert.NotEqual() Failure
  Expected: Not 0
  Actual:   0
  Stack Trace:
     at Microsoft.DotNet.CoreSetup.Test.HostActivation.StandaloneAppActivation.Running_AppHost_with_GUI_Reports_Errors_In_Window() in F:\workspace\_work\1\s\src\test\HostActivationTests\StandaloneAppActivation.cs:line 263
```

This sometimes fails my PR validation too. I don't think it's flaky to the point of being critical (it's not too common, and hitting retry has always worked in my experience) but of course we don't want stuff like this to build up or delay release blockers.

Based on the name of the test alone (`Running_AppHost_with_GUI_Reports_Errors_In_Window`) I wouldn't doubt this being a machine issue, but I'm not familiar with what can go wrong with this kind of dialogue box test.

/cc @vitek-karas

vitek-karas:
Since we don't have enough info to know why the tests were failing before, for now I'm increasing the timeout and adding better diagnostics.

The one failure mentioned in #6117 is interesting in that there are two tests which do the same thing (only small variation), and only one failed while the other worked. Which supports the theory of too short timeout.

Fixes #6117

### DIFF
--- a/src/test/HostActivationTests/StandaloneAppActivation.cs
+++ b/src/test/HostActivationTests/StandaloneAppActivation.cs
@@ -320,16 +320,22 @@ namespace Microsoft.DotNet.CoreSetup.Test.HostActivation
         [DllImport("user32.dll")]
         private static extern bool EnumThreadWindows(int dwThreadId, EnumThreadWindowsDelegate plfn, IntPtr lParam);
 
-        private IntPtr WaitForPopupFromProcess(Process process, int timeout = 5000)
+        private IntPtr WaitForPopupFromProcess(Process process, int timeout = 30000)
         {
             IntPtr windowHandle = IntPtr.Zero;
+            StringBuilder diagMessages = new StringBuilder();
             while (timeout > 0)
             {
                 foreach (ProcessThread thread in process.Threads)
                 {
                     // Note we take the last window we find - there really should only be one at most anyway.
                     EnumThreadWindows(thread.Id,
-                        (hWnd, lParam) => { windowHandle = hWnd; return true; }, IntPtr.Zero);
+                        (hWnd, lParam) => {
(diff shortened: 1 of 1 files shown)


src/test/HostActivationTests/StandaloneAppActivation.cs
```diff
@@ -341,6 +347,10 @@ private IntPtr WaitForPopupFromProcess(Process process, int timeout = 5000)
                 timeout -= 100;
             }
 
+            Assert.True(
```
danmosemsft: When I've had tests that seemed to be timing out, but I didn't know whether they were hung or just needed more timeout, I used the following scheme
1. If test completes before X seconds, mark it as pass. Otherwise set a flag.
2. If test completes before X+N seconds, fail, logging X+N
3. If test has not completed after X+N seconds, fail, logging that it's hung.
Eg., I might use 30 seconds for X and 90 seconds for N.
Then next time it fails I can see whether it might need a longer X, or it's actually likely hung.

PR #6198, opened 2019-04-30
Build WindowsDesktop targeting pack installer and package assets for Linux and macOS

dagood:
The SDK currently expects the same targeting packs to available regardless of platform, and there's no known reason to omit WindowsDesktop bits from non-Windows. The WindowsDesktop targets only run on Windows, but it's possible to avoid the targets and compile against the reference DLLs directly.

### DIFF
--- a/src/pkg/projects/windowsdesktop/pkg/dir.props
+++ b/src/pkg/projects/windowsdesktop/pkg/dir.props
@@ -15,12 +15,6 @@
     <FrameworkListFrameworkName>$(FrameworkPackageName)</FrameworkListFrameworkName>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <BuildDebPackage>false</BuildDebPackage>
-    <BuildRpmPackage>false</BuildRpmPackage>
-    <GeneratePkg>false</GeneratePkg>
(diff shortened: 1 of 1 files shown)


johnbeisner:
@dagood 
@dsplaisted 
What is the status of this PR?

dagood:
Waiting on a final decision in the "Offline SDK restores for non-Windows" email thread. This still seems to be the pragmatic solution--doing it "right" (download these bits separately from the gallery in the very unlikely case someone on non-Windows has a practical use for them) has significant cost.

dsplaisted:
We decided not to support building projects using WindowsDesktop (SDK or Shared Framework) on non-Windows operating systems for now.  See https://github.com/dotnet/cli/issues/10842 and https://github.com/dotnet/cli/issues/11410

dagood:
Thanks for the update! It sounds like this bit in particular is what's addressing the offline non-Windows non-WD scenario:

> https://github.com/dotnet/cli/issues/11410 On non-Windows, don't pre-emptively download WindowsDesktop targeting or runtime packs (which we normally do in case we encounter a transitive FrameworkReference)

PR #6246, opened 2019-05-02
[release/2.1] Generalize runtime deps Deb/RPM packages

dagood:
### Description

Adds openSUSE 15.0 `dotnet-runtime-deps` support (https://github.com/dotnet/core-setup/issues/5628) and generally expands `dotnet-runtime-deps` applicability by making the package dependencies more inclusive.

* On RPM package distros, depend on the `libicu` "capability" that a `libicu##` package provides.
* On Debian package distros, use "or" dependencies that list all known versions of libicu and libssl.

Using "or" dependencies for Debian packages lets us collapse the now-identical Deb packages into one. To do this, remove JSON configs and pipebuild steps, and update the PublishRid list.

Cherry picked from https://github.com/dotnet/core-setup/pull/6070 and https://github.com/dotnet/core-setup/pull/6213.

### Customer Impact

Without this change, .NET Core can't be installed on openSUSE 15.0 through RPM packages.

The Debian package changes have two effects:
* Reduce dev and release maintenance required for new distros and versions.
* Makes the Deb packages forgiving if a user has the wrong package repository set up. E.g. the packages for Ubuntu 16.04 also install fine on Ubuntu 19.04. (This occurs fairly often based on issues filed on the dotnet/core repo.)

### Regression?

No.

### Risk

Not significant. I tested install through to `dotnet --info` in Docker containers for each distro. The build pipeline changes are routine.


vivmishra:
Approved for 2.1.12

PR #6366, opened 2019-05-07
Add support for call to Register/Unregister COM delegate

AaronRobinsonMSFT:
Fixes https://github.com/dotnet/coreclr/issues/24394

See https://github.com/dotnet/coreclr/pull/24452

### DIFF
--- a/src/corehost/cli/comhost/comhost.cpp
+++ b/src/corehost/cli/comhost/comhost.cpp
@@ -42,14 +42,14 @@ struct com_activation_context
     void **class_factory_dest;
 };
 
-using com_activation_fn = int(STDMETHODCALLTYPE*)(com_activation_context*);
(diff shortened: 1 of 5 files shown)


AaronRobinsonMSFT:
cc @elinor-fung @vitek-karas

AaronRobinsonMSFT:
@vitek-karas We definitely can, but there is going to be a tough issue here and I am not sure how worth it it is. Calling the `DllRegisterServer` or `DllUnregisterServer` is going to - by design - attempt access to the registry. This means we would need to mock out the COM host itself just to avoid hitting the registry in that path _or_ run the test elevated. I am doubtful this is worth it the pain, but am open to other opinions.

I guess there is another option, but I would like to chat before going down this route. Instead of always writing to HKLM, the COMhost DLL could detect if it is in an elevated process and if it is not write to HKCU - which does not require elevated access. This will have the unfortunate aspect of impacting global machine state though.

Thoughts?

PR #6515, opened 2019-05-22
Update Microsoft.NetCore.Targets dependency to latest 3.0.0 version

wtgodbe:
Reacts to https://github.com/dotnet/corefx/pull/37600 - allows Core-Setup to depend on the latest version of the `Targets` package from 3.0.0, rather than being pinned on the 2.0.0 version.

CC @dagood @joperezr @ericstj


wtgodbe:
Should unblock https://github.com/dotnet/core-setup/pull/6502

ericstj:
We were just matching what we did in the past and were planning to delete it, but the SDK created a scenario for us to keep it (due to lacking nuget feature).

PR #6541, opened 2019-05-25
Alter the way comhost looks for the loose .clsidmap file

AaronRobinsonMSFT:
cc @vitek-karas


AaronRobinsonMSFT:
>Just thinking how hard it would be to add a test, but we probably don't have the infra tools to do this easily.

@vitek-karas Agreed. Basically we would have to do the very thing we don't ever want to do in this repo.

vitek-karas:
@AaronRobinsonMSFT Good point - I didn't even realize that... definitely don't then.

PR #6576, opened 2019-05-30
Single-File: MemoryMap the bundle file

swaroop-sridhar:
This changes makes a few changes in preparation for
loading IL files from the single-file bundle.

**Memory Mapping**

At startup, the entire bundle-file is memory mapped.
The individual embedded files are then extracted out to actual files on disk on startup.
However, this change will facilitate later checkins to:
* Loading files directly from bundle (ex: IL files using PELoader, config files using memory-streams, etc ).
* Extracting out files to disk lazily

Further details are available in [this document](https://github.com/dotnet/designs/blob/master/accepted/single-file/design.md#the-host)

**Refactoring**

* Static bundle_runner: The bundle_runner is made a static class that holds a global pointer
  to the memory-mapped application bundle, and meta-data about the contents.
  This facilitates the implementation of call-backs for accessing content directly from the bundle,
  as explained in [this document](https://github.com/dotnet/designs/blob/master/accepted/single-file/direct.md)

* A few new classes are created by factoring out functionality from bundle_runner
   * Extractor: Functionality for extracting contents of a bundle to files on disk.
   * Reader: This is a helper class that implements sequentional reading on memory-mapped bundle file.
   * Util: File handling utilities that assist in processing the bundle.

* Naming: bundle::bundle_runner is renamed as bundle::runner to remove the redundancy.
  Similarly, the newly refactored classes are named bundle::extractor and bundle::reader.

**Other Issues**

 If the current AppHost is not a bundle, process_manifest_footer() returns a bool value
  to indicate the same, rather than throwing StatusCode::AppHostExeNotBundle.
  This ensures that we do not throw every time a non-bundled app-host runs, which
  is undesirable (especially while debugging).


src/corehost/cli/apphost/bundle/extractor.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (!pal::getenv(_X("DOTNET_BUNDLE_EXTRACT_BASE_DIR"), &m_extraction_dir))
+    {
+        if (!pal::get_temp_directory(m_extraction_dir))
+        {
+            trace::error(_X("Failure processing application bundle."));
+            trace::error(_X("Failed to determine location for extracting embedded files"));
+            throw StatusCode::BundleExtractionFailure;
+        }
+
+        append_path(&m_extraction_dir, _X(".net"));
```
vitek-karas: In .NET Core we typically use `dotnet` as the name which refers to the product when used by file/directory names. Maybe I'm biased, but if I would see `.net` I would think .NET Framework.

src/corehost/cli/apphost/bundle/extractor.cpp
```diff
@@ -0,0 +1,150 @@
...
+// If DOTNET_BUNDLE_EXTRACT_BASE_DIR is not set in the environment, the 
+// base directory defaults to $TMPDIR/.net
+void extractor_t::determine_extraction_dir()
+{
+    if (!pal::getenv(_X("DOTNET_BUNDLE_EXTRACT_BASE_DIR"), &m_extraction_dir))
+    {
+        if (!pal::get_temp_directory(m_extraction_dir))
+        {
+            trace::error(_X("Failure processing application bundle."));
+            trace::error(_X("Failed to determine location for extracting embedded files"));
```
vitek-karas: Nit - period at the end of the sentence

src/corehost/cli/apphost/bundle/extractor.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    trace::info(_X("Files embedded within the bundled will be extracted to [%s] directory"), m_extraction_dir.c_str());
+}
+
+// Compute the working extraction location for this process, before the 
+// extracted files are committed to the final location
+// m_working_extraction_dir = $DOTNET_BUNDLE_EXTRACT_BASE_DIR/<app>/<proc-id-hex>
+void extractor_t::determine_working_extraction_dir()
+{
+    m_working_extraction_dir = get_directory(m_extraction_dir);
```
vitek-karas: This relies on the correct order of external calls... which is very fragile. Instead this should call `extraction_dir()`.

src/corehost/cli/apphost/bundle/manifest.h
```diff
@@ -5,44 +5,44 @@
...
-#pragma pack(push, 1)
-        struct
-        {
-            uint32_t major_version;
-            uint32_t minor_version;
-            int32_t num_embedded_files;
-            int8_t bundle_id_length_byte_1;
-        } m_data;
-#pragma pack(pop)
+        const manifest_header_fixed_t* m_data;
```
vitek-karas: In `file_entry` you make a copy of the memory mapped data to avoid IO later on. Here you don't. I don't know how big of a problem the IO can be, but we should be consistent. So either one or the other, but everywhere.

src/corehost/cli/apphost/bundle/reader.h
```diff
@@ -0,0 +1,69 @@
...
+            return ptr;
+        }
+
+        size_t read_path_length();
+        void read_path_string(pal::string_t &str);
+
+    private:
+        const int8_t* m_ptr;
+        const int8_t* m_base_ptr;
+    };
```
vitek-karas: We must store the length of data available beyond the base pointer. And then all operations in this class must check that they don't overflow the buffer.

vitek-karas: It should also use secure arithmetic to do the math - it absolutely must avoid overflows. I know that in theory this is not processing untrusted data, but even with trusted data one can get overflows which end up crashing the process. The crash itself might be exploitable.

src/corehost/cli/apphost/bundle/runner.h
```diff
@@ -0,0 +1,40 @@
...
+#include "error_codes.h"
+
+namespace bundle
+{
+    class runner_t
+    {
+    public:
+        static StatusCode extract();
+
+        static pal::string_t bundle_path;
```
vitek-karas: I think it would be better to:
* expose everything as methods
* be consistent - `extraction_path` is a method, `bundle_path` is a field.

I would also prefer `get_bundle_path` and so on...

src/corehost/cli/apphost/bundle/runner.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+using namespace bundle;
+
+manifest_t* runner_t::s_manifest = nullptr;
+int32_t runner_t::s_num_embedded_files = 0;
+pal::string_t runner_t::bundle_path;
+pal::string_t runner_t::s_bundle_id;
+pal::string_t runner_t::s_extraction_path;
+int8_t* runner_t::s_bundle_map = nullptr;
+size_t runner_t::s_bundle_length = 0;
```
vitek-karas: Can we instead make the runner instantiable and have a single global of type runner_t (or something similar)? Or maybe create a class just for the global state.
It would read nicer than what looks like a random collection of global variables.

src/corehost/cli/apphost/bundle/runner.cpp
```diff
@@ -0,0 +1,108 @@
...
+        throw StatusCode::BundleExtractionIOError;
+    }
+}
+
+// Checks if this host binary has a valid bundle footer.
+// If so, it sets reader to the position of the manifest header, and returns true.
+// If not, returns false.
+bool runner_t::process_manifest_footer(reader_t &reader)
+{
+    reader.set_offset(s_bundle_length - sizeof(manifest_footer_t));
```
vitek-karas: secure math - if this underflows we're... screwed  :-)

src/corehost/cli/apphost/bundle/runner.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+        reader_t reader(s_bundle_map);
+        
+        // Determine if the current executable is a bundle
+        //
+        //  If the current AppHost is a bundle, it's layout will be 
+        //    AppHost binary 
+        //    Embedded Files: including the app, its configuration files, 
+        //                    dependencies, and possibly the runtime.
+        //    Bundle Manifest Footer
```
vitek-karas: This only mentions the footer - there's also the header and then the file entries.

src/corehost/common/pal.windows.cpp
```diff
@@ -76,6 +76,46 @@ bool pal::touch_file(const pal::string_t& path)
     return true;
 }
 
+void* pal::map_file_readonly(const pal::string_t& path, size_t &length)
```
vitek-karas: Tabs -> Spaces

src/corehost/common/pal.windows.cpp
```diff
@@ -76,6 +76,46 @@ bool pal::touch_file(const pal::string_t& path)
...
+	void *address = MapViewOfFile(map, FILE_MAP_READ, 0, 0, 0);
+
+	if (map == NULL)
+	{
+		trace::warning(_X("Failed to map file. MapViewOfFile(%s) failed with error %d"), path.c_str(), GetLastError());
+		CloseHandle(file);
+		return nullptr;
+	}
+
+	return address;
```
vitek-karas: I think the function needs a comment about the fact that there's no way to close the mapping (or the file).
I wonder if this is actually OK for the bundler. Especially in this stage where everything is extracted to disk - we would be keeping potentially large amount of memory blocked by the file mapping. And we would unnecessary stress on the memory manager by using all those pages (to extract the files to disk), but never freeing those. Eventually they would get evicted from physical ram, but it feel unnecessary.
We should either close the mapping (and thus map part by part probably), or if there's a way tell the memory managed to discard the mapped pages we know we're not going to need anymore.

src/corehost/common/pal.unix.cpp
```diff
@@ -57,6 +57,37 @@ bool pal::touch_file(const pal::string_t& path)
...
+	struct stat buf;
+	if (fstat(fd, &buf) != 0)
+	{
+		trace::warning(_X("Failed to map file. fstat(%s) failed with error %d"), path.c_str(), errno);
+        close(fd);
+        return nullptr;
+	}
+
+	length = buf.st_size;
+	void* address = mmap(nullptr, , PROT_READ, MAP_SHARED, fd, 0);
```
vitek-karas: You're probably missing the `length` parameter. This probably won't even compile.

src/corehost/common/pal.unix.cpp
```diff
@@ -57,6 +57,37 @@ bool pal::touch_file(const pal::string_t& path)
...
+
+	if(address == nullptr)
+	{
+		trace::warning(_X("Failed to map file. mmap(%s) failed with error %d"), path.c_str(), errno);
+        close(fd);
+        return nullptr;
+	}
+
+	close(fd);
```
vitek-karas: Hmm - here we close the file, on Windows we don't... is there a reason?

src/corehost/corehost.cpp
```diff
@@ -112,13 +112,13 @@ int exe_start(const int argc, const pal::char_t* argv[])
         requires_v2_hostfxr_interface = true;
     }
 
-    bundle::bundle_runner_t extractor(host_path);
-    StatusCode bundle_status = extractor.extract();
+    bundle::runner_t::bundle_path = host_path;
```
vitek-karas: I personally would prefer the runner to be an instance. And then pass the pass as an argument to the extract... setting a field like this doesn't feel right.

src/corehost/cli/apphost/bundle/extractor.cpp
swaroop-sridhar: The directory to be extracted to is a topic of ongoing discussion; and I would like to address it separately. 

This PR doesn't change the extraction directory, it keeps the extraction location as described in the [design](https://github.com/dotnet/designs/blob/master/accepted/single-file/extract.md).

src/corehost/cli/apphost/bundle/manifest.h
swaroop-sridhar: As noted in the comments [here](https://github.com/dotnet/core-setup/pull/6576/files#diff-3d1ffec12bf784039f08581e68f7533fR41), the difference is in the usage pattern between the two kinds of headers. 

* I don't think we should read file-entry data from the bundle-file every time it is needed (ex: when loading an IL image in the middle of execution). It may cause additional file I/O. 

* The bundle header/footer on the other hand, is only read during startup. Of course, we can make a copy of the data here, but it is unnecessary, and will be discarded immediately.

The two are different structures with different lifetimes. So, I don't see why they need to be treated similarly.

src/corehost/common/pal.windows.cpp
swaroop-sridhar: We keep the mapping file open, because it's necessary to access parts of the file later (tbd in further stages). 

I don't understand the concern here.  Is it:
* Running out of virtual addresses? 
    * Even on 32 bit architectures, if the app is so large that it is running short of virtual addresses, I it is not a good candidate to be published as a single-exe. 
    * We can unmap the file after extraction to disk now, but such large apps will start to fail once we try to load files directly from bundle. 
    * Alternatively, we can map parts of the file (which is tantamount to the existing `fread-fwrite` solution). But I don't think we should optimize for this case.
* Paging ? 
    * The bundle is mapped as `O_RDONLY`. So, there is no overhead to evicting the pages -- they're just gone when physical memory is needed. 
    * I don't see why these pages should be released upfront, that should handled by the OS as needed.

src/corehost/common/pal.unix.cpp
swaroop-sridhar: Sorry, I'll fix this.
I carved out this branch from a larger change that attempts to load IL from the bundle. But I missed a fix that was out of sequence.

src/corehost/common/pal.unix.cpp
swaroop-sridhar: The [manual](http://man7.org/linux/man-pages/man2/mmap.2.html) for `mmap()` says: "After the mmap() call has returned, the file descriptor, fd, can be closed immediately without invalidating the mapping." So, we release the file-descriptor after the mapping.

The [manual](https://docs.microsoft.com/en-us/windows/desktop/api/memoryapi/nf-memoryapi-createfilemappingw) for `CreateFileMapping` seems to say the same thing (now that I've read it again) but in a less straightforward way. I'll update the Windows code to close the file handle.

src/corehost/cli/apphost/bundle/runner.cpp
swaroop-sridhar: I thought about this option: making `runner_t` instantiable, and having a separate `bundle_info_t` holding global data. However, most of the above fields will end up being in transferred/copied to `bundle_info_t`.  Alternatively, we can have a global object of `runner_t` type.

Anyway, I'll change `runner_t` to be an instantiable class for now. 
We can create the global static data when it is actually used by a later change.

src/corehost/corehost.cpp
swaroop-sridhar: https://github.com/dotnet/core-setup/pull/6576#discussion_r289160318

src/corehost/cli/apphost/bundle/runner.h
swaroop-sridhar: This will be subsumed by https://github.com/dotnet/core-setup/pull/6576#discussion_r289160318

src/corehost/cli/apphost/bundle/reader.h
swaroop-sridhar: I'll add a bounds check to this class because it is easy enough to do (and the cost of the check is likely not significant compared to the following I/O operations).

However, we have a lot of unmanaged code in the runtime where there are no explicit checks on every access (ex: PE Loader) .

src/corehost/common/pal.windows.cpp
vitek-karas: The concern is both. I'm not saying that either is blocking, but I would like us to consider improving this. The runtime (and this is part of the runtime effectively) always tries to be as efficient as possible to give most resources to the app itself.

VM space usage is something which some apps may be sensitive to, if nothing else, people often just look at the memory usage of an app and complain if it's too high. I agree that mapping each file separately is probably too much of a hassle (but it should still be better than the current read/write). That said when we're creating the bundle we know pretty well which files we're going to extract and which will be potentially read from memory. We could order the files in the bundle into two groups and then memory map the MSIL as a whole. For the files which will be spilled to disk I would probably go with memory mapping per file.

Paging - I agree that the OS will have no problem evicting the pages, the point was that for the files which were extracted we know for a fact that the memory will never be accessed again. It would be a very nice to tell OS about that fact. Otherwise OS will treat these pages the same as any other RO pages in the process and potentially chose RO pages which will be needed again for eviction over those we know will never be used.

Both these points are relatively subtle, but both are something we typically tend to look into as part of being "in the runtime".

src/corehost/cli/apphost/bundle/manifest.h
vitek-karas: Makes sense - maybe just add a comment... thanks!

src/corehost/cli/apphost/bundle/reader.h
vitek-karas: This makes me nervous. I do agree that we don't support loading untrusted assemblies and as such we don't expect being attacked directly through that mechanism. That said I would definitely like us to be secure anyway unless it causes sever performance issues, in which case we should have a discussion about it.

Basically my position is: Native buffer overruns are evil. Even if they're not security issues (which often they are in really crazy ways), they are a nightmare to debug. So we should avoid them as much as possible.

src/corehost/common/pal.windows.cpp
swaroop-sridhar: I think we should unmap the bundle file in the following cases:
* When the apphost is not a bundle
* When the apphost runs in self-extractor mode
    * Currently always
    * In later stages, when [`ExtractAllFiles`](https://github.com/dotnet/designs/blob/master/accepted/single-file/extract.md#extraction-configurations) property is set. 

I'll make this change. 

However, when files are actually used directly from the bundle, I'm not sure if mapping/unmapping in parts is warranted. I say this because:

* At startup, only the pages that are actually extracted out to disk will be touched, the rest of the file will only take up virtual-addresses, not the memory backing those addresses. So, this will not count towards memory footprint of the app.
* Mapping/unmapping in parts causes additional system call overhead during startup.
* As you mentioned, we could consider splitting the bundle into two parts -- (i) Files to be extracted, and (ii) Files to be used directly from bundle, plus meta-data -- and load them separately. But as implementation progresses to further stages, most of the files will be in the second set that lives through the process. 
* Regarding paging, I suggest that we measure the impact (if any) on some large apps once assemblies are actually loaded from the bundle. We can take mitigating steps if necessary.

swaroop-sridhar:
The latest commit rebased the changes against master, accounting for the following checkins which had conflicts: 
* https://github.com/dotnet/core-setup/commit/0dd08fe1414c690d3243739f3be4cd35f84f7dde#diff-5c2eb952c0365ec0c860557785f5d26fR201
* https://github.com/dotnet/core-setup/commit/bd497d99a0f029ebed76e5398b1db769c4cc0d3c#diff-9171b749b284f5183d0b0412f7a0bb64
* https://github.com/dotnet/core-setup/commit/19b99b3babd8cf3ff8a99089eb2f64e782b75b95#diff-9171b749b284f5183d0b0412f7a0bb64
* https://github.com/dotnet/core-setup/commit/552dba193c0f21a03687d2f14861c7a4dcf8dd0f#diff-9171b749b284f5183d0b0412f7a0bb64
*https://github.com/dotnet/core-setup/commit/e02bb0a4fa18355495e6af653de92d75d5a26ad3#diff-9171b749b284f5183d0b0412f7a0bb64 

WIP: Address some feedback on the PR.

src/corehost/common/pal.h
```diff
@@ -146,24 +146,25 @@ namespace pal
...
 
-    inline bool mkdir(const pal::char_t* dir, int mode) { return CreateDirectoryW(dir, NULL) != 0; }
-    inline bool rmdir (const pal::char_t* path) { return RemoveDirectoryW(path) != 0; }
-    inline int rename(const pal::char_t* old_name, const pal::char_t* new_name) { return ::_wrename(old_name, new_name); }
-    inline int remove(const pal::char_t* path) { return ::_wremove(path); }
+    inline bool mkdir(const char_t* dir, int mode) { return CreateDirectoryW(dir, NULL) != 0; }
+    inline bool rmdir (const char_t* path) { return RemoveDirectoryW(path) != 0; }
+    inline int rename(const char_t* old_name, const char_t* new_name) { return ::_wrename(old_name, new_name); }
+    inline int remove(const char_t* path) { return ::_wremove(path); }
+    inline void unmap_file(void* addr, size_t length) { return UnmapViewOfFile(addr) != 0; }
```
vitek-karas: The function returns `void` - so it can't report a failure - also the body now contains a `return` in a function which is `void`.

swaroop-sridhar:
@vitek-karas  I've made the changes you've suggested; please take a look. Thanks.

PR #6614, opened 2019-05-31
[release/2.1] Remove item metadata attributes for MSBuild 14

dagood:
The official build uses MSBuild 14 on Windows. When it tries to parse this file using arbitrary attribute names to add metadata to items, it fails. This was introduced in https://github.com/dotnet/core-setup/pull/6246 because it was only tested on Linux, where a newer MSBuild is used because it's built with .NET Core.


dagood:
The PR validation build on the original changes was actually green for Windows--I suspect the Jenkins definition was using a newer MSBuild. (Although the logs are gone now.) The new AzDO CI uses `15.9.20+g88f5fadfbe for .NET Core` so it wouldn't catch it either.

PR #6676, opened 2019-06-05
Fix building tizen rootfs

kbaladurin:


kbaladurin:
@alpencolt @gbalykov @jkotas PTAL

jkotas:
release/3.0 still takes integrations from master. It is sufficient to submit this to master only.

PR #6677, opened 2019-06-05
Fix building tizen rootfs

kbaladurin:


kbaladurin:
@alpencolt @gbalykov @jkotas PTAL

jkotas:
Note that we are switching over to shared cross-compilation scripts. The master copy of these scripts lives in https://github.com/dotnet/arcade/tree/master/eng/common/cross and the scripts are mirrored to all repos. We expect that the cross directory to be deleted at some point.

Could you please submit any Tizen fixes also to https://github.com/dotnet/arcade/tree/master/eng/common/cross if they are not there already?

kbaladurin:
Thank you! I've submitted PR with these changes to arcade project: https://github.com/dotnet/arcade/pull/2969

Issue #4149, opened 2018-05-14
"Failed to initialize CoreCLR, HRESULT: 0x80004005" when $TMPDIR is not writeable/nonexistent


jacobrillema:
## Steps to reproduce
```bash
sudo apt-get install dotnet-sdk-2.1.300-rc1-008673
dotnet --version

sudo dotnet --version
```
## Expected  behavior
```bash
dotnet --version
2.1.300-rc1-008673

sudo dotnet --version
2.1.300-rc1-008673
```

## Actual behavior
```bash
dotnet --version
Failed to initialize CoreCLR, HRESULT: 0x80004005

sudo dotnet --version
2.1.300-rc1-008673
```

## Environment data
```bash
dotnet --info
Failed to initialize CoreCLR, HRESULT: 0x80004005

Host (useful for support):
  Version: 2.1.0-rc1
  Commit:  eb9bc92051

.NET Core SDKs installed:
  2.1.300-rc1-008673 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.0-rc1-final [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.0-rc1-final [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.0-rc1 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download

lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.4 LTS
Release:        16.04
Codename:       xenial

uname -a
Linux BSS-IMM 4.4.0-17134-Microsoft #48-Microsoft Fri Apr 27 18:06:00 PST 2018 x86_64 x86_64 x86_64 GNU/Linux
```

danmosemsft:
So it works fine under sudo? That's interesting. If you use `strace` is it possible to see what we are failing to access?

jacobrillema:
I just tried running 2.1.200 SDK install and it works fine without sudo. I ran the strace again on the RC candidate and here is the area I think it is failing on.

```bash
591 14020 unlink("/mnt/c/Users/jrillema/AppData/Local/Temp/clr-debug-pipe-14020-209726-in") = -1 ENOENT (No such file or directoy>
[...]

PR #6692, opened 2019-06-05
Validate if ${TMPDIR} is pointing to a directory in Unix systems

lpereira:
While not an actual fix for #4149, this adds a little bit of robustness when trying to determine the path to a temporary directory.

(is_directory() isn't called for /var/tmp and /tmp because the full path qualifier, including the trailing slash, are also passed to realpath(), which should take care of this check.)

### DIFF
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -265,12 +265,33 @@ bool pal::get_default_servicing_directory(string_t* recv)
     return true;
 }
 
+static bool is_directory(const pal::string_t& path)
+{
+    if (path.empty())
+    {
+        return false;
+    }
+
+    struct stat buffer;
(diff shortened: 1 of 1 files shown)


dagood:
@janvorli @lpereira @vitek-karas Can some review this and get it merged?

janvorli (approved):
LGTM, thank you!

ViktorHofer:
What's the status here?

ViktorHofer:
OK just saw that this issue has the post-consolidation label. Sorry for the noise.

maryamariyan:
Thank you for your contribution. As announced in #27549 the dotnet/runtime repository will be used going forward for changes to this code base. Closing this PR as no more changes will be accepted into master for this repository. If you’d like to continue working on this change please move it to dotnet/runtime.

Issue #6607, opened 2019-05-31
Add automatic verification that all WindowsDesktop assemblies are assigned a Profile


dagood:
The initial implementation of `Profile` classification is loose, so new files, renames, and deletes coming from the WinForms and WPF repos will silently break the existing `Profile`s.

In the short term (this issue), we should enforce that each assembly we get has a classification.

In the long term, this info should not be stored in Core-Setup. It should flow from the original repos in some way so that changes don't have to be manually fixed up in Core-Setup after the fact.

---

Background on `Profile`: https://github.com/dotnet/core-setup/issues/6210.
https://github.com/dotnet/core-setup/pull/6573 implemented classification.

Issue #6210, opened 2019-05-01
Add "Profile" classification to WindowsDesktop FrameworkList.xml


dagood:
The files in WindowsDesktop are for WPF and WinForms (and integration), but at design time, seeing all of these in intellisense causes a poor experience. See https://github.com/dotnet/cli/issues/10536 for more details.

To help solve this, the WD targeting pack `FrameworkList.xml` should classify each DLL so they can be filtered. Add a `Profiles` attribute, see https://github.com/dotnet/cli/issues/10536#issuecomment-488871926 for values.

The list makes more sense to maintain in Core-Setup because it's where the packages are combined, rather than listing all the files somewhere deeper in the SDK.

@ericstj did a pass of manual classification and got these results:

WindowsDesktop File | Classification
-- | --
Accessibility.dll |  
DirectWriteForwarder.dll | WPF
Microsoft.Win32.Registry.dll |  
Microsoft.Win32.SystemEvents.dll |  
PenImc_cor3.dll | WPF
PresentationCore-CommonResources.dll | WPF
PresentationCore.dll | WPF
PresentationFramework-SystemCore.dll | WPF
PresentationFramework-SystemData.dll | WPF
PresentationFramework-SystemDrawing.dll | WPF
PresentationFramework-SystemXml.dll | WPF
PresentationFramework-SystemXmlLinq.dll | WPF
PresentationFramework.Aero.dll | WPF
PresentationFramework.Aero2.dll | WPF
PresentationFramework.AeroLite.dll | WPF
PresentationFramework.Classic.dll | WPF
PresentationFramework.dll | WPF
PresentationFramework.Luna.dll | WPF
PresentationFramework.Royale.dll | WPF
PresentationNative_cor3.dll | WPF
PresentationUI.dll | WPF
ReachFramework.dll | WPF
System.CodeDom.dll |  
System.Configuration.ConfigurationManager.dll |  
System.Design.dll | WindowsForms
System.Diagnostics.EventLog.dll |  
System.DirectoryServices.dll |  
System.Drawing.Common.dll | WindowsForms
System.Drawing.Design.dll | WindowsForms
System.Drawing.Design.Primitives.dll | WindowsForms
System.IO.FileSystem.AccessControl.dll |  
System.Media.SoundPlayer.dll |  
System.Printing.dll | WPF
System.Security.AccessControl.dll |  
System.Security.Cryptography.Cng.dll |  
System.Security.Cryptography.Pkcs.dll |  
[...]

PR #6698, opened 2019-06-05
Require 1:1 WindowsDesktop Profile classification

dagood:
Change `FrameworkListFileProfile` from a loose mapping to a 1:1 mapping. This means any mismatches will break upgrade PRs rather than (potentially) going unnoticed. Resolves https://github.com/dotnet/core-setup/issues/6607.

A decent number of the classifications weren't needed. Some files probably went away since the table in https://github.com/dotnet/core-setup/issues/6210 was made, and resource DLLs and native files aren't in the targeting pack.

`<FrameworkListFileProfile Include="WindowsFormsIntegration.dll" />` means that this file is known to correctly have no `Profile` (because it's for integration).

/cc @vatsan-madhavan, @rladuca, @dsplaisted, @nguerrera

### DIFF
--- a/src/pkg/projects/windowsdesktop/pkg/dir.props
+++ b/src/pkg/projects/windowsdesktop/pkg/dir.props
@@ -23,12 +23,8 @@
 
   <ItemGroup Condition="'$(PackageTargetRuntime)' == ''">
     <FrameworkListFileProfile Include="Accessibility.dll" Profile="WindowsForms;WPF" />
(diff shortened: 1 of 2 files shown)


dagood:
I'm going ahead with the merge because it's a simple change, and if it gets stale, there's a risk that WinForms or WPF will change a bit and conflict with this PR causing official build breaks when I do eventually merge it.

PR #6760, opened 2019-06-10
Improvements to the component dependency resolution tests

vitek-karas:
* Move these under the DependencyResolution (since they belong there)
* Refactor to avoid code duplication
* Change it to not use published real builds of components, but instead simply create mocks on disk (the tests never actually run the components)
* Change where the tests put files on disk
* Reduce file copying for tests which don't need to modify the prepared components

This change does NOT modify what the test actually do. The tested functionality and how it's tested has remained the same.

I'm planning to add new tests which rely on the ability to carefully modify the `.deps.json` of the tested component (and the shared framework it resolves against). Without the move to the mocks this would be rather hard to achieve. With the mocks, the new tests can now use the new `NetCoreAppBuilder` to craft the special `.deps.json` easily without breaking the rest of the component.


vitek-karas:
Since I moved the file github doesn't even show the diff. That said the diff would be all green/red anyway so it's probably better to just look at the new version of the tests if the style makes sense. As noted above the functionality remained the same.

PR #6781, opened 2019-06-11
Component dependency resolution aware of frameworks

vitek-karas:
Makes component dependency resolution (that is `AssemblyDependencyResolver`) aware of frameworks.
From a functionality perspective this means that components will correctly override assemblies in the frameworks. So if the component contains an assembly A which is also in framework, A will be resolved either from the component or the framework based on version match.
* If the component assembly version is higher the one from the component will be used
* If the one in framework is higher or equal version, it will be used.

This matches the behavior when resolving assemblies in an app during startup.

Note that this doesn't mean that `AssemblyDependencyResolver` will resolve framework assemblies all up. Right now it is intentionally restricted to only resolve assemblies which are references by the component itself. That said this change will make it relatively easy to enable framework assembly resolution as well if we decide to do so.

Changes made:
* Store the list of frameworks on the `hostpolicy_init` object. The stored list is `const` since we need to be able to access it from multiple threads at the same time. The list is only populated once the frameworks are processed during the first context init. After this, the frameworks should not change at all.
* Regarding memory management of frameworks nothing has changed, the `hostpolicy_init` already stored the frameworks, but as a non-const vector, the change is to simply add a const version of the vector - this makes it "Easy" to prove that accesses to this are thread safe.
* Add a new constructor for `deps_resolver` which creates it from already processed frameworks (those from the new const list). In this case there's no need for `.deps.json` parsing and other processing, so the new constructor simply initializes the resolver.
* Change the component dependency resolution entrypoint to construct a list containing the component and all the frameworks (except the app) and pass that list to the resolver (so that it has all the frameworks).
* Modify resolver to be able to limit resolved assemblies to a specific level - currently it's used as either "resolve everything" (when running app), or "resolve just the app" (during component resolution - the component act's an app).
* Some very minor cleanup. Note that there are lot of opportunities to improve the codebase in the dependency resolver, this change doesn't attempt to make any of those.

Test changes:
* Refactor existing component dependency resolution tests to make it easier to reuse.
* Implement new tests which validate the assembly resolution between the app/component and the frameworks, specifically with regard to version resolution. The tests are written for both component and app scenarios (the behaviors should be the same).

### DIFF
--- a/src/corehost/cli/fx_definition.h
+++ b/src/corehost/cli/fx_definition.h
@@ -43,13 +43,24 @@ private:
 };
 
 typedef std::vector<std::unique_ptr<fx_definition_t>> fx_definition_vector_t;
+typedef std::vector<const fx_definition_t*> fx_definition_const_vector_t;
 
 static const fx_definition_t& get_root_framework(const fx_definition_vector_t& fx_definitions)
 {
     return *fx_definitions[fx_definitions.size() - 1];
 }
 
+static const fx_definition_t& get_root_framework(const fx_definition_const_vector_t& fx_definitions)
+{
+    return *fx_definitions[fx_definitions.size() - 1];
+}
+
 static const fx_definition_t& get_app(const fx_definition_vector_t& fx_definitions)
+{
+    return *fx_definitions[0];
+}
+
+static const fx_definition_t& get_app(const fx_definition_const_vector_t& fx_definitions)
 {
     return *fx_definitions[0];
 }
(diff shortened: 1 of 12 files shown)


vitek-karas:
I also wrote a stress test to validate the multi-threading somewhat:
* 20 threads - each 1000 iterations
* Each constructing a new `AssemblyDependencyResolver` over a component which does need framework resolution
* Validate that it resolves the assembly from the framework

It ran for several minutes without any issues and successfully finished.

lpereira:
> I also wrote a stress test to validate the multi-threading somewhat:
> 

To test multi-threading stuff, I like running things under ThreadSanitizer.  It's able to detect most data races.

vitek-karas:
Abandoning this as per discussion in https://github.com/dotnet/core-setup/issues/6826. In short, there's no way to make this work for self-contained apps right now, and we'd rather have consistent behavior across all app types.

The tests from this change are still valid (although with modified expected outcomes) and they are part of https://github.com/dotnet/core-setup/pull/6919.

PR #6799, opened 2019-06-12
Bundler: Check for duplicate file entries.

swaroop-sridhar:
If the input to the bundler contains multiple entries
with the same relative path, we currently generate a bundle
with a larger size with some unusable files.

Therefore, check this condition when generating bundles, and
throw an error.

The extractor also checks this condition when extracting
malformed bundles.


swaroop-sridhar:
@nguerrera, I changed the code to use the Linq grouping API. Thanks.

PR #6814, opened 2019-06-13
Small improvements to Unix PAL

lpereira:
This PR makes some small changes to the Unix version of the PAL:

- Uses `fstatat(2)` instead of building a temporary `std::string` while reading directories
- Uses `access(2)` to determine the existence of a file/directory (no need to validate/fill in a `struct stat`)
- Construct the `pal::string_t` entry in `pal::readdir()` in place using `emplace_back()`

### DIFF
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -811,11 +811,12 @@ static void readdir(const pal::string_t& path, const pal::string_t& pattern, boo
                 continue;
             }
 
-            pal::string_t filepath(entry->d_name);
-            if (filepath != _X(".") && filepath != _X(".."))
+            if (!strcmp(entry->d_name, ".") || !strcmp(entry->d_name, ".."))
             {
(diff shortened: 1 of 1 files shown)


src/corehost/common/pal.unix.cpp
```diff
@@ -816,11 +806,12 @@ static void readdir(const pal::string_t& path, const pal::string_t& pattern, boo
                 continue;
             }
 
-            pal::string_t filepath(entry->d_name);
-            if (filepath != _X(".") && filepath != _X(".."))
+            if (!strcmp(entry->d_name, _X(".")) || !strcmp(entry->d_name, _X("..")))
```
janvorli: A nit - we can remove the _X(...) wrappers here. We know that the strcmp needs plain 8 bit char strings

lpereira: Done. Force-pushed with the changes, and rebased to current HEAD.

PR #6831, opened 2019-06-16
Add AppHost Binary transformation Utilities to HostModel

swaroop-sridhar:
The code that transforms AppHost binary in order to update the
path to the App.dll binary lives in multiple repositories
(SDK, CLI and partly in core-setup tests).

This change adds the code to HostModel library, so that other
copies can be removed.

This also facilitates implementing a similar mechanism to
locate the offset of the bundle header within the binary
(so that the bundle headers are not required to live at the end
of the file).


src/managed/Microsoft.NET.HostModel/AppHost/AppUpdater.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            if (intermediateAssembly != null && appHostIsPEImage)
+            {
+                if (ResourceUpdater.IsSupportedOS())
+                {
+                    // Copy resources from managed dll to the apphost
+                    new ResourceUpdater(appHostDestinationFilePath)
+                        .AddResourcesFromPEImage(intermediateAssembly)
+                        .Update();
```
wli3: make it a throw instead of logging

src/managed/Microsoft.NET.HostModel/AppHost/AppUpdater.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <param name="appHostDestinationFilePath">The destination path for desired location to place, including the file name</param>
+        /// <param name="appBinaryFilePath">Full path to app binary or relative path to the result apphost file</param>
+        /// <param name="windowsGraphicalUserInterface">Specify whether to set the subsystem to GUI. Only valid for PE apphosts.</param>
+        /// <param name="intermediateAssembly">Path to the intermediate assembly, used for copying resources to PE apphosts.</param>
+        public static void UpdateAppPath(
+            string appHostSourceFilePath,
+            string appHostDestinationFilePath,
+            string appBinaryFilePath,
+            bool windowsGraphicalUserInterface = false,
+            string intermediateAssembly = null)
```
wli3: a better name than intermediateAssembly

src/managed/Microsoft.NET.HostModel/AppHost/AppUpdater.cs
```diff
@@ -0,0 +1,133 @@
...
+                        throw new BinaryUpdateException($"Unable to use '{appHostSourceFilePath}' as application host executable as it does not contain the expected placeholder byte sequence '{AppBinaryPathPlaceholder}' that would mark where the application name would be written");
+                    }
+
+                    appHostIsPEImage = BinaryUtils.IsPEImage(accessor);
+
+                    if (windowsGraphicalUserInterface)
+                    {
+                        if (!appHostIsPEImage)
+                        {
+                            throw new BinaryUpdateException($"Unable to use '{appHostSourceFilePath}' as application host executable because it's not a Windows executable for the CUI (Console) subsystem");
```
wli3: throw them in distinguished error types

src/managed/Microsoft.NET.HostModel/AppHost/ResourceUpdater.cs
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        private static void ThrowExceptionForLastWin32Error()
+        {
+            throw new HResultException(Marshal.GetHRForLastWin32Error());
+        }
+
+        private static void ThrowExceptionForInvalidUpdate()
+        {
+            throw new InvalidOperationException("Update handle is invalid. This instance may not be used for further updates");
```
wli3: this type of developer exception is ok to be just in english

jkoritzinsky:
Do we want to use the new non-OS-bound PE resource APIs that @davidwrighton did for xplat IBCMerge or do we want to wait on that until a future PR/post-3.0?

swaroop-sridhar:
@wli3, thanks for the comments, I've made the changes now.

src/managed/Microsoft.NET.HostModel/AppHost/AppNameTooLongException.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+
+namespace Microsoft.NET.HostModel.AppHost
+{
+    /// <summary>
+    /// Given app file name is longer than 1024 bytes
+    /// </summary>
+    public class AppNameTooLongException : AppHostUpdateException
+    {
+        public string LongName;
```
wli3: nit: maybe just expose {get;}

wli3:
:+1:

swaroop-sridhar:
> Do we want to use the new non-OS-bound PE resource APIs that @davidwrighton did for xplat IBCMerge or do we want to wait on that until a future PR/post-3.0?

I think we should bring the code over as-is. If the resource-updater APIs can be made public, then we can change it.

src/managed/Microsoft.NET.HostModel/AppHost/AppNameTooLongException.cs
swaroop-sridhar: Fixed it in this commit: https://github.com/dotnet/core-setup/pull/6831/commits/ff295f51bdbdaf941a053279ed7bd7f470f5529f

swaroop-sridhar:
@dotnet-bot  test dotnet.core-setup (Build_OSX release) 
@dotnet-bot  test dotnet.core-setup (Build_Linux_x64_glibc debug)

nguerrera:
> Do we want to use the new non-OS-bound PE resource APIs

Ooh. Yes, please!

> I think we should bring the code over as-is. If the resource-updater APIs can be made public, then we can change it.

OK.

davidwrighton:
Yes, I would continue this PR without pulling in the non OS bound PE resource apis. The work that I did around the non OS bound PE resource apis should be just about drop in compatible with the OS api surface, so it should be easy, but its a large pile of really terrible code which could be destabilizing, and would certainly distract from the core thrust of this PR. If you're interested in doing that, it should be a separate PR.

PR #6866, opened 2019-06-19
Fix infinite loop in traversing thunk chunks when checking if thunks are installed.

jkoritzinsky:
We weren't actually traversing the linked list of `bootstrap_thunk_chunk`s in `are_thunks_installed_for_module`, so if there were multiple IJW assemblies loaded into memory, we'd get an infinite loop when trying to load the second one.

This fixes the infinite loop.

### DIFF
--- a/src/corehost/cli/ijwhost/ijwthunk.cpp
+++ b/src/corehost/cli/ijwhost/ijwthunk.cpp
@@ -183,6 +183,7 @@ bool are_thunks_installed_for_module(pal::dll_t instance)
         {
             return true;
         }
+        currentChunk = currentChunk->GetNext();
     }
 
     return false;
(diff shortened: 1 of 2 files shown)


src/corehost/cli/ijwhost/ijwthunk.cpp
```diff
@@ -183,6 +183,7 @@ bool are_thunks_installed_for_module(pal::dll_t instance)
         {
             return true;
         }
+        currentChunk = currentChunk->GetNext();
```
AaronRobinsonMSFT: This is a bit unsettling. How was this missed during development? Was this path never exercised?

jkoritzinsky: This path is only hit when loading 2 or more IJW modules into the same image. I had only tested loading one IJW module into a DLL at a time since the only way I could get the initial testing at the time was with manually hex-editing the binary to refer to the new host instead of the old host.

jkoritzinsky:
This code is not compiled on Linux_x64_Rhel6, so it's definitely not the cause of the test timeout. I'm going to merge this in.

PR #6884, opened 2019-06-20
Do not throw during startup if not running a bundled executable

### LINKED ISSUE #6883: apphost.exe throws internally C++ exception on every run
Using Preview6 create a new console app (`dotnet new console`) and build it (`dotnet build`). Then run the produced `.exe` under a debugger letting stop on all C++ first chance exceptions. The exception is thrown from the bundle parser:
```
>	ca2.exe!_CxxThrowException(void * pExceptionObject, const _s__ThrowInfo * pThrowInfo) Line 133	C++
 	ca2.exe!bundle::manifest_footer_t::read(struct _iobuf *)	C++
 	ca2.exe!bundle::bundle_runner_t::extract(void)	C++
 	ca2.exe!exe_start(int,wchar_t const * * const)	C++
 	ca2.exe!wmain()	C++
```

Since this will happen on startup of every application which is not a single-file bundle, it affects perf of pretty much everything. We need to make the initial check for "bundle or not" super efficient so that we don't pay for the fact that we use the same apphost for both normal apps and single-file apps.

lpereira:
This fixes #6883.  This requires some further love and dedication, but will do for now so we can stabilize 3.0 further.

Also plugs a memory leak while reading the footer by allocating the space for the footer in the stack rather than from the heap.

### DIFF
--- a/src/corehost/cli/apphost/bundle/bundle_runner.cpp
+++ b/src/corehost/cli/apphost/bundle/bundle_runner.cpp
@@ -9,33 +9,63 @@
 
 using namespace bundle;
 
-void bundle_runner_t::seek(FILE* stream, long offset, int origin)
+StatusCode bundle_runner_t::try_seek(FILE* stream, long offset, int origin)
 {
     if (fseek(stream, offset, origin) != 0)
     {
         trace::error(_X("Failure processing application bundle; possible file corruption."));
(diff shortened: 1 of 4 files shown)


lpereira:
Closing this PR after @swaroop-sridhar's comments.

PR #6885, opened 2019-06-21
Enable singing of single-file bundles

swaroop-sridhar:
This change removes the requirement that the bundle-signature
be found at the end of the file, so that tools like signtool
can append their custom information at the end of the file.

The layout of .net core bundles was

AppHost Code
Bundled Files
Manifest header
Manifest entries
Bundle footer
<EOF>

The bundle-footer is now removed, its functionality is served by a
special global in apphost (Bundle-Placeholder) similar to the way
APP-DLL name is encoded in the AppHost.

The normal unbundled apphost has a default value for the bundle-placeholder.
If the apphost is a single-file bundle, the bundler tool (in HostModel) updates
the value of the variable on disk to a different value, and encodes the location
of the manifest header in it.

This way, there is no requirement that the bundle meta-data must remain at
the end of the file.

This change also simplifies the test for whether an apphost is a bundle:
we only check a value in the loaded binary, rather than performing file reads.


src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
```diff
@@ -88,37 +85,66 @@ public static class HostWriter
...
+                    int position = BinaryUtils.SearchInFile(accessor, bundleSignature);
+                    if(position == -1)
+                    {
+                        bundleHeaderOffset = 0;
+                        return false;
+                    }
 
-            // Copy apphost to destination path so it inherits the same attributes/permissions.
-            File.Copy(appHostSourceFilePath, appHostDestinationFilePath, overwrite: true);
+                    bundleHeaderOffset = accessor.ReadInt64(position - sizeof(Int64));
```
AustinWise: Perhaps this method should if `bundleHeaderOffset` is zero, like `marker_t::is_bundle()` does.

src/corehost/cli/apphost/bundle/bundle_runner.cpp
```diff
@@ -166,28 +167,9 @@ void bundle_runner_t::reopen_host_for_reading()
...
-    header_offset = footer->manifest_header_offset();
-    return true;
-}
-
-void bundle_runner_t::process_manifest_header(int64_t header_offset)
-{
-    seek(m_bundle_stream, header_offset, SEEK_SET);
+    seek(m_bundle_stream, marker_t::header_offset(), SEEK_SET);
 
     manifest_header_t* header = manifest_header_t::read(m_bundle_stream);
```
lpereira: Since `header` is not stored anywhere, and it's allocated in `manifest_header_t`, the memory leaks.  It might be a better idea to allocate on the stack here and have `manifest_header_t::read()` take a point to it as a parameter.

lpereira (changes requested):

src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
swaroop-sridhar: @AustinWise, change is in this commit: https://github.com/dotnet/core-setup/pull/6885/commits/75f19262c4d87f3ed75e1b08e2e7c26fd304cd23

src/corehost/cli/apphost/bundle/bundle_runner.cpp
swaroop-sridhar: @lpereira  please see this commit
https://github.com/dotnet/core-setup/pull/6885/commits/bd597daecedfba1995832d66d720b37d35b08c39

src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
AustinWise: Thanks, that should make for a more clear error message.

vitek-karas (approved):
LGTM

src/corehost/cli/apphost/bundle/bundle_runner.cpp
```diff
@@ -302,22 +273,11 @@ StatusCode bundle_runner_t::extract()
...
-
-        int64_t manifest_header_offset;
-
-        if (!process_manifest_footer(manifest_header_offset))
-        {
-            return StatusCode::AppHostExeNotBundle;
-        }
-        process_manifest_header(manifest_header_offset);
+        // Read the bundle header
+        seek(m_bundle_stream, marker_t::header_offset(), SEEK_SET);
```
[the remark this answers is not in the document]
lpereira: It's often a good idea to replace `seek()`+`read()` with `pread()` if supported by the system.  This way, if reading with an invalid offset, the `pread()` call will fail.

src/corehost/cli/apphost/bundle/header.h
```diff
@@ -0,0 +1,51 @@
...
+        {
+            uint32_t major_version;
+            uint32_t minor_version;
+            int32_t num_embedded_files;
+            int8_t bundle_id_length_byte_1;
+        } m_data;
+#pragma pack(pop)
+        pal::string_t m_bundle_id;
+
+        static const uint32_t m_current_major_version = 0;
```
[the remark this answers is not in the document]
lpereira: If it's static, should it be prefixed with `m_`?

src/corehost/cli/apphost/bundle/bundle_runner.h
```diff
@@ -50,10 +52,9 @@ namespace bundle
         void extract_file(file_entry_t* entry);
 
         FILE* m_bundle_stream;
+        header_t* m_header;
```
lpereira: What is destroying `m_header`?  The memory leak is still here.  A few options here, in order from the most preferred to the least preferred:

- Declare this inline as `header_t m_header`. No need to allocate a new struct from the heap. Goes away as soon as `bundle_runner_t` goes away.
- Make this a `std::unique_ptr<header_t>` instead.
- Have a `delete m_header` in `bundle_runner_t::~bundle_runner_t()`.

lpereira (changes requested):

src/corehost/cli/apphost/bundle/header.h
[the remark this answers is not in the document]
swaroop-sridhar: No private statics should be `s_`, but this field can just be a const; so I'll remove the annotations.

src/corehost/cli/apphost/bundle/header.cpp
```diff
@@ -0,0 +1,43 @@
...
+    {
+        trace::error(_X("Failure processing application bundle."));
+        trace::error(_X("Bundle header version compatibility check failed"));
+
+        throw StatusCode::BundleExtractionFailure;
+    }
+
+    // bundle_id is a component of the extraction path
+    size_t bundle_id_length = 
+        bundle_runner_t::get_path_length(header->m_data.bundle_id_length_byte_1, stream);
```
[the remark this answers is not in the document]
swaroop-sridhar: `bundle_id` is actually a part of a path (that is, it should be path-compatible, and conform to length limitation). The `get_path_length` method only reads two bytes in a length encoding (because of `MAX_PATH`) restriction. So, that's why I had called it `get_path_length` instead of `get_string_length` 

I can rewrite the function more generically, call it `read_string_length` and do the `MAX_PATH` check elsewhere -- but it doesn't serve much purpose other than the symmetry with `read_string()` function.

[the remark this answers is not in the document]
vitek-karas: OK - it was just a nit anyway... :-)

src/corehost/cli/apphost/bundle/bundle_runner.cpp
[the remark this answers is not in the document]
swaroop-sridhar: @vitek-karas  the `seek()` function above already checks for failure, and throws the bundleIOerror exception.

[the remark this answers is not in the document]
swaroop-sridhar: @lpereira  I agree that the PAL can be improved by adding `pread()`. 
But I think this code will soon change to use memory-mapping of the whole bundle file instead of seek-read-write, etc. So, I left it as-is for now.

src/corehost/cli/apphost/bundle/bundle_runner.h
swaroop-sridhar: I added a destructor to the `bundle_runner_t`.
I did not inline the manifest, because when files are directly processed from the bundle, the manifest will survive beyond the `bundle_runner`  object.
I also wanted to keep the interface of `header_t` similar to `manifest_t` and the managed code in `HostModel`. Thanks.

swaroop-sridhar:
The win-x64 debug failure says `System.IO.IOException : There is not enough space on the disk : 'D:\a\1\s\bin\tests\win-x64.Debug\8\StandaloneApp\publish\StandaloneApp.exe'` 
The error happens when the fixture is cloned for use, likely to be a testing issue.

dagood:
I hit the `There is not enough space on the disk` errors in my work to migrate to Arcade, too, and moving off the tiny hosted agents onto the Helix queues worked for me: https://github.com/dotnet/core-setup/commit/cbe76ff2c47e20accd927efa65abcf3cdf514cf3. IMO feel free to do so here.

swaroop-sridhar:
Thanks @dagood ; I'll try that (in a separate change).

swaroop-sridhar:
The Build_Windows_x64 debug leg passed with the test disabled.
A few more legs are waiting for machines to run on, but have previously passed. 
https://dev.azure.com/dnceng/public/_build/results?buildId=239197 

So, I'll merge the change, so that the fix is available in the upcoming preview.

src/corehost/cli/apphost/bundle/marker.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "pal.h"
+#include "trace.h"
+#include "utils.h"
+
+using namespace bundle;
+
+int64_t marker_t::header_offset()
+{
+    // Contains the bundle_placeholder default value at compile time.
+    // If this is a single-file bundle, the last 8 bytes are replaced
```
nxtn, line 15: I believe it is the **first** 8 bytes that are replaced?

vitek-karas: Good catch - thanks a lot!
https://github.com/dotnet/runtime/pull/44874

Issue #5045, opened 2019-01-25
Should M.E.DependencyModel only target netstandard2.0+ going forward?


ahsonkhan:
M.E.DependencyModel currently supports several TFMs which means we have to maintain the use of two JSON APIs (System.Text.Json for netstandard2.0+ and Newtonsoft.Json for < netstandard2.0).

https://github.com/dotnet/core-setup/blob/63a01b08e5d1d1a6b8544f598b3d3bda76e6e424/src/managed/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.csproj#L6

We should consider only targeting netstandard2.0 for the upcoming major release and simplify our dependencies.

VS requires 4.7.2 which is compatible with netstandard2.0.

This effort goes in conjunction with https://github.com/dotnet/announcements/issues/90 and helps remove the dependency on Json.NET more cleanly: https://github.com/dotnet/core-setup/issues/3311

Related PRs:
- https://github.com/dotnet/core-setup/pull/5009
- https://github.com/dotnet/core-setup/pull/5010

cc @joshfree, @glennc, @eerhardt, @terrajobst, @jaredpar, @ericstj, @davidfowl, @livarcocc

jaredpar:
In case it's relevant Roslyn decided to move to netstandard20 only dropping our netstandard1.3 platforms in the NuPkg. This was a result of VS moving to require net472 hence making netstandard20 an option. The benefits of targeting it were high enough we felt it justified the dropping support for netstandard1.3.

terrajobst:
> The benefits of targeting it were high enough we felt it justified the dropping support for netstandard1.3.

That sounds a like a non-sequitur to me. You can offer .NET Standard 2.0 without dropping 1.x. Presumably your unstated position is that you don't want the support burden of doing both, so if you had to pick only one, you felt justified with choosing 2.0?

terrajobst:
> Correct. The burden of maintaining netstandard1.1 is quite high in our setup.

That's fair.

eerhardt:
[...]

PR #6937, opened 2019-06-26
Switch from source to binary package for S.Text.Json

ericstj:
For now this is just a port without removing the abstraction as suggested in https://github.com/dotnet/core-setup/issues/5045.

I'll make a follow up PR that addresses https://github.com/dotnet/core-setup/issues/5045 but I wanted to make sure this one gets in.

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
@@ -96,21 +96,7 @@ namespace Microsoft.Extensions.DependencyModel
             return items.ToArray();
         }
 
-        public void Skip()
-        {
-            if (IsTokenTypeProperty())
-            {
-                _reader.Read();
-            }
-
-            if (_reader.TokenType == JsonTokenType.StartObject || _reader.TokenType == JsonTokenType.StartArray)
-            {
-                int depth = _reader.CurrentDepth;
-                while (_reader.Read() && depth <= _reader.CurrentDepth)
-                {
-                }
-            }
-        }
+        public void Skip() => _reader.Skip();
 
         public string ReadAsString()
         {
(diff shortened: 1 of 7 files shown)


ericstj:
Looks like I need to fix some tests too.  Looking.

ericstj:
ok, I think this should be ready to go now.

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
```diff
@@ -160,8 +146,13 @@ public bool ReadAsBoolean(bool defaultValue)
 
         private static Exception CreateUnexpectedException(ref Utf8JsonReader reader, string expected)
         {
+            // Replace with public API once https://github.com/dotnet/corefx/issues/34768 is fixed
+            object boxedState = reader.CurrentState;
+            long lineNumber = (long)(typeof(JsonReaderState).GetField("_lineNumber")?.GetValue(boxedState) ?? -1);
+            long bytePositionInLine = (long)(typeof(JsonReaderState).GetField("_bytePositionInLine")?.GetValue(boxedState) ?? -1);
```
dagood: I'm assuming this will go away naturally when the abstraction is removed for https://github.com/dotnet/core-setup/issues/5045, otherwise maybe we should track it separately.

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
```diff
@@ -96,21 +96,7 @@ public string[] ReadStringArray()
...
-
-            if (_reader.TokenType == JsonTokenType.StartObject || _reader.TokenType == JsonTokenType.StartArray)
-            {
-                int depth = _reader.CurrentDepth;
-                while (_reader.Read() && depth <= _reader.CurrentDepth)
-                {
-                }
-            }
-        }
+        public void Skip() => _reader.Skip();
```
joperezr: This is more of a general question, but why do we want to have our own copy of a JsonReader and Writer? If what we want is to actually get the document back, should we use JsonDocument.Parse and let that call the reader?

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
ericstj: It won't go away with #5045, it'll need an issue to track removal.  I'll file one.

ericstj:
Merged despite failing linux legs.

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
ericstj: This is lower level code that cares about perf and wants to directly use the reader/writer.

PR #6943, opened 2019-06-26
Allow ijwhost to work without requiring a runtimeconfig.json for dll being loaded

batzen:
This PR will allow ijwhost to work without requiring a runtimeconfig.json for dll being loaded.

This is made possible by adding the runtime_config_file to the properties of hostpolicy_init and reusing this in ijwhost if the calculated config file does not exist.
Due to this we can just reuse an existing loaded runtime.
I also thought that adding the runtimeconfig.json file to the properties makes sense as the deps file is also already added.

The use case for this are, at least but not limited to, tools like https://github.com/snoopwpf/snoopwpf which inject themselves into foreign processes.
Please note that the current version of snoop uses c++/cli and is not compatible with .net core 3.0 and that this PR tries to pave to way to get away from c++/cli in that project as it's way easier to have pure managed code (plus some tool to produce the modified assemblies exporting functions) than managing c++/cli projects.
You can see how it works with "pure" managed code, but currently only working with the full .net framework instead of core, in https://github.com/batzen/snoopwpf/tree/NETCore3.

This PR is related to #6100 (which got moved to https://github.com/dotnet/coreclr/issues/24318).

Notes:
For this to fully work out of the box the DLLMain/ExeMain of the library being loaded has to point to ijwhost.dll instead of mscoree.dll.
I did this by using a modified version of dnlib as i need those changes there anyway to create a fully managed dll with exported function.
You can find the PR at https://github.com/0xd4d/dnlib/pull/281.

I also prepared a demo project creating such dlls at https://github.com/batzen/NativeToManagedTest. 
To test this out you have to:
- clone the NativeToManagedTest repository
- build it (via dotnet build)
- checkout this PR
- build it 
- copy hostfxr.dll and ijwhost.dll to the "NativeToManagedTest\bin\Debug\x64\netcoreapp3.0" directory 
- copy hostpolicy.dll (make a backup of the original one before this step) to the dotnet shared framework directory being referenced in the NativeToManagedTest projects
- run WPFTestApp.exe
- click the button labeled "Call managed native"
- two message boxes should be shown saying "Hi from method ..."

Questions:
- I had a hard time figuring out how to use my self compiled hostpolicy.dll instead of the one from the shared framework. Do you have any hints on how to achieve this easily? Placing hostfxr.dll beside my .exe was enough for that one to be picked up.
- How and where should i write tests for this? The only tests for ijwhost i could find are in coreclr and not in core-setup.

Also adding @jeffschwMSFT @vitek-karas @AaronRobinsonMSFT as requested.

### DIFF
--- a/src/corehost/cli/ijwhost/ijwhost.cpp
+++ b/src/corehost/cli/ijwhost/ijwhost.cpp
@@ -39,6 +39,44 @@ pal::hresult_t get_load_in_memory_assembly_delegate(pal::dll_t handle, load_in_m
 
             *config_path_out = std::move(config_path_local);
 
+            // if file does not exist, try to get the one used for an existing context (if there is any)
+            if (!pal::realpath(&config_path_local))
+            {
+                trace::info(_X("Failed to resolve .runtimeconfig.json path of the current mixed-mode module [%s]"), mod_path.c_str());
+                trace::info(_X("Trying to get .runtimeconfig.json from existing context"));
+
+                pal::dll_t out_fxr;
+                pal::string_t out_fxr_path;
+
+                if (pal::get_loaded_library(LIBFXR_NAME, "hostfxr_get_runtime_property_value", &out_fxr, &out_fxr_path))
+                {
+                    hostfxr_get_runtime_property_value_fn get_runtime_property_value = reinterpret_cast<hostfxr_get_runtime_property_value_fn>(pal::get_symbol(out_fxr, "hostfxr_get_runtime_property_value"));
+
+                    const pal::char_t *value;
+                    int rc = get_runtime_property_value(nullptr, _X("APP_CONTEXT_RUNTIME_CONFIG_FILE"), &value);
+
+                    if (rc == StatusCode::Success)
+                    {
+                        pal::string_t config_path_from_context = value;
+                        *config_path_out = std::move(config_path_from_context);
+
+                        trace::info(_X("Got .runtimeconfig.json [%s] from existing context"), config_path_out->c_str());
+                        return StatusCode::Success;
+                    }
+                    else
+                    {
+                        trace::error(_X("Getting APP_CONTEXT_RUNTIME_CONFIG_FILE from existing context failed with [%i]"), rc);
+                    }
+                }
+                else
+                {
+                    trace::error(_X("Could not get .runtimeconfig.json from existing context as [%s] was not loaded"), LIBFXR_NAME);
+                }
+
+                trace::error(_X("Failed to resolve .runtimeconfig.json path of the current mixed-mode module [%s] and also failed to get it from an existing context"), mod_path.c_str());
+                return StatusCode::InvalidConfigFile;
+            }
+
             return StatusCode::Success;
         },
         delegate
--- a/src/corehost/cli/ijwhost/ijwhost.cpp
+++ b/src/corehost/cli/ijwhost/ijwhost.cpp
@@ -30,12 +30,13 @@ pal::hresult_t get_load_in_memory_assembly_delegate(pal::dll_t handle, load_in_m
             pal::string_t mod_path;
             if (!pal::get_module_path(handle, &mod_path))
             {
-                trace::error(_X("Failed to resolve full path of the current mixed-mode module [%s]"), host_path.c_str());
+                trace::error(_X("Failed to resolve full path of the current mixed-mode module [%s]"), mod_path.c_str());
                 return StatusCode::LibHostCurExeFindFailure;
             }
 
             pal::string_t config_path_local { strip_file_ext(mod_path) };
             config_path_local.append(_X(".runtimeconfig.json"));
+
             *config_path_out = std::move(config_path_local);
 
             return StatusCode::Success;
--- a/src/corehost/cli/fxr/corehost_init.cpp
+++ b/src/corehost/cli/fxr/corehost_init.cpp
@@ -16,12 +16,14 @@ void make_cstr_arr(const std::vector<pal::string_t>& arr, std::vector<const pal:
 corehost_init_t::corehost_init_t(
     const pal::string_t& host_command,
     const host_startup_info_t& host_info,
+    const pal::string_t &runtime_config_file,
     const pal::string_t& deps_file,
     const pal::string_t& additional_deps_serialized,
     const std::vector<pal::string_t>& probe_paths,
     const host_mode_t mode,
     const fx_definition_vector_t& fx_definitions)
     : m_tfm(get_app(fx_definitions).get_runtime_config().get_tfm())
+    , m_runtime_config_file(runtime_config_file)
     , m_deps_file(deps_file)
     , m_additional_deps_serialized(additional_deps_serialized)
     , m_is_framework_dependent(get_app(fx_definitions).get_runtime_config().get_is_framework_dependent())
@@ -93,6 +95,7 @@ const host_interface_t& corehost_init_t::get_host_init_data()
         hi.fx_ver = _X("");
     }
 
+    hi.runtime_config_file = m_runtime_config_file.c_str();
     hi.deps_file = m_deps_file.c_str();
     hi.additional_deps_serialized = m_additional_deps_serialized.c_str();
     hi.is_framework_dependent = m_is_framework_dependent;
--- a/src/corehost/cli/fxr/corehost_init.h
+++ b/src/corehost/cli/fxr/corehost_init.h
@@ -17,6 +17,7 @@ private:
     std::vector<const pal::char_t*> m_clr_keys_cstr;
     std::vector<const pal::char_t*> m_clr_values_cstr;
     const pal::string_t m_tfm;
+    const pal::string_t m_runtime_config_file;
     const pal::string_t m_deps_file;
     const pal::string_t m_additional_deps_serialized;
     bool m_is_framework_dependent;
@@ -40,6 +41,7 @@ public:
     corehost_init_t(
         const pal::string_t& host_command,
         const host_startup_info_t& host_info,
+        const pal::string_t &runtime_config_file,
         const pal::string_t& deps_file,
         const pal::string_t& additional_deps_serialized,
         const std::vector<pal::string_t>& probe_paths,
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -260,26 +260,26 @@ namespace
     int read_config(
         fx_definition_t& app,
         const pal::string_t& app_candidate,
-        pal::string_t& runtime_config,
+        pal::string_t& runtime_config_file,
         const runtime_config_t::settings_t& override_settings)
     {
-        if (!runtime_config.empty() && !pal::realpath(&runtime_config))
+        if (!runtime_config_file.empty() && !pal::realpath(&runtime_config_file))
         {
-            trace::error(_X("The specified runtimeconfig.json [%s] does not exist"), runtime_config.c_str());
+            trace::error(_X("The specified runtimeconfig.json [%s] does not exist"), runtime_config_file.c_str());
             return StatusCode::InvalidConfigFile;
         }
 
         pal::string_t config_file, dev_config_file;
 
-        if (runtime_config.empty())
+        if (runtime_config_file.empty())
         {
             trace::verbose(_X("App runtimeconfig.json from [%s]"), app_candidate.c_str());
             get_runtime_config_paths_from_app(app_candidate, &config_file, &dev_config_file);
         }
         else
         {
-            trace::verbose(_X("Specified runtimeconfig.json from [%s]"), runtime_config.c_str());
-            get_runtime_config_paths_from_arg(runtime_config, &config_file, &dev_config_file);
+            trace::verbose(_X("Specified runtimeconfig.json from [%s]"), runtime_config_file.c_str());
+            get_runtime_config_paths_from_arg(runtime_config_file, &config_file, &dev_config_file);
         }
 
         app.parse_runtime_config(config_file, dev_config_file, override_settings);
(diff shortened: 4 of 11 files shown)


vitek-karas:
You mention that ideally your tool would be 100% managed code - so I assume there's really no need for C++/CLI and thus `ijwhost` in that world.  
Assuming you have 100% managed code, you can inject it into a running process using the new native hosting APIs. Low level detailed docs are [here](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/native-hosting.md), a sample is [here](https://github.com/dotnet/samples/tree/master/core/hosting/HostWithHostFxr).  
If you have a way to run native code in the target process (which I assume you must have in order for the snoop to work in some way), then you can use the native hosting APIs to get the hostfxr in the process, which in turn exposes the `hostfxr_initialize_for_runtime_config` which will get you a context for your managed code through which you can ask for the runtime delegate - this can load the managed assembly into the process (in isolated ALC) and gets you a native function pointer to a managed method of your choosing. You can then call the function pointer to get to your managed code and do "whatever you want".

This approach still requires `.runtimeconfig.json` for your managed component - this is intentional, the `.runtimeconfig.json` should specify the minimal framework requirements (in your case `Microsoft.NETCoreApp` and `Microsoft.WindowsDesktop.App`) the minimum versions supported. You can use roll forward settings to tweak the compatibility ranges [spec](https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md) [technical doc](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/framework-version-resolution.md). Ideally you would build your managed component using the new properties introduced in https://github.com/dotnet/sdk/pull/3305 - in your case `EnableDynamicLoading=true` - that will create the `.runtimeconfig.json` for you and will automatically default to `rollForward=LatestMinor` which is probably correct for your use case.

I would try to avoid the approach where you component doesn't declare minimum framework requirements (And thus no `.runtimeconfig.json`) - it's just not reliable - if things are not right you would get all kinds of weird exceptions trying to run the managed code. With the `.runtimeconfig.json` you will get a clear error message stating why it didn't work. As such even if you have to go with IJW, I would still argue that you should have `.runtimeconfig.json` - you can make it very "flexible" via the roll forward settings, but having a minimum required framework versions just feels like a good idea no matter what.

Answers:
* How to test with locally built `hostpolicy`: When you fully build the core-setup repo you should get a folder like this `\bin\tests\win-x64.Debug\sharedFrameworkPublish`. This is a "xcopy" install of .NET Core runtime (`Microsoft.NETCore.App` only) which has the hosting bits from your local build (so `hostfxr` and `hostpolicy`). You can add the `Microsoft.WindowsDesktop.App` to it by extracting also locally produced `\bin\win-x64.Debug\packages\windowsdesktop-runtime-3.0.0-preview8-27827-0-win-x64.zip` (the version numbers will differ) to the `shared` folder in the local install (and into the right `Microsoft.WindowsDesktop.App/<Version>` subfolder. With this you can then set `DOTNET_ROOT=\bin\tests\win-x64.Debug\sharedFrameworkPublish` and also `DOTNET_MULTILEVEL_LOOKUP=0` and run your `.exe` app - it will now run only on the locally built frameworks (and hosting layer).
* Tests for IJW are currently missing - mostly because we still don't have the necessary C++ compiler changes to actually produce the correct .NET Core 3.0 compatible C++/CLI assemblies. Depending on what the tests should do, they would either use some custom built managed app (Which the XUnit tests simply runs) or they would get integrated into the `nativehost` native test project, which can call APIs on the `hostfxr` directly (for example the suggested native hosting APIs above are tested this way). Again the XUnit tests end up simply executing the test project.

batzen:
My tool is 100% managed code, but with dllexport (vtfixup etc.).
Which requires ijwhost.

> if you have a way to run native code in the target process (which I assume you must have in order for the snoop to work in some way)

Snoop is used to spy on the UI of any WPF process. You can use snoop, as an example, to spy on the UI of Visual Studio without any modifications required to the code of Visual Studio.
And that's the problem. I don't have any native code in the target process. The code is 100% managed + dllexport.
Maybe you should have a look at the code at https://github.com/batzen/snoopwpf/blob/NETCore3/Shared/InjectorHelper.cs and https://github.com/batzen/snoopwpf/blob/NETCore3/ManagedInjector/Injector.cs.
The code in InjectorHelper is responsible for injecting a callback into a foreign process.
The code in Injector is the code being called in the foreign process through the callback we established through InjectorHelper.
If you know some way to call the hosting API in a foreign process to load my library there i wouldn't need the ijwhost.

I don't want to use runtimeconfig.json as that would mean i'd dictate, to some degree, the runtime version it should use and i would also get a new ALC wouldn't i?
The problem with roll forward settings is that i can't say "use this minimum framework version, but roll forward to the framework version the current/existing context is using if there is any". Or is there some way to achieve this?

We had to get around framework differences in snoop when needing to support .net 3.5 and 4.0 at the same time.
This was solved by detecting which framework version the target process used and then injected the appropriate version of our injection dll.
That's also the way i'd like to solve this in the future as it allows finer grained control over supported features/framework version differences.

Thanks for answering my questions. That will make testing/fiddling a lot easier for me.

src/corehost/cli/ijwhost/ijwhost.cpp
```diff
@@ -30,14 +30,53 @@ pal::hresult_t get_load_in_memory_assembly_delegate(pal::dll_t handle, load_in_m
...
+            {
+                trace::info(_X("Failed to resolve .runtimeconfig.json path of the current mixed-mode module [%s]"), mod_path.c_str());
+                trace::info(_X("Trying to get .runtimeconfig.json from existing context"));
+
+                pal::dll_t out_fxr;
+                pal::string_t out_fxr_path;
+
+                if (pal::get_loaded_library(LIBFXR_NAME, "hostfxr_get_runtime_property_value", &out_fxr, &out_fxr_path))
+                {
+                    hostfxr_get_runtime_property_value_fn get_runtime_property_value = reinterpret_cast<hostfxr_get_runtime_property_value_fn>(pal::get_symbol(out_fxr, "hostfxr_get_runtime_property_value"));
```
AaronRobinsonMSFT: If the type is properly declared on the right side, feel free to use `auto` on the left. These function `typedef`s and names tend to get out of control.

AaronRobinsonMSFT:
>I don't want to use runtimeconfig.json as that would mean i'd dictate, to some degree, the runtime version it should use and i would also get a new ALC wouldn't i?

That is true. However an example of a case where this would be bad would be if the user only has a pre-3.0 .NET Core installed. In that instance a pre-3.0 runtime could be loaded and will more than likely not be able to satisfy your applications requirements and crash. Specifying a minimum runtime (i.e. 3.0) would avoid this. I _think_ this is the kind of scenario @vitek-karas had in mind - please correct me if I am wrong.

vitek-karas:
Thanks a lot @batzen for the explanation. I looked at the mechanism used to inject code into the target process - interesting, didn't know it's possible to do that 😄 . So if I read it correctly:
* The mechanism can get a native function (the windows message hook handler) loaded into the target process and run it
* Ideally you want to implement that function in C# - so you use IJW to export a native entry point from your assembly
* This native entry point is used as the windows message hook handler

@AaronRobinsonMSFT is right in the comment about declaring minimum required version. Without runtime config the tool would try to load your assembly into any .NET Core process, regardless of version requirements. Overall I don't think it would actually break anything, just the experience might be rather weird:
* If the assembly relies on runtime features which are not present, it might fail in unexpected ways
* If the assembly has dependencies on versions of assemblies which are not available in the target process (typically too low versions) it might fail
* If the target process does not have the dependency assemblies (for example in your case you expect WPF to exist in the target process) it would fail
* Any failure in the hooked assembly would probably show up as a really weird behavior, either crash or weird exceptions and so on

There is one other consideration for the future. Let's say this is enabled in 5 and your code uses it and so on... now 6 comes along and since it's a major version the .NET Core would have a breaking change in it. Without declaring framework versions and allowed roll forward behaviors, your component might break when loaded into a process which has version 6 loaded. Unlike the above samples where the loaded version is too low, in this case there would be no indication of the problem, now assembly load failures and so on, it's just that something would not work as expected.

For these reasons I think it's just cleaner to declare the minimum required frameworks and their versions and the desired compatibility band (via roll forward). The error experience is probably not gonna be much better (IJW does not have a good way to report errors to the caller), but at least it would fail just during the injection process and not later on at random points.

About the code changes:
If the goal is to allow loading IJW assemblies without `.runtimeconfig.json` I think there is a cleaner way to solve this which would benefit other scenarios as well.
We could modify `hostfxr_initialize_for_runtime_config` to consider the runtime config path as optional, or we could introduce a new export like `hostfxr_initialize_for_loaded_runtime` which would basically take no parameters. The idea is that:
* If there already is a runtime in the process the hosting layer already knows the frameworks and other configuration
* In that case the runtime config passed to `hostfxr_initialize_for_runtime_config` is only used for validation, none of the information in it will affect the loaded runtime in any way
* If the desire is to basically skip that validation we can simply make the runtime config optional and if it's not present, skip the validation steps.
* The rest of the code paths would be unaffected since they already work with the loaded runtime alone
* Calling the `hostfxr_initialize_for_runtime_config` without runtime config (or the new API) if there's no runtime loaded in the process would cause a failure - since without runtime config there's no way for us to find and load a runtime (this is not your scenario, in your case you very likely want to fail anyway).

This would effectively allow the "no runtime config" scenario for any usage of the hosting layer, not just IJW. Eventually we could enable it for COM and WinRT hosting as well. The direct native hosting would get it immediately.

From a technical perspective such solution looks really clean and I would have no problem with it. That said I'm still not 100% I want to enable it. I just fear that we would make it too easy for users to get into bad situations... But if others think having this support would be beneficial I'll be OK with that as well 😄

batzen:
Thinking about this a bit more, I don't have an issue with the runtimeconfig at all.
It's more the fact that there is no way to express:
"Hey, i have this as my minimum runtime version, but i am fine with whatever version is loaded if it satisfies my minimum requirement. If the current loaded version does not satisfy my requirement fail."
What if we add that?
That would also benefit comhost i think.

We can already express:
- Minimum required version
- Roll forward via default roll forward rules if no runtime is loaded

We'd just need an option to say:
- Roll forward to loaded version

It would then be possible to have something like:
```
{
  "runtimeOptions": {
    "tfm": "netcoreapp3.0",
    "framework": {
      "name": "Microsoft.WindowsDesktop.App",
      "version": "3.0.0-preview6-27804-01",
      "rollForward": "LatestMinor",
      "useLoadedRuntimeIfPresent": "true"
    }
  }
}
```
I guess i should create a new PR for the changes needed to make that possible.

Besides that:
Would you be fine with the rest of the PR? Aside from the IJWHost changes, for which i could create a new PR. I mean exposing the used runtimeconfig file for the current context via hostfxr properties might be useful in general.
I am asking because the transport of the runtimeconfig through the layers alone would allow me to create my own ijwhost.dll and simply ship that while working on the enhanced rollForward mechanism.

AaronRobinsonMSFT:
I am fine in principle with these changes. The `ijwhost` work seems appropriate to me. I would like @elinor-fung to take a look at these changes though, she will be back next week. Please consider me signed off. Thanks.

batzen:
So is there anything more for me to do, or do i just have to wait for next week?
How about the failing unit tests? They run successfully on my machine. Did i miss something?

vitek-karas:
Given your scenario I actually don't see a need for any change at all. The current roll forward should already support what you're looking for.

The roll forward setting can be thought of as having two parts:
* version range
* how to pick a version in the range

So for example `3.0.0 rollForward=Minor` means anything above 3.0.0 but lower then 4.0.0 (`Minor` means no change in major version number). It also means closest available version.

When the hosting layer determines compatibility with already loaded runtime, it only uses the version range, it ignores the other part. If the loaded runtime "fits" the minimum version and the specified version range, it will allow it.

So for example if your component specified `3.0.0 LatestMinor` and the runtime in the process is `3.1.0` it will happily work. If the runtime was `5.0.0`, then it would fail. The `Latest` part is completely ignored and it would work the same for both `Minor` and `LatestMinor`.

If you want to have a component which runs on anything above a given version, then specify `3.0.0 LatestMajor` (or `Major`, but components should specify `Latest` for consistency as it's important to those which can initialize a runtime).

The check when the runtime is already loaded is done here: https://github.com/dotnet/core-setup/blob/c76e79f4c0c38227f3beda3ccfb60c8cd7c30d9f/src/corehost/cli/fxr/fx_muxer.cpp#L626

and then here: https://github.com/dotnet/core-setup/blob/c76e79f4c0c38227f3beda3ccfb60c8cd7c30d9f/src/corehost/cli/fxr/fx_resolver.cpp#L548

I don't have a big problem with reporting the location of the runtime config outside of the `hostfxr`, but it feels unnecessary. Also there's an easier way to do that by adding the new runtime property to the list of properties reported to the hostpolicy. The collection is filled here: https://github.com/dotnet/core-setup/blob/c76e79f4c0c38227f3beda3ccfb60c8cd7c30d9f/src/corehost/cli/fxr/corehost_init.cpp#L55

The code currently reports any properties which are collected from runtime config files (app or framework). If we really need to add another, simply add one to the list after those from config files.

But again - given the scenario as I understand it, I don't see any need for hosting changes... but I could easily be wrong.

batzen:
Haven't thought of the easy way... Will try if it works the way you suggested next weekend.

batzen:
Next weekend was a bit optimistic.
I finally had time to try this and i can verify that it works.
I just have to place a runtimeconfig.json beside the assembly i want to inject and it works as you described.
So there is no need to change anything.

Should i close this issue then?

AaronRobinsonMSFT:
@batzen thanks for the confirmation. I will close the PR for you.

vitek-karas:
Just wanted to add that with Preview7 it should be relatively easy to build the component with the `.runtimeconfig.json` just add `<EnableDynamicLoading>true</EnableDynamicLoading>` to your project.

batzen:
That already worked in preview 6 ;-)

vitek-karas:
Right - I'm really bad at remembering what ships where :-)

batzen:
And when :-) It's moving so fast that it's quite difficult to keep track.

PR #6946, opened 2019-06-26
RFC: Only load resources with locale tags if it matches system locale

lpereira:
As it stands, I don't know if there are side effects to this change; it'll probably break some applications.  However, it can significantly reduce the amount of system calls performed by starting some applications with the muxer.

The idea is to not load resources tagged with a specific locale if it doesn't match the current environment.  For instance, if the current environment has "en-US" as a culture, it'll unconditionally load resources without any "locale" tag, but load conditionally only those that match either "en-*" ("en-US", "en-GB", etc.) or "en" as locale.

The current code is *very* hackish and possibly will only work on Unix systems, as it relies on the LANG environment variable; it's just a PoC.

However, it shows promising results; by counting the syscalls with "strace -qfc -etrace=%file" (so, only syscalls related to file I/O, to avoid tracing calls to futex(2) and whatnot that are unrelated to this
change), these are the results:

```
TemplateApp	#SyscallsBefore	#SyscallsAfter	Diff
console		19742		15739		-4003 (~25%)
webapp		27843		23839		-4004 (~16%)
```

Where the TemplateApp column lists programs created with `dotnet new`. The webapp program has been modified so that its Main() function does nothing bug print "Finished loading", instead of waiting for connections to serve.

(Each program has been executed 3 times under strace, and the best number of syscalls have been chosen.  By tracing only file I/O-related syscalls, the numbers are pretty stable.)

Running each program under time(1) 4 times, by dropping the caches every time, I got these results (on my fast workstation, with fast SSDs, etc), these were the average values:

```
TemplateApp	SysTimeBefore	SysTimeAfter	Diff
console		0.191s		0.169s		-0.022s
webapp		0.195s		0.184s		-0.011s
```

From 10ms to 20ms shaved from initialization isn't all that bad; this is of course on a machine that's mostly idle, with fast I/O, plenty of free RAM, etc.

### DIFF
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -287,6 +287,80 @@ bool deps_json_t::process_runtime_targets(const json_value& json, const pal::str
     return true;
 }
 
+static bool matches_current_locale(const pal::string_t& locale)
+{
+    // This is a TERRIBLE HACK! Just to test the idea.
+    //
+    // getenv("LANG") should really go to the PAL under something like
+    // pal::get_current_locale(), and it should do the platform-specific
+    // massaging there.
+    //
+    // Also, this might trigger loading certain assets that might not really
+    // relate to the current locale.  For instance, Chinese assets for the
+    // traditional script might be loaded even though the current locale
+    // uses the simplified script.  Without an actual locale fallback graph,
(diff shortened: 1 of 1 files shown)


dagood:
Is this PR active?

lpereira:
> Is this PR active?

Not anymore.  I'm closing it.

PR #6967, opened 2019-06-27
[dev/arcade-migration] Fix tests for CI platforms, add PR validation

dagood:
Changes are segmented into commits for easier review.


eng/jobs/windows-build.yml
```diff
@@ -206,25 +75,31 @@ jobs:
         displayName: Publish build assets
         condition: and(succeeded(), eq(variables._BuildConfig, 'Release'))
 
+    - template: steps/check-space-powershell.yml
```
MichaelSimons: Can you explain the scenario for why this exists here?  Are there plans to use it elsewhere?

dagood: Ah, I copied it over from source-build because I suspected the machines were running out of space, causing test failures. (This is why the job is no longer on the hosted pool.) Removing this diag step.

PR #6968, opened 2019-06-27
HostModel:Enable TestWithAdditionalContentAfterBundleMetadata

swaroop-sridhar:
The above test was disabled in PR #6885 because the lab runs
showed failures because of "not enough disk space"

This PR tries to enable the test, and move the job to not
use the hosted pool, as suggested [here](https://github.com/dotnet/core-setup/pull/6885#issuecomment-505998739).


dagood:
I already did the pool change as part of the Arcade SDK migration:

https://github.com/dotnet/core-setup/blob/732ef42a49e82b1bfbe1363fa6799b34d7e8ef34/eng/jobs/windows-build.yml#L13-L20

I think you should be able to rebase this PR to just add the test back in.

PR #6989, opened 2019-06-29
Make Microsoft.NET.HostModel.AppHost.ResourceUpdater public

swaroop-sridhar:
Make Microsoft.NET.HostModel.AppHost.ResourceUpdater class public,
so that it can be consumed from the SDK repo.


swaroop-sridhar:
CC @wli3

PR #7040, opened 2019-07-02
Update timeouts to 120 minutes for all jobs

wtgodbe:
Jobs are failing in CI/official builds with timeouts pretty frequently recently. Bumping up the timeouts to 120 to get around this. Will port this to 3.0 as well.

@safern @dagood PTAL

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,7 +97,6 @@ jobs:
     portableBuild: true
     skipTests: true
     targetArchitecture: arm
-    timeoutInMinutes: 120
 
 - template: /eng/jobs/bash-build.yml
   parameters:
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -198,13 +208,15 @@ jobs:
...
     targetArchitecture: x64
+    timeoutInMinutes: 120
 
 # Windows x86
 - template: /eng/jobs/windows-build.yml
   parameters:
     buildFullPlatformManifest: true
     displayName: Build_Windows_x86
     targetArchitecture: x86
+    timeoutInMinutes: 120
```
safern: Given that you set the default value of the parameter in the templates, you shouldn't need to pass down the value unless you want to override it to a different value from 120

eng/jobs/osx-build.yml
```diff
@@ -1,5 +1,8 @@
+parameters:
+  timeoutInMinutes: 120
```
safern: Do you need so much time for OSX? Do we sign on OSX?

Issue #7039, opened 2019-07-02
Duplicate entry for libclrjit.so in RuntimeList.xml for linux-arm.


peterhuene:
## Steps to reproduce

1. Install a .NET Core SDK version `3.0.100-preview7-012769` or later.
2. `mkdir repro && cd repro`
3. `dotnet new web`
4. `dotnet publish -r linux-arm`

## Expected  behavior

The application publishes correctly.

## Actual behavior

```
Microsoft.NET.Sdk.targets(168,5): error MSB4018: The "GenerateDepsFile" task failed unexpectedly. [repro.csproj]
Microsoft.NET.Sdk.targets(168,5): error MSB4018: System.ArgumentException: Can not add property libclrjit.so to Newtonsoft.Json.Linq.JObject. Property with the same name already exists on object. [repro.csproj]
```

This is caused by a duplicate for `libclrjit.so` in the `RuntimeList.xml` file:

```xml
<File Type="Native" Path="runtimes/linux-arm/native/libclrjit.so" FileVersion="0.0.0.0" />
...
<File Type="Native" Path="runtimes/x64_arm/native/libclrjit.so" FileVersion="0.0.0.0" />
```

The `x64_arm/native/libclrjit.so` probably does not belong in the list of files to publish for a self-contained application.

We're tracking making the SDK handle this gracefully with dotnet/cli#11705.

## Environment data

```
$ dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-preview7-012769
 Commit:    3d1e17251c

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.14
 OS Platform: Darwin
 RID:         osx.10.14-x64
 Base Path:   /usr/local/share/dotnet/sdk/3.0.100-preview7-012769/

Host (useful for support):
  Version: 3.0.0-preview7-27901-06
  Commit:  df2b1489de

.NET Core SDKs installed:
  2.2.103 [/usr/local/share/dotnet/sdk]
  2.2.104 [/usr/local/share/dotnet/sdk]
  3.0.100-preview5-011568 [/usr/local/share/dotnet/sdk]
  3.0.100-preview7-012430 [/usr/local/share/dotnet/sdk]
  3.0.100-preview7-012769 [/usr/local/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.2.1 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.All]
[...]

PR #7041, opened 2019-07-02
[master] Leave any cross-targeting file off RuntimeList.xml

dagood:
For https://github.com/dotnet/core-setup/issues/7039.

Detect the `_` in e.g. `runtimes/x64_arm/native/libclrjit.so` and don't include it in `RuntimeList.xml`.

I would prefer to change the input `TargetFilePrefixes` to use `runtimes/linux-arm/` rather than `runtimes/`, but that would take more validation work than the task-side workaround. I'll file an issue to follow up.

I'm running a mock build now to make sure this works.

Submitting a port PR to `release/3.0` shortly.

/cc @wtgodbe @dsplaisted

### DIFF
--- a/tools-local/tasks/CreateFrameworkListFile.cs
+++ b/tools-local/tasks/CreateFrameworkListFile.cs
@@ -104,6 +104,18 @@ namespace Microsoft.DotNet.Build.Tasks
 
                 string path = Path.Combine(f.TargetPath, f.Filename).Replace('\\', '/');
 
+                if (path.StartsWith("runtimes/"))
+                {
+                    var pathParts = path.Split('/');
+                    if (pathParts.Length > 1 && pathParts[1].Contains("_"))
+                    {
+                        // This file is a runtime file with a "rid" containing "_". This is assumed
+                        // to mean it's a cross-targeting tool and shouldn't be deployed in a
+                        // self-contained app. Leave it off the list.
+                        continue;
+                    }
+                }
+
                 var element = new XElement(
                     "File",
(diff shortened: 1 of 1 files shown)


dsplaisted:
Are we sure that `_` is always going to mean a cross-targeted JIT?  Is it possible to exclude this file by marking the input item with metadata?

dagood:
Confirmed that the entry is removed from the runtime list (and only this entry) for:
```
Microsoft.NETCore.App.Runtime.linux-arm
Microsoft.NETCore.App.Runtime.linux-arm64
Microsoft.NETCore.App.Runtime.linux-musl-arm64
Microsoft.NETCore.App.Runtime.win-arm
Microsoft.NETCore.App.Runtime.win-arm64
```

The mock build hasn't finished any non-arm packages yet to check if they're ok. Very low risk though.

> Are we sure that `_` is always going to mean a cross-targeted JIT?

No, this is a reason I don't like this workaround. It is true for our current RIDs, though.

> Is it possible to exclude this file by marking the input item with metadata?

Yes, but the current solution I'd like is to simply exclude anything in `runtime/` that doesn't match the desired RID. I'm pretty sure these files aren't special in the current build logic, so I'd have to add a place to find these files and add the metadata, and that IMO would break more easily.

dagood:
Confirmed the lists for `win-x64`, `linux-x64`, and `osx-x64` are good in the mock build.

dagood:
Filed https://github.com/dotnet/core-setup/issues/7046 to improve the filter.

PR #7042, opened 2019-07-02
[release/3.0] Leave any cross-targeting file off RuntimeList.xml

dagood:
Clean port of https://github.com/dotnet/core-setup/pull/7041:

> For #7039.
> 
> Detect the `_` in e.g. `runtimes/x64_arm/native/libclrjit.so` and don't include it in `RuntimeList.xml`.
> 
> I would prefer to change the input `TargetFilePrefixes` to use `runtimes/linux-arm/` rather than `runtimes/`, but that would take more validation work than the task-side workaround. I'll file an issue to follow up.
> 
> I'm running a mock build now to make sure this works.
> 
> /cc @wtgodbe @dsplaisted


dagood:
Merging: notes about validation in https://github.com/dotnet/core-setup/pull/7041.

PR #7134, opened 2019-07-09
Use tracing infrastructure to log realpath() and getcwd() errors

lpereira:
We have a nice tracing infrastructure; let's use it instead of sending messages to stderr with perror().  Also, if possible, give a little bit more context.

This is probably only for 5.0 as will change `dotnet` and it's too late for that at this point.


vitek-karas:
Feel free to merge this now.

PR #7135, opened 2019-07-09
Retry extraction of bundle to avoid issues with locked files by AV

vitek-karas:


lpereira:
> ~~640KB~~50s should be enough for everybody.

Seems to be a robust workaround.  Even though it's in a platform-independent area, this shouldn't affect platforms where AV software is less common (which, coincidentally, are also platforms where files can be moved around even if some process have them open.)

src/corehost/cli/apphost/bundle/bundle_runner.cpp
```diff
@@ -312,21 +312,41 @@ StatusCode bundle_runner_t::extract()
...
-                return StatusCode::Success;
+                bool should_retry = errno == EACCES;
+                if (can_reuse_extraction())
+                {
+                    // Another process successfully extracted the dependencies
+
+                    trace::info(_X("Extraction completed by another process, aborting current extracion."));
+
+                    remove_directory_tree(m_working_extraction_dir);
+                    return StatusCode::Success;
```
AaronRobinsonMSFT: Should we be calling `fclose(m_bundle_stream)` here? All other cases cause that to happen.

lpereira: Good catch. Yes, seems like it, even though the original code didn't do it in this path.

dasMulli:
50s is actually good, McAffee's default timeout for on-access scanning is 45s, so the files "should" be unlocked 45s after writing 🤞

PR #7171, opened 2019-07-11
Fixes couple of places in the bundle runner where we might leak memory.

vitek-karas:
Note that in reality this is a non-issue: The leaks are only present if something fails and in case of failure the process will immediately exit anyway.

I know that @swaroop-sridhar has plans to rewrite most of this using file mapping, so the code as-is will mostly go away. I did this mostly for my own education on how to write this correctly. And since I had the changes read, we might as well merge it.

### DIFF
--- a/src/corehost/cli/apphost/bundle/manifest.h
+++ b/src/corehost/cli/apphost/bundle/manifest.h
@@ -17,7 +17,7 @@ namespace bundle
     class manifest_t
     {
     public:
-        std::list<file_entry_t> files;
+        std::vector<file_entry_t> files;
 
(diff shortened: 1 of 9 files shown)


vitek-karas:
This is ready for official review... anybody?

src/corehost/cli/apphost/bundle/manifest.h
```diff
@@ -17,13 +17,9 @@ namespace bundle
     class manifest_t
     {
     public:
-        manifest_t()
-            :files()
-        {}
+        std::list<file_entry_t> files;
```
AaronRobinsonMSFT: Use `vector<T>` instead.

PR #7210, opened 2019-07-15
[dev/arcade-migration] Add simple sharedfx package test project, fix PlatformManifest creation

dagood:
Adds some simple tests for the packages (as final product nupkgs) and fixes platform manifest generation and inclusion. Also:
* Simplifies the way test variables are passed in so I could debug them in VS and remove a workaround, per https://github.com/dotnet/arcade/issues/3077.
* Fix the WindowsDesktop packages building on unexpected platforms. (Was breaking the tests.)
* Disable Arcade's `use_installed_dotnet_cli` and `use_global_nuget_cache` to make the build and dev experience more consistent. This is copied from dotnet/coreclr.

### DIFF
--- a/src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
+++ b/src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
@@ -1,0 +1,166 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.DotNet.CoreSetup.Test;
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Xml.Linq;
+using Xunit;
+
+namespace Microsoft.DotNet.CoreSetup.Packaging.Tests
+{
+    public class NuGetArtifactTester : IDisposable
+    {
+        public static NuGetArtifactTester Open(
+            RepoDirectoriesProvider dirs,
+            string project,
+            string id = null)
+        {
+            var tester = OpenOrNull(dirs, project, id);
+            Assert.NotNull(tester);
+            return tester;
+        }
+
+        public static NuGetArtifactTester OpenOrNull(
+            RepoDirectoriesProvider dirs,
+            string project,
+            string id = null)
+        {
+            id = id ?? project;
+
+            string nuspecPath = Path.Combine(
+                dirs.BaseBinFolder,
+                project,
+                dirs.Configuration,
+                "specs",
+                $"{id}.nuspec");
+
+            if (!File.Exists(nuspecPath))
+            {
+                return null;
+            }
+
+            PackageIdentity builtIdentity = new NuspecReader(nuspecPath).GetIdentity();
+
+            string nupkgPath = Path.Combine(
+                dirs.BaseArtifactsFolder,
+                "packages",
+                dirs.Configuration,
+                "Shipping",
+                $"{builtIdentity}.nupkg");
+
+            // If the nuspec exists, the nupkg should exist.
+            Assert.True(File.Exists(nupkgPath));
+
+            return new NuGetArtifactTester(nupkgPath);
+        }
+
+        private readonly PackageArchiveReader _reader;
+
+        public NuGetArtifactTester(string file)
+        {
+            _reader = new PackageArchiveReader(ZipFile.Open(file, ZipArchiveMode.Read));
+        }
+
+        public void Dispose()
+        {
+            _reader.Dispose();
+        }
+
+        public void IsTargetingPack()
+        {
+            IsFrameworkPack();
+
+            Assert.NotEmpty(_reader.GetFiles("ref"));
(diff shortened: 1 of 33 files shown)


src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/WindowsDesktopTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        private readonly RepoDirectoriesProvider dirs = new RepoDirectoriesProvider();
+
+        [Fact]
+        public void WindowsDesktopTargetingPackIsValid()
+        {
+            using (var tester = NuGetArtifactTester.OpenOrNull(
+                dirs,
+                "Microsoft.WindowsDesktop.App.Ref"))
+            {
+                if (ShouldExist)
```
MichaelSimons: IMO the readability of this could be improved.  I suspect most new readers of this code will be forced to look at the implementation to figure out what this means.

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
```diff
@@ -0,0 +1,159 @@
...
+                $"{id}.nuspec");
+
+            if (!File.Exists(nuspecPath))
+            {
+                return null;
+            }
+
+            PackageIdentity builtIdentity = new NuspecReader(nuspecPath).GetIdentity();
+
+            string path = Path.Combine(
```
MichaelSimons: (nit) from a readability perspective, I think naming this nupkgPath would help since there is already another `path` related variable previously declared.

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            _reader.Dispose();
+        }
+
+        public void IsTargetingPack()
+        {
+            IsFrameworkPack();
+
+            Assert.NotEmpty(_reader.GetFiles("ref"));
+            Assert.Empty(_reader.GetFiles("runtimes").Concat(_reader.GetFiles("lib")));
```
MichaelSimons: It seems like it would be easier to track down issues if the asserts were more fine grained.  Why not break out `Assert.Empty(_reader.GetFiles("lib"));`

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            HasOnlyTheseDataFiles("data/RuntimeList.xml");
+
+            ContainsFrameworkList("RuntimeList.xml");
+        }
+
+        public void HasOnlyTheseDataFiles(params string[] files)
+        {
+            Assert.Equal(
+                files,
```
MichaelSimons: In practice all specified files with be alphabetized?

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/WindowsDesktopTests.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.DotNet.CoreSetup.Packaging.Tests
+{
+    public class WindowsDesktopTests
+    {
+        private readonly RepoDirectoriesProvider dirs = new RepoDirectoriesProvider();
+
+        [Fact]
+        public void WindowsDesktopTargetingPackIsValid()
+        {
+            using (var tester = NuGetArtifactTester.OpenOrNull(
```
MichaelSimons: I think a code comment would be helpful calling out the `OrNull` scenario.

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/WindowsDesktopTests.cs
dagood: Renamed to `CurrentRidShouldCreateNupkg` (pointless to be generic about it until there's a reason to be).

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
dagood: My reasoning was that combining them reduces the onion peeling if both runtime and lib files started showing up. I believe the assert shows the strings, so it's diagnosable fine. Changing, though--splitting this up also makes it more readable IMO and it's unlikely this is a practical issue.

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
dagood: These files come from the calling test code, so it's up to test authoring to make sure they're alphabetized--but this isn't made clear anywhere so I'll rework this. Alphabetization was just a trick to do a set equality assert.

src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/NuGetArtifactTester.cs
dagood: > I believe the assert shows the strings

Not true!

PR #7233, opened 2019-07-16
[dev/arcade-migration] Improve dev scenarios

dagood:
Improved dev scenarios that were common with the old infrastructure:

Provide `/p:Subset=...` that works similarly to the old `run.cmd` args that built only certain projects.
* Old: `.\build.cmd -corehost-build`, new: `.\build.cmd /p:Subset=corehost`
* `/p:Subset=` is not great to type, but handling it in MSBuild lets us avoid maintaining a custom build script with arg parsing/passthrough specific to Core-Setup and duplicating the list of subsets.
  * If `-foo` is super desirable above `/p:Subset=foo`, we can add custom build scripts that turn known `-foo` subset args into `/p:Subset=foo` MSBuild args. This also gives us a place to add to `-help`/`--help`. (Arcade help is not extensible without a custom script.)
  * @vitek-karas, curious what you think here.

Fix readme table generation.
* Used the Subset infra to make a new easy build command to put in the readme.

Build managed projects and tests in VS, and run tests with VS Test Explorer.
* This was already mostly working. I updated some references that were pointing to old .NET Core versions to make the solution build without errors/warnings in my VS install.
* Removed `visual-studio-devenv.cmd` (obsolete) and updated doc.


eng/Build.props
```diff
@@ -1,45 +1,83 @@
...
+  <PropertyGroup>
+    <SubsetToLower>$(Subset.ToLowerInvariant())</SubsetToLower>
+  </PropertyGroup>
+
   <!--
     Projects for Arcade to build.
 
+    Supports '/p:Subset=<desired subset string>' dev build argument. If not set, all projects run.
+    Otherwise, each subset only runs if it is contained in the desired subset string. Each subset
+    listed in the arg should be separated by a '-' to avoid unintended matches.
```
MichaelSimons: I'm not following how a `-` avoids unintended matches?  The matching logic is somewhat _fuzzy_ because of the use of `contains`.  Aren't there are a number of scenarios which could have unintended matches?  e.g. 'mismanaged`  I'm not saying there is anything wrong with that but the comment is somewhat misleading.

eng/ShowNoProjectToBuildError.proj
```diff
@@ -0,0 +1,16 @@
+<Project>
+
+  <!--
+    Display an error if no projects will be built. Echo back the Subset variable if set: a typo in
+    that variable is the only anticipated way to hit this situation.
+  -->
+  <Target Name="Build">
+    <PropertyGroup>
+      <Message>No ProjectToBuild items found</Message>
+      <Message Condition="'$(Subset)' != ''">$(Message) in subset matching '$(Subset)'</Message>
```
MichaelSimons: It is a little unfortunate that a typo of a single subset provides a meaningful error UX but there isn't one when specifying a set - e.g. corehost-tast (typo of corehost-test)

dagood:
Cool, thanks.

> would be nice to have a doc which describes how to use these and what are the various values about.

I'm thinking I'll move this bit of `Build.props` into the root, calling it `Subsets.props`, add a big doc comment, and add a link in the readme. Seems likely to stay up to date that way.

eng/ShowNoProjectToBuildError.proj
dagood: Yeah... I'll see if I can do this better. One limitation is that right now I don't have a list of the valid subsets, but I think it's worthwhile to add that.

eng/Build.props
dagood: Yeah, this should say *some* unintended matches, and even then I don't have any practical examples. This is specifically about two subset names combining to form some unexpected name.

So, say, `corehost` and `estimate`  are `corehostestimate`, which contains `test`.

But if I can make the validation better (your other comment) this will become a requirement rather than just a convention.

dagood:
> add a link in the readme.

On second thought, I'm not actually sure this is a great idea without having the basics documented yet. I did doc work in `Subsets.props` but IMO there's no obvious place to link to it from.

I did get the functionality so the build validates each subset in the list individually, and added a `help` psuedo-subset that tells you about them.

```
  Accepted Subset values (see Subsets.props for more info):
  - CoreHost
      The .NET Core Host projects. This includes all the native code in Core-Setup.
  - Managed
      The managed .NET projects. This includes PlatformAbstractions, DependencyModel, and HostModel.
  - depproj
      The dependency projects. These gather shared framework files and run crossgen on them to turn them into ready-to-run (R2R) assemblies for the current platform.
  - pkgproj
      The packaging projects. These produce NETCoreApp and WindowsDesktop assets: NuGet packages, installers, zips, and Linux packages.
  - Installer
      Generates additional installers. This produces the shared frameworks and their installers.
  - Test
      The test projects. Note that building this doesn't execute tests: you must also pass the '-test' argument.
  - RegenerateReadmeTable [only runs on demand]
      Regenerates the table of asset links in the README.md file.

C:\git\core-setup-arcade\eng\Build.props(55,5): error : Choose a subset to use, or do not specify a subset to perform the full build. [C:\git\core-setup-arcade\.packages\microsoft.dotnet.arcade.sdk\1.0.0-beta.19364.1\tools\Build.proj]
```

vitek-karas:
One place which could use parts of this info is the end of this doc: https://github.com/dotnet/core-setup/blob/master/Documentation/building/windows-instructions.md

Ideally we would also add another one for Linux and Mac...

vitek-karas:
There's also https://github.com/dotnet/core-setup/blob/master/Documentation/building/Running-Tests-In-VisualStudio-IDE.md which needs updating or outright deletion (and adding the two lines which still aply to the other doc)

dagood:
I updated [Running-Tests-In-VisualStudio-IDE.md](https://github.com/dagood/core-setup/blob/dev-scenarios/Documentation/building/Running-Tests-In-VisualStudio-IDE.md) already, actually. 😄 I did miss `windows-instructions.md` though, I disregarded it because it looked like just prereqs. I added a note to the bottom about subsets and cleaned up a couple things.

Issue #7137, opened 2019-07-09
Latest DependencyModel writes / as \u0002f breaks all but bleeding edge versions of xunit + `dotnet test`


nguerrera:
Deps files come out like this:

```
      "Microsoft.CodeCoverage\u002f16.0.1": {
        "runtime": {
          "lib\u002fnetcoreapp1.0\u002fMicrosoft.VisualStudio.CodeCoverage.Shim.dll": {
            "assemblyVersion": "15.0.0.0",
            "fileVersion": "16.0.28223.3002"
          }
        }
      },
      "Microsoft.CSharp\u002f4.0.1": {
```

Where \u0002f could be /

This breaks `dotnet test` on `dotnet new xunit` unless the xunit version used has https://github.com/xunit/visualstudio.xunit/commit/7553f4fa3b67884ccdda597f5429c5cdf2809d77 from yesterday.

Once https://github.com/dotnet/corefx/issues/37192 is available, DependencyModel should use it to only escape what's truly necessary. At a minimum / should be allowed through.

cc @ericstj @steveharter

nguerrera:
Putting this in 3.0 milestone because we will have to pin the version of dependency model to a preview 6 build and we can't ship like that.

nguerrera:
Looks like it + escaping is also a problem (so / fix is not sufficient). Ideally, the escaping would mimic what newtonsoft.json was doing with 2.2 dependency model.

GrabYourPitchforks:
How is this breaking? This is perfectly valid JSON. In theory every single character could be "\uXXXX" escaped and it would still be a legitimate, well-formed payload. All JSON consumers need to anticipate "\uXXXX" escape sequences in the payload and handle them appropriately, otherwise they're not spec-compliant.

nguerrera:
It's not about theoretical spec compliance. It's about the practical fact that every project in the world will have to take a new xunit to use 3.0 SDK because until a few days ago the private json reader in xunit was not spec compliant around this. 

It will break users that upgrade to 3.0 SDK without upgrading xunit. In this emprical sense, it is a breaking change.

I could maybe live with that (though it's not my call) if the failure was at all obvious, but thanks to my biggest pet peeve of all time, somebody is catching IndexOutOfRangeException and it manifests as "no tests discovered", which googling will very likely lead you on to a wild goose chase through the 42000 ways you can get this error. :(

[...]

PR #7241, opened 2019-07-17
Use a relaxed JavaScriptEncoder for writing JSON to disk.

ahsonkhan:
Fixes https://github.com/dotnet/core-setup/issues/7137

This should also resolve the concerns brought up in https://github.com/microsoft/vstest/issues/2079.

~**Waiting on a newer dependency update from corefx (newer than https://github.com/dotnet/core-setup/pull/7240 - https://github.com/dotnet/core-setup/pull/7240/commits/fe9de03c1c91e3facf17973782a53f82266e56dd - `4.6.0-preview8.19366.8`) that has this fix in - https://github.com/dotnet/corefx/pull/39550, but preparing the PR in advance.**~

~This won't compile/pass CI until the dependencies are updated.~

Edit: Dependencies updated in https://github.com/dotnet/core-setup/pull/7243

One thing we should consider doing (especially after `DependencyModel` targets netstandard2.0 only), is re-write some of the tests to **not** use `JObject` from `Newtonsoft.Json` as part of the verification (and maybe even be JSON-library-agnostic). This difference in escaping wasn't being captured by existing tests because:
1) The existing tests didn't really have Base-64 encoded text for sha512 (or any string containing characters like + that would get escaped).
2) The verification would still use `JObject` which would skip escaping them anyway (particularly the Save method).
https://github.com/dotnet/core-setup/blob/6bc158d4eb36a742a7926f85f12a0d3b8b3e9c0a/src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs#L12-L28

Using `Utf8JsonReader` and grabbing the unmodified/unescaped `ValueSpan` for the comparison would be a good solution. Alternatively, do direct string comparisons and match the whole input to the expected output json file.

cc @nguerrera, @GrabYourPitchforks, @wtgodbe, @ericstj, @tmds, @dagood

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Utf8JsonWriter.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Utf8JsonWriter.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.IO;
+using System.Text.Encodings.Web;
 using System.Text.Json;
 
 namespace Microsoft.Extensions.DependencyModel
@@ -22,7 +23,11 @@ namespace Microsoft.Extensions.DependencyModel
             }
             using (var bufferWriter = new ArrayBufferWriter())
             {
-                var options = new JsonWriterOptions { Indented = true };
+                // Custom encoder is required to fix https://github.com/dotnet/core-setup/issues/7137
+                // Since the JSON is only written to a file that is read by the SDK (and not transmitted over the wire),
+                // it is safe to skip escaping certain characters in this scenario 
+                // (that would otherwise be escaped, by default, as part of defense-in-depth, such as +).
+                var options = new JsonWriterOptions { Indented = true, Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping };
                 var jsonWriter = new Utf8JsonWriter(bufferWriter, options);
                 WriteCore(context, new UnifiedJsonWriter(jsonWriter));
                 bufferWriter.CopyTo(stream);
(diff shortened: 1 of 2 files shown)


src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Utf8JsonWriter.cs
```diff
@@ -22,7 +23,11 @@ public void Write(DependencyContext context, Stream stream)
             }
             using (var bufferWriter = new ArrayBufferWriter())
             {
-                var options = new JsonWriterOptions { Indented = true };
+                // Custom encoder is required to fix https://github.com/dotnet/core-setup/issues/7137
+                // Since the JSON is only written to a file that is read by the SDK (and not transmitted over the wire),
+                // it is safe to skip escaping certain characters in this scenario 
+                // (that would otherwise be escaped, by default, as part of defense-in-depth, such as +).
+                var options = new JsonWriterOptions { Indented = true, Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping };
```
ericstj: Is this any more unsafe than what Json.Net did before?

ahsonkhan:
Any feedback, or is this good to merge?

src/managed/Microsoft.Extensions.DependencyModel/DependencyContextWriter.Utf8JsonWriter.cs
ahsonkhan: Nope.

ahsonkhan:
Picked up for the release/3.0 branch: https://github.com/dotnet/core-setup/pull/7257

Issue #7253, opened 2019-07-17
[dev/arcade-migration] Package content diffs after Arcade migration


dagood:
I plan to fix these:

`Microsoft.NETCore.App.versions.txt` no longer has Core-Setup hash, and is missing CoreCLR (in the runtime package txt files).

Commit hash no longer appears in pkgproj nupkg descriptions. (The managed projects now have a `repository` element containing this information, but pkgproj-based projects don't.)

`Microsoft.NETCore.App.Internal` (legacy netcoreapp framework package) has some differences in `build/netcoreapp3.0/`:
  * It doesn't include `Microsoft.NETCore.App.Internal.props` or the platform manifest. With a trivial fix to put these in the nupkg, I see these further diffs:
  * `PackageConflictPreferredPackages` property is now empty (just a `;`).
  * `_Microsoft_NETCore_App_Internal_Version` is renamed `_Microsoft_NETCore_App_Version`
  * `PackageConflictPlatformManifests` points to `netcoreapp.PlatformManifest.txt` (the original name of the file, not the name it has in the package).
  * *These are legacy packages used for consumption in tests. This difference doesn't affect Core-Setup's tests. I'm on the fence about fixing it (vs leaving the files out entirely) but I want to reduce the migration risk on CoreCLR and ASP.NET Core.*

Shared frameworks:
* WindowsDesktop shared framework has some files with Microsoft.NETCore.App naming that should be Microsoft.WindowsDesktop.App.
* Both shared frameworks include a reference to `Microsoft.DotNet.Build.Tasks.Packaging` in their deps json files.
* `.version` file has no commit hash.

---

I don't plan to fix these before merging the migration:

Platform manifests have symbol files in them now, in general.
  * *I believe this is just a little useless data that should be trimmed for size, but has no effect.*

dagood:
Fixed by https://github.com/dotnet/core-setup/pull/7259

PR #7259, opened 2019-07-17
[dev/arcade-migration] Fix package and shared framework diffs

dagood:
Fixes https://github.com/dotnet/core-setup/issues/7253

### DIFF
--- a/src/pkg/packaging-tools/framework.dependency.targets
+++ b/src/pkg/packaging-tools/framework.dependency.targets
@@ -144,6 +144,12 @@
       <PermitDllAndExeFilesLackingFileVersion>true</PermitDllAndExeFilesLackingFileVersion>
     </PropertyGroup>
 
+    <!--
+      Generate the props file using placeholders for __PackageId__ and __PreferredPackages__. Each
+      pkgproj that depends on this depproj needs to fill it in with unique info. The depproj's copy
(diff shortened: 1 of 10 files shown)


src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -143,11 +145,11 @@
     </PropertyGroup>
 
     <GenerateFileVersionProps Files="@(SharedFrameworkRuntimeFiles)"
-                              PackageId="$(FrameworkPackageName)"
+                              PackageId="__PackageId__"
```
dseefeld: Why these items are now hardcoded here?

dagood: They aren't, this is a templating token that's replaced in the pkgproj. The commit + commit message might shed some light: https://github.com/dotnet/core-setup/pull/7259/commits/cc3c81b6caec5947f03f02ecb306610236e82fee.

dseefeld: Doh!  If I would have read a little further, I would see where the replacement happens.  Thanks.

dagood: Would actually be a good place for a comment, on second thought. It's only in context in this PR, not when people read the file itself later.

PR #7272, opened 2019-07-18
yaml updates for building snaps

leecow:
yaml updates for building snaps.


vivmishra:
Which yaml file is to be used for 3.0 Preview? 

- dotnet-sdk-3/snap/snapcraft.yaml
- dotnet-sdk/snap/snapcraft.yaml

PR #7278, opened 2019-07-18
update .gitignore for snap builds

leecow:


.gitignore
```diff
@@ -274,3 +274,9 @@ config.ps1
 # VS generated files
 launchSettings.json
 
+# Snapcraft files
+.snapcraft
+*.snap
+parts/
+prime/
+stage/
```
dagood: \<nit> Should keep a newline at the end to be nice.

leecow: added

PR #7288, opened 2019-07-19
[master] Merge Arcade SDK migration

dagood:
Merges the migration from BuildTools to the Arcade SDK into `master`.

I had to make some changes to handle the 3.0 => 5.0 update, see the last three commits for that. (Resolved merge conflicts and made tests work despite the 3.0 SDK not supporting the 5.0 runtime.)

I believe all scenarios from the "Core-setup poll - what tools do you use?" mail thread are covered, with the exception of a `clean.cmd` which Arcade doesn't have underlying support for: https://github.com/dotnet/arcade/issues/906. (Just use `git clean -xdf` or `rm -rf artifacts` depending on what you want to clean up.)

A vanilla Visual Studio instance now works to run tests, see https://github.com/dotnet/core-setup/blob/3b8cebc9528342972f0229314aa375f9b09d0038/Documentation/building/Running-Tests-In-VisualStudio-IDE.md for more info. `dotnet vstest ...` also works without having to set up env vars due to the same underlying change.

After merging this, I'll be working on merging it into `release/3.0` as well.


dagood:
Current conflicts are trivial--version update conflicts. I'll resolve during my PR merge to avoid fighting against the stream of auto-update PRs coming in.

Directory.Build.props
```diff
@@ -137,12 +137,7 @@
     <DebugType Condition="'$(DebugType)' == ''">Portable</DebugType>
   </PropertyGroup>
 
-  <!--
-    TODO: (arcade) Enable sourcelink.
-    Always disable sourcelink for now:
-    error MSB4057: The target "InitializeSourceControlInformationFromSourceControlManager" does not exist in the project. [/work/src/corehost/build.proj]
-  -->
-  <PropertyGroup Condition="'$(DisableSourceLink)' == 'true' or true">
+  <PropertyGroup Condition="'$(DisableSourceLink)' == 'true'">
```
dseefeld: Where does DisableSourceLink get set?

dagood: By the user passing in a global property, there's no scenario for it normally.

src/pkg/Directory.Build.targets
```diff
@@ -10,7 +10,11 @@
   -->
   <Target Name="GetInstallerBrandingNames"
           DependsOnTargets="GetProductVersions">
-    <PropertyGroup Condition="'$(PreReleaseVersionLabel)' != '' and $(PreReleaseVersionLabel.StartsWith('preview'))">
+    <PropertyGroup
```
dseefeld: Why have this special case for preview rather than updating both PreReleaseVersionLabel and ReleaseBrandSuffix for preview release in eng/Versions.props?

dagood: This logic let me remove the `previewX` `Preview X` duplication in `eng/Versions.props`, and I considered making it also work for `alphaX` `Preview X`, but for expedience I'm going to let the `alpha1` vs. `Alpha 1` duplication live on for now.

Ultimately, I want to make this handle `[a-z]+[0-9]+` to get rid of the duplication and allow anything.

PR #7289, opened 2019-07-19
Use access(2) to implement pal::file_exists()

lpereira:
access(2) doesn't have to validate and copy the stat buffer; it should be
slightly cheaper.


lpereira:
I had forgotten I had sent this patch already in #6814. I'm merging that instead, as it was already approved. Thanks for the review, @janvorli!

Issue #7218, opened 2019-07-15
Add framework list metadata to prevent WPF theme assemblies from being auto-referenced


nguerrera:
The targeting pack framework list for Microsoft.WindowsDesktop.App should add ReferencedByDefault="false" to the following assemblies:

PresentationFramework.Aero
PresentationFramework.Aero2,
PresentationFramework.AeroLite
PresentationFramework.Classic
PresentationFramework.Luna
PresentationFramework.Royale

See https://github.com/dotnet/sdk/issues/3265 and dotnet/designs#81

cc @dagood

dagood:
How does this differ from having no `Profile` assignment?

Maybe my question is how the "Integration" assemblies, the ones with no profile per https://github.com/dotnet/cli/issues/10536#issuecomment-488871926, end up being referenced, vs. how these are referenced.

nguerrera:
Replied here: https://github.com/dotnet/core-setup/issues/7344#issuecomment-514354724

dagood:
Done by https://github.com/dotnet/core-setup/pull/7346 for preview 8. (Missed closing this issue earlier.)

PR #7320, opened 2019-07-22
Add ReferencedByDefault=false to theme assembiles in WindowsDesktop framework list

dagood:
For https://github.com/dotnet/core-setup/issues/7218. Adds support for `ReferencedByDefault` to the `FrameworkListFileProfile` items, now called `FrameworkListFileClass` items.

I considered implementing this with some pattern matching (anything that begins with `PresentationFramework.` and isn't `PresentationFramework.dll`) but this profile list is hand-edited and I don't think the added complexity would be worthwhile.

/cc @nguerrera

### DIFF
--- a/src/pkg/projects/Directory.Build.targets
+++ b/src/pkg/projects/Directory.Build.targets
@@ -297,7 +297,7 @@
 
     <CreateFrameworkListFile
       Files="@(File)"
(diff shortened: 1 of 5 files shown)


nguerrera:
Thanks. FYI, @dsplaisted. We still have to make the change on our side this week.

Issue #7290, opened 2019-07-19
Fix up list of Windows compatibility pack libraries included in WindowsDesktop targeting pack and runtime


dagood:
There are some bits of the compat pack that are very commonly used in ports to .NET Core WindowsDesktop, and we'd like to include in the WD targeting pack (and runtime) for ease of use.

## Expose in Desktop SDK

Assembly | Justification
-- | --
Microsoft.Win32.Registry.AccessControl | ACLs makes sense on Windows
System.Diagnostics.EventLog |  
System.Diagnostics.PerformanceCounter |  
System.DirectoryServices |  
System.IO.FileSystem.AccessControl |  
System.IO.Pipes.AccessControl | ACLs makes sense on Windows
System.Threading.AccessControl | ACLs makes sense on Windows

## Do not expose (remove)

Assembly | Justification
-- | --
System.Security.Permissions | Legacy tech

dagood:
> System.IO.Pipes.AccessControl

The situation here looks a little strange, so I want to confirm. New CoreFX builds aren't producing a package for this library, you can see this in the 3.0 preview6 Windows compat pack dependencies, for example:

https://www.nuget.org/packages/Microsoft.Windows.Compatibility/3.0.0-preview6.19303.8

```
...
System.IO.Packaging (>= 4.6.0-preview6.19303.8)
System.IO.Pipes.AccessControl (>= 4.5.1)
System.IO.Ports (>= 4.6.0-preview6.19303.8)
...
```

@ericstj The assembly is currently shipped "runtime only" in NETCoreApp. The other AccessControl bits have a freshly built nupkg available for me to use for refs in WindowsDesktop. Is CoreFX missing packaging for `System.IO.Pipes.AccessControl`, or has it just not needed any changes and I should use 4.5.1?

ericstj:
The package for System.IO.Pipes.AccessControl was removed with https://github.com/dotnet/corefx/commit/4e76a6c02552422ce07967f4249acad3b71b80ef.  You can use the 4.5.1 package.  What happens WRT implementation when you do this? So long as we omit it, should be fine.  @JeremyKuhne @carlossanlop

dagood:
[...]

PR #7324, opened 2019-07-22
Fix up the set of Windows compatibility libraries exposed in WindowsDesktop

dagood:
For https://github.com/dotnet/core-setup/issues/7290.

This PR assumes that we should use `System.IO.Pipes.AccessControl` `4.5.1`. (https://github.com/dotnet/core-setup/issues/7290#issuecomment-513926657)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -88,6 +88,10 @@
     <MicrosoftDiaSymReaderNativePackageVersion>1.7.0</MicrosoftDiaSymReaderNativePackageVersion>
     <!-- Infrastructure and test-only. -->
     <MicrosoftSourceLinkVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkVersion>
+    <!-- TODO: Sort these into the above. These are out of the way at the moment to make it easier to resolve conflicts. -->
+    <MicrosoftWin32RegistryAccessControlVersion>5.0.0-alpha1.19372.3</MicrosoftWin32RegistryAccessControlVersion>
+    <SystemDiagnosticsPerformanceCounter>5.0.0-alpha1.19372.3</SystemDiagnosticsPerformanceCounter>
+    <SystemIOPipesAccessControl>4.5.1</SystemIOPipesAccessControl>
   </PropertyGroup>
   <!--Package names-->
   <PropertyGroup>
--- a/src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
+++ b/src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
@@ -53,11 +53,21 @@
     <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="DirectWriteForwarder" />
   </ItemGroup>
 
-  <Target Name="GetNETCoreAppIgnoredReference" BeforeTargets="VerifyClosure">
+  <Target Name="GetNETCoreAppIgnoredReference"
+          BeforeTargets="VerifyClosure">
     <MSBuild
+      Condition="'$(PackageTargetRuntime)' == ''"
       Projects="../../netcoreapp/src/netcoreapp.depproj"
       Properties="Configuration=netcoreapp"
       Targets="GetReferenceFilenames">
+      <Output TaskParameter="TargetOutputs" ItemName="IgnoredReference" />
+    </MSBuild>
+
+    <MSBuild
+      Condition="'$(PackageTargetRuntime)' != ''"
+      Projects="../../netcoreapp/src/netcoreapp.depproj"
+      Properties="Configuration=netcoreapp"
+      Targets="GetReferenceCopyLocalPathsFilenames">
       <Output TaskParameter="TargetOutputs" ItemName="IgnoredReference" />
     </MSBuild>
   </Target>
--- a/src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
+++ b/src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
@@ -51,9 +51,6 @@
 
     <!-- We don't have a ref assembly for DirectWriteForwarder -->
     <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="DirectWriteForwarder" />
-
-    <!-- We want to intentionally exclude the ref assembly for System.Security.Permissions. -->
-    <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="System.Security.Permissions" />
   </ItemGroup>
 
   <Target Name="GetNETCoreAppIgnoredReference" BeforeTargets="VerifyClosure">
--- a/src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
+++ b/src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
@@ -51,6 +51,9 @@
 
     <!-- We don't have a ref assembly for DirectWriteForwarder -->
     <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="DirectWriteForwarder" />
+
+    <!-- We want to intentionally exclude the ref assembly for System.Security.Permissions. -->
+    <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="System.Security.Permissions" />
   </ItemGroup>
 
   <Target Name="GetNETCoreAppIgnoredReference" BeforeTargets="VerifyClosure">
--- a/src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
+++ b/src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
@@ -16,7 +16,6 @@
     <PackageReference Include="System.Drawing.Common" Version="$(SystemDrawingCommonVersion)" />
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="$(SystemIOFileSystemAccessControlVersion)" />
     <PackageReference Include="System.IO.Packaging" Version="$(SystemIOPackagingVersion)" />
-    <PackageReference Include="System.IO.Pipes.AccessControl" Version="$(SystemIOPipesAccessControl)" />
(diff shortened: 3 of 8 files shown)


eng/Versions.props
```diff
@@ -88,6 +88,10 @@
     <MicrosoftDiaSymReaderNativePackageVersion>1.7.0</MicrosoftDiaSymReaderNativePackageVersion>
     <!-- Infrastructure and test-only. -->
     <MicrosoftSourceLinkVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkVersion>
+    <!-- TODO: Sort these into the above. These are out of the way at the moment to make it easier to resolve conflicts. -->
```
dagood: I'll do these with my merge commit. (If this PR is the way to go.)

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
```diff
@@ -5,28 +5,31 @@
...
+    <PackageReference Include="System.Threading.AccessControl" Version="$(SystemThreadingAccessControlVersion)" />
     <PackageReference Include="System.Windows.Extensions" Version="$(SystemWindowsExtensionsPackageVersion)" />
   </ItemGroup>
 
   <ItemGroup>
-    <RuntimeOnlyPackageReference Include="System.Diagnostics.EventLog" Version="$(SystemDiagnosticsEventLogVersion)" />
-    <RuntimeOnlyPackageReference Include="System.DirectoryServices" Version="$(SystemDirectoryServicesVersion)" />
-    <RuntimeOnlyPackageReference Include="System.IO.FileSystem.AccessControl" Version="$(SystemIOFileSystemAccessControlVersion)" />
-    <RuntimeOnlyPackageReference Include="System.Threading.AccessControl" Version="$(SystemThreadingAccessControlVersion)" />
+    <RuntimeOnlyPackageReference Include="System.Security.Permissions" Version="$(SystemSecurityPermissionsVersion)" />
```
dagood: I removed this entirely with my first commit, but that isn't right: this package is depended on by a few of the others, so the ref assembly was pulled in anyway. This caused CI failure when the ref didn't have `Profile` classification. I have to include this package reference to mark the exclusion.

src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
```diff
@@ -51,6 +51,9 @@
 
     <!-- We don't have a ref assembly for DirectWriteForwarder -->
     <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="DirectWriteForwarder" />
+
+    <!-- We want to intentionally exclude the ref assembly for System.Security.Permissions. -->
+    <IgnoredReference Condition="'$(PackageTargetRuntime)' == ''" Include="System.Security.Permissions" />
```
ericstj: Careful here: who's referencing this?  I would have expected that reference to be gone.  We should wait to remove it until it is.

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
```diff
@@ -5,28 +5,34 @@
...
+    <PackageReference Include="System.Threading.AccessControl" Version="$(SystemThreadingAccessControlVersion)" />
     <PackageReference Include="System.Windows.Extensions" Version="$(SystemWindowsExtensionsPackageVersion)" />
   </ItemGroup>
 
   <ItemGroup>
-    <RuntimeOnlyPackageReference Include="System.Diagnostics.EventLog" Version="$(SystemDiagnosticsEventLogVersion)" />
-    <RuntimeOnlyPackageReference Include="System.DirectoryServices" Version="$(SystemDirectoryServicesVersion)" />
-    <RuntimeOnlyPackageReference Include="System.IO.FileSystem.AccessControl" Version="$(SystemIOFileSystemAccessControlVersion)" />
-    <RuntimeOnlyPackageReference Include="System.Threading.AccessControl" Version="$(SystemThreadingAccessControlVersion)" />
+    <RefOnlyPackageReference Include="System.IO.Pipes.AccessControl" Version="$(SystemIOPipesAccessControl)" />
```
ericstj: Why did you need to explicitly make it RefOnly?  Can we instead make conflict resolution do the work to remove anything that would conflict with the base shared-framework and lose?

src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
dagood: I see... when I remove the ignore, the list is:

```
Assembly 'System.Xaml' is missing dependency 'System.Security.Permissions'
Assembly 'System.Configuration.ConfigurationManager' is missing dependency 'System.Security.Permissions'
Assembly 'System.Security.Cryptography.Xml' is missing dependency 'System.Security.Permissions'
Assembly 'WindowsBase' is missing dependency 'System.Security.Permissions'
Assembly 'System.Windows.Forms.Design.Editors' is missing dependency 'System.Security.Permissions'
Assembly 'System.Drawing' is missing dependency 'System.Security.Permissions'
Assembly 'System.DirectoryServices' is missing dependency 'System.Security.Permissions'
```

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
dagood: AFAIK we don't have tooling like that yet, I created https://github.com/dotnet/core-setup/issues/7340 to track that. Definitely seems possible.

ericstj: What do you mean by "tooling like that"  It should be built into the SDK.  You just need to set a property that points to the platform manifest you generated in this repository.

ericstj: Scratch that, an item.  Try just setting PackageConflictPlatformManifests item.

src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
ericstj: We need to determine why this stuff is pulling it in.  Dangling type-forwards are OK, actual type references are not.

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
dagood: That's exactly what I mean, wasn't aware of that. :smile: I'll give it a try.

src/pkg/projects/windowsdesktop/pkg/Microsoft.WindowsDesktop.App.pkgproj
dagood: We haven't had an update from WPF for a while, and it looks like there are some relevant changes we don't have yet: https://github.com/dotnet/wpf/compare/ea72094ebe0de0da03c591752559739d6fb67f03...master.

dagood: Ah... `dotnet-wpf-int` isn't producing for 5.0 yet. Filed https://github.com/dotnet/core-setup/issues/7343 to figure out how to get these updated. For now, I'll keep `System.Security.Permissions` so the rest can be mergeable.

ericstj: That's not the only cause for this.  Some of these other dependencies are concerning.  I suspect we'll need to suppress this, since even once everything is fixed WindowsBase will have typeforwards to System.Security.Permission, but we need to examine all the references before suppressing to find out if there is other work to do to safely omit this.

dagood: Filed https://github.com/dotnet/core-setup/issues/7344 to follow up on this, I'll keep this PR to what seems doable before then. Thanks for the explanation, this makes sense. (I don't know what the best way to check the references is to get this done myself.)

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
dagood: Seems to work as expected! Thanks!

I had to change `GetNETCoreAppIgnoredReference` a bit for closure validation: when verifying the runtime package, it ignores netcoreapp *refs*, not *runtimes*, causing [a number of missing reference errors](https://gist.github.com/dagood/2d9ea2ed775edf884e2ff6f3dfc66bf0). I think that was just a bug that happened to rely on the asset duplication we had.

dagood:
I diffed the data files (runtimelist/frameworklist) to get a summary of the results: https://gist.github.com/dagood/5f328e6148acd7cc8ee1afcc6c60f814. All the removals are in the NETCoreApp sfx, other changes look like what we intended.

dagood:
I didn't rerun crossgen in my local build, which looks like it's failing now. Looking into it.

ericstj:
It's due to WindowsBase removal:
```
  Error generating PDB for 'F:\workspace\_work\1\s\artifacts\bin\win-x64.Debug\crossgen\windowsdesktop\runtimes/win-x64/lib/netcoreapp5.0/PresentationFramework.Aero.dll': Could not load type 'System.Windows.DependencyObject' from assembly 'WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
  Microsoft (R) CoreCLR Native Image Generator - Version 4.5.30319.0
  Copyright (c) Microsoft Corporation.  All rights reserved.
```

dagood:
I'm going to apply this PR to `release/3.0` directly to make progress. https://github.com/dotnet/wpf/issues/1351 tracks fixing the WindowsBase conflict resolution problem for 5.0. (Some discussion about what's going on there in the gist: https://gist.github.com/dagood/5f328e6148acd7cc8ee1afcc6c60f814.)

dleeapho:
Should probably move this PR to the new dotnet/windowsdesktop repo.

dagood:
This PR makes changes to some files now in the sharedfx tooling SDK and relies on the fact that netcoreapp and windowsdesktop are both in this repo, so I can't actually move the PR over. However, I opened https://github.com/dotnet/windowsdesktop/issues/7 to track this and I'm closing this PR.

PR #7347, opened 2019-07-23
[release/3.0] Fix post-arcade-migration preview8 build version: lower than pre-arcade

dagood:
Fix for ASP.NET Core auto-upgrade failure: https://github.com/dotnet/core-sdk/pull/3377.

Set `/p:_BuildNumberYY=28` in all official build commands so that we produce `28xxx-xx` versions, which is higher than the last BuildTools-based build, `27919-09`.

This is a nasty workaround: using global properties to prevent the MSBuild script from doing what it intends to do, and even further, `_`-prefixed props are meant to be private.

I don't see any obvious way to bump up the Arcade-generated version number: https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets.

I hope there's a better way I'm missing. If not, we will have to take care to remove this before the first build of preview 9 or it's stuck there as well. (Ideally I'd have a way to apply a workaround that would automatically deactivate in preview 9, but I don't know if it's reasonable to implement that in AzDO yaml.)

Running a mock official build offline to see how it behaves end to end.

/cc @nguerrera

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -4,6 +4,12 @@
     <MajorVersion>3</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
+    <!--
+      TODO: When changing this to preview9, remove /p:_BuildNumberYY=28 workaround in the official
+      build yml files. These were added because we migrated to the Arcade SDK during preview8, and
+      needed to raise the version number to preserve continuity. In preview9 we should remove these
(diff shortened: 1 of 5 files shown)


nguerrera:
Thanks!

dagood:
The mock build publish step is still waiting for test signing, but the artifacts themselves look fine. I'm going to go ahead and merge this to get things unblocked for now.

PR #7359, opened 2019-07-24
[release/3.0] Remove Linux package signing (RPM and Deb packages)

dagood:
This was added during the Arcade SDK migration, but we don't know for sure that signing is working and Core-SDK is unable to validate the signatures. Remove it to unblock dependency flow.

Planning to merge this as soon as validation is green to unblock. See https://github.com/dotnet/core-sdk/pull/3377#issuecomment-514442844.


dagood:
Setback: CI hit an unusual unrelated infra failure. I'm going to go ahead and merge and watch the official build. This is extremely low risk.

Issue #6507, opened 2019-05-22
Add package overrides to targeting packs


dsplaisted:
We should add a text file to the targeting packs that specifies which packages (and versions) should be overridden.  This is used to help optimize conflict resolution.  This information was previously provided in .targets files in the corresponding NuGet packages or in the .NET SDK itself.  (https://github.com/dotnet/sdk/issues/2937 has more details).

The information should go in a `data/PackageOverrides.txt` file with each package and version on a separate line, separated by a pipe.  For example (from ASP.NET Core):

```
Microsoft.Extensions.Caching.Abstractions|3.0.0-preview6.19262.1
Microsoft.Extensions.Caching.Memory|3.0.0-preview6.19262.1
Microsoft.Extensions.Configuration.Abstractions|3.0.0-preview6.19262.1
```

@ericstj can you help us figure out what the overridden packages should be for Microsoft.NETCore.App, Microsoft.WindowsDesktop.App, and NETStandard.Library?  We could just use the lists we used for previous versions, although I would guess there are some new NuGet packages that are also bundled in .NET Core 3, and we might want to override the WindowsCompat packages.

ericstj:
/cc @wtgodbe @eerhardt 

I believe the purpose of this feature is only to include the dead-ended packages which have been absorbed into the framework packages.  For .NETStandard2.1 this would be the shims which were added in 2.1.  NETCoreApp should be the same.  

For the rest I believe we wish to continue letting conflict resolution of file versions decide which versions to include since that is all mechanical and easier for people to reason about.

dagood:
It *looks* like this is a duplicate of https://github.com/dotnet/core-setup/issues/3607, albeit with a different spec. I found the background info there pretty useful, so please correct me if I'm wrong. 😄

> For .NETStandard2.1 this would be the shims which were added in 2.1. NETCoreApp should be the same.

Is it right to interpret this as basically copying the lists in [Microsoft.NET.DefaultPackageConflictOverrides.targets](https://github.com/dotnet/sdk/blob/ec2f3741092f164daeb023e7caa4364c810461e2/src/Tasks/Common/targets/Microsoft.NET.DefaultPackageConflictOverrides.targets) into Core-Setup?

> For the rest I believe we wish to continue letting conflict resolution of file versions decide which versions to include since that is all mechanical (and thus generateable / toolable in the shared framework tooling) and easier for people to reason about.

[...]

PR #7369, opened 2019-07-24
Add targeting pack data/PackageOverrides.txt files

dagood:
Adds `data/PackageOverrides.txt` to the NETCoreApp and NETStandard targeting packs. Content is static and not expected to change.

For https://github.com/dotnet/core-setup/issues/6507.

I'll port to 3.0 immediately after merging to unblock SDK work depending on this.

/cc @dsplaisted @ericstj


dagood:
It's not clear to me if there's potential for this to happen in 3.0+:

> if more package overrides are created and we see a performance degradation again.

This PR didn't add any package overrides, so I didn't see if it makes sense to generate the list. Maybe we can clean up that related issue though. I'll ping it with a question.

PR #7372, opened 2019-07-24
[release/3.0] Fix up the set of Windows compatibility libraries exposed in WindowsDesktop

dagood:
Port of https://github.com/dotnet/core-setup/pull/7324 to `release/3.0`. For https://github.com/dotnet/core-setup/issues/7290.

The WindowsBase version issue is not a problem in `release/3.0`. WindowsDesktop has:

```
WindowsBase.dll|Microsoft.WindowsDesktop.App.Runtime.win-x64|4.0.0.0|4.800.19.37306
```

and NETCoreApp has:

```
WindowsBase.dll|Microsoft.NETCore.App.Runtime.win-x64|4.0.0.0|4.700.19.37208
```

---

After the port, a new error exposed an issue where `netcoreapp.depproj` was building for `win-x86`, which doesn't match the set platform ,`x64`. In `master`, this issue didn't result in an error, maybe because the Arcade SDK version is different. I think it probably caused undesired behavior that just wasn't as easy to see. To fix this, I changed how `PlatformManifestProjectReference` works a bit:

* Before: when `windowsdesktop.depproj` builds `netcoreapp.depproj` *inside* the RID-specific MSBuild call, the forced `RuntimeIdentifier` prop causes the `netcoreapp.depproj` build to fail.
* Fix: I avoid doing this by instead using `OrderProjectReference`, which builds *before* the RID-specific package build, so the build runs with the default `RuntimeIdentifier` and it hits the MSBuild cache.

### DIFF
--- a/src/pkg/packaging-tools/packaging-tools.targets
+++ b/src/pkg/packaging-tools/packaging-tools.targets
@@ -136,6 +136,23 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="ResolvePlatformManifestProjectReferences"
+          Condition="'@(PlatformManifestProjectReference)' != ''"
+          BeforeTargets="ResolveProjectReferences">
+    <MSBuild
+      Projects="@(PlatformManifestProjectReference)"
+      Targets="GetDataFiles"
+      SkipNonexistentTargets="true">
+      <Output TaskParameter="TargetOutputs" ItemName="ProjectReferenceDataFile" />
+    </MSBuild>
+
+    <ItemGroup>
+      <PackageConflictPlatformManifests
+        Include="@(ProjectReferenceDataFile)"
+        Condition="'%(ProjectReferenceDataFile.PlatformManifestFile)' == 'true'" />
+    </ItemGroup>
+  </Target>
+
   <!--
     By default, shared frameworks are generated based on the package. This can be overridden by the
     project, and in the future, the runtime pack is likely to be used to assemble the sfx.
--- a/src/pkg/packaging-tools/packaging-tools.targets
+++ b/src/pkg/packaging-tools/packaging-tools.targets
@@ -147,16 +147,20 @@
     </MSBuild>
 
     <ItemGroup>
-      <!--
-        Add the platform manifest, if it exists. It may not exist if this is, for example, an x64
-        GetFilesToPackage build running under TargetArchitecture=x86.
-      -->
-      <PackageConflictPlatformManifests
+      <ProjectReferencePlatformManifestFile
         Include="@(ProjectReferenceDataFile)"
-        Condition="
-          '%(ProjectReferenceDataFile.PlatformManifestFile)' == 'true' and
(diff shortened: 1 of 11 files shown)


ericstj:
Looks to me like something is still going wrong with fetching the platform manifest.  Builds are failing because x86 isn't finding x64 manifest, and vice-versa.

src/pkg/packaging-tools/packaging-tools.targets
```diff
@@ -147,9 +147,15 @@
     </MSBuild>
 
     <ItemGroup>
+      <!--
+        Add the platform manifest, if it exists. It may not exist if this is, for example, an x64
+        GetFilesToPackage build running under TargetArchitecture=x86.
+      -->
```
ericstj: Do you also want to ensure you get at least one?

dagood: Yeah... not really happy with this fix. I'm looking if I can make this target only run where it should rather than making it flexible. (If I can find the right place to put the check to make sure one's found, then I imagine I can just get this target to run in that situation.)

dagood: I think I misunderstood, I need to clarify--this target only ever expects to find 1 or 0, the issue is that it's finding the place a platform manifest *would* exist if the other arch were built. This exists check would make sure it doesn't try to use a platform manifest that wasn't built, but it runs the risk of not erroring out if this target fails to find 1 when it should have.

ericstj: I see, I had expected this to only run 1:1 on the cases where we needed to locate the platform manifest.

dagood: Realized I was looking at this the wrong way: the platform manifest is for all platforms, not just x86 or x64 in particular, so the path was wrong and 1 is always expected even with the broad `BeforeTargets="ResolveProjectReferences"`. Fixing it up in this direction (and some related issues I iterated on to get the sfx and runtime pack how we expect).

dagood:
List diffs: https://gist.github.com/dagood/c13bf4309bab2a76da0d066f04840738

Same as last time, but:

* `WindowsBase.dll` isn't removed. 🎉 

* `System.Drawing.dll` isn't removed.
   * I think also 🎉 for the same reason, but it didn't come up before.
      ```
      System.Drawing.dll|Microsoft.NETCore.App.Runtime.win-x86|4.0.0.0|4.700.19.37208
      System.Drawing.dll|Microsoft.WindowsDesktop.App.Runtime.win-x86|4.0.0.0|4.800.19.37302
      ```

dagood:
Reference for the WinForms version of `System.Drawing.dll` being intended to override CoreFX's, indicating that the above change is good: https://github.com/dotnet/winforms/issues/407.

dagood:
Merged by https://github.com/dotnet/core-setup/commit/5ebe41990870f4e10c75b4696a72efbf36b95d52. (I resolved the TODOs in that commit.)

PR #7389, opened 2019-07-25
[release/3.0] Fix release/3.0 official build Latest channel

dagood:
I didn't account for the latest channel branding change in the Arcade migration, so both `release/3.0` and `master` are copying to the `master` latest channel and the 3.0 readme column is out of date. This PR fixes it.


dagood:
FYI @hoyosjs, thanks for letting me know about this problem. 🙂

Issue #7396, opened 2019-07-25
Build command won't work on preview visual studio versions


tlakollo:
## Steps to reproduce
Install Visual Studio 2019 Preview, run the build.cmd in core-setup and it will fail with error:
Error: Visual Studio 2017 or 2019 required
Please see https://github.com/dotnet/core-setup/tree/master/Documentation/building/windows-instructions.md for build instructions.
Although clearly Visual Studio is installed, the code that checks for installation path only checks for released versions, with the following code snippet:

```
if exist %_VSWHERE% (
  for /f "usebackq tokens=*" %%i in (`%_VSWHERE% -latest -property installationPath`) do set _VSCOMNTOOLS=%%i\Common7\Tools
)
if not exist "%_VSCOMNTOOLS%" set _VSCOMNTOOLS=%VS140COMNTOOLS%
if not exist "%_VSCOMNTOOLS%" goto :MissingVersion
```
## Expected  behavior
A -Prerelease tag should be included in case we fail to locate the Visual Studio release install path, then if both fail we can go to the MissingVersion tag. 

cc: @dagood

PR #7398, opened 2019-07-25
Support Visual Studio Preview Install Path discovery at Build

tlakollo:
Included changes to support Install path discovery at build in core-setup
Issue #7396


dagood:
CoreFX simply allows prereleases, on the other hand: https://github.com/dotnet/corefx/blob/a71f32dc4c0df15c76a6ab2c2cd6120db7932602/src/Native/build-native.cmd#L54.

PR #7424, opened 2019-07-26
[release/3.0] Add missing Microsoft.Win32.Registry.AccessControl Version.Details.xml entry

dagood:
This Dependency entry wasn't added when the version prop was added in https://github.com/dotnet/core-setup/pull/7372.

Quick fix to make sure Maestro++ upgrades it along with the other packages.

/cc @MichaelSimons


dagood:
Merging: this file is not used by the build.

MichaelSimons:
LGTM

PR #7436, opened 2019-07-26
[release/3.0] Fix PlatformManifestProjectReference ordering to fix flakiness

dagood:
Props file import order meant `OrderProjectReference Include="@(PlatformManifestProjectReference)"` was being evaluated before PlatformManifestProjectReference was set. Solve this by moving the include to a new GetDefaultOrderProjectReferences target so the static prop order doesn't matter.

Fixes Windows job CI flakiness introduced with https://github.com/dotnet/core-setup/pull/7372.

An alternative would be to change where the `PlatformManifestProjectReference` is defined to put it before it's needed, but I find relying on static order makes the build harder to understand and debug, and less flexible.

Somehow this didn't affect the initial PR... a lot of (all?) other 3.0 PRs are currently hitting this issue in CI, though.

### DIFF
--- a/src/pkg/packaging-tools/framework.dependency.targets
+++ b/src/pkg/packaging-tools/framework.dependency.targets
@@ -468,7 +468,8 @@
           DependsOnTargets="ResolveReferences"
           Returns="@(ReferenceCopyLocalPaths -> '%(Filename)')" />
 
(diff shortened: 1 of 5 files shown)


dagood:
Merging to unblock CI.

Issue #7412, opened 2019-07-26
InvalidCastException and ArgumentException in Microsoft.Extensions.DependencyModel Preview 7


albahari:
I've just received two different bug reports from LINQPad 6 users since upgrading Microsoft.Extensions.DependencyModel from Preview 4 to Preview 7 yesterday.

The exceptions are thrown when calling DependencyContextJsonReader.Read(Stream stream).

I can't give you the .deps.json files that caused the exceptions because the users in question haven't included their contact details, however, I can provide the full stack trace for each. I believe it's unlikely that the .deps.json files are corrupt, but can't rule this out. These files will almost certainly have been generated by Visual Studio / MSBuild.

For now, I'm reverting to preview4 of Microsoft.Extensions.DependencyModel so that users can continue to reference .NET Core assemblies reliably in LINQPad 6. However, if you're unable to determine the cause I can consider returning to preview 7 and waiting for a user to report more information.

First, the InvalidCastException:

```
InvalidCastException - Unable to cast object of type 'System.Int32' to type 'System.Int64'.
Source=Microsoft.Extensions.DependencyModel

   at Microsoft.Extensions.DependencyModel.UnifiedJsonReader.CreateUnexpectedException(Utf8JsonReader& reader, String expected)
   at Microsoft.Extensions.DependencyModel.UnifiedJsonReader.ReadStringArray()
   at Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.ReadCompilationOptions(UnifiedJsonReader& reader)
   at Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.ReadCore(UnifiedJsonReader reader)
   at Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.Read(Utf8JsonReader jsonReader)
   at Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.Read(Stream stream)
```

Second, the ArgumentException:

```
ArgumentException - version
Source=Microsoft.Extensions.DependencyModel

   at Microsoft.Extensions.DependencyModel.Dependency..ctor(String name, String version)
   at Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.ReadTargetLibraryDependencies(UnifiedJsonReader& reader)
   at Microsoft.Extensions.DependencyModel.DependencyContextJsonReader.ReadTargetLibrary(UnifiedJsonReader& reader, String targetLibraryName)
[...]

PR #7510, opened 2019-07-30
Fix how we determine line/column from a JsonReaderState

vitek-karas:
Previously these fields were public and so simple `.GetField("name")` worked. Now they are internal, so we need to use the overload which takes `BindingFlags` and ask for private reflection.

The manifestation of this bug is that instead of throwing `FormatException` with details about what is wrong, we throw `InvalidCastException` with cryptic message (can't cast `Int32` to `Int64`).

Related to #7412

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Reflection;
 using System.Text.Json;
 
 namespace Microsoft.Extensions.DependencyModel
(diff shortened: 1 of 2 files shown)


src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
```diff
@@ -148,8 +149,8 @@ private static Exception CreateUnexpectedException(ref Utf8JsonReader reader, st
         {
             // Replace with public API once https://github.com/dotnet/corefx/issues/34768 is fixed
             object boxedState = reader.CurrentState;
-            long lineNumber = (long)(typeof(JsonReaderState).GetField("_lineNumber")?.GetValue(boxedState) ?? -1);
-            long bytePositionInLine = (long)(typeof(JsonReaderState).GetField("_bytePositionInLine")?.GetValue(boxedState) ?? -1);
+            long lineNumber = (long)(typeof(JsonReaderState).GetField("_lineNumber", BindingFlags.Instance | BindingFlags.NonPublic)?.GetValue(boxedState) ?? -1);
```
ahsonkhan: > Previously these fields were public

They were never public but since we were previously referencing the source package, we had access to internal fields (we now depend on the regular binary package: https://github.com/dotnet/core-setup/pull/6937, so encapsulation works as intended).
https://github.com/dotnet/corefx/blob/f84927d4a85444e63ede2da38a49eff7116790db/src/System.Text.Json/src/System/Text/Json/Reader/JsonReaderState.cs#L19-L20

In which case, I am wondering, did this work as intended before? cc @ericstj 
https://github.com/dotnet/core-setup/blob/2ddb6f1effeb875ebcef8a7537e5befa4aeca8ea/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs#L151-L152

If not, did the `.GetField(...).GetValue(...)` call just silently fail?

src/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonReaderTest.cs
```diff
@@ -697,6 +697,20 @@ public void ReadsCompilationOptions()
             context.CompilationOptions.WarningsAsErrors.Should().Be(true);
         }
 
+        [Fact]
+        public void FailsToReadInvalidDefines()
+        {
+            Assert.Throws<FormatException>(() => Read(
```
ahsonkhan: nit: Consider verifying that the expected line number and byte position in line are actually showing up in the message

src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
vitek-karas: If simple `GetField(name)` would not work the code would always fail (it basically returns null in that case and then tries to cast it to long). So I assume if you reference the package reflection gives you direct access.

PR #7527, opened 2019-07-31
[release/3.0] Fix how we determine line/column from a JsonReaderState (#7510)

vitek-karas:
#### Description
When there's a problem parsing the `.deps.json` in `Microsoft.Extensions.DependencyModel` the code should throw a `FormatException` which includes line/column numbers of the invalid token.
Previously we were using `Newtonsoft.Json` to parse the file which exposes line/column information on the reader as a public API and it worked like this.
In 3.0 we switched to `System.Text.Json` which does not expose line/column information as a public API. To workaround this limitation, we use private reflection to get to the fields which store this information. The way we call the private reflection is wrong though and we end up getting null which then fails to cast to a number. So the end result is that the product throws `InvalidCastException` with no indication of the original error condition.

#### Customer Impact
Hard to diagnose mysterious failures - there's no way for the customer to tell that the `.deps.json` was invalid - other then guessing it based on the callstack of the exception.
If the customer relied on `FormatException` in the application code (catching it for example), a different exception type is thrown now breaking the expected behavior of the application.

#### Regression?
This is a regression in 3.0

#### Risk
Pretty low, the code changes only affect code path which is executed when there's a problem parsing `.deps.json`. This in itself is a relatively rare scenario. The change itself is just a modification of existing logic there are no code flow changes.

Related to #7412

### DIFF
--- a/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
+++ b/src/managed/Microsoft.Extensions.DependencyModel/UnifiedJsonReader.Utf8JsonReader.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Reflection;
 using System.Text.Json;
 
 namespace Microsoft.Extensions.DependencyModel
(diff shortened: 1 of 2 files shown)


sergiy-k:
Approved. This is a regression and a customer reported issue.

sergiy-k:
Please note that you still need to go through the Ask Mode process before you can merge this change. Thank you.

leecow:
Tactics approved - do not merge until 8/5

wtgodbe:
Branch is open, merging

PR #7547, opened 2019-08-02
Stop signing apphost.exe and comhost.dll in nupkg

dagood:
The nupkg is recursively signed, so leaving `apphost.exe` and `comhost.dll` off the list of files to sign is not enough, and they still end up signed. We also need to set up `FileSignInfo` to tell recursive signing what to do.

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -18,7 +18,10 @@
     <!-- Find bundle artifacts, which need multiple stages to fully sign. -->
     <BundleInstallerEngineArtifact Include="$(ArtifactsPackagesDir)**/*engine.exe" />
     <BundleInstallerExeArtifact Include="$(ArtifactsPackagesDir)**/*.exe" />
-  </ItemGroup>
+
+    <!-- apphost and comhost template files are not signed, by design. -->
+    <FileSignInfo Include="apphost.exe;comhost.dll" CertificateName="None" />
+  </ItemGroup>>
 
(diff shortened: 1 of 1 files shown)


dagood:
What breaks when these files are signed? Could you maybe file a issue with details to start with?

dagood:
I'm going to go ahead and merge: this has no affect on PR validation builds, and I want to see it in action in an official build.

vitek-karas:
@elinor-fung can you please file the issue - since you played with it in detail.

The broken scenario is: Build an app with app.exe - try to sign the app.exe with my own cert - fails.

eng/Signing.props
```diff
@@ -18,7 +18,10 @@
     <!-- Find bundle artifacts, which need multiple stages to fully sign. -->
     <BundleInstallerEngineArtifact Include="$(ArtifactsPackagesDir)**/*engine.exe" />
     <BundleInstallerExeArtifact Include="$(ArtifactsPackagesDir)**/*.exe" />
-  </ItemGroup>
+
+    <!-- apphost and comhost template files are not signed, by design. -->
+    <FileSignInfo Include="apphost.exe;comhost.dll" CertificateName="None" />
+  </ItemGroup>>
```
dagood: Extra `>`. Whoops. Fixing.

dagood: Fixed by https://github.com/dotnet/core-setup/pull/7548.

Issue #6370, opened 2019-05-08
WindowDesktop Runtime archives [zips] are not rooted @ 'shared'


johnbeisner:
windowsdesktop-runtime-[version]-win-x86.zip
windowsdesktop-runtime-[version]-win-x64.zip
windowsdesktop-runtime-[version]-win-arm.zip
windowsdesktop-runtime-[version]-win-arm64.zip

When unzipping a WindowDesktop Runtime archive, it unzips to a "Microsoft.WidowsDesktop.App" folder instead of the expected "shared\Microsoft.WidowsDesktop.App" folder as do the SDK, Runtime, Asp.Net Runtime archives.

This should be standardized before the install scripts [install-dotnet.ps1] introduce support for installations of the WindowDesktop Runtime: https://github.com/dotnet/cli/issues/11115

Related: https://github.com/dotnet/cli/pull/11298

johnbeisner:
@dagood what is the status of this work - we have a dependent issue.

dagood:
It's in the backlog. Should be an easy change, but since Core-SDK isn't taking 5.0 and Core repos don't branch for previews, it will take a while for it to be able to flow.

dagood:
Actually, with https://github.com/dotnet/core-sdk/pull/3741 coming in soon, 5.0 will be flowing to the SDK shortly. I went ahead and submitted https://github.com/dotnet/core-setup/pull/7581 to make this change.

dagood:
Done with https://github.com/dotnet/core-setup/pull/7581 and ported to 3.0 with https://github.com/dotnet/core-setup/pull/7584.

PR #7581, opened 2019-08-06
Add root "shared" dir to WindowsDesktop zip

dagood:
Also fix a small bug where both the new and old infrastructure were creating a zip for the NETCoreApp sfx. This wasn't causing a problem in parallel because they are ordered, but it made it unclear to a reader which infra was responsible for the zip.

This will need porting to 3.0.

For https://github.com/dotnet/core-setup/issues/6370.


dagood:
@johnbeisner PTAL and send a review when you're ready to react to the Core-SDK PR break this will/should cause after it builds. (I imagine sometime after https://github.com/dotnet/core-sdk/pull/3741 is merged.)

PR #7584, opened 2019-08-06
[release/3.0] Add root "shared" dir to WindowsDesktop zip

dagood:
#### Description

This PR is to address https://github.com/dotnet/core-setup/issues/6370 in 3.0 Preview 9:

> When unzipping a WindowDesktop Runtime archive, it unzips to a "Microsoft.WidowsDesktop.App" folder instead of the expected "shared\Microsoft.WidowsDesktop.App" folder as do the SDK, Runtime, Asp.Net Runtime archives.
> 
> This should be standardized before the install scripts [install-dotnet.ps1] introduce support for installations of the WindowDesktop Runtime: dotnet/cli#11115

There is also a small, related infra fix to help maintainability. https://github.com/dotnet/core-setup/pull/7581:

> Also fix a small bug where both the new and old infrastructure were creating a zip for the NETCoreApp sfx. This wasn't causing a problem in parallel because they are ordered, but it made it unclear to a reader which infra was responsible for the zip.

#### Customer Impact

Without this change, adding WindowsDesktop support to `dotnet-install.ps1` would require logic in the script specific to each shared framework. This is undesirable, and work to add support is currently considered blocked on standardizing the layout.

`dotnet-install.ps1` support would be useful to install specific versions of WindowsDesktop on dev and CI machines, especially for testing scenarios. (https://github.com/dotnet/cli/issues/11115)

#### Regression?

No.

#### Risk

Low. The infra change is trivial, but dependent repos do need to update to account for it. If they don't, WindowsDesktop may end up in `<dotnet home>/shared/shared/Microsoft.WindowsDesktop.App/...`, breaking installs that use the zip. @johnbeisner is aware of this for Core-SDK.

This change was done in `master` with PR https://github.com/dotnet/core-setup/pull/7581, mitigating the risk. (This is a clean cherry-pick.) As of writing, it hasn't been ingested, and we should make sure that works before moving forward with this PR.

/cc @dleeapho

### DIFF
--- a/src/pkg/packaging-tools/packaging-tools.targets
+++ b/src/pkg/packaging-tools/packaging-tools.targets
@@ -88,9 +88,15 @@
       <CompressedArchiveFile>$(AssetOutputPath)$(InstallerFileNameWithoutExtension)$(CompressedFileExtension)</CompressedArchiveFile>
     </PropertyGroup>
 
-    <!-- By default, a shared framework only has files in shared/, so archive that. -->
+    <!--
+      Set the directory containing contents to include in the archive (zip/tarball). This default
+      means the "shared" dir is inside the archive and it can be extracted directly into the dotnet
+      home dir. This allows the .NET Core Runtime archive to include files outside the "shared" dir,
+      such as the host, resolver, and license files, without special handling vs. WindowsDesktop and
+      ASP.NET Core sharedfx.
+    -->
     <PropertyGroup>
-      <SharedFrameworkArchiveSourceDir Condition="'$(SharedFrameworkArchiveSourceDir)' == ''">$(SharedFrameworkLayoutDir)shared/</SharedFrameworkArchiveSourceDir>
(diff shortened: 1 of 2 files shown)


wtgodbe:
Branch is open, feel free to merge once this is approved by tactics (CC @danmosemsft in case I forget to bring up core-setup tomorrow)

dagood:
> As of writing, it hasn't been ingested, and we should make sure that works before moving forward with this PR.

Status: this hasn't happened, `master` dependency flow is waiting on https://github.com/dotnet/core-sdk/pull/3741.

We don't **need** to wait for this. It just means that the breaking change will require a semi-novel fix in Core-SDK 3.0 to unblock dependency flow.

wtgodbe:
Approved by tactics, merge on CI is green

dagood:
The validation build is green, AzDO just didn't update the GitHub checks. (And the only way to make AzDO retry it is to rerun every single job.) Merging.

Issue #7597, opened 2019-08-07
Publishing large single file executables fails on systems with on-access scan


dasMulli:
## Steps to reproduce

* Use a system with on-access scanning enabled (e.g. Enterprise laptop with McAffee Endpoint Security), no exclusion for `dotnet.exe`
* Create an app with lots of external DLLs and some front-end assets as well (Angular App)
* Publish it as single file self-contained app to an empty directory or from clean state.

## Expected  behavior

Publish works

## Actual behavior

```
C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(801,5): error MSB4018: The "GenerateBundle" task failed unexpectedly. [path\to.csproj]
C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(801,5): error MSB4018: System.IO.IOException: The process cannot access the file 'C:\path\to\built.exe' because it is being used by another process. [path\to.csproj]
C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(801,5): error MSB4018:    at System.IO.FileSystem.OpenHandle(String fullPath, Boolean asDirectory) [path\to.csproj]
C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(801,5): error MSB4018:    at System.IO.FileSystem.SetLastWriteTime(String fullPath, DateTimeOffset time, Boolean asDirectory) [path\to.csprojj]
C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(801,5): error MSB4018:    at System.IO.File.SetLastWriteTimeUtc(String path, DateTime lastWriteTimeUtc) [path\to.csproj]
C:\Program Files\dotnet\sdk\3.0.100-preview7-012821\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Publish.targets(801,5): error MSB4018:    at Microsoft.NET.HostModel.AppHost.HostWriter.SetAsBundle(String appHostPath, Int64 bundleHeaderOffset) [path\to.csproj]
[...]

PR #7617, opened 2019-08-08
HostModel: Retry updates on IO error

swaroop-sridhar:
The HostModel implements services for updating the AppHost DLL
- Write App-dll name
- Set GUI/CUI bit
- Read/Write BundleHeader marker

These updates involve multiple file open/close operations.
An Antivirus scanner may intercept in-between and cause these operations
to fail with IO-Error.

So, this change adds a few retries to the IO operations, similar to
https://github.com/dotnet/core-setup/pull/7135

Fix for #7597

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
@@ -2,10 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.IO.MemoryMappedFiles;
 using System.Text;
+using System.Threading;
 
 namespace Microsoft.NET.HostModel.AppHost
 {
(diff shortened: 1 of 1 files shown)


src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
```diff
@@ -21,6 +21,32 @@ public static class HostWriter
...
+        /// An Antivirus scanner may intercept in-between and lock the file, 
+        /// causing the operations to fail with IO-Error.
+        /// So, the operations are retried a few times on IOException.
+        /// </summary>
+        /// <param name="func">The action to retry on IO-Error</param>
+        private static void RetryOnIOError(Action func)
+        {
+            uint numberOfRetries = 32;
+
+            for(uint i = 1; i <= numberOfRetries; i++)
```
vitek-karas: Nit: add space after `for`.
Similarly below for `catch`.

src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
```diff
@@ -21,6 +21,32 @@ public static class HostWriter
...
+        /// The HostModel implements several services for updating the AppHost DLL.
+        /// These updates involve multiple file open/close operations.
+        /// An Antivirus scanner may intercept in-between and lock the file, 
+        /// causing the operations to fail with IO-Error.
+        /// So, the operations are retried a few times on IOException.
+        /// </summary>
+        /// <param name="func">The action to retry on IO-Error</param>
+        private static void RetryOnIOError(Action func)
+        {
+            uint numberOfRetries = 32;
```
vitek-karas: This might be too low - we would wait at the maximum 6.4 seconds. On spinning drives that's not that long. The extractor waits up to 50 seconds (and several people commented that it feels about right).

swaroop-sridhar: I increased the iteration count to 256, the wait time is now 51.2s

swaroop-sridhar:
>> Also - I assume you tested this with the AV emulator.

Yes I tested with AVEmu (adjusting it for file-type) on a large single-exe.

swaroop-sridhar:
CC: @nguerrera

nguerrera:
Hmm, seems there's now duplication between https://github.com/dotnet/sdk/pull/3500 and this.

cc @peterhuene

swaroop-sridhar:
Yes if we combine both changes, there will be extra retries.

There are two points to note: 
* The fix in the SDK changes only the AppHost DLL update, while this HostModel fix is more generic, and applies single-exe update as well, which is the problem in #7135
* This fix in HoseModel is more fine-grained -- it retries individual IO operations (around each open/close operation), rather than whole updates. I think this is better wrt avoiding a race. For example, retries around the HostModel APIs may fail if the AV interjects between file open/close operations in HostModel implementation each time.
* The change in SDK provides customization for retry count; do we need this to be externally visible? 

Suppose we take in both the fixes -- its probably not too bad wrt execution (but it is extra code in two repos). It doesn't impact the usual case, where the update is expected to happen in a single try.
In the worst case, the HostModel code will retry each IO block 512 times with 200ms delay. In the worst case, it throws an exception after 2 minutes. The default retry count in the SDK is 2, so, it will retry the entire operation after 1 sec.

peterhuene:
After discussing this with @swaroop-sridhar, I think the best course of action is to set the default retry count on the task to `0`.  This would ensure that we rely, by default, on the retries internal to `HostWriter`, which are already substantial and we wouldn't want to try them again by default.  It would also mean that users can, for whatever reason, ask the task to retry the whole operation by setting some build properties.

@nguerrera what do you think about that plan of action?

nguerrera:
Seems reasonable.

PR #7638, opened 2019-08-09
[release/3.0] HostWriter: Retry updates on IO error (#7617)

swaroop-sridhar:
# HostWriter: Retry updates on IO error

### Original Commit
This change ports the following commits from master branch: 
https://github.com/dotnet/core-setup/commit/f19be99463d48d2fae98030d26c5f394a18d7df6 (#7617)
https://github.com/dotnet/core-setup/commit/77adbdd3f78f1584b9e4f230fee322efb8ec9e0a (#7665)

### Tracking Issue
#7597

### Customer Scenario 
* This is an app _build time_ failure scenario, not a runtime failure. 
* It was reported by an external customer. 

The customer tried to publish a sizable app as a single-file in an environment that had McCafe Antivirus scan-on-access protection turned on. 

### Problem 
`dotnet publish` crashed because it couldn't modify the AppHost.exe file, since it was opened/locked by the AntiVirus.

### Mitigation
Retry the update to AppHost.exe after waiting a few milliseconds.
512 attempts are done in 200ms intervals.

### Details
The HostModel implements services for updating the AppHost DLL
- Write App-dll name
- Set GUI/CUI bit
- Read/Write BundleHeader marker

These updates involve multiple file open/close operations. An Antivirus scanner may intercept in-between and cause these operations to fail with IO-Error.

So, this change adds a few retries to the IO operations, similar to #7135

### Validation

* Local testing with AntiVirus emulation
* Sent instructions to the customer who reported the issue, requesting them to validate the fix. Awaiting their response.
* Standard lab tests on the PR.

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
@@ -2,10 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.IO.MemoryMappedFiles;
 using System.Text;
+using System.Threading;
 
 namespace Microsoft.NET.HostModel.AppHost
 {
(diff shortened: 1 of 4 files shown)


swaroop-sridhar:
CC: @peterhuene @vitek-karas  @nguerrera

src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
```diff
@@ -21,6 +21,32 @@ public static class HostWriter
...
+        /// The HostModel implements several services for updating the AppHost DLL.
+        /// These updates involve multiple file open/close operations.
+        /// An Antivirus scanner may intercept in-between and lock the file, 
+        /// causing the operations to fail with IO-Error.
+        /// So, the operations are retried a few times on IOException.
+        /// </summary>
+        /// <param name="func">The action to retry on IO-Error</param>
+        private static void RetryOnIOError(Action func)
+        {
+            uint numberOfRetries = 256;
```
AaronRobinsonMSFT: We should match the expressed timing in https://github.com/dotnet/core-setup/pull/7135. Iterations = 500 and sleep is 100ms.

swaroop-sridhar:
## HostModel: Retry ResourceUpdate on Win32 File-Lock error

I added another related commit to the PR:  https://github.com/dotnet/core-setup/commit/77adbdd3f78f1584b9e4f230fee322efb8ec9e0a (#7665) to extend the retry-facility to resource-update operation after failure in Win32 file operations.

For Win32 PE files, the SDK copies resources to the AppHost binary from the intermediate assembly.
This update is performed using native Win32 operations. This change retries the resource update if the operation fails because the file is locked (say because of AntiVirus scan).

Issue #7567, opened 2019-08-05
Update master branch NETStandard.Library.Ref to produce 2.2 (for 5.0)


dagood:
@nguerrera pointed out in https://github.com/dotnet/core-sdk/pull/3741#discussion_r310339695 that core-setup `master` still produces 2.1 `NETStandard.Library.Ref`. dotnet/standard now produces 2.2 from its `master` branch (https://github.com/dotnet/standard/pull/1344), which core-setup should probably match.

A consideration is that we should move this package to dotnet/standard, https://github.com/dotnet/standard/issues/1209, but we don't have the prereqs to do so right now. It might be useful to just do this small amount of work to get it flowing.

/cc @wtgodbe

wtgodbe:
CC @JunTaoLuo @dougbu. @dagood, this is blocking Asp.Net repos since they're consuming `2.1.0-alpha` versions of this package from master, which the SDK automatically rolls forward to the latest on nuget (`preview7`), which doesn't contain some new API that they rely on.

JunTaoLuo:
Would appreciate an eta when available since our master branches are blocked currently.

dagood:
Is changing the package version enough, or do you know if there's an assumption somewhere that the major.minor of the package is the same as the major.minor of the TFM? https://github.com/dotnet/core-setup/pull/7646 is just the package version. If needed, Core-Setup can move stuff into the right place without much effort, but don't want to do that unless needed.

If infra problems clear up, no reason this can't go in today.

nguerrera:
Why are your branches blocked. I saw something about NuGet picking a different older version somewhere but that sounds like the issue with packages disappearing if that’s what the blocker is. If the alpha package is found it should work.

nguerrera:
Please don’t bump the version to 2.2 with 2.1 tfm

nguerrera:
> they're consuming 2.1.0-alpha versions of this package from master, which the SDK automatically rolls forward to the latest on nuget (preview7), which doesn't contain some new API that they rely on.

The SDK does no such roll forward of targeting pack versions. NuGet rolls forward package versions if the exact version is not found in some cases, and warns.

I saw something about packages going missing, are we sure that's not what is happening. Can you share a pointer to a failing build or binlog?

dagood:
Filed https://github.com/dotnet/standard/issues/1421 to get to 2.2 with consistency.

The missing package issues are tracked at https://github.com/dotnet/arcade/issues/3620.

[...]

PR #7646, opened 2019-08-09
Update NETStandard.Library.Ref to 2.2

dagood:
For https://github.com/dotnet/core-setup/issues/7567.

Updates the package version to 2.2. Note that the targeting pack contents are still in `ref/netstandard2.1/`: `NETStandard.Library` has them in `build/netstandard2.1/ref/`.

/cc @JunTaoLuo


dagood:
@nguerrera, will this break something in the 5.0 SDK, that the `NETStandard.Library.Ref` package version will be 2.2 but the TFM inside is `netstandard2.1`?

dagood:
Closing per https://github.com/dotnet/core-setup/issues/7567#issuecomment-519969731, having version mismatch is bad.

PR #7658, opened 2019-08-09
SharedHost always installs to 'ProgramFiles'\dotnet

ghost:
When installing the SharedHost; copy all files to the traditional default install location: 'ProgramFiles'\dotnet even when installing somewhere else.

This is part of the work to allow SDKs, when installing with VS, to install to a different volume when VS is doing the same.

**User Scenerio:**
When a user has installed a 2.1 SDK or similar SDK first so that the PATH environmental variable has 'ProgramFiles'\dotnet in a first position. Then the user installs a 3.0 SDK thru VS to a different volume with VS so that 'VSSharedVolume'\dotnet is in a second position. We want the 3.0 SharedHost to be on the PATH in the first position. Therefore, installing the SharedHost to both 'VSSharedVolume'\dotnet and 'ProgramFiles'\dotnet will achieve this.

Reference: https://github.com/dotnet/core-sdk/issues/2174

@dotnet/dotnet-cli

### DIFF
--- a/src/pkg/packaging-tools/windows/product/product.wxs
+++ b/src/pkg/packaging-tools/windows/product/product.wxs
@@ -49,9 +49,6 @@
     <Directory Id="TARGETDIR" Name="SourceDir">
       <Directory Id="$(var.Program_Files)">
         <Directory Id="DOTNETHOME" Name="dotnet" />
-        <?ifdef DefineDefaultInstallationDir ?>
-          <Directory Id="PROGRAMFILES_DOTNET" Name="dotnet" />
-        <?endif?>
       </Directory>
     </Directory>
   </Fragment>
--- a/src/pkg/packaging-tools/windows/product/product.wxs
+++ b/src/pkg/packaging-tools/windows/product/product.wxs
@@ -49,6 +49,9 @@
     <Directory Id="TARGETDIR" Name="SourceDir">
       <Directory Id="$(var.Program_Files)">
         <Directory Id="DOTNETHOME" Name="dotnet" />
+        <?ifdef DefineDefaultInstallationDir ?>
+          <Directory Id="PROGRAMFILES_DOTNET" Name="dotnet" />
+        <?endif?>
       </Directory>
     </Directory>
   </Fragment>
--- a/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
+++ b/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
@@ -8,7 +8,9 @@
     <ComponentGroup Id="InstallSharedHostandDetectionKeys">
 
       <Component Id="cmpCoreHost" Directory="DOTNETHOME" Guid="{45399BBB-DDA5-4386-A2E9-618FB3C54A18}" >
-        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe" />
+        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe">
+          <CopyFile Id="copyFileCoreHostExe" DestinationProperty="$(var.Program_Files)dotnet" />
+        </File>
         <RegistryKey Root="HKLM" Key="SOFTWARE\dotnet\Setup\InstalledVersions\$(var.Platform)\sharedhost">
           <RegistryValue Action="write" Name="Version" Type="string" Value="$(var.NugetVersion)"/>
         </RegistryKey>
@@ -16,8 +18,12 @@
       </Component>
 
       <Component Id="cmpLicenseFiles" Directory="DOTNETHOME" Guid="{A61CBE5B-1282-4F29-90AD-63597AA2372E}">
-        <File Id="fileLicenseTxt" KeyPath="yes" Source="$(var.HostSrc)\LICENSE.txt" />
-        <File Id="fileThirdPartyNoticesTxt" Source="$(var.HostSrc)\ThirdPartyNotices.txt" />
+        <File Id="fileLicenseTxt" KeyPath="yes" Source="$(var.HostSrc)\LICENSE.txt">
+          <CopyFile Id="copyFileLicenseTxt" DestinationProperty="$(var.Program_Files)dotnet" />
+        </File>
+        <File Id="fileThirdPartyNoticesTxt" Source="$(var.HostSrc)\ThirdPartyNotices.txt">
+          <CopyFile Id="copyFileThirdPartyNoticesTxt" DestinationProperty="$(var.Program_Files)dotnet" />
+        </File>
       </Component>
 
     </ComponentGroup>
--- a/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
+++ b/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
@@ -22,7 +22,11 @@
 
       <Component Id="cmpCoreHost" Directory="DOTNETHOME" Guid="{45399BBB-DDA5-4386-A2E9-618FB3C54A18}" >
         <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe">
-          <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="[$(var.Program_Files)]dotnet" />
+          <?if $(var.Platform)=x86?>
+            <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="[ProgramFilesFolder]dotnet" />
+          <?elseif $(var.Platform)=x64?>
+            <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="[ProgramFiles64Folder]dotnet" />
+          <?endif?>
         </File>
         <RegistryKey Root="HKLM" Key="SOFTWARE\dotnet\Setup\InstalledVersions\$(var.Platform)\sharedhost">
           <RegistryValue Action="write" Name="Version" Type="string" Value="$(var.NugetVersion)"/>
@@ -32,10 +36,18 @@
 
       <Component Id="cmpLicenseFiles" Directory="DOTNETHOME" Guid="{A61CBE5B-1282-4F29-90AD-63597AA2372E}">
         <File Id="fileLicenseTxt" KeyPath="yes" Source="$(var.HostSrc)\LICENSE.txt">
-          <CopyFile Id="copyFileLicenseTxt" DestinationDirectory="[$(var.Program_Files)]dotnet" />
+          <?if $(var.Platform)=x86?>
+            <CopyFile Id="copyFileLicenseTxt" DestinationDirectory="[ProgramFilesFolder]dotnet" />
+          <?elseif $(var.Platform)=x64?>
+            <CopyFile Id="copyFileLicenseTxt" DestinationDirectory="[ProgramFiles64Folder]dotnet" />
+          <?endif?>
         </File>
         <File Id="fileThirdPartyNoticesTxt" Source="$(var.HostSrc)\ThirdPartyNotices.txt">
-          <CopyFile Id="copyFileThirdPartyNoticesTxt" DestinationDirectory="[$(var.Program_Files)]dotnet" />
+          <?if $(var.Platform)=x86?>
+            <CopyFile Id="copyFileThirdPartyNoticesTxt" DestinationDirectory="[ProgramFilesFolder]dotnet" />
+          <?elseif $(var.Platform)=x64?>
+            <CopyFile Id="copyFileThirdPartyNoticesTxt" DestinationDirectory="[ProgramFiles64Folder]dotnet" />
+          <?endif?>
         </File>
       </Component>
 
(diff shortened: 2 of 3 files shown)


joeloff (approved):

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
```diff
@@ -8,16 +8,22 @@
     <ComponentGroup Id="InstallSharedHostandDetectionKeys">
 
       <Component Id="cmpCoreHost" Directory="DOTNETHOME" Guid="{45399BBB-DDA5-4386-A2E9-618FB3C54A18}" >
-        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe" />
+        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe">
+          <CopyFile Id="copyFileCoreHostExe" DestinationProperty="$(var.Program_Files)dotnet" />
```
nguerrera: I assume the system is smart enough to handle the common case where the two destinations are the same?

I don't feel qualified to review this code, but I approve of the change.

dagood: I don't see how `DestinationProperty="$(var.Program_Files)dotnet"` works... `Program_Files` is `ProgramFilesFolder` (or 64 bit) at this point, so `DestinationProperty` would be e.g. `ProgramFilesFolderdotnet`. Then it's evaluated as a property at install-time?

dagood (commented):
Side note, it looks like your git config (or GitHub configuration) isn't right, your account isn't linked to the commit.

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
johnbeisner: I have testing this case: 'the two destinations are the same' and it is handled by the installer gracefully and without error.

dagood: I see `bundle.wxs` doing something similar:

```xml
<Variable Name="DOTNETHOME" Type="string" Value="[$(var.Program_Files)]dotnet" bal:Overridable="no" />
```

The `[]` makes that one make sense to me.

joeloff: yes, when the targets for FileCopy are the same, it only executes once per the documentation:
_If the specified destination directory is the same as the directory containing the original file and the name for the proposed source file is the same as the original, then no action takes place._

joeloff: The $(var) should resolve to [ProgramFilesFolder] (with the square brackets). That in turn should resolve to the built-in property which should have a trailing slash AFAIK, but it would be good to confirm.

dagood: This is the current way the var is defined, the same in the wxi files I found in Core-SDK and ASP.NET Core:

https://github.com/dotnet/core-setup/blob/0e6effe229e0913dcd71bc20d63faac8cd34391b/src/pkg/packaging-tools/windows/variables.wxi#L17

dagood: Depending on the trailing slash seems safe based on this, assuming the rest goes right: https://stackoverflow.com/q/2733628.

johnbeisner: It appears from the log that the property has a trailing slash:
MSI (s) (2C:68) [15:23:57:956]: Dir (target): Key: ProgramFiles64Folder	, Object: C:\Program Files\

johnbeisner:
[MSI74cd7.LOG](https://github.com/dotnet/core-setup/files/3488251/MSI74cd7.LOG)

vitek-karas:
I have no idea about the exact changes (no expert in installer tech), but the intent of the change looks good to me.

Maybe one nit: it would be nice to document this somewhere and in that place also mention what happens if I install on a clean machine into a custom location.  If I understand it correctly we would still install the SharedHost into the program files... I don't see a problem with it, but we should document that it happens and why.

johnbeisner:
@dagood are we good to merge this change?

dagood (changes requested):
It doesn't seem to work in my VM. I set a custom location, which it installs to, but nothing is put in the default location. A `/l*v log.txt` has this, which looks like it could be an error code, but I can't seem to find it:

```
...
Action 12:11:08: DuplicateFiles. Creating duplicate files
Action start 12:11:08: DuplicateFiles.
DuplicateFiles: File: Creating duplicate files,  Directory: ,  Size: 
MSI (s) (10:90) [12:11:08:622]: Note: 1: 2756 2: ProgramFiles64Folderdotnet 
MSI (s) (10:90) [12:11:08:622]: Note: 1: 2756 2: ProgramFiles64Folderdotnet 
MSI (s) (10:90) [12:11:08:622]: Note: 1: 2756 2: ProgramFiles64Folderdotnet 
Action ended 12:11:08: DuplicateFiles. Return value 1.
...
```

https://gist.github.com/dagood/50a9e937a39272758edfff56cf02336d

This seems related to the question I asked earlier--the trailing `\` wasn't my concern: https://github.com/dotnet/core-setup/pull/7658#discussion_r312664737.

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
joeloff: Oh, wait I see what's going on. So the variable is used to resolve the <Directory> However, for Copy file you need to provide the actual folder.

So you want something like DestinationProperty="[ProgramFiles]dotnet". The $(var) reference is a preprocessor variable so it will literally resolve the authoring to DestinationProperty="ProgramFilesdotnet"

joeloff (changes requested):

joeloff (approved):

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
johnbeisner: Fixed.

johnbeisner:
@vitek-karas 
Can you review my documentation.

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
```diff
@@ -7,17 +7,36 @@
...
+
+      On a 'clean' machine when installing into a custom location; the installation behavior is the same.
+      Rationale:
+        1. the legacy SDK can be subsequently installed
+        2. the user runs 'dotnet' commands directly against 'ProgramFiles'\dotnet\dotnet.exe -->
+
       <Component Id="cmpCoreHost" Directory="DOTNETHOME" Guid="{45399BBB-DDA5-4386-A2E9-618FB3C54A18}" >
-        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe" />
+        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe">
+          <CopyFile Id="copyFileCoreHostExe" DestinationProperty="[$(var.Program_Files)]dotnet" />
```
dagood: The change to `DestinationDirectory` is missing.

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
```diff
@@ -7,17 +7,36 @@
...
+      User Scenerio:
+      When a user has installed a 2.1 SDK or similar SDK first so that the PATH environmental variable has 'ProgramFiles'\dotnet in a first position.
+      Then the user installs a 3.0 SDK thru VS to a different volume with VS so that 'VSSharedVolume'\dotnet is in a second position.
+      We want the 3.0 SharedHost to be on the PATH in the first position. Therefore, installing the SharedHost to both 'VSSharedVolume'\dotnet and
+      'ProgramFiles'\dotnet will achieve this.
+
+      On a 'clean' machine when installing into a custom location; the installation behavior is the same.
+      Rationale:
+        1. the legacy SDK can be subsequently installed
+        2. the user runs 'dotnet' commands directly against 'ProgramFiles'\dotnet\dotnet.exe -->
```
dagood: Can you wrap this to something more modest? I go with 100.

On the other hand, I feel like we'd want this in a markdown doc (unfortunately don't know where 😕) so it's easier to point devs confused by the behavior to this. Maybe there's a dotnet/designs doc we can link to from a code comment, something like that?

dagood (changes requested):

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
johnbeisner: Fixed.

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
vitek-karas: The description looks good, but I agree with @dagood that we should document this somewhere we can point external people to- source code isn't really it.

heaths (approved):

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
```diff
@@ -7,17 +7,36 @@
...
+
       <Component Id="cmpCoreHost" Directory="DOTNETHOME" Guid="{45399BBB-DDA5-4386-A2E9-618FB3C54A18}" >
-        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe" />
+        <File Id="fileCoreHostExe" KeyPath="yes" Source="$(var.HostSrc)\dotnet.exe">
+          <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="[$(var.Program_Files)]dotnet" />
```
dagood: Failing to compile in PR validation:

> ...\host.wxs(25): error CNDL0014: The CopyFile/@DestinationDirectory attribute's value, '[ProgramFiles64Folder]dotnet', is not a legal identifier.  Identifiers may contain ASCII characters A-Z, a-z, digits, underscores (_), or periods (.).  Every identifier must begin with either a letter or an underscore.

I missed the importance of the rest of @joeloff's comment:

> You want this in the actual authoring after the preprocessor finishes:
> 
> `<Directory Id="ProgramFilesFolder">`
> `<CopyFile DestinationDirectory="[ProgramFilesFolder]">`
> 
> This will be the resolved value of ProgramFilesFolder

But that doesn't make sense with the error message above, is that meant to be something like this?

```xml
<Directory Id="ProgramFilesFolder">
  <CopyFile DestinationDirectory="dotnet">
```

johnbeisner: I'm going with my first non-elegant solution:          

```
<?if $(var.Platform)=x86?>
            <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="[ProgramFilesFolder]dotnet" />
          <?elseif $(var.Platform)=x64?>
            <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="[ProgramFiles64Folder]dotnet" />
          <?endif?>
```

dagood: That looks like it'll get identical output as before once the preprocessor runs. 😕

johnbeisner: DestinationDirectory should be DestinationProperty

Set this value to a property that will have a value that resolves to the full path of the destination directory. The property does not have to exist in the installer database at creation time; it could be created at installation time by a custom action, on the command line, etc. This attribute cannot be specified in conjunction with DestinationDirectory.

johnbeisner: I needed to pre-define a 'ProgramFiles'\dotnet directory and then use that property in the CopyFiles. I've defined "PROGRAMFILES_DOTNET" conditionally in 'product.wxs' and then we can use it in 'host.wxs'

src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
dagood: Filed https://github.com/dotnet/core-setup/issues/7737 to follow up on this.

src/pkg/packaging-tools/windows/product/product.wxs
```diff
@@ -49,6 +49,9 @@
     <Directory Id="TARGETDIR" Name="SourceDir">
       <Directory Id="$(var.Program_Files)">
         <Directory Id="DOTNETHOME" Name="dotnet" />
+        <?ifdef DefineDefaultInstallationDir ?>
+          <Directory Id="PROGRAMFILES_DOTNET" Name="dotnet" />
+        <?endif?>
```
dagood: I would prefer adding this as an unconditional `Fragment` in `src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs`: this directory doesn't have a clear meaning in context, and it makes this highly shared file more complex. For example, over here:

https://github.com/dotnet/core-setup/blob/b44e5a25b376cfa6e0581cd69af05ce655d72f40/src/pkg/packaging-tools/windows/product/breadcrumbstorefolder.wxs#L28-L37

IMO this is fine to follow up on later, though, for the sake of getting something that we know works merged.

dagood (approved):
Thanks! I haven't tested this version myself, but it makes a lot more sense how it's meant to work.

src/pkg/packaging-tools/windows/product/product.wxs
johnbeisner: @dagood I went ahead and made your recommended change and re-tested. I'll merge when CI is green.

PR #7665, opened 2019-08-10
HostModel: Retry ResourceUpdate on Win32 File-Lock error

swaroop-sridhar:
On Win32 PE files, the SDK copies resources on the AppHost binary
from the intermediate assembly.

This update is performed using native Win32 operations.
This change retries the resource update if the operation fails
because the file is locked (say because of AntiVirus scan).

This change is similar to: https://github.com/dotnet/core-setup/pull/7617
which handles failures in managed code via IOException.

The Retry logic is factored out to the RetryUtil class. Currently the
HostWriter is the only client for the class. But it is made public
because other components (like Bundler) may use it in future.

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
@@ -10,7 +10,7 @@ using System.Threading;
 namespace Microsoft.NET.HostModel
 {
     /// <summary>
-    /// HowtModel library implements several services for updating the AppHost DLL.
+    /// HostModel library implements several services for updating the AppHost DLL.
     /// These updates involve multiple file open/close operations.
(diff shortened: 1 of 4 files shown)


swaroop-sridhar:
@peterhuene  @wli3 please review.
CC: @vitek-karas.

src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
```diff
@@ -0,0 +1,71 @@
...
+using System;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using System.Text;
+using System.Threading;
+
+namespace Microsoft.NET.HostModel
+{
+    /// <summary>
+    /// HowtModel library implements several services for updating the AppHost DLL.
```
peterhuene:
```suggestion
    /// HostModel library implements several services for updating the AppHost DLL.
```

Issue #5645, opened 2019-04-02
Relative path in deps.json is ignored for assemblies from the app


vitek-karas:
The `.deps.json` can specify relative path for any asset. The directory portion of that path is ignored for normal assemblies which are part of the app though.

Repro:
```
dotnet new console
dotnet add package Newtonsoft.Json
dotnet build
```

Now in the `bin/Debug/netcoreapp3.0` create a `sub` directory and move the `Newtonsoft.Json.dll` into it.
Edit the `app.deps.json` to point to `sub/Newtonsoft.Json.dll` like this:

```json
{
  "runtimeTarget": {
    "name": ".NETCoreApp,Version=v3.0",
    "signature": "6596ed3d155c0211dab1fd90ba9d71b725c973ba"
  },
  "compilationOptions": {},
  "targets": {
    ".NETCoreApp,Version=v3.0": {
      "Newtonsoft.Json/12.0.1": {
        "runtime": {
          "sub/Newtonsoft.Json.dll": {
          }
```

Run the app via `dotnet app.dll`.

**Expected:**
The host should be able to find the `Newtonsoft.Json.dll` since it exists in the right relative path.

**Actual:**
The host fails to find the file, and it fails with error like this:

```
Error:
  An assembly specified in the application dependencies manifest (app.deps.json) was not found:
    package: 'Newtonsoft.Json', version: '12.0.1'
    path: 'sub/Newtonsoft.Json.dll'
```

Which is very misleading, as the path contains the subdirectory, but the host doesn't use it to look up the file. This can be observed in the trace:
```
Processing TPA for deps entry [Newtonsoft.Json, 12.0.1, sub/Newtonsoft.Json.dll]
  Considering entry [Newtonsoft.Json/12.0.1/sub/Newtonsoft.Json.dll], probe dir [], probe fx level:0, entry fx level:0
    Local path query did not exist F:\AppModel\repro\DepsJsonRelative\bin\Debug\netcoreapp3.0\Newtonsoft.Json.dll
```

liesauer:
hi, i am working on an awesome tool called [NetCoreBeauty](https://github.com/nulastudio/NetCoreBeauty) that makes the .net core application directory looks beauty recently, and got the same problem. here's the things happends, if i `cd` into app base directory, and then run my app with command `./myapp` or just double-click it, everything works properly, but when i try to run my app at other directory such as `~` with command `/path/to/my/app`, it fails to load the assemblies, and then i found something interesting.
```
[...]

Issue #7534, opened 2019-08-01
is there any reasons that merging working dir not app base dir into probing path?


liesauer:
https://github.com/dotnet/core-setup/blob/6db018b7fc5c104810f5a53b538841004e099713/src/corehost/cli/fxr/fx_muxer.cpp#L214

see https://github.com/dotnet/core-setup/issues/5645#issuecomment-516764531 too.

liesauer:
it should be supposed to merge app base dir.

vitek-karas:
I don't know - but it's unlikely to change now. Changing it would be a significant breaking change. Also note that it's not just in the `fx_muxer` you linked above, the same behavior applies to other places - basically everywhere the call to `pal::realpath` is made.

liesauer:
i edited this line https://github.com/dotnet/core-setup/blob/6db018b7fc5c104810f5a53b538841004e099713/src/corehost/cli/fxr/fx_muxer.cpp#L217

to
```cpp
pal::string_t probe_path = host_info.dotnet_root.c_str() + path;
```

everything seems working properly, no `pal::realpath` changes needed,  so i don't think this solution is a breaking change.

vitek-karas:
There are multiple ways to specify additional probing paths, one such way is command line: `--additionalprobingpath`. So if somebody has a script which does something along these lines:
```
cd MyExtensionsDirectory
/absolute/path/to/app/myapp --additionalprobingpath LoggingExtension
```
With your proposed change such script would stop working. Before your change it would look for the `LoggingExtension` under the `MyExtensionsDirectory`. After your change it would look for it in the `myapp`.

Similar argument can be made about other ways to specify additional probing paths. I personally agree that at least those probing paths specified in the `.runtimeconfig.dev.json` should be considered relative to the app. On the other hand the command line option would make sense to be relative to current directory (since that's how most command line arguments work in general if they specify relative paths). Regardless, the current behavior is at least consistent across multiple ways to specify the same thing... and the breaking change nature of such modification would make this a very hard sell.

[...]

PR #7743, opened 2019-08-16
fix additionalProbingPaths resolver

liesauer:
this fixed dotnet/core-setup#5645

### DIFF
--- a/src/corehost/cli/runtime_config.cpp
+++ b/src/corehost/cli/runtime_config.cpp
@@ -97,14 +97,14 @@ bool runtime_config_t::parse_opts(const json_value& opts)
     {
         if (probe_paths->second.is_string())
         {
-            m_probe_paths.insert(m_probe_paths.begin(), probe_paths->second.as_string());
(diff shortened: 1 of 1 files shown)


vitek-karas:
As described in the discussion to https://github.com/dotnet/core-setup/issues/7534 such a change could break applications. Currently the paths are resolved against current directory, with this change they would be resolved against application directory.
While I agree that ideally they should resolve against application directory, the code has not worked like that in multiple shipped versions, and we can't change it now.

Additionally this change is in `hostfxr` which is shared globally by all installed .NET Core versions on the machine - so any behavioral change affects all .NET Core versions - which makes changes to this component even more complicated.

I still don't understand why the proposed fix for https://github.com/dotnet/core-setup/issues/5645 doesn't help here. It would add the application base directory as another probe path and correctly append the relative path specified by the `.deps.json`. As far as I can tell that should fully resolve issues you're seeing with your scenario (trying to move some of the files from self-contained app to a subfolder).

vitek-karas:
Also please note that this change breaks existing tests - which shows that it might break applications as well.

dagood:
@vitek-karas what is the status of this PR? It's on my radar because it's somewhat old now.

PR #7755, opened 2019-08-16
Enabling ci legs for 3.1 branch

Anipik:


Anipik:
The fix is mostly taken from https://github.com/dotnet/core-setup/commit/3b8cebc9528342972f0229314aa375f9b09d0038

eng/Versions.props
```diff
@@ -9,7 +9,7 @@
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
     <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
     <!-- Disable SemVer v2 until after 3.0.0. -->
-    <SemanticVersioningV1 Condition="&#xA;      $(MajorVersion) &lt; 3 or&#xA;      '$(MajorVersion).$(MinorVersion).$(PatchVersion)' == '3.0.0'">true</SemanticVersioningV1>
+    <SemanticVersioningV1 Condition="&#xA;      $(MajorVersion) &lt; 3 or&#xA;      '$(MajorVersion).$(MinorVersion).$(PatchVersion)' == '3.1.0'">true</SemanticVersioningV1>
```
dagood: Modifying this line breaks the plan, and prevents us from changing to semver v2 until the next preview. #6825

dagood: Maybe you didn't find https://github.com/dotnet/core-setup/commit/cafcf54a8caccf388715618943cb641fd49f912f#diff-fec826feae04e51c0d94076385408bdc ?

Directory.Build.targets
```diff
@@ -11,6 +11,14 @@
   <UsingTask TaskName="GetTargetMachineInfo" AssemblyFile="$(LocalBuildToolsTaskFile)" />
   <UsingTask TaskName="RegenerateReadmeTable" AssemblyFile="$(LocalBuildToolsTaskFile)" />
 
+  <PropertyGroup>
+    <!--
+      Define this here (not just in Versions.props) because the SDK resets it
+      unconditionally in Microsoft.NETCoreSdk.BundledVersions.props.
+    -->
+    <NETCoreAppMaximumVersion>$(MajorVersion).$(MinorVersion)</NETCoreAppMaximumVersion>
```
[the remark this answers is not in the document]
Anipik: I added it for the TestProjects.targets but it is required here as well for .pkgproj

eng/Versions.props
```diff
@@ -90,6 +90,7 @@
     <MicrosoftDiaSymReaderNativePackageVersion>1.7.0</MicrosoftDiaSymReaderNativePackageVersion>
     <!-- Infrastructure and test-only. -->
     <MicrosoftSourceLinkVersion>1.0.0-beta2-18618-05</MicrosoftSourceLinkVersion>
+    <MNAVersion>3.0.0-preview8-28379-02</MNAVersion>
```
Anipik: @dagood i have to specify as it was being set to empty. This and setting the NetCoreAppMaximum fixes the issue.

dagood: This can't be a specific value, it varies per build. You shouldn't have to set it, it's set here:

https://github.com/dotnet/core-setup/blob/9528017ba529995cc42697abc1e3c59258303192/src/test/PrepareTestAssets/PrepareTestAssets.proj#L39-L47

Issue #6679, opened 2019-06-05
"Uninstall" button truncated on runtime installation wizard with Japanese language


AriNuer:
## Steps to reproduce

1. Open the runtime installation (.NET core runtime-3.0.0-preview6-27804-01) in Japanese OS.
2. Observe the "Uninstall" button.

## Expected  behavior

"Uninstall"  button text is integer visible.

## Actual behavior

"Uninstall" button is truncated.

![111](https://user-images.githubusercontent.com/42136111/58933545-295a2e00-879a-11e9-9173-5bdfecd53863.png)

## Environment data
win10-x64 with Japanese

Issue #6532, opened 2019-05-24
"Install" button truncated in shared framework installer in Japanese language


dagood:
Spotted in SDK installer by https://github.com/dotnet/cli/issues/11397, also reproduces with the shared framework installer.

## Steps to reproduce

1. Open the .NET Core shared framework installer in Japanese.
   * (`.\dotnet-runtime-3.0.0-preview6-27710-0-win-x64.exe -lang 1041`)
2. Observe the "インストール" button.

## Expected  behavior

Install button text is entirely visible.

## Actual behavior

![image](https://user-images.githubusercontent.com/12819531/58347367-5534f580-7e23-11e9-8722-9341d028c7a4.png)

/cc @johnbeisner @dleeapho

PR #7757, opened 2019-08-16
Increase bundle's button size: avoid loc clipping

dagood:
Fix button label clipping in Japanese localization. For https://github.com/dotnet/core-setup/issues/6532 and https://github.com/dotnet/core-setup/issues/6679. Fix copied from Core-SDK, https://github.com/dotnet/core-sdk/pull/2123.

Confirmed via `.\dotnet-runtime-5.0.0-dev-win-x64.exe -lang 1041`.

### DIFF
--- a/src/pkg/projects/netcoreapp/sfx/bundle.thm
+++ b/src/pkg/projects/netcoreapp/sfx/bundle.thm
@@ -15,7 +15,7 @@
 
         <Text Name="HelpHeader" X="11" Y="80" Width="-11" Height="30" FontId="2" DisablePrefix="yes">#(loc.HelpHeader)</Text>
         <Text Name="HelpText" X="20" Y="115" Width="-11" Height="-35" FontId="3" DisablePrefix="yes">#(loc.HelpText)</Text>
-        <Button Name="HelpCancelButton" X="-11" Y="-11" Width="75" Height="23" TabStop="yes" FontId="0">#(loc.HelpCloseButton)</Button>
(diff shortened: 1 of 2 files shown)


MichaelSimons:
@dagood - Any chance you could include a screen shot?  Ideally both english and Japanese.  I see you asked on the core-sdk thread but I didn't see a response.  Are there UI design guidelines we should be reviewing?

dagood:
Sure:

| Install | install_en | install_jp |
| --- | --- | --- |
| Uninstall | uninstall_en | uninstall_jp |


> I see you asked on the core-sdk thread but I didn't see a response. Are there UI design guidelines we should be reviewing?

There are the accessibility guidelines (with https://accessibilityinsights.io/docs/en/windows/overview/), but other than that, I'm not aware of any.

MichaelSimons:
My primary concern is using fixed widths and localization never work well together in practice over the long term.

From my days on working on VS Connected Services, I wonder what the VS UX designers would think about this.  My hunch is they would not like the wider buttons.  In VS all wizard buttons like this were supposed to have a 75 pixel min width IIRC.  I have no WiX experience, it looks like they only support fixed width content.  The other feature that I don't see is theme support.  Having to touch this many many places for controls that are supposed to have the same appearance is not ideal.

In the end I have a feeling your proposed/copied changes are the only option in which I have no objection to the changes 😃

dagood:
Everything you're saying is true, and more. Looking forward to ditching the default burn bundle engine.

PR #7762, opened 2019-08-17
Make it more obvious what constitutes an invalid directory.

pi1024e:
There is a line of code that returns with an exclamation mark operating on a series of boolean statements that would be better off negated individually.


vitek-karas:
@lpereira can you please review this one?

vitek-karas:
@pi1024e thanks a lot for the fix!

PR #7764, opened 2019-08-18
[release/3.0] Create WindowsDesktop shared framework bundle installer

dagood:
For https://github.com/dotnet/core-setup/issues/6384.

#### Description

Create a Windows Desktop Runtime (Shared Framework) installer. This supports running WindowsDesktop apps published as FDD/FDE.

Cherry-picks https://github.com/dotnet/core-setup/pull/7540 for WindowsDesktop installer implementation.
Cherry-picks https://github.com/dotnet/core-setup/pull/7758 to include localization.

#### Customer Impact

Without this artifact, a user must install the full SDK to run WPF/WinForms apps on .NET Core, and in some cases, both x86 and x64 SDKs. (https://github.com/dotnet/winforms/issues/1624)

A few times I've been asked why installing the ".NET Core Runtime" doesn't make WPF and WinForms apps work. Having a "Windows Desktop Runtime" artifact listed on the same page will make the requirement somewhat discoverable.

#### Regression?

No. This is a new artifact.

#### Risk

Minimal risk to existing scenarios. The refactoring of .NET Core Runtime bundle installer generation has been working fine in `master`.

Scenario risks:
* The WindowsDesktop installer doesn't include the .NET Core Runtime, so there may be new confusion from customers why they need to install both the Windows Desktop Runtime and .NET Core Runtime to run WPF/WinForms apps. There is future work to create a support bundle that includes .NET Core Runtime: https://github.com/dotnet/core-setup/issues/7763.
* The installer doesn't have finalized branding. See https://github.com/dotnet/core-setup/issues/7530.

---

/cc @vatsan-madhavan @zsd4yr @merriemcgaw @grubioe @richlander @leecow @dleeapho

PR #7766, opened 2019-08-19
SharedHost always installs to 'ProgramFiles'\dotnet; 3.0

ghost:
When installing the SharedHost; copy all files to the traditional default install location: 'ProgramFiles'\dotnet even when installing somewhere else.

This is part of the work to allow SDKs, when installing with VS, to install to a different volume when VS is doing the same.

**User Scenerio:**
When a user has installed a 2.1 SDK or similar SDK first so that the PATH environmental variable has 'ProgramFiles'\dotnet in a first position. Then the user installs a 3.0 SDK thru VS to a different volume with VS so that 'VSSharedVolume'\dotnet is in a second position. We want the 3.0 SharedHost to be on the PATH in the first position. Therefore, installing the SharedHost to both 'VSSharedVolume'\dotnet and 'ProgramFiles'\dotnet will achieve this.

Reference: dotnet/core-sdk#2174
Similar: https://github.com/dotnet/core-setup/pull/7658

@dotnet/dotnet-cli

### DIFF
--- a/src/pkg/packaging/windows/host/host.wxs
+++ b/src/pkg/packaging/windows/host/host.wxs
@@ -30,21 +30,40 @@
         <Directory Id="TARGETDIR" Name="SourceDir">
             <Directory Id="$(var.Program_Files)">
                 <Directory Id="DOTNETHOME" Name="dotnet"/>
+                <Directory Id="PROGRAMFILES_DOTNET" Name="dotnet" />
             </Directory>
         </Directory>
     </Fragment>
+
+      <!-- When installing the SharedHost; copy all files to the traditional default install location: 'ProgramFiles'\dotnet even when installing into a custom location.
+      User Scenerio:
+      When a user has installed a 2.1 SDK or similar SDK first so that the PATH environmental variable has 'ProgramFiles'\dotnet in a first position.
+      Then the user installs a 3.0 SDK thru VS to a different volume with VS so that 'VSSharedVolume'\dotnet is in a second position.
+      We want the 3.0 SharedHost to be on the PATH in the first position. Therefore, installing the SharedHost to both 'VSSharedVolume'\dotnet and
+      'ProgramFiles'\dotnet will achieve this.
+      On a 'clean' machine when installing into a custom location; the installation behavior is the same.
+      Rationale:
+        1. the legacy SDK can be subsequently installed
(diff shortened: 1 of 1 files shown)


src/pkg/packaging/windows/host/host.wxs
```diff
@@ -30,21 +30,42 @@
...
+         User Scenerio:
+         When a user has installed a 2.1 SDK or similar SDK first so that the PATH environmental variable has 'ProgramFiles'\dotnet in a first position.
+         Then the user installs a 3.0 SDK thru VS to a different volume with VS so that 'VSSharedVolume'\dotnet is in a second position.
+         We want the 3.0 SharedHost to be on the PATH in the first position. Therefore, installing the SharedHost to both 'VSSharedVolume'\dotnet and
+         'ProgramFiles'\dotnet will achieve this.
+         On a 'clean' machine when installing into a custom location; the installation behavior is the same.
+      
+         Rationale:
+           1. the legacy SDK can be subsequently installed
+           2. the user runs 'dotnet' commands directly against 'ProgramFiles'\dotnet\dotnet.exe -->
```
dagood: Could you leave this the same as it is in `master` to avoid unnecessary differences?

https://github.com/dotnet/core-setup/blob/8a5f710c474a955ae82ce449fb9185a781782558/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs#L10-L21

dagood (approved):
https://github.com/dotnet/core-setup/pull/7764 is an in-progress PR porting the new shiny infra, which is why the files are in currently in a different place with more duplication. The impl in this PR looks good. Let's merge this PR before https://github.com/dotnet/core-setup/pull/7764 once they have approval, I'll deal with merge conflicts over there.

src/pkg/packaging/windows/host/host.wxs
johnbeisner: Done.

johnbeisner:
**Description**
When installing the SharedHost; copy all files to the traditional default install location: 'ProgramFiles'\dotnet even when installing somewhere else. This is part of the work to allow SDKs, when installing with VS, to install to a different volume when VS is doing the same.

**Customer Impact**
When a user installs 16.3 Visual Studio with the 3.0 SDK to a non-operating system volume, the 3.0 SharedHost will also install to 'ProgramFiles'\dotnet

**Regression**
No.

**Risk**
Low.

johnbeisner:
image

PR #7777, opened 2019-08-21
Bundles: Align assemblies at a 4K boundary

swaroop-sridhar:
Align assemblies at a 4K boundary to facilitate their load
directly from the (memory-mapped) bundle by the runtime.

Assemblies are written with 4K alignment to facilitate loading from bundle
Remaining files (native binaries and other files) are written without alignment.

The unaligned files are written first, followed by the aligned files,
and finally the bundle manifest.

Further work is required to order the files better in order to minimize bundle size.

Bundle size change due to this change:
Console self-contained win10-x64: Increased by 0.3 MB (67.95 -> 68.26)
WebAPI self-contained win10-x64: Increased by 0.5 MB (84.53 -> 85.10)

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/Bundle/Bundler.cs
+++ b/src/managed/Microsoft.NET.HostModel/Bundle/Bundler.cs
@@ -129,21 +129,6 @@ namespace Microsoft.NET.HostModel.Bundle
         }
 
         /// <summary>
-        /// Determine whether files of this type will be aligned
-        /// in the bundle.
-        /// <paramref name="type">
-        /// The FileType to examine
-        /// </paramref>
-        /// <returns>
-        /// Whether a file of this type will be aligned
-        /// </returns>
-        /// </summary>
-        public static bool NeedsAlignment(FileType type)
-        {
-            return type == FileType.IL || type == FileType.Ready2Run;
-        }
-
-        /// <summary>
         /// Generate a bundle, given the specification of embedded files
         /// </summary>
         /// <param name="fileSpecs">
@@ -175,7 +160,7 @@ namespace Microsoft.NET.HostModel.Bundle
                 throw new ArgumentException("Invalid input specification: Must specify the host binary");
             }
 
-            if(fileSpecs.GroupBy(file => file.BundleRelativePath).Where(g => g.Count() > 1).Any())
+            if (fileSpecs.GroupBy(file => file.BundleRelativePath).Where(g => g.Count() > 1).Any())
             {
                 throw new ArgumentException("Invalid input specification: Found multiple entries with the same BundleRelativePath");
             }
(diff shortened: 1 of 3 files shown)


swaroop-sridhar:
I checked with @nguerrera that changes to HostModel in the master branch doesn't affect the SDK for 3.0 release.

swaroop-sridhar:
CC: @janvorli  @davidwrighton

src/managed/Microsoft.NET.HostModel/Bundle/Bundler.cs
```diff
@@ -191,7 +216,27 @@ public string GenerateBundle(IReadOnlyList<FileSpec> fileSpecs)
                     using (FileStream file = File.OpenRead(fileSpec.SourcePath))
                     {
                         FileType type = InferType(fileSpec.BundleRelativePath, file);
-                        long startOffset = AddToBundle(bundle, file, type);
+
+                        if(NeedsAlignment(type))
```
vitek-karas:
```suggestion
                        if (NeedsAlignment(type))
```

src/test/BundleTests/Microsoft.NET.HostModel.Bundle.Tests/BundlerConsistencyTests.cs
```diff
@@ -143,6 +143,26 @@ public void AllBundledFilesAreExtracted()
...
+
+            var hostName = BundleHelper.GetHostName(fixture);
+            var appName = Path.GetFileNameWithoutExtension(hostName);
+            string publishPath = BundleHelper.GetPublishPath(fixture);
+            var bundleDir = BundleHelper.GetBundleDir(fixture);
+
+            var bundler = new Bundler(hostName, bundleDir.FullName);
+            bundler.GenerateBundle(BundleHelper.GetPublishPath(fixture));
+
+            foreach(var fileEntry in bundler.BundleManifest.Files)
```
vitek-karas:
```suggestion
            foreach (var fileEntry in bundler.BundleManifest.Files)
```

src/test/BundleTests/Microsoft.NET.HostModel.Bundle.Tests/BundlerConsistencyTests.cs
```diff
@@ -143,6 +143,26 @@ public void AllBundledFilesAreExtracted()
...
+            {
+                Assert.True(!Bundler.NeedsAlignment(fileEntry.Type) ||
```
vitek-karas: I think the test should not rely on the product to tell it what needs alignment... since we know exactly what app we're building here, the test should relatively easily list all the files to be aligned.

swaroop-sridhar:
I discussed this issue further with @janvorli. The 4K alignment is strictly not necessary in order to map the contents of the bundle for loading. 

MAPMapViewOfFile already maps sections that are not  4K aligned (by mapping extra bytes before/after each section, possibly redundantly). With some additional work to CorUnix::InternalMapViewOfFile and MAPMapViewOfFile, loading from single-exe bundles can also be made to work without the requirement for 4K alignment (at least as well as the section-loading currently works). 

I'll still check this change in, so that we get a HostModel package that is 4K aligned for experimentation/prototyping. Once the changes to map.cpp are implemented, we can revert this change.

PR #7821, opened 2019-08-23
[release/3.0] Fix bundle installer signing

dagood:
#### Description

#7817. The bundle installers fail to install with `0x80070001 - Incorrect function`. The most likely direct cause of that error is the burn engine not being signed. There are several MSBuild authoring problems related to signing that this PR fixes: runtime MSIs are not signed due to a build order issue, and the burn bundle engine is not reattached to the bundle installer.

A validation real-signed build shows that this PR resolves the issue: https://github.com/dotnet/core-setup/issues/7817#issuecomment-524388499.

Clean cherry-pick of https://github.com/dotnet/core-setup/pull/7820 to `release/3.0`.

#### Customer Impact

The .NET Core Runtime and Windows Desktop Runtime bundle installers currently fail to install, with a cryptic error message.

#### Regression?

Yes, the .NET Core Runtime bundle installer in particular. The changes made to produce the Windows Desktop Runtime bundle inadvertently caused this problem for both bundle installers.

#### Risk

Low. The authoring errors are simple to correct.

### DIFF
--- a/signing/SignBurnBundleFiles.proj
+++ b/signing/SignBurnBundleFiles.proj
@@ -4,7 +4,7 @@
   <Target Name="ReattachAllEnginesToBundles"
           BeforeTargets="RunArcadeSigning">
     <MSBuild
-      Projects="@(ProjectToBuild -> WithMetadataValue('SignPhase', 'Bundle'))"
+      Projects="@(ProjectToBuild -> WithMetadataValue('SignPhase', 'BundleInstallerFiles'))"
       Targets="ReattachEngineToBundle" />
   </Target>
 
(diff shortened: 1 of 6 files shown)


dagood:
Working on the CI errors, I have a local repro. It has something to do with the dependency order changes made for https://github.com/dotnet/core-setup/pull/7372 that made it into `release/3.0` but haven't been merged into `master` yet.

dagood:
The cycle was that the MSI project thinks it needs to sign MSIs before it can build MSIs. This is clearly wrong, so I cut that dependency. Ensuring MSIs are signed is done by the bundle project tooling, where it should be:

https://github.com/dotnet/core-setup/blob/10bdf83b80772ddd94eb838d24edd0fb3babb5c2/src/pkg/packaging-tools/windows/wix.targets#L158-L163

Issue #7572, opened 2019-08-05
Disable Microsoft.WindowsDesktop.App.Ref when building from source (source-build)


dagood:
In 3.0 preview 7, it was reported that during source-build, Core-Setup created a `Microsoft.WindowsDesktop.App.Ref` that had to be deleted manually: https://github.com/dotnet/source-build/issues/1123. It should not be created--source-build isn't targeting Windows, the only place the output would work.

dagood:
Fixed with https://github.com/dotnet/core-setup/pull/7897, ported to 3.0 with https://github.com/dotnet/core-setup/pull/7910.

PR #7910, opened 2019-08-30
[release/3.0] Fix source-build patch and workaround arg

dagood:
#### Description

For https://github.com/dotnet/core-setup/issues/7572. Ports https://github.com/dotnet/core-setup/pull/7897.

Incorporates source-build patch that was added to exclude WindowsDesktop from NuGet restore: [0002-Try-ExcludeFromSourceBuild-in-windowsdesktop.patch](https://github.com/dotnet/source-build/blob/249b034a11bf0f0fcb6e264cb5c933265687777a/patches/core-setup/0002-Try-ExcludeFromSourceBuild-in-windowsdesktop.patch)

Adds logic to determine RestoreAllBuildRids based on DotNetBuildFromSource, so that source-build doesn't have to pass in the `/p:RestoreAllBuildRids=false` override argument.

#### Customer Impact

Reduced source-build turnaround time due to lower maintenance cost.

#### Regression?

No.

#### Risk

Minimal. Changes depend on `DotNetBuildFromSource` being set, which is never true for Microsoft builds. If this doesn't end up working in source-build, it can be patched around.

---

Pending a decision on whether merging this makes sense in 3.0 at this point.

### DIFF
--- a/src/pkg/packaging-tools/framework.sharedfx.targets
+++ b/src/pkg/packaging-tools/framework.sharedfx.targets
@@ -18,9 +18,15 @@
     </PropertyGroup>
   </Target>
 
+  <!--
+    Set up the PackageReference based on the pkgproj output at the right point in the build process.
+    The BeforeTargets makes it try to run even when the project is excluded from a build from
+    source, so this has to be handled directly.
+  -->
   <Target Name="GetPackageReference"
           DependsOnTargets="GetProductVersions"
-          BeforeTargets="CollectPackageReferences">
+          BeforeTargets="CollectPackageReferences"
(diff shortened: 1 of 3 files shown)


dagood:
> Pending a decision on whether merging this makes sense in 3.0 at this point.

I think that we should merge it. It doesn't add any burden to source-build: it doesn't conflict with the existing source-build patch (the change is applied to a different file) and if source-build continues to pass `/p:RestoreAllBuildRids=false`, that doesn't cause any failures. It lets source-build remove the patch and property whenever it's useful to do so.

@crummel @dleeapho @MichaelSimons any objection?

MichaelSimons:
+1 on merging in 3.0

dleeapho:
+1 for 3.0

crummel:
That sounds good to me as well.

Issue #8100, opened 2019-09-09
Fix infra to allow stabilized builds


dagood:
The known problem right now is that the VS insertion packages end up with stable versions, even though they are nonshipping. This causes problems because the packages fail to push to the VS feed: mutation is not allowed.

dagood:
Done: ported to 3.0 by https://github.com/dotnet/core-setup/pull/8188.

PR #8105, opened 2019-09-09
Use IsShipping=false to keep VS insertion nupkgs unique

dagood:
For https://github.com/dotnet/core-setup/issues/8100.

In Arcade, NonShipping packages always have unstable (`-{prerelease}-{buildnum}`) versions. This is necessary to push VS insertion packages to the VS feed, because mutation is not allowed and we may need to build a stabilized build multiple times. Core-Setup infra structure didn't properly support this in two ways:
* The insertion package generation infra runs inside the pkgproj that also produces the MSI. This means they are both shipping, or both nonshipping. The MSI needs to be shipping and the VS insertion package needs to be nonshipping.
  * Solved by using an MSBuild task with shipping set to `false` for the insertion nupkg.
* `IsShipping` isn't supported in Core-Setup's `GetProductVersions` target.
  * Solved by adding it. Also support the related `SuppressFinalPackageVersion` property in case it's necessary in the future.
* `PushToAzureDevOpsArtifacts` is missing `IsStableBuild="$(IsStableBuild)"`, which is necessary to make the correct asset manifest.
  * Solved by adding it. It is missing from the [yaml stages publish doc](https://github.com/dotnet/arcade/blob/a82e129e6df744f26e920284ab098fb5c2c13418/Documentation/CorePackages/YamlStagesPublishing.md) but present in [Arcade's usage of the task](https://github.com/dotnet/arcade/blob/694d59f090b743f894779d04a7ffe11cbaf352e7/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj#L169).

(This PR is a cherry-pick of changes I'm making to a stabilization validation branch based on 3.0.)

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -33,7 +33,12 @@
   <Target Name="GetProductVersions">
     <PropertyGroup>
       <IncludePreReleaseLabelInPackageVersion Condition="'$(DotNetFinalVersionKind)' != 'release'">true</IncludePreReleaseLabelInPackageVersion>
+      <IncludePreReleaseLabelInPackageVersion Condition="'$(SuppressFinalPackageVersion)' == 'true'">true</IncludePreReleaseLabelInPackageVersion>
+      <IncludePreReleaseLabelInPackageVersion Condition="'$(IsShipping)' != 'true'">true</IncludePreReleaseLabelInPackageVersion>
+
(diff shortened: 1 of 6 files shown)


dagood:
I jumped the gun on this--an earlier validation build that used `SuppressFinalPackageVersion` instead of `IsShipping` worked with the internal build, but now tests fail with this change. (Failures repro locally.) Looking into it.

dagood:
This set of changes are good, but there's more servicing TODO: https://github.com/dotnet/core-setup/issues/8112.

PR #8143, opened 2019-09-10
Add license to project file

barkerd427:
I'm a total noob when it comes to dotnet. Feel free to reject.

I was running gitlab's built-in license management tool, and no license was found for this package. Upon reading the docs, this seems like the preferred method for NuGet packages moving forward. My assumption is that this will update NuGet and I will get the correct license information in the future. If this works, I'm happy to update other csproj files in other dotnet repos.

I believe this can be skip ci please


barkerd427:
Hmmm, these builds seem to inject PackageLicenseFile from somewhere. Not sure if that isn't configured for the package in question, but the file model also makes it really hard for license management systems to understand which license is actually being used and whether it has been modified.

dagood:
Thanks for the attempt, I think you'll be better off filing an issue with some information about what GitLab is not able to find, though. The package does have a license, e.g. https://www.nuget.org/packages/Microsoft.DotNet.PlatformAbstractions/3.0.0-preview9-19423-09/License. Also, you may be interested in https://github.com/dotnet/arcade.

PR #8186, opened 2019-09-11
[release/3.0] Fix NonShipping packages to always be unstable

dagood:
#### Description

For #8112. Arcade publishing requires that NonShipping packages have an unstable version, because they are published to a static feed that doesn't allow mutation.

Ports https://github.com/dotnet/core-setup/pull/8155 to `release/3.0`.

> Change `<IsShippingPackage>false</IsShippingPackage>` to `<IsShipping>false</IsShipping>`: this is what I did to get the nonshipping packages unstable before we get the Arcade fix change [dotnet/arcade#3909](https://github.com/dotnet/arcade/pull/3909), and I don't know a reason to revert it.
> 
> Fix the `Microsoft.NETCore.App.Internal` dependency on `Microsoft.NETCore.DotNetHostPolicy` to be correct regardless of stable/nonstable versioning, by adding a `VersionFromProject` feature similar to `VersionProp`.
> 
> Add `PrepareStabilizedLegacyPackages` in `PrepareTestAssets.proj` that kicks in when shipping packages are stable. It uses a new task `CopyNupkgAndChangeVersion` to create a copy of the nonshipping packages in an intermediate directory that is fed into test restore. When the versions are the same, like here in `master`, the packages are left alone.

#### Customer Impact

Allows 3.0 stable-versioned builds. Without this, Arcade can't process the build.

#### Regression?

No.

#### Risk

Minimal. The workaround is a fairly simple way to let our tests run as they would normally, with fully stabilized versions, while satisfying Arcade requirements.

### DIFF
--- a/src/pkg/projects/Directory.Build.targets
+++ b/src/pkg/projects/Directory.Build.targets
@@ -16,11 +16,34 @@
 
   <!--
     For any Dependency items with a VersionProp, set it to the property by that name given by the
-    version generation target.
+    version generation target. For any with a VersionFromProject, query the ProductVersion from that
+    project file and use it as the dependency's version.
   -->
   <Target Name="SetCustomPackageDependencyVersions"
           BeforeTargets="GetPackageDependencies"
           DependsOnTargets="GetProductVersions">
+    <!--
+      Generate a VersionProp name for each dependency with VersionFromProject. The batched MSBuild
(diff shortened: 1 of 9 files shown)


dagood:
I missed porting https://github.com/dotnet/core-setup/issues/8100 as part of this, recreating.

Issue #7846, opened 2019-08-26
Create AppHost MSIs that install to other RID's SDK location


dagood:
This is for C++/CLI support, specifically for scenarios like building a win-x86 app using the win-x64 SDK. Visual Studio and the SDK bundle installer must support this.

C++/CLI projects don't support nuget restore, so the AppHost packs need to be in the `DOTNETHOME\packs` dir for the SDK to find them. (There is a side effort to try to get this working, but we may not be able to wait for that.)

This *could* be solved in dotnet/core-sdk by having that repo extract the AppHost nupkgs to assemble a new MSI (or multiple MSIs) that install the data to the desired locations. Core-SDK does produce zips this way.

There are two other solutions we thought of that would live in Core-Setup:

> 1. Create “cross target” msis in core-setup. So for example, a 32 bits msi installs to the 64bits location.
>    * We need extra 6 more msis, and on SDK side, we need to config them to installers and VS
>    * New msis are 
>      * Install to 64bits location: win-x86, win-arm, win-arm64
>      * Install to 32bits location: win-x64, win-arm, win-arm64
> 2. Core-setup to have a new step to gather all apphosts from different RIDs’ to create a new apphosts bundle.

The first one looks the most appealing:
* We don't have to implement a "join" stage/job that brings together all the bits.
* Doesn't add any movement of bits between stages.
* Avoids making the build YAML more complicated.

IMO this should live in Core-Setup so that it remains clear where the AppHost installers are produced and the versioning can be the same between the normal AppHost MSIs and the cross-RID ones.

/cc @wli3 @livarcocc @johnbeisner @dleeapho @MichaelSimons

wli3:
also @nguerrera

wli3:
On SDK side:
All in archives, all in installers, and **VS**

wli3:
On core-setup: 

Wrap msi with nupkg and pubish to VS feed

dagood:
Merged into 3.1 with https://github.com/dotnet/core-setup/pull/8295. Build going with the change: https://dev.azure.com/dnceng/internal/_build/results?buildId=361681&view=results.

PR #8292, opened 2019-09-19
Generate cross-arch AppHost pack MSIs

dagood:
For https://github.com/dotnet/core-setup/issues/7846.

Creates and publishes cross target MSIs and VS insertion nupkgs.

New MSIs and nupkgs are:
  * Install to x64 SDK location: `x64_x86`, `x64_arm`, `x64_arm64`
  * Install to x86 SDK location: `x86_x64`, `x86_arm`, `x86_arm64`

Note that there are no new product binaries here, this is simply a new set of installers because for each install location, we need a new MSI.

Validation build ongoing as of writing: https://dev.azure.com/dnceng/internal/_build/results?buildId=360394

Porting to 3.1 after this goes in.

/cc @wli3 @nguerrera @livarcocc @johnbeisner @dleeapho

### DIFF
--- a/eng/jobs/windows-build.yml
+++ b/eng/jobs/windows-build.yml
@@ -73,10 +73,6 @@ jobs:
           publishFeedCredentials: 'DevDiv - VS package feed'
         condition: and(
           succeeded(),
-          or(
-            eq(variables['TargetArchitecture'], 'x64'),
-            eq(variables['TargetArchitecture'], 'x86')
-          ),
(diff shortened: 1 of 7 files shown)


dagood:
Thanks. Merging the functionality into `master`. There is some offline discussion about using a different separator than `_` due to VS insertion authoring complications--at the moment I plan to resolve that before porting to 3.1.

PR #8296, opened 2019-09-20
Revert "[master] Update dependencies from dotnet/corefx (#8266)"

dagood:
This reverts https://github.com/dotnet/core-setup/pull/8266.

That PR removed `TEST_RESTORE_SOURCES_INSERTION_LINE`, which normally causes test restore to fail. It looks like it didn't fail in this case because an internal build published packages with the same versions due to a coincidental build number match. The green build allowed the PR to merge, but now it's breaking builds where this coincidence isn't occurring.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -6,6 +6,12 @@
   </fallbackPackageFolders>
   <packageSources>
     <clear />
+    <!--
+      'src/test/PrepareTestAssets/PrepareTestAssets.proj' generates a NuGet.config file using this
+      one as a template. The following line is a marker to insert the test restore sources.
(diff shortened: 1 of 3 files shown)


dagood:
Filed https://github.com/dotnet/core-setup/issues/8297 to track making the tests not do that, with some more description.

dagood:
FYI @wtgodbe @mmitche, it's possible for Maestro++ updates to sneak through in Core-Setup and break builds if the public CI build number matches a successful official build's number. Something to be aware of until https://github.com/dotnet/arcade/issues/3938 is deployed.

PR #8311, opened 2019-09-23
Change post-build yaml param dependsOn => validateDependsOn to match Arcade

dagood:
https://github.com/dotnet/arcade/pull/3986/files#diff-853fb537e588caa8b548253bdcf21cd8 broke the Core-Setup build by removing the `validate` dependency:

image

https://dev.azure.com/dnceng/internal/_build/results?buildId=363134

This PR should restore it by renaming the param to what it was changed to. This was a breaking change for Core-Setup because of the way I had to implement the custom publish steps without Arcade support.

/cc @markwilkie

### DIFF
--- a/eng/stages/publish.yml
+++ b/eng/stages/publish.yml
@@ -19,7 +19,7 @@ stages:
 # Stages-based publishing entry point
 - template: /eng/common/templates/post-build/post-build.yml
   parameters:
-    dependsOn:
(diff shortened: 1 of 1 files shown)


dagood:
Build running with the fix: https://dev.azure.com/dnceng/internal/_build/results?buildId=363729&view=results.

PR #8312, opened 2019-09-24
Try switching core-setup to use VS2019.

jkoritzinsky:
Validate that core-setup can build on the VS2019 queue.


jkoritzinsky:
Since this build succeeded, I believe that core-setup is ready for VS2019. I've opened #8318 to update our instructions to state as such.

PR #8313, opened 2019-09-25
Update service branding

Anipik:


eng/Versions.props
```diff
@@ -3,13 +3,13 @@
...
-    <PreReleaseVersionLabel>rc2</PreReleaseVersionLabel>
+    <PatchVersion>1</PatchVersion>
+    <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
     <!-- Enable to remove prerelease label. -->
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
     <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
     <!-- Disable SemVer v2 until after 3.0.0. -->
-    <SemanticVersioningV1 Condition="&#xA;      $(MajorVersion) &lt; 3 or&#xA;      '$(MajorVersion).$(MinorVersion).$(PatchVersion)' == '3.0.0'">true</SemanticVersioningV1>
+    <SemanticVersioningV1 Condition="&#xA;      $(MajorVersion) &lt; 3 or&#xA;      '$(MajorVersion).$(MinorVersion).$(PatchVersion)' == '3.0.1'">true</SemanticVersioningV1>
```
dagood: This change prevents us from producing semver v2 outputs in 3.0.1, not desirable from my understanding. Unfortunate that Darc super-escaped all this and put it on a horizontal scrollbar. 😛

If you really do want to do this, we shouldn't bump 3.0.0 to 3.0.1, rather we should change the logic to forward-compatible logic to keep semver v2 disabled in 3.1.x.

/cc @jeffschwMSFT @sdmaclea 

https://github.com/dotnet/core-setup/issues/6825

PR #8330, opened 2019-09-26
[release/3.0] Upgrade SDK toolset from 3.0 preview6 => GA 

dagood:
#### Description

Upgrades the SDK used to build Core-Setup from 3.0 preview 6 to GA.

Handle SDK target name change from ResolveFrameworkReferences to ProcessFrameworkReferences in https://github.com/dotnet/sdk/commit/a55022eb89ef3d0fcecbcb6b53883181928e7030.

#### Customer Impact

Unblocks Arcade auto-update PRs like https://github.com/dotnet/core-setup/pull/8317.

#### Regression?

No.

#### Risk

There may be undetected additional breaks that need followup PRs to fix.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview6-012264"
+    "dotnet": "3.0.100"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.19461.7"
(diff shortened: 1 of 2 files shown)


dagood:
My guess as to why this this doesn't repro in `master`, which already uses `3.0.100`, is that the 3.0 SDK only knows about a 3.0 ASP.NET Core runtime pack, not a 5.0 runtime pack, so it never tries to restore one in the first place when building a `netcoreapp5.0` test project. I'm going to port this fix to `master` as futureproofing.

PR #8432, opened 2019-10-02
A couple small source-build fixes

crummel:
- Exclude WindowsDesktop from source-build.
- Pin NuGet and MSBuild packages to reference versions.

### DIFF
--- a/src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
+++ b/src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
@@ -1,4 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
+  </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.DotNet.Wpf.DncEng" Version="$(MicrosoftDotNetWpfDncEngPackageVersion)" />
--- a/tools-local/tasks/core-setup.tasks.csproj
+++ b/tools-local/tasks/core-setup.tasks.csproj
@@ -15,16 +15,17 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NuGet.ProjectModel" Version="$(NugetProjectModelPackageVersion)" />
+    <PackageReference Include="NuGet.ProjectModel" Version="4.9.4" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.1.0" />
   </ItemGroup>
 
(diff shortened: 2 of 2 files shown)


src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
```diff
@@ -1,4 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
dagood: This should have been fixed by https://github.com/dotnet/core-setup/commit/939f0a7116eee51f44cb4ea5396d781c2ab1c6ce: https://github.com/dotnet/core-setup/blob/412e8f8ebb1fb24e41940fed7f3eb99d7a2d3028/src/pkg/projects/windowsdesktop/Directory.Build.props#L21-L25

Do you know for sure if the patch is still necessary after that? (No harm adding a potentially redundant exclude if it would take a lot of effort to check, though.)

tools-local/tasks/core-setup.tasks.csproj
```diff
@@ -15,16 +15,17 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="NuGet.ProjectModel" Version="$(NugetProjectModelPackageVersion)" />
+    <PackageReference Include="NuGet.ProjectModel" Version="4.9.4" />
```
dagood: These are pinned as far as Maestro++/Darc are concerned: https://github.com/dotnet/core-setup/blob/412e8f8ebb1fb24e41940fed7f3eb99d7a2d3028/eng/Versions.props#L71-L78

I take it we need to do more here because the PVP is inadvertently changing the value? (And the API version change broke the build or something?)

I would prefer to continue to keep track of this version in Versions.props, and instead add a prefix to the property name to avoid the PVP. I'll push a quick commit to show what I mean.

(Ideally this isn't actually required--in that case, would be nice to not change this at all. 😄)

dagood:
@crummel can you take a look at my additions?

src/pkg/projects/windowsdesktop/src/windowsdesktop.depproj
crummel: It shouldn't be necessary anymore, thanks.

tools-local/tasks/core-setup.tasks.csproj
crummel: I like the way you're doing it here, I'm going to use this in some other repos too.  This was at least required in the 3.0.1 preview I merged today, if you think it should work now I can give it a try.

dagood: Yeah, I think the PR as it is now should work the same as the earlier patch to deep hard-code the versions. Validating it would be great if it's not too much trouble. 🙂

crummel: Yup, this is working locally for me.  Thanks for the improvements!

PR #8448, opened 2019-10-03
Enable using Live-live builds of CoreCLR and CoreFX with Core-Setup

jkoritzinsky:
Currently, there is no ability in the core-setup repo to consume a local coreclr or corefx build. This PR enables building the shared framework from local builds of coreclr and corefx.

To build with a locally built coreclr, pass in the `CoreCLROverridePath` property: `/p:CoreCLROverridePath=path/to/coreclr/bin/Product/OS.Arch.Configuration`. The Core-Setup build will swap out the package-restored CoreCLR files with the local CoreCLR files from the output Product directory.

To build with a locally built corefx, pass in the `CoreFXOverridePath` property: `/p:CoreFXOverridePath=path/to/corefx/artifacts/packages/Configuration`. The Core-Setup build will swap out the package-restored CoreFX files with local CoreFX files found from the nuspec for the Microsoft.Private.CoreFX.NETCoreApp package (since this is the only listing in CoreFX of where to find all of the files that make up the shared framework).

### DIFF
--- a/Documentation/building/local-coreclr-corefx.md
+++ b/Documentation/building/local-coreclr-corefx.md
@@ -1,0 +1,29 @@
+# Build Core-Setup with a Local CoreCLR or CoreFX
+
+## Testing with private CoreCLR bits
+
+Generally the Core-Setup build system gets the CoreCLR from a NuGet package which gets pulled down and correctly copied to the various output directories by building `src\pkg\projects\netcoreapp\src\netcoreapp.depproj` which gets built as part of `build.cmd/sh`. For folks that want to do builds and test runs in corefx with a local private build of coreclr you can follow these steps:
+
+1. Build CoreCLR and note your output directory. Ex: `\coreclr\bin\Product\Windows_NT.x64.Release\` Note this will vary based on your OS/Architecture/Flavor and it is generally a good idea to use Release builds for CoreCLR when building Core-Setup and the OS and Architecture should match what you are building in Core-Setup.
+2. Build Core-Setup either passing in the `CoreCLROverridePath` property or setting it as an environment variable:
+
+```batch
+build.cmd /p:CoreCLROverridePath=d:\git\coreclr\bin\Product\Windows_NT.x64.Release\
+```
+
+By convention the project will look for PDBs in a directory under `$(CoreCLROverridePath)/PDB` and if found will also copy them. If not found no PDBs will be copied. If you want to explicitly set the PDB path then you can pass `CoreCLRPDBOverridePath` property to that PDB directory.
+Once you have updated your CoreCLR you can run tests however you usually do and it should be using your copy of CoreCLR.
+If you prefer, you can use a Debug build of System.Private.CoreLib, but if you do you must also use a Debug build of the native portions of the runtime, e.g. coreclr.dll. Tests with a Debug runtime will execute much more slowly than with Release runtime bits.
+
+## Testing with private CoreFX bits
+
+Similarly to how Core-Setup consumes CoreCLR, Core-Setup also consumes CoreFX from a NuGet package which gets pulled down and correctly copied to the various output directories by building `src\pkg\projects\netcoreapp\src\netcoreapp.depproj` which gets built as part of `build.cmd/sh`. To test with a local private build of CoreFX, you can follow these steps:
+
+1. Build CoreFX and note your configuration (Ex. Debug, Checked, Release).
+2. Build Core-Setup either passing in the `CoreFXOverridePath` property or setting it as an environment variable (assuming your CoreFX repo is located at `d:\git\corefx`):
+
+```batch
+build.cmd /p:CoreFXOverridePath=d:\git\corefx\artifacts\packages\Debug
+```
+
+The Core-Setup build will resolve the correct libraries to use to build the shared framework based on the restored packages and will locate the files in the output directory based on the nuspec file of the shared-framework package in the local CoreFX build.
--- a/Documentation/building/local-coreclr-corefx.md
+++ b/Documentation/building/local-coreclr-corefx.md
@@ -13,7 +13,7 @@ build.cmd /p:CoreCLROverridePath=d:\git\coreclr\bin\Product\Windows_NT.x64.Relea
 
 By convention the project will look for PDBs in a directory under `$(CoreCLROverridePath)/PDB` and if found will also copy them. If not found no PDBs will be copied. If you want to explicitly set the PDB path then you can pass `CoreCLRPDBOverridePath` property to that PDB directory.
 Once you have updated your CoreCLR you can run tests however you usually do and it should be using your copy of CoreCLR.
-If you prefer, you can use a Debug build of System.Private.CoreLib, but if you do you must also use a Debug build of the native portions of the runtime, e.g. coreclr.dll. Tests with a Debug runtime will execute much more slowly than with Release runtime bits.
+If you prefer, you can use a Debug build of System.Private.CoreLib, but if you do you must also use a Debug build of the native portions of the runtime, e.g. coreclr.dll. Tests with a Debug runtime will execute much more slowly than with Release runtime bits. This override is only enabled for building the NETCoreApp shared framework.
 
 ## Testing with private CoreFX bits
 
@@ -26,4 +26,4 @@ Similarly to how Core-Setup consumes CoreCLR, Core-Setup also consumes CoreFX fr
 build.cmd /p:CoreFXOverridePath=d:\git\corefx\artifacts\packages\Debug
 ```
 
-The Core-Setup build will resolve the correct libraries to use to build the shared framework based on the restored packages and will locate the files in the output directory based on the nuspec file of the shared-framework package in the local CoreFX build.
+The Core-Setup build will resolve the correct libraries to use to build the NETCoreApp shared framework based on the restored packages and will locate the files in the output directory based on the nuspec file of the shared-framework package in the local CoreFX build. This override does not enable using a local CoreFX build when creating the WindowsDesktop shared framework.
--- a/Documentation/building/local-coreclr-corefx.md
+++ b/Documentation/building/local-coreclr-corefx.md
@@ -8,7 +8,7 @@ Generally the Core-Setup build system gets the CoreCLR from a NuGet package whic
 2. Build Core-Setup either passing in the `CoreCLROverridePath` property or setting it as an environment variable:
 
 ```batch
-build.cmd /p:CoreCLROverridePath=d:\git\coreclr\bin\Product\Windows_NT.x64.Release\
+build.cmd /p:CoreCLROverridePath=d:\git\coreclr\bin\Product\Windows_NT.x64.Release
 ```
 
 By convention the project will look for PDBs in a directory under `$(CoreCLROverridePath)/PDB` and if found will also copy them. If not found no PDBs will be copied. If you want to explicitly set the PDB path then you can pass `CoreCLRPDBOverridePath` property to that PDB directory.
--- a/src/pkg/packaging-tools/framework.dependency.targets
+++ b/src/pkg/packaging-tools/framework.dependency.targets
@@ -40,7 +40,8 @@
       <!-- include all doc files -->
       <_docFilesToPackage Include="%(FilesToPackage.RootDir)%(FilesToPackage.Directory)**\%(FilesToPackage.FileName).xml" />
 
-      <_docFilesToPackage Condition="'$(CoreFXOverridePath)' != ''" Include="@(FilesToPackage->'$(CoreFXOverridePath)/../../bin/docs/%(FileName).xml')" />
+      <_coreFXOverrideDocFiles Condition="'$(CoreFXOverridePath)' != ''" Include="@(FilesToPackage->'$(CoreFXOverridePath)/../../bin/docs/%(FileName).xml')" />
+      <_docFilesToPackage Include="@(_coreFXOverrideDocFiles)" Condition="Exists('%(Identity)')" />
 
       <FilesToPackage Include="@(_docFilesToPackage)">
         <TargetPath Condition="'%(_docFilesToPackage.TargetPath)' != ''">ref/$(PackageTargetFramework)/%(RecursiveDir)</TargetPath>
--- a/src/pkg/packaging-tools/framework.dependency.targets
+++ b/src/pkg/packaging-tools/framework.dependency.targets
@@ -53,6 +53,31 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="OverrideRuntimeFilesFromPackageResolve" Condition="'$(CoreCLROverridePath)' != ''" BeforeTargets="GetFilesFromPackageResolve">
+    <Error Condition="!Exists('$(CoreCLROverridePath)')" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not exist." />
+    <PropertyGroup>
+      <CoreCLRPDBOverridePath Condition="'$(CoreCLRPDBOverridePath)' == '' and Exists('$(CoreCLROverridePath)/PDB')">$(CoreCLROverridePath)/PDB</CoreCLRPDBOverridePath>
+    </PropertyGroup>
+    <ItemGroup>
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/*.*" />
+      <CoreCLRFiles Condition="'$(CoreCLRPDBOverridePath)' != ''" Include="$(CoreCLRPDBOverridePath)/*.pdb;$(CoreCLRPDBOverridePath)/*.dbg" />
+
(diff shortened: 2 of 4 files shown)


Documentation/building/local-coreclr-corefx.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Testing with private CoreCLR bits
+
+Generally the Core-Setup build system gets the CoreCLR from a NuGet package which gets pulled down and correctly copied to the various output directories by building `src\pkg\projects\netcoreapp\src\netcoreapp.depproj` which gets built as part of `build.cmd/sh`. For folks that want to do builds and test runs in corefx with a local private build of coreclr you can follow these steps:
+
+1. Build CoreCLR and note your output directory. Ex: `\coreclr\bin\Product\Windows_NT.x64.Release\` Note this will vary based on your OS/Architecture/Flavor and it is generally a good idea to use Release builds for CoreCLR when building Core-Setup and the OS and Architecture should match what you are building in Core-Setup.
+2. Build Core-Setup either passing in the `CoreCLROverridePath` property or setting it as an environment variable:
+
+```batch
+build.cmd /p:CoreCLROverridePath=d:\git\coreclr\bin\Product\Windows_NT.x64.Release\
```
dagood: Does that trailing `\` work? The other command doesn't have one and the code adds another `\` so I'm guessing this is just out of date. (In my experience another alternative is switching the trailing `\` to `/` for Windows.)

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -40,8 +40,11 @@
       <!-- include all doc files -->
       <_docFilesToPackage Include="%(FilesToPackage.RootDir)%(FilesToPackage.Directory)**\%(FilesToPackage.FileName).xml" />
 
+      <_coreFXOverrideDocFiles Condition="'$(CoreFXOverridePath)' != ''" Include="@(FilesToPackage->'$(CoreFXOverridePath)/../../bin/docs/%(FileName).xml')" />
```
dagood: I'd prefer if `CoreFXOverridePath` were normalized rather than having to keep track of this prop being nonstandard.

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -40,8 +40,11 @@
...
+      <_docFilesToPackage Include="@(_coreFXOverrideDocFiles)" Condition="Exists('%(Identity)')" />
+
       <FilesToPackage Include="@(_docFilesToPackage)">
-        <TargetPath>ref/$(PackageTargetFramework)/%(RecursiveDir)</TargetPath>
+        <TargetPath Condition="'%(_docFilesToPackage.TargetPath)' != ''">ref/$(PackageTargetFramework)/%(RecursiveDir)</TargetPath>
```
dagood: What is this condition doing? I don't remember what the state of these items would be well enough to figure it out.

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
 
+  <Target Name="OverrideRuntimeFilesFromPackageResolve" Condition="'$(CoreCLROverridePath)' != ''" BeforeTargets="GetFilesFromPackageResolve">
+    <Error Condition="!Exists('$(CoreCLROverridePath)')" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not exist." />
+    <PropertyGroup>
+      <CoreCLRPDBOverridePath Condition="'$(CoreCLRPDBOverridePath)' == '' and Exists('$(CoreCLROverridePath)/PDB')">$(CoreCLROverridePath)/PDB</CoreCLRPDBOverridePath>
+    </PropertyGroup>
+    <ItemGroup>
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/*.*" />
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/Redist/**/*.dll" />
+      <CoreCLRFiles Condition="'$(CoreCLRPDBOverridePath)' != ''" Include="$(CoreCLRPDBOverridePath)/*.pdb;$(CoreCLRPDBOverridePath)/*.dbg" />
```
dagood: Missing `[.dylib].dwarf` for macOS?

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+
+      <OverriddenRuntimeFiles Include="@(ReferenceCopyLocalPaths)" Condition="'@(CoreCLRFiles->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)'">
+        <CoreCLRFile>@(CoreCLRFiles)</CoreCLRFile>
+      </OverriddenRuntimeFiles>
+
+      <LongNameDacFile Include="@(ReferenceCopyLocalPaths)" Condition="$([System.String]::Copy('%(FileName)').StartsWith('mscordaccore_'))">
+        <TargetPath>%(PathInPackage)</TargetPath>
+      </LongNameDacFile>
+
+      <ShortNameDacFileOverride Include="@(CoreCLRFiles)" Condition="'%(FileName)%(Extension)' == 'mscordaccore.dll'" />
```
dagood: Need the same for SOS and long-name SOS?

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+      <ReferenceCopyLocalPaths Remove="@(OverriddenRuntimeFiles)" />
+      <ReferenceCopyLocalPaths Include="@(OverriddenRuntimeFiles->Metadata('CoreCLRFile'))" />
+      
+    </ItemGroup>
+
+    <Error Condition="'@(CoreCLRFiles)' == ''" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not contain any files." />
+  </Target>
+
+  <PropertyGroup>
+    <CoreFXSharedFrameworkPackageName>Microsoft.Private.CoreFx.NETCoreApp</CoreFXSharedFrameworkPackageName>
```
dagood: Why out here rather than in the target? (And why in the targets file rather than props file, if a static property is actually necessary?)

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+
+  <Target Name="OverrideFrameworkFilesFromPackageResolve" Condition="'$(CoreFXOverridePath)' != ''" BeforeTargets="GetFilesFromPackageResolve">
+    <Error Condition="!Exists('$(CoreFXOverridePath)')" Text="The path provided to CoreFXOverridePath ($(CoreFXOverridePath)) does not exist." />
+
+    <PropertyGroup>
+      <CoreFXSharedFrameworkPackageSpec>$(CoreFXOverridePath)/specs/$(CoreFXSharedFrameworkPackageName).nuspec</CoreFXSharedFrameworkPackageSpec>
+      <CoreFXSharedFrameworkImplementationPackageSpec Condition="'$(PackageRID)' != ''">$(CoreFXOverridePath)/specs/runtime.$(PackageRID).$(CoreFXSharedFrameworkPackageName).nuspec</CoreFXSharedFrameworkImplementationPackageSpec>
+    </PropertyGroup>
+
+
```
dagood: (Nit) Extra line

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+      </OverriddenRuntimeFiles>
+
+      <OverriddenRuntimeFiles>
+        <IsNative Condition="$([System.String]::new('%(Identity)').ToLowerInvariant().Replace('\', '/').Contains('/native/'))">true</IsNative>
+      </OverriddenRuntimeFiles>
+
+      <ReferenceCopyLocalPaths Remove="@(OverriddenRuntimeFiles)" />
+      <ReferenceCopyLocalPaths Include="@(OverriddenRuntimeFiles->Metadata('CoreCLRFile'))" />
+      
+    </ItemGroup>
```
dagood: (Nit) Extra line before close tag.

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+
+    <PropertyGroup>
+      <CoreFXSharedFrameworkPackageSpec>$(CoreFXOverridePath)/specs/$(CoreFXSharedFrameworkPackageName).nuspec</CoreFXSharedFrameworkPackageSpec>
+      <CoreFXSharedFrameworkImplementationPackageSpec Condition="'$(PackageRID)' != ''">$(CoreFXOverridePath)/specs/runtime.$(PackageRID).$(CoreFXSharedFrameworkPackageName).nuspec</CoreFXSharedFrameworkImplementationPackageSpec>
+    </PropertyGroup>
+
+    
+    <Error Condition="!Exists('$(CoreFXSharedFrameworkPackageSpec)')" Text="The nuspec for the reference Microsoft.Private.CoreFx.NETCoreApp package does not exist." />
+    <Message Condition="'$(CoreFXSharedFrameworkImplementationPackageSpec)' != '' And !Exists('$(CoreFXSharedFrameworkImplementationPackageSpec)')"
+      Text="The nuspec for the implementation runtime.$(PackageRID).Microsoft.Private.CoreFx.NETCoreApp package does not exist. Falling back to packaged CoreFX." />
```
dagood: When is it desirable to let this happen? Would it be better to error out to catch the maybe more common case of the CoreFX build failing to produce it, but allow passing another property if it's truly what you want?

Should the message have `Importance="High"` so the user has a chance of seeing it in the logs?

(All depends on scenario, I guess.)

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+      <Output TaskParameter="Result" ItemName="CoreFXReferenceItems" />
+    </XmlPeek>
+
+    <XmlPeek
+      Condition="'$(CoreFXSharedFrameworkImplementationPackageSpec)' != '' And Exists($(CoreFXSharedFrameworkImplementationPackageSpec))"
+      Namespaces="&lt;Namespace Prefix='nuget' Uri='http://schemas.microsoft.com/packaging/2013/01/nuspec.xsd' /&gt;"
+      Query="nuget:package/nuget:files/nuget:file/@src"
+      XmlInputPath="$(CoreFXSharedFrameworkImplementationPackageSpec)">
+      <Output TaskParameter="Result" ItemName="CoreFXReferenceCopyLocalItems" />
+    </XmlPeek>
```
dagood: I imagine there's improvement to be made here... CoreFX laying out these files in a way Core-Setup can consume, and the CoreFX nuget package being phased out.

src/pkg/packaging-tools/framework.dependency.targets
```diff
@@ -53,6 +56,100 @@
...
+    
+    <ItemGroup>
+      <CoreFXReferenceItems NuGetPackageId="Microsoft.Private.CoreFx.NETCoreApp" />
+      <CoreFXReferenceCopyLocalItems NuGetPackageId="runtime.$(PackageRID).Microsoft.Private.CoreFx.NETCoreApp" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <OverriddenFrameworkReferenceFiles Include="@(Reference)" Condition="'@(CoreFXReferenceItems->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)' And '@(CoreFXReferenceItems->'%(NuGetPackageId)')' == '%(NuGetPackageId)'">
```
dagood: Can you split this? (I keep lines to 100 columns when I can personally, but for this PR, this and `OverriddenFrameworkImplementationFiles` are the only bits I think are really affecting readability.)

```xml
<OverriddenFrameworkReferenceFiles
  Include="@(Reference)"
  Condition="
    '@(CoreFXReferenceItems->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)' And
    '@(CoreFXReferenceItems->'%(NuGetPackageId)')' == '%(NuGetPackageId)'">
```

src/pkg/packaging-tools/framework.dependency.targets
dagood: Whoops, nope, forgot SOS is no longer built in CoreCLR.

src/pkg/packaging-tools/framework.dependency.targets
jkoritzinsky: If we normalize `CoreFXOverridePath` to the artifacts directory, we'd need a way to specify which configuration of CoreFX to use. We could just require that we use a matching configuration of CoreFX, but I don't know if that's what we want to do here.

src/pkg/packaging-tools/framework.dependency.targets
jkoritzinsky: Yes that is something we'll want to do once we consolidate the repos. This just happens to be the only way that works at the moment with how the CoreFX build works.

src/pkg/packaging-tools/framework.dependency.targets
dagood: Wow, reading this back, I was very unclear. I only meant how this is `$(CoreFXOverridePath)/..` rather than the conventional `$(CoreFXOverridePath)..`. I guess I was in the mindset of slashes because I just wrote the comment above. I think the rest of the line is fine for this point in time. Sorry.

src/pkg/packaging-tools/framework.dependency.targets
dagood: I see, I think this is fine then. Nice to have the indicator in the binlog when digging deep but not something that needs to be shown during every build. Thanks.

Side note, since source-build *needs* to do this without other platforms' output, it disables the feature that makes this target run on other RIDs:

https://github.com/dotnet/core-setup/blob/ce971b238196b08e9423bf9a5bfd192b8f7e1e0b/src/pkg/packaging-tools/framework.dependency.targets#L148-L149

That solution is a gap in the artifact, so probably not something we can do in the Microsoft build without decent planning.

Documentation/building/local-coreclr-corefx.md
jkoritzinsky: Yep this is out of date. I copied this from the CoreFX docs initially.

src/pkg/packaging-tools/framework.dependency.targets
jkoritzinsky: I think this was from a different experiment I tried earlier. I'll remove it.

jkoritzinsky:
Re the fact that this is moving to a shared SDK in Arcade: we might want to extract out the override infra I'm adding here to be a separate local targets file. I can do that in this PR if you want. I don't know if we need the ability to manually crossgen winforms/wpf against a local CoreCLR/CoreFX once the windows-desktop portion of this repo is moved out. That way, any changes we have to make to this override infra after repo consolidation wouldn't have to be propagated through from arcade.

dagood:
I like that idea, good note about the Arcade propagation delay.

src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
```diff
@@ -0,0 +1,118 @@
...
+  <Target Name="OverrideRuntimeFilesFromPackageResolve" Condition="'$(CoreCLROverridePath)' != ''" BeforeTargets="GetFilesFromPackageResolve">
+    <Error Condition="!Exists('$(CoreCLROverridePath)')" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not exist." />
+    <PropertyGroup>
+      <CoreCLRPDBOverridePath Condition="'$(CoreCLRPDBOverridePath)' == '' and Exists('$(CoreCLROverridePath)/PDB')">$(CoreCLROverridePath)/PDB</CoreCLRPDBOverridePath>
+    </PropertyGroup>
+    <ItemGroup>
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/*.*" />
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/Redist/**/*.dll" />
+      <CoreCLRFiles Condition="'$(CoreCLRPDBOverridePath)' != ''"
+        Include="$(CoreCLRPDBOverridePath)/*.pdb;$(CoreCLRPDBOverridePath)/*.dbg;$(CoreCLRPDBOverridePath)/*.dylib;$(CoreCLRPDBOverridePath)/*.dylib.dwarf" />
```
dagood: I guess I wasn't clear here--`.dylib` => `.dylib.dwarf` is like `.so` => `.so.dbg`, so this is what I meant:

```suggestion
        Include="$(CoreCLRPDBOverridePath)/*.pdb;$(CoreCLRPDBOverridePath)/*.dbg;$(CoreCLRPDBOverridePath)/*.dwarf" />
```

jkoritzinsky:
I've updated this PR to scope all of the changes that enable the override to only the NETCoreApp shared framework. This should enable moving the shared-framework generation code to the SDK easily and allow us to iterate on the implementation of the NETCoreApp usage of local builds without having to go through arcade.

jkoritzinsky:
cc: @jaredpar @jashook

Issue #8422, opened 2019-10-02
PublishSingleFile generates corrupted file


marek-safar:
I've tried to use `PublishSingleFile` for building my project targetting `win-x64` RID, unfortunately, despite of generating over 64 MB file, it failed to run with:

```
Failure processing application bundle; possible file corruption.
Invalid FileEntry detected.
A fatal error was encountered. Could not extract contents of the bundle
```

The same issue happens when trying to target other RIDs, such as `linux-x64`. The resulting app also doesn't launch there.

I'm not sure what else I can provide to help tackle down this issue, so please feel free to clone my **[repo](https://github.com/JustArchiNET/ArchiSteamFarm)**, switch to `mono-linker-crash-tests` branch and `dotnet publish ArchiSteamFarm -c Release -o out -r win-x64 -f netcoreapp3.0 /p:PublishSingleFile=true`. Thank you.

_Moved from https://github.com/mono/linker/issues/763_

/cc @JustArchi

vitek-karas:
The bug is that the bundle extractor fails on empty files (glaring oversight from us):
https://github.com/dotnet/core-setup/blob/a73fff116b052c187c231bf6664f389c5b42cc8f/src/corehost/cli/apphost/bundle/file_entry.cpp#L14

This checks that the file size is `>0` which is wrong, it should be `>=0`.

In the repro project, the bundle contains several files `.gitkeep` which are empty. These are added with a wildcard as `Content` items to the project. The bundler will correctly bundle them into single-exe, but at runtime it fails on the above check.

Simple repro:
```
dotnet new console
echo. 2>empty.txt
```

Add this to the project file:
``` XML
  <ItemGroup>
    <Content Include="empty.txt">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Content>
  </ItemGroup>
```

Now
```
dotnet publish -r win-x64 --self-contained false /p:PublishSingleFile=true
<run the produced .exe>
```

JustArchi:
Thank you a lot for finding the root cause, I appreciate your help with this 👍.

vitek-karas:
@JustArchi Thanks a lot for reporting this - I must admit I have no idea how we missed it - but it's great we know now, and can fix it.
I'll try to take the fix for 3.1 (hopefully will be approved). But it definitely won't make it into 3.0 in any way or form.

[...]

PR #8462, opened 2019-10-04
Fix bundle failing to extract empty file.

vitek-karas:
The file entry validity check marked empty files as invalid. This is wrong, the file size must be non-negative number.

Added a test and fixed the condition.

This is the master version of a fix for #8422. Once merged I will port this to 3.1


swaroop-sridhar:
@vitek-karas  did you trip on this scenario on a real app or test?

PR #8471, opened 2019-10-04
BundleExtraction: Determine extraction path based on writability.

### LINKED ISSUE #7940: Single EXE extractor path w/ AWS Lambda
## Steps to reproduce
Make a .net binary with /p:PublishSingleFile=true. Place this binary in an aws lambda bundle and invoke it. 

## Expected  behavior
It should extract all deps to a temporary directory and execute.

## Actual behavior
It attempts to extract to /var/tmp but fails as only /tmp is writable in lambda.

## Suggestion to fix
To get out of the box runability without setting DOTNET_BUNDLE_EXTRACT_BASE_DIR i propose to change this to /tmp/.net . Or perhaps there should be multiple attempts at common locations (/var/tmp, /tmp, same directory as binary, etc)

swaroop-sridhar:
## Summary 
On Unix systms where $TMPDIR is not set, temp-directory is currently
determined as:

/var/tmp/... (if it exists)
/tmp (if not)

However, this causes failures on systems where /var/tmp exists, but is not
writable (as is the case in AWS lambda).

This change fixes the issue by adding a writability check.

## Testing
* Tested on Ubuntu by turning off write permissions on `/var/tmp` the app works by extracting to `/tmp` instead,
* Once the change is checked in, and an SDK is built with it, I'll request @stevemk14ebr to verify that the fix solves the problem in executing hos AWS lambda function.

## Further considerations
* Once the fix is verified, I intent to port the change to 3.1 branch.
* A further consideration was to add the current directory (or the directory where the single-file app resides) to the list of candidate directories where extraction is performed. 
    * However this change was deemed significantly depart from 3.0 behavior (ex: directory containing extracted files shows up on desktop), and not a suitable in 3.1 (bugfix) release.
    * This change can be considered for .net 5, if extraction is still necessary at that point.

## Tracking
Fixes #7940

### DIFF
--- a/src/corehost/cli/apphost/bundle/extractor.cpp
+++ b/src/corehost/cli/apphost/bundle/extractor.cpp
@@ -23,6 +23,7 @@ void extractor_t::determine_extraction_dir()
         {
             trace::error(_X("Failure processing application bundle."));
             trace::error(_X("Failed to determine location for extracting embedded files."));
+            trace::error(_X("DOTNET_BUNDLE_EXTRACT_BASE_DIR is not set, and temp-directory doesn't exist or is not readable/writable."));
             throw StatusCode::BundleExtractionFailure;
         }
 
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -297,26 +297,32 @@ bool pal::get_default_servicing_directory(string_t* recv)
     return true;
 }
 
+bool is_read_write_able_directory(pal::string_t& dir)
+{
+    return pal::realpath(&dir) &&
+           (access(dir.c_str(), R_OK | W_OK | X_OK) == 0);
+}
+
 bool pal::get_temp_directory(pal::string_t& tmp_dir)
 {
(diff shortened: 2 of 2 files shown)


swaroop-sridhar:
> We should add a test for this - I suspect we don't have any for the entire fallback logic...
> The `TMPDIR` to `/var/tmp` fallback should be relatively easy to test, so we should do that. And also testing the writebility of `TMPDIR` should be reasonably simple, so we should add that. The other combinations would require containers/root access (since we would be making system-wide changes).

@vitek-karas, we actually don't fallback from `$TMPDIR` to any other directory. POSIX compatible OSes have  `$TMPDIR` set (OSX sets it system-wide).  I added the writability check here for symmetry, but even if we don't check here, extraction will fail in the next step with IO error.

The only fallback is `/var/tmp` to `/tmp` -- which is hard to test on the OS-configurations used in the lab.

`chmod` on `$TMPDIR` or `/var/tmp` requires root access. So, I didn't add this test.

Issue #5732, opened 2019-04-08
Move WindowsDesktop shared framework generation into another repo


dagood:
https://github.com/dotnet/core-setup/issues/5219 tracks moving WindowsDesktop gathering into Core-Setup, but it's mostly for pragmatic reasons. Targeting pack, runtime pack, and shared framework tooling exists in Core-Setup today and is not easily shareable, and using Core-Setup allows us to dissolve DotNet-Trusted in a reasonable timeframe.

Once tooling is shareable (#4773, #4776, ...?) the WindowsDesktop shared framework should be built in a different repo. Broadly speaking, this is similar to how the ASP.NET Core shared framework builds in a different repo.

This specifically allows testing against the current netcoreapp shared framework in WinForms and WPF. Both shared frameworks being built by one repo introduces a testing cycle, so the dependency has to be some N-1 (or N-x). If WPF and WinForms feed into a separate repo, this breaks the cycle:

```
+-> wpf +------------+
|                    v
+-> winforms +--> core-setup +
|                    ^       |
+-> corefx +---------+       |
|                            |
+-----netcore-sfx------------+


corefx +--> core-setup +--> wpf +----+-> windowsdesktop
  ^                    |             |
  +---netcore-sfx------+--> winforms +
```

(Right now it's a bit worse than the diagram: WPF and WinForms test on the netcoreapp in the CLI output, which is normally even older.)

The most impactful aspect of this is how it will help the WPF and WinForms tests. Thanks @ericstj for pointing this out.

/cc @zsd4yr @vatsan-madhavan

dagood:
It turns out this also helps with branching... https://github.com/dotnet/core-setup/issues/7343.

vatsan-madhavan:
Is the plan to refactor and expose the tooling through the Arcade SDK?

dagood:
Yep, https://github.com/dotnet/core-setup/issues/4787 tracks that. I'm hoping we can get past the point where splitting a repo means duplicating all the infra inside it (like CLI => Core-Setup all that time ago).

dagood:
@vatsan-madhavan, @zsd4yr, and I chatted about this offline, **dotnet/windowsdesktop** seems like a reasonable new repo name.

This also unlocks some other improvements that can be made over time: there's some code that would be much better off moved to dotnet/windowsdesktop, like project templates and the windowsdesktop SDK. (Hence no `-setup` suffix!)

dagood:
[...]

Issue #4787, opened 2018-11-19
Centralize  common setup related tools and targets to Arcade


dleeapho:

natemcmaster:
👀 aspnet might benefit from this too. We have a set of setup related tools for building the aspnetcore shared framework and installers.

dagood:
Merged with https://github.com/dotnet/core-setup/pull/8547.

PR #8494, opened 2019-10-07
Remove WindowsDesktop from Core-Setup, now dotnet/windowsdesktop

dagood:
This shared framework will now be produced from dotnet/windowsdesktop instead of this repo. https://github.com/dotnet/core-setup/issues/5732.

The removed files here were already copied over by https://github.com/dotnet/windowsdesktop/pull/1/.

 (Note: https://github.com/dotnet/windowsdesktop will be opened very soon, and may show up as a 404 until then if you don't have permissions.)

I have an example Core-SDK PR to change where it gets the WD sharedfx from from over at https://github.com/dotnet/core-sdk/pull/5161.


dagood:
> https://github.com/dotnet/core-setup/blob/6f12a57e3c1a5e0bc592745b2648b041819ae4ed/src/test/Microsoft.DotNet.CoreSetup.Packaging.Tests/WindowsDesktopTests.cs

Already deleted this with the second commit. 😄 https://github.com/dotnet/core-setup/pull/8494/commits/4178fd8f2f7cc8bfe47ad7215e132a4e0890afce Leaving this in actually causes a failure, as it should.

> core-setup/src/pkg/packaging-tools/framework.dependency.targets

This file will be deleted when I switch Core-Setup to the tooling package, no need to clean up refs in that dir. https://github.com/dotnet/core-setup/issues/4787

dagood:
> (Note: https://github.com/dotnet/windowsdesktop will be opened very soon, and may show up as a 404 until then if you don't have permissions.)

To be specific, I plan to open the WD repo before merging this PR.

PR #8497, opened 2019-10-08
Specify sdk version, use nuget cache

ViktorHofer:
Submitting targeted PRs to align core-setup, corefx and coreclr.

- Use the user's nuget cache for local builds 
Other dotnet repositories like CoreFx are already using the user's
nuget cache for local builds vs the repo local cache for CI builds
and no problems were reported. Switching to remove differences in builds
- Enforce a minimum dotnet sdk version
Specify a minimum dotnet sdk version in the sdk section global.json
which is honored by the Desktop msbuild as well. Set a roll-forward
policy so that a globally installed sdk with a more recent version can
be used when invoking a project directly.
- Use global dotnet installation if matching
Arcade guarantees that the requested version specified in global.json
is installed.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,4 +1,9 @@
 {
+  "sdk": {
+    "version": "3.0.100",
+    "allowPrerelease": true,
+    "rollForward": "major"
+  },
   "tools": {
     "dotnet": "3.0.100"
   },
--- a/eng/configure-toolset.ps1
+++ b/eng/configure-toolset.ps1
@@ -1,7 +1,3 @@
-# Disable using the globally installed SDK. Using the global install can cause
-# roll-forward to a newer SDK that may not work.
-$script:useInstalledDotNetCli = $false
-
 # Always use the local repo packages directory instead of the user's NuGet cache
 # to keep the same between "ci" and non-"ci" builds. If the efficiency gain is
 # required and it's worth maintaining the different types of build, this can be
--- a/eng/configure-toolset.ps1
+++ b/eng/configure-toolset.ps1
@@ -1,5 +1,0 @@
-# Always use the local repo packages directory instead of the user's NuGet cache
-# to keep the same between "ci" and non-"ci" builds. If the efficiency gain is
-# required and it's worth maintaining the different types of build, this can be
-# removed.
-$script:useGlobalNuGetCache = $false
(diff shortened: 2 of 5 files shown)


ViktorHofer:
@jkoritzinsky do you think you can help with this change? I don't have enough context to diagnose the failing tests. I looks like there are assumptions that don't hold with my changes.

jkoritzinsky:
I have the fixes for the failing tests locally as part of porting back the changes for consolidation. I'll split them out and add them to your PR.

dagood:
FYI @vitek-karas about the new test-local `global.json`. Seems like a reasonable fix to me.

dagood:
> * Use global dotnet installation if matching
>   Arcade guarantees that the requested version specified in global.json
>   is installed.

My experience is that it does ensure it's installed, but it may use a different version anyway based on roll-forward rules. 😄 https://github.com/dotnet/arcade/issues/3764#issuecomment-532769966

This can cause a problem in Core-Setup in particular because we have somewhat deeper dependencies on the SDK targets than usual, so breaking changes are more frequent. Once time I updated VS and the new SDK broke my local build, for example.

jkoritzinsky:
To expand on the test-local `global.json`: the test expect that the roll-forward policy found by the tests is "latest". They break if the rollforward policy isn't a "latest*" style policy.

ViktorHofer:
> My experience is that it does ensure it's installed, but it may use a different version anyway based on roll-forward rules.

My experience is that it only picks the globally installed one if the version matches EXACTLY: https://github.com/dotnet/arcade/blob/master/eng/common/tools.ps1#L132. The roll-forward property in it is at the moment completely ignored by arcade. I doubt it would cause troubles but as a thoughtful developer I would never say never 😅

dagood:
Either way, I think it's ok to merge it because I think we generally want roll-forward to work and should make any reasonable fixes necessary. But when I read that script, I see it picking the globally installed *dotnet directory* when it detects *a* matching version of the SDK there, but later on it leaves it up to `dotnet` which SDK it eventually uses for the `dotnet msbuild` command.

ViktorHofer:
Right, because the PATH is mutated: https://github.com/dotnet/arcade/blob/master/eng/common/tools.ps1#L152

global.json
```diff
@@ -1,4 +1,9 @@
 {
+  "sdk": {
+    "version": "3.0.100",
```
ViktorHofer: This will require devs to have that version of a later version installed locally. In example when opening with VS or directly `dotnet msbuilding` a project. That shouldn't be a problem though.

Issue #8504, opened 2019-10-08
Include .NET Core Runtime in WindowsDesktop bundle


dagood:
To run a .NET Core WPF/WinForms app, you need to install both the .NET Core Runtime and WindowsDesktop Runtime. (As well as the correct x86/x64 bitness.) This is unexpected and is a significant pain point when trying to deploy FDE WinForms and WPF apps.

This issue tracks the incremental step of including the .NET Core Runtime dependency in the installer bundle for the WindowsDesktop Runtime.

This will bring our current matrix of 4 installers needed to run WindowsDesktop apps:

|   | x64 | x86 |
| --- | --- | --- |
| .NET Core Runtime | x | x |
| .NET Desktop Runtime | x | x |

down to 2:

|   | x64 | x86 |
| --- | --- | --- |
| .NET Desktop Runtime | x | x |

Then, later, we may be able to combine the x86/x64 ones as well by completing https://github.com/dotnet/core-setup/issues/7763. (This is more difficult from an implementation perspective.)

/cc @KathleenDollard @leecow @dleeapho @MichaelSimons @rowanmiller

KathleenDollard:
👍 Yeah!

dagood:
This is merged for 3.1 preview 2 with PR https://github.com/dotnet/core-setup/pull/8508!

Waiting to gather data/feedback before potentially porting to the 3.0 servicing branch as well. (Not tracking that in this issue.)

Implementing it in 5.0+ has to be done separately in the dotnet/windowsdesktop repo, https://github.com/dotnet/windowsdesktop/issues/20 tracks that.

cup:
https://github.com/dotnet/core-setup/issues/8862

Issue #8368, opened 2019-09-28
.NET Core runtime for Desktop does not include the actual runtime and it doesn't block or prompt for it


Symbai:
1. Created a plain .NET Core 3.0 WPF project
2. Rebuild in release mode
3. Copied the files in the release folder to a new computer which is missing .NET Core at all
![image](https://user-images.githubusercontent.com/14368203/65813024-c2d0a400-e1be-11e9-8fb0-c154b3c273ad.png)
4. Visited https://dotnet.microsoft.com/download on this computer and downloaded the `Download .NET Core Runtime for desktop apps` and installed it
5. Tried to start the program, nothing happens. 

The event viewer says:
```
Description: A .NET Core application failed.
Application: WpfApp12.exe
Path: C:\Users\TestAccount\Desktop\WpfApp12.exe
Message: A fatal error occurred. The required library hostfxr.dll could not be found.
If this is a self-contained application, that library should exist in [C:\Users\TestAccount\Desktop\].
If this is a framework-dependent application, install the runtime in the global location [C:\Program Files\dotnet] or use the DOTNET_ROOT environment variable to specify the runtime location or register the runtime location in [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x64\InstallLocation].
```

//edit: `dotnet publish` creates >290 files even when not self contained with an overall size of 150mb. My thought was that installing the runtime would install these >290 files and avoid having me to ship a 150mb application. Similar to .NET Framework. If that's not the case then: What exactly is the runtime installer for? And whats the purpose of self-contained when even the not self-contained application would force me to ship >290 files and >150mb?

danmosemsft:
Does `C:\Program Files\dotnet\host\fxr\3.0.0\hostfxr.dll` exist?

dagood:
> 4. Visited https://dotnet.microsoft.com/download on this computer and downloaded the `Download .NET Core Runtime for desktop apps` and installed it
> 5. Tried to start the program, nothing happens.

Did you also install the .NET Core Runtime? Missing it would explain hostfxr not existing and nothing happening. This message on the WindowsDesktop Runtime download page is meant to point to it:

![image](https://user-images.githubusercontent.com/12819531/65821256-cf71ee00-e1f8-11e9-9ad7-80498daa72f3.png)

https://github.com/dotnet/core-setup/issues/8363 is an earlier issue about this.

[...]

Issue #7763, opened 2019-08-18
Create WindowsDesktop + .NET Core support bundle


dagood:
Similar to the "ASP.NET Core/.NET Core Runtime & Hosting Bundle", there should be one for WindowsDesktop that sets up everything needed to run a WPF or WinForms app on .NET Core. It should include both x86 and x64, and both the .NET Core Runtime and Windows Desktop Runtime.

dagood:
We have new plans in 3.1 to use a single SDK installer with options to install various components including this runtime, with dependency information to simplify the process. @leecow, can you add a pointer to the issue where that's tracked (when available)?

With that plan, it's unclear whether this artifact is necessary--closing for now.

dagood:
There's some uncertainty about the 3.1 installer fulfilling this scenario, so I'm adding some detail about what makes this difficult:

The x86 and x64 installers are built in parallel on different machines. However, all the MSIs need to be present on a single machine during the official build to bundle them. There are two main ways I have on my mind to do this:
1. Add a "join" step to the build, where the MSIs are all downloaded from the parallel legs and assembled into the bundle.
    * This seems fairly low risk, because the join part of the build would be fairly isolated from the existing build.
2. Build x86 and x64 on the same machine.
   * I don't think this is a useful approach for the current situation.
      * I believe ASP.NET Core's Support Bundle uses this approach, but they have had time to validate it and used it for a while.
   * I have concerns that this will give us bad results (some build outputs not being differentiated per arch), and this adds considerable risk.
    * This also would roughly double the product build time.

---

There is an alternative that skips that... produce one support bundle per arch. Users would need to figure out whether they need x86 or x64, or install both, but it would be quick to implement in Core-Setup.

PR #8508, opened 2019-10-09
[release/3.1] Include .NET Core Runtime in WindowsDesktop bundle

dagood:
#### Description

https://github.com/dotnet/core-setup/issues/8504: Includes the .NET Core Runtime MSIs inside the .NET Core Desktop Bundle (WindowsDesktop).

#### Customer Impact

The current state of 3.0+ is that you need to install *both* the .NET Core Runtime Bundle and the .NET Core Desktop Bundle in order to run a WinForms or WPF FDE/FDD app. This causes a lot of confusion, because:
1. It seems normal to assume that the Desktop Bundle will give you all you need to run .NET Core Desktop apps, but it doesn't.
2. There is text on the dot.net site that tells you you also need to install .NET Core Runtime, but it is very easy to miss. https://github.com/dotnet/core-setup/issues/8368

These are solved by including .NET Core Runtime inside the .NET Core Desktop Bundle. The user still needs to select the correct architecture(s) to install, but this will be mitigated by creating an improved download page specific to the .NET Core Desktop Runtime Bundle. https://github.com/dotnet/core-setup/issues/7763 tracks also combining the architectures.

3. When you try to launch an FDE WPF/WinForms app without .NET Core Runtime installed, it fails, but doesn't inform the user (only writes to the event log). https://github.com/dotnet/core-setup/issues/8222

This becomes harder to hit because it won't be possible to run the .NET Core Desktop Bundle without also getting the .NET Core Runtime. However, the lack of feedback will still occur if the user tries to launch an FDE app without the prereq runtimes installed, so this is still an important issue.

#### Regression?

No.

#### Risk

Low. We have not fully tested and considered the effects of including the .NET Core Runtime with the .NET Core Desktop Bundle, however this is same approach as the SDK Bundle: the general principle is known and has worked for a long time. (**Edit: The comments explore some interesting scenarios, and they don't raise any concerns.**)

Tested by running `dotnet new winforms` and `dotnet new wpf` FDE apps on a VM with only the .NET Core Desktop Bundle installed, and both worked.

/cc @KathleenDollard @rowanmiller @vatsan-madhavan @richlander

### DIFF
--- a/src/pkg/projects/windowsdesktop/sfx/Microsoft.WindowsDesktop.App.Bundle.bundleproj
+++ b/src/pkg/projects/windowsdesktop/sfx/Microsoft.WindowsDesktop.App.Bundle.bundleproj
@@ -10,6 +10,11 @@
 
   <ItemGroup>
     <BundleComponentReference Include="Microsoft.WindowsDesktop.App.SharedFx.sfxproj" />
+
+    <!-- Include .NET Core Runtime MSIs. -->
+    <BundleComponentReference Include="..\..\netcoreapp\sfx\Microsoft.NETCore.App.SharedFx.sfxproj" />
+    <BundleComponentReference Include="..\..\Microsoft.NETCore.DotNetHostResolver\Microsoft.NETCore.DotNetHostResolver.pkgproj" />
+    <BundleComponentReference Include="..\..\Microsoft.NETCore.DotNetHost\Microsoft.NETCore.DotNetHost.pkgproj" />
   </ItemGroup>
 
   <ItemGroup>


vatsan-madhavan:
What happens if the bundle .NET Core runtime is already installed? 

- Does that portion of the setup become a no-op and the WindowsDesktop portion continue as before? 
- Does the bundled .NET Core runtime setup run again, and overwrite or repair the previous installation... 
- Something else?

vatsan-madhavan:
What happens if the WindowsDesktop runtime is already installed (say, using the dotnet install script, not the installer), but the bundled .NET Core runtime is not installed yet ? 

- Will the bundled .NET Core runtime get installed, but the WindowsDesktop runtime installer no-op and yield to the previously installed copy of the same version of WindowsDesktop runtime? 
- Will neither .NET Core runtime (bundled) nor the WindowsDesktop runtime be installed (the whole setup will yield to the fact that the exact same version of WindowsDesktop runtime is already installed)
- Both the bundled .NET Core runtime and the WindowsDesktop runtime will be installed. WindowsDesktop runtime will overwrite the one previously installed by dotnet install script. 
- Something else?

rladuca:
What are the implications (if any) of this with respect to building WindowsDesktop in the separate repo?  Do we not build installers there?  Does the runtime installer have to flow to the WindowsDesktop repo in some sort of loop?

vatsan-madhavan:
> What are the implications (if any) of this with respect to building WindowsDesktop in the separate repo? Do we not build installers there? Does the runtime installer have to flow to the WindowsDesktop repo in some sort of loop?

@dagood, am I correct in thinking that the the same core-setup build used to produce both WindowsDesktop and .NET Core installers previously, and now you are simply chaining two of those build artifacts together ?

dagood:
> What happens if the bundled .NET Core runtime is already installed?
> 
> * Does that portion of the setup become a no-op and the WindowsDesktop portion continue as before?

Yep, the way this works is that each MSI has its own identity and dependencies on each one are ref-counted. The .NET Core Runtime Bundle has one reference to it, and installing the .NET Core Desktop Bundle will add its own reference to those specific component MSIs.

This does have the consequence that if you upgrade WindowsDekstop but not .NET Core Runtime, then you end up with both the old and new one installed SxS. (This is the same as having an old Runtime Bundle present and installing a new SDK Bundle.)

> What happens if the WindowsDesktop runtime is already installed (say, using the dotnet install script, not the installer), but the bundled .NET Core runtime is not installed yet ?

The dotnet install script, to my knowledge, always writes to some user-writable dir (a repo, user profile, ...) and never interacts with the global install. I don't think this *exact* situation can happen. But I think it's similar if someone has, say, `3.1.0-alpha1` installed without this change and upgrades to `3.1.0-preview2` that does have this change.

> * Will the bundled .NET Core runtime get installed, but the WindowsDesktop runtime installer no-op and yield to the previously installed copy of the same version of WindowsDesktop runtime?

Yes, ref counting should sort this out fine in general.

> @dagood, am I correct in thinking that the the same core-setup build used to produce both WindowsDesktop and .NET Core installers previously, and now you are simply chaining two of those build artifacts together ?

That's right, I need to file an issue to reimplement this for 5.0+ where WindowsDesktop is split out.

> Does the runtime installer have to flow to the WindowsDesktop repo in some sort of loop?

There must be a flow from Core-Setup => WindowsDesktop, but this shouldn't cause a loop. This dependency is already required for other reasons anyway, like making sure the reference closure is valid.

dagood:
> [...] if someone has, say, `3.1.0-alpha1` installed without this change and upgrades to `3.1.0-preview2` that does have this change.

This is the situation we'll be in if we port this to `release/3.0`, (3.0.0 doesn't have this change, 3.0.1 would) so I tested it.
1. I built a bundle without this PR and another with it.
2. When I install the first one, I only have on disk the WindowsDesktop runtime.
3. When I then upgrade to the newer bundle, it installs the new version of both the .NET Core Runtime and WindowsDesktop, and uninstalls the old WindowsDesktop. (Per ref counts.)
   * We never ship multiple MSIs that contain the same runtime, which keeps ref counting reliable. On top of that, we always produce a new version of the runtimes in new builds, so I don't expect edge cases like mismatching runtime versions in a bundle anyway.

leecow:
Tactics approved

PR #8543, opened 2019-10-15
HostWriter: Remove signature on MAC host

swaroop-sridhar:
In order to conform with Catalina notarization requirements, the template apphost
will be shipped after codesigning.

This commit implements a change to remove the code-signature when customizing the apphost for an app.

Testing:
- Tested the SDK locally using a signed version of apphost
- Tested that unsigned host can be re-signed later
- Tested that stripping out the signature on apphost binary via HostModel produces exactly the same binary as codesign --remove-signature.

TODO:
* SDK changes for consuming the AppHostMachOFormatException
* Port the change to 3.1, backport to earlier versions
* For .net 5, factor out PE specific code in BinaryUtils into a separate file (similar to MacOUtil).

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
@@ -87,14 +87,26 @@ namespace Microsoft.NET.HostModel.AppHost
                 }
             }
 
+            void RemoveSignatureIfMachO()
+            {
+                MachOUtils.RemoveSignature(appHostDestinationFilePath);
+            }
+
+            void SetLastWriteTime()
+            {
+                // Memory-mapped write does not updating last write time
+                File.SetLastWriteTimeUtc(appHostDestinationFilePath, DateTime.UtcNow);
+            }
+
             try
             {
                 RetryUtil.RetryOnIOError(RewriteAppHost);
 
                 RetryUtil.RetryOnWin32Error(UpdateResources);
 
-                // Memory-mapped write does not updating last write time
-                RetryUtil.RetryOnIOError(() => File.SetLastWriteTimeUtc(appHostDestinationFilePath, DateTime.UtcNow));
+                RetryUtil.RetryOnIOError(RemoveSignatureIfMachO);
+
+                RetryUtil.RetryOnIOError(SetLastWriteTime);
             }
             catch (Exception ex)
             {
--- a/src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
@@ -1,0 +1,310 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Microsoft.NET.HostModel.AppHost
+{
+    internal static class MachOUtils
+    {
+        // The MachO Headers are copied from 
+        // https://opensource.apple.com/source/cctools/cctools-870/include/mach-o/loader.h
+        //
+        // The data fields and enumerations match the structure definitions in the above file,
+        // and hence do not conform to C# CoreFx naming style.
+
+        enum Magic : uint
+        {
+            MH_MAGIC = 0xfeedface,
+            MH_CIGAM = 0xcefaedfe,
+            MH_MAGIC_64 = 0xfeedfacf,
+            MH_CIGAM_64 = 0xcffaedfe
+        }
+
+        enum FileType : uint
+        {
+            MH_EXECUTE = 0x2
+        }
+
+#pragma warning disable 0649
+        struct MachHeader
+        {
+            public Magic magic;
+            public int cputype;
+            public int cpusubtype;
+            public FileType filetype;
+            public uint ncmds;
+            public uint sizeofcmds;
+            public uint flags;
+            public uint reserved;
+
+            public bool Is64BitExecutable()
+            {
+                return magic == Magic.MH_MAGIC_64 && filetype == FileType.MH_EXECUTE;
+            }
+
+            public bool IsValid()
+            {
+                switch (magic)
+                {
+                    case Magic.MH_CIGAM:
+                    case Magic.MH_CIGAM_64:
+                    case Magic.MH_MAGIC:
+                    case Magic.MH_MAGIC_64:
+                        return true;
+
+                    default:
+                        return false;
+                }
+            }
+        }
+
+        enum Command : uint
+        {
+            LC_SYMTAB = 0x2,
+            LC_SEGMENT_64 = 0x19,
+            LC_CODE_SIGNATURE = 0x1d,
+        }
+
+        struct LoadCommand
+        {
+            public Command cmd;
+            public uint cmdsize;
+        }
+
+        // The linkedit_data_command contains the offsets and sizes of a blob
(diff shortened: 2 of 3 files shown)


swaroop-sridhar:
@wli3  Please review.

swaroop-sridhar:
I'm not sure how to add a test-case for the PR, because it requires `codesign` to run with the machine identiry. Once a signed apphost is available in the runtime package, this code will execute on every build.

swaroop-sridhar:
CC: @jeffschwMSFT

swaroop-sridhar:
CC: @nguerrera

nguerrera:
@wli3

nguerrera:
> SDK changes for consuming the AppHostMachOFormatException

We don't expect this to happen unless the apphost is corrupted, right?

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
```diff
@@ -0,0 +1,300 @@
...
+                        while (*(p + len) != 0 && len++ < 16) ;
+                        return Encoding.UTF8.GetString(p, len);
+                    }
+                }
+            }
+        }
+
+#pragma warning restore 0649
+
+        unsafe public static bool IsMachOImage(string filePath)
```
wli3: Where is this method used? Cannot find it

wli3: I am looking for a somewhere says "only if a mac apphost, remove signature "

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
```diff
@@ -0,0 +1,300 @@
...
+
+using System;
+using System.IO;
+using System.IO.MemoryMappedFiles;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Microsoft.NET.HostModel.AppHost
+{
+    public static class MacOUtils
```
nguerrera:
```suggestion
    public static class MachOUtils
```

I haven't seen it written as mac-o, is that common?

nguerrera: Does this need to be public? Is it for tests? Consider internals visible to tests.

swaroop-sridhar: That's right, I'll rename it. Thanks.

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
```diff
@@ -0,0 +1,300 @@
...
+                using (var mappedFile = MemoryMappedFile.CreateFromFile(stream,
+                                                                        mapName: null,
+                                                                        capacity: 0,
+                                                                        MemoryMappedFileAccess.ReadWrite,
+                                                                        HandleInheritability.None,
+                                                                        leaveOpen: true))
+                {
+                    using (var accessor = mappedFile.CreateViewAccessor())
+                    {
+                        byte* file = (byte*)accessor.SafeMemoryMappedViewHandle.DangerousGetHandle();
```
nguerrera: Elsewhere we have had to add PointerOffset, I'm not vlear on when that can be non-zero. Also, is it better to use AcquirePointer, ReleasePointer than casting DangerousGetHandle()?

src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
```diff
@@ -87,14 +87,26 @@ void UpdateResources()
                 }
             }
 
+            void RemoveSignatureIfMachO()
+            {
+                MacOUtils.RemoveSignature(appHostDestinationFilePath);
```
nguerrera: Is the if part missing here?

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
swaroop-sridhar: > Where is this method used? Cannot find it

I originally had written the code as 
```
if (IsMachOImage())
{
    RemoveSignature()
}
```
But this required mapping the file twice, so moved the functionality to within `RemoveSignature`. I left behind the `IsMachOImage()` as a service, but will remove it until there is a customer for it.

swaroop-sridhar: > I am looking for a somewhere says "only if a mac apphost, remove signature "

It is here within `RemoveSignature`
https://github.com/dotnet/core-setup/pull/8543/files#diff-3916d24dd8c65a1fb887a20b556e9223R209-R213

swaroop-sridhar:
@nguerrera  I've made the changes you suggested, thanks.

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
```diff
@@ -0,0 +1,302 @@
...
+
+            public string SegName
+            {
+                get
+                {
+                    fixed (byte* p = segname)
+                    {
+                        int len = 0;
+                        while (*(p + len) != 0 && len++ < 16) ;
+                        return Encoding.UTF8.GetString(p, len);
```
lpereira: Is it possible for this to throw given a corrupted file?

lpereira: (Might want to throw a new AppHostMachOFormatException if this throws a decoding error.)

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
```diff
@@ -0,0 +1,302 @@
...
+                                // Remove the signature blob (note for truncation)
+                                signatureSize = (uint)(fileEnd - symtabEnd);
+
+                                // Adjust the __LINKEDIT segment load command
+                                linkEdit->filesize -= signatureSize;
+
+                                // codesign --remove-signature doesn't reset the vmsize.
+                                // Setting the vmsize here makes the output bin-equal with the original 
+                                // unsigned apphost (and not bin-equal with a signed-unsigned-apphost).
+                                linkEdit->vmsize = linkEdit->filesize;
```
lpereira: I wonder if this confuses tools such as `size` or whatever is the equivalent of `readelf` for Mach-O.

swaroop-sridhar: * I've verified that `size` and `otool` work fine on the generated apphost.
* The generated host is bit-equivalent to the original apphost (before it was signed at all).

So, I think the updated are OK.

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
swaroop-sridhar: I've fixed this, thanks.

src/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs
```diff
@@ -0,0 +1,310 @@
...
+                            }
+
+                            Verify(header->Is64BitExecutable(), MachOFormatError.Not64BitExe);
+
+                            file += sizeof(MachHeader);
+                            SegmentCommand64* linkEdit = null;
+                            SymtabCommand* symtab = null;
+                            LinkEditDataCommand* signature = null;
+
+                            for (int i = 0; i < header->ncmds; i++)
```
lpereira: Might want to declare `i` as `uint` too (`ncmds` is also unsigned).

swaroop-sridhar: Fixed this too.

swaroop-sridhar:
@nguerrera, @wli3  any further thoughts on this PR? 
Otherwise will merge it into 5 and then port it to 3.1 branch. Thanks.

PR #8547, opened 2019-10-15
Use shared tooling to build the .NET Core shared framework

dagood:
For https://github.com/dotnet/core-setup/issues/4787.

Use `Microsoft.DotNet.Build.SharedFramework.Sdk`, the new location of shared framework tooling. This mirrors the way the WindowsDesktop sharedfx is created, see https://github.com/dotnet/windowsdesktop/pull/1. Split into 3 commits for review, although it's not a huge PR either way.

Adds the `sfxproj` subset because it's now possible due to changes in how the restore works. This also brings it in line with dotnet/windowsdesktop.

Note that this is a very basic level of shared tooling, and I wouldn't consider it reasonable for other repos to try to use it yet. https://github.com/dotnet/arcade/issues/2704 still tracks that.

This succeeds in an internal build, with only expected build-to-build diffs.


dagood:
Artifacts look good, only found the expected core-setup version/commit hash differences.

PR #8551, opened 2019-10-16
HostModel: Bundler: Use invariant culture when normalizing file ext

swaroop-sridhar:
The Bundler excludes PDB files by default. When looking for PDB files,
the extension is normalized by using ToLower().
This change fixes the issue to use invariant culture instead.


swaroop-sridhar:
@dagood Please review.

PR #8593, opened 2019-10-18
Changes to prepare for consolidation

jkoritzinsky:
- Define a `SetupRoot` MSBuild property to represent the root of the core-setup repo. It will point to the root folder of the core-setup files in the consolidated repo.
- Make some projects SDK projects so they pull in `Directory.Build.props/Directory.Build.targets`.
- Make the root directory configurable for the native build.
- Rename LICENSE.TXT to LICENSE to match the final name in the consolidated repo.
- Add MSBuild property definitions to configure the Microsoft.DotNet.Build.Tasks.SharedFramework.Sdk for this repo and the consolidated repo.


jkoritzinsky:
Looks like I might have missed a spot in the Arcade SDK. I'll put out a PR there to fix the issue.

jkoritzinsky:
Actually the SDK version hasn't been updated yet. That should be automatic... I'll update it manually.

dagood:
Hmm... I submitted a PR to Maestro++ a while ago to enable auto-updates for this SDK. I'll see why that isn't (apparently) working. https://github.com/dotnet/arcade-services/pull/662

eng/Version.Details.xml
```diff
@@ -123,7 +123,7 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>b47a997282b4d347cb079d26515ea4147e2ac548</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Build.SharedFramework.Sdk" Version="5.0.0-beta.19515.1">
+    <Dependency Name="Microsoft.DotNet.Build.SharedFramework.Sdk" Version="5.0.0-beta.19517.8">
```
dagood: Oh, whoops. Do you mind including this fix?

```suggestion
    <Dependency Name="Microsoft.DotNet.Build.Tasks.SharedFramework.Sdk" Version="5.0.0-beta.19517.8">
```

jkoritzinsky: That'll do it haha. I'll add this in.

Directory.Build.props
```diff
@@ -44,9 +44,10 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!-- ProjectDir needs to be set for build-tools and legacy usages. New usages should use RepoRoot -->
-    <ProjectDir>$(RepoRoot)</ProjectDir>
-    <SourceDir>$(RepoRoot)src/</SourceDir>
+    <SetupRoot>$(RepoRoot)</SetupRoot>
+    <!-- ProjectDir needs to be set for build-tools and legacy usages. New usages should use SetupRoot. -->
+    <ProjectDir>$(SetupRoot)</ProjectDir>
```
ViktorHofer: I think we can remove this property now, right?

src/corehost/build.cmd
```diff
@@ -137,8 +144,6 @@ goto :Failure
 :: Build the project created by Cmake
 set __msbuildArgs=/p:Platform=%__BuildArch% /p:PlatformToolset="%__PlatformToolset%"
 
-cd %__rootDir%
```
ViktorHofer: why don't you need the CD anymore?

src/corehost/build.sh
```diff
@@ -148,12 +144,26 @@ while [ "$1" != "" ]; do
         --stripsymbols)
             __cmake_defines="${__cmake_defines} -DSTRIP_SYMBOLS=true"
             ;;
+        --rootdir)
+           shift
+           RootRepo=$1
```
ViktorHofer: nit usually we do the shift after accessing the variable. not sure if this works.

src/pkg/projects/Directory.Build.props
```diff
@@ -58,7 +63,7 @@
     <!-- Pkgproj requires its depproj crossgen outputs to be signed.-->
     <OrderProjectReference
       Condition="'$(MSBuildProjectExtension)' == '.pkgproj'"
-      Include="$(RepoRoot)signing\SignBinaries.proj" />
+      Include="$(SetupRoot)signing\SignBinaries.proj" />
```
ViktorHofer: nit: do you think we could try moving the signing dir now into eng?

Directory.Build.props
jkoritzinsky: There might still be some usages of it that need to be updated. I’ll go do a pass next week and remove them.

src/corehost/build.sh
dagood: That's accompanied by the difference between `$1` and `$2`. Either works, this just happens to be the file's style.

src/corehost/build.cmd
jkoritzinsky: I couldn't find any reason we needed the cd. We weren't using any relative paths in this script and we already `pushd` for calling `gen-buildsys-win.bat`, which is the only script that needs a specific working directory.

ViktorHofer:
The renaming of the LICENSE.TXT to LICENSE needs to be revert asap if the Microsoft.NETCore.App assets (package, ...) are hard-linking to this file. cc @dagood @jkoritzinsky

ViktorHofer:
we can keep the LICENSE.TXT name anyway, initially I thought the renaming is necessary for GH to detect the right license but that was wrong.

dagood:
Yes, the license link on https://www.nuget.org/packages/Microsoft.NETCore.App.Ref/3.0.0 (and probably more) is a 404 now: https://github.com/dotnet/core-setup/blob/master/LICENSE.TXT. I'll put up a PR.

PR #8596, opened 2019-10-18
[release/3.1] BundleExtraction: Improve extraction path computation

### LINKED ISSUE #7940: Single EXE extractor path w/ AWS Lambda

swaroop-sridhar:
# HostWriter: BundleExtraction: Improve extraction path computation

### Tracking Issue
Fixes #7940

### Customer Scenario

Single-file apps fail to run on AWS lambda because `/var/tmp` is not writable.

### Problem

On Unix systms where $TMPDIR is not set, temp-directory is currently
determined as:

`/var/tmp/...` (preferred if it exists, because )
`/tmp/...` (if /var/tmp/ is not found)

However, this causes failures on systems where `/var/tmp` exists, but is not writable (as is the case in AWS lambda).

### Fix

This change fixes the problem with a writability check when choosing between `/var/tmp` and '/tmp'

### Testing
* Tested on Ubuntu by turning off write permissions on /var/tmp the app works by extracting to /tmp instead.
* I've requested the developer who filed the issue to verify that the fix solves the problem in his AWS lambda function. 

### Master Branch
Commit: https://github.com/dotnet/core-setup/commit/b2bff948e10f7632a40799808f4d7c74b912df80
PR: https://github.com/dotnet/core-setup/pull/8471

### DIFF
--- a/src/corehost/cli/apphost/bundle/bundle_runner.cpp
+++ b/src/corehost/cli/apphost/bundle/bundle_runner.cpp
@@ -179,6 +179,8 @@ void bundle_runner_t::determine_extraction_dir()
         {
             trace::error(_X("Failure processing application bundle."));
             trace::error(_X("Failed to determine location for extracting embedded files"));
+            trace::error(_X("DOTNET_BUNDLE_EXTRACT_BASE_DIR is not set, and temp-directory doesn't exist or is not readable/writable."));
+
             throw StatusCode::BundleExtractionFailure;
         }
 
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -265,26 +265,32 @@ bool pal::get_default_servicing_directory(string_t* recv)
     return true;
 }
 
+bool is_read_write_able_directory(pal::string_t& dir)
+{
+    return pal::realpath(&dir) &&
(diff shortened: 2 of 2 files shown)


MeiChin-Tsai:
approved for 3.1

karataliu:
This seems not work for multiple users:
The first user who trigger extractor will crate /var/tmp/.net and takes ownership.
If another user triggers extractor, '/var/tmp/.net' will not be writable then.

swaroop-sridhar:
Thanks for reporting this @karataliu. I've filed #8882 track this issue.
The `DOTNET_BUNDLE_EXTRACT_BASE_DIR` can be set to configure where the contents of single-file apps are extracted; this is probably the best way to configure extraction-location on Unix systems that don't set `$TMPDIR`. Thanks.

PR #8621, opened 2019-10-21
Move signing dir under eng/. Remove usages of ProjectDir.

jkoritzinsky:
Follow up on feedback to #8593


jkoritzinsky:
Can you kick one off?

dagood:
Sure, here: https://dev.azure.com/dnceng/internal/_build/results?buildId=395946&view=results.

ViktorHofer:
Looks like the internal build failed.

dagood:
The failure is the Microsoft.NET.HostModel test project being unable to access `Microsoft.NET.HostModel.dll`, I would guess because the signing process is trying to replace it.

Looking back at failed official builds, I see it happening at least once before.

I need to have the tests wait for signing to complete--this bug has been there for a while, apparently, it just hasn't happened enough for me to see it. Filed https://github.com/dotnet/core-setup/issues/8626.

Running a retry on that validation build to try to get past that flakiness and get on to the rest of the infra.

dagood:
Retry worked. Thanks!

PR #8622, opened 2019-10-21
[release/3.1] HostModel: Use invariant culture when normalizing file extension

swaroop-sridhar:
The Bundler excludes PDB files by default. When looking for PDB files,
the extension is normalized by using ToLower().
This change fixes the issue to use invariant culture instead.

There is no known customer problem because of this issue. 
The anticipated problem is that some files may incorrectly get excluded from the single-file bundle if they look like PDB files in the current-culture. 
So, proactively fixing it.


swaroop-sridhar:
We decided to take the fix only for .net 5.

PR #8641, opened 2019-10-22
Refactor TestProjects.props/targets lookup for consolidation

jkoritzinsky:
Move TestProjects.* to src/test. Update resolution of these files to go through RepoDirectoriesProvider instead of though a complicated dynamic discovery.


dagood:
I see that same test failure hit here in an official build too--I hit retry on this PR. I'm filing an issue to track this (apparent) flakiness.

PR #8644, opened 2019-10-23
[release/3.1] Set apphost and bundle file permission to 755

swaroop-sridhar:
# Unix: Set apphost and bundle file permission to 755

When building a .net core 3 app, the SDK currently simply copies the apphost template (including its permissions) from the install-location.

### Customer Impact

This caused two problems in Unix systems:
* If the dotnet install location is write-protected, the build fails when SDK tries update the apphost (to set the app-path, etc.)  (#8511)
* The built apphost can only be run by the owner (#7062)

### Solution

This change explicitly sets the file permissions of the Apphost in the SDK (via Microsoft.NET.HostModel package) when building on Unix systems to fix the above issues.

### Notes
This fix was contributed by @am11 -- thank you very much.

### Testing 
* Manual testing in 3.1 branch to check that the permissions are set in Unix systems.
* Automated tests are checked into the master branch -- has a dependency on the newer arcade SDK, and hence the test is not ported here.

### Master branch
PR: https://github.com/dotnet/core-setup/pull/8510
Commit: https://github.com/dotnet/core-setup/commit/a6a9202dc186f622f5df7125b12d95d9107c193a

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/HostWriter.cs
@@ -2,8 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.ComponentModel;
 using System.IO;
 using System.IO.MemoryMappedFiles;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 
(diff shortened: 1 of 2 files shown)


swaroop-sridhar:
CC: @vitek-karas, @lpereira

MeiChin-Tsai:
approved for 3.1

PR #8655, opened 2019-10-23
Resolve test assets location via a test context variable.

jkoritzinsky:
Pass the path to the test assets as a test context variable so we don't have to update the path manually when we move to a consolidated repo.


src/test/Directory.Build.targets
```diff
@@ -84,6 +84,7 @@
       <TestContextVariable Include="MNA_TFM=$(NETCoreAppFramework)" />
       <TestContextVariable Include="DOTNET_SDK_PATH=$(DotNetRoot)" />
       <TestContextVariable Include="TEST_PROJECTS_PROPS_TARGETS_FOLDER=$(MSBuildThisFileDirectory)" />
```
dagood: I think it would actually be a good idea to move `TestProjects.props/targets` into the `Assets/TestProjects` dir and remove `TEST_PROJECTS_PROPS_TARGETS_FOLDER` (usage now covered by `TEST_ASSETS`). They really are test assets, they were just in a weird place before because they were discovered by dir traversal.

Not a necessary change, but would be nice if you have time. 🙂

jkoritzinsky: I'll add that to this PR.

PR #8659, opened 2019-10-24
Proposal: Publish nethost static library

NextTurn:
See discussion in #7114 

Using "s" suffix as a temporary workaround to distinguish between static/import libraries.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -6,6 +6,8 @@
     <LibSuffix>.so</LibSuffix> 
     <LibSuffix Condition="'$(OSGroup)' == 'Windows_NT'">.dll</LibSuffix>
     <LibSuffix Condition="'$(OSGroup)' == 'OSX'">.dylib</LibSuffix>
+    <StaticLibSuffix>s.a</StaticLibSuffix>
+    <StaticLibSuffix Condition="'$(OSGroup)' == 'Windows_NT'">s.lib</StaticLibSuffix>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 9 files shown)


dagood:
The packaging infra work looks good, nice. I'm not familiar with the consequences of shipping this, e.g. signing, delivering both the dynamic `nethost.lib` and static `nethosts.lib` on Windows, and the best way to resolve naming issues to get rid of the temp `s` suffix.

@vitek-karas @elinor-fung

vitek-karas:
I looked at what we do for other libraries. For example CRT prepends `lib` to for the static ones: https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-library-features?view=vs-2019

So in that case it would be `nethost.lib` for the dynamic one and `libnethost.lib` for the static one on Windows. For Linux/Max I think it would be good to be consistent, so `libnethost.a` there as well - but I don't know how other products handle this - @lpereira may know.

In any case it's an explicit action to use this in a build, so people can pick one easily and there should be no naming issues I'm aware of.

As for the signing: I don't know what our policies are shipping unsigned code, but since it is not an executable itself, I think it should be fine. @GrabYourPitchforks would know for sure I think.

/cc @jeffschwMSFT @swaroop-sridhar

NextTurn:
The `lib` prefix looks good to me.

lpereira:
On Unix, libraries are usually always prefixed with "lib" and postfixed with something like ".so" or ".dylib" or ".a" (in case of static libraries).  Which one gets picked depends on how the compiler driver is instructed to call the linker (with e.g. the `-static` parameter in GCC and Clang).

In other words, I find it kind of confusing to have the static library having the prefix and the shared library without it: to link with the static library, you'd pass `-lnethost` (with potentially a `-L` somewhere), but to link with the shared library, you'd need to pass the path to the shared library instead.

vitek-karas:
@lpereira  On Linux/Mac the shared library does have the prefix - it should be (already) called:
* `libnethost.so`
* `libnethost.dylib`

Now we would be basically adding `libnethost.a` as well.

PR #8678, opened 2019-10-27
Add value of multilevel lookup to log

danmosemsft:
eg
```
--- Resolving FX directory, name 'Microsoft.NETCore.App' version '3.0.0'
DOTNET_MULTILEVEL_LOOKUP is set to 0
Multilevel lookup is false
Searching FX directory in [C:\Program Files\dotnet]
```

I was trying to figure out how to get it to load my private coreclr. Evidently setting this is not enough, but nevertheless, it ought to be logged.

### DIFF
--- a/src/corehost/common/utils.cpp
+++ b/src/corehost/common/utils.cpp
@@ -272,7 +272,9 @@ bool multilevel_lookup_enabled()
     {
         auto env_val = pal::xtoi(env_lookup.c_str());
         multilevel_lookup = (env_val == 1);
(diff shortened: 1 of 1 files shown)


vitek-karas:
As for the "Can't make it use my private CoreCLR": If you're running an app via executable (app.exe), then you need to set `DOTNET_ROOT` to point to your private install location.

danmosemsft:
Thanks I’ll try DOTNET_ROOT. I ended up creating a 3.0.9999 folder instead.

PR #8709, opened 2019-10-31
Prepare Core-Setup for repo consolidation

ViktorHofer:
Changes:
- Moved Directory.Build.src&targets into src to prepare for the collapsed directory structured that we discussed offline (same thing already happened in corefx). React to the collapsed structure in the msbuild files as well (ie SetupRoot -> InstallerProjectRoot, which is now under src in core-setup).
- Collapse yml files into eng/pipelines/ for easier migration
- Make tools-local independent of core-setup's Directory.Build.props&targets and rename the core-setup specifics to installer.
- Import the runtime Directory.Build.props&targets by auto-detection.
- Remove a few stable properties (ie in DisableSourceControlManagement.targets)
- Use common arcade props (NugetPackageRoot, ArtifactsBinDir, ArtfactsObjDir instead of the repo specifics PackagesDir, BinDir and ObjDir). Also merged the same change in Arcade: https://github.com/dotnet/arcade/commit/291b2730e6183ea9868f38c6d1f5b7752c5d289d
- Move eng/signing to src/signing as it depends on core-setup specifics

As soon as we get this change in, I will update the runtime repo to make core-setup usable in it.

cc @jaredpar


ViktorHofer:
Official build: https://dev.azure.com/dnceng/internal/_build/results?buildId=409306

eng/Build.props
```diff
@@ -74,8 +74,11 @@
           BeforeTargets="Execute"
           Inputs="@(RepoTasksSrc)"
           Outputs="$(RepoTasksOutputFile)">
+    <ItemGroup>
+      <RepoTaskProjects Include="$(RepoTasksDir)**\*.csproj" />
+    </ItemGroup>
     <MSBuild
```
dagood: (Style nit) Missing newline between itemgroup and task.

eng/DisableSourceControlManagement.targets
```diff
@@ -1,9 +1,5 @@
 <Project>
 
-  <PropertyGroup>
-    <DisableSourceControlManagementTargetsImported>true</DisableSourceControlManagementTargetsImported>
```
dagood: Is this related to:

> Remove a few stable properties (ie in DisableSourceControlManagement.targets)

What is a stable property? This is just an include guard AFAIK.

src/Directory.Build.props
```diff
@@ -1,5 +1,404 @@
...
+    <DisableBuildToolsRoslynVersion>true</DisableBuildToolsRoslynVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <InstallerProjectRoot>$(MSBuildThisFileDirectory)</InstallerProjectRoot>
+    <SigningToolsDir>$(InstallerProjectRoot)signing\</SigningToolsDir>
+    <LicenseFile>$(RepoRoot)LICENSE.TXT</LicenseFile>
+  </PropertyGroup>
+
+  <!-- TODO: Remove when consolidated. -->
```
dagood: I'd recommend a specific marker that can easily be found with a global search and burned down:

```
<!-- TODO: (Consolidation) Remove when consolidated. -->
```

(`TODO: (Consolidation)` being the search term there. In this case the rest of the comment is a little redundant, but there's room for some freeform extra info where it makes sense.)

src/Directory.Build.props
```diff
@@ -1,5 +1,404 @@
 <Project>
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., Directory.Build.props))\Directory.Build.props" />
+
```
dagood: Would have been nice to split meaningful changes into a separate commit to make this more reviewable. 😕

src/publish/Directory.Build.targets
```diff
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project>
+  <Import Project="..\Directory.Build.targets" />
```
dagood: Why the order change? Are the targets in this file overriding something in the above target files now?

src/signing/Directory.Build.props
```diff
@@ -1,6 +1,6 @@
 <Project>
 
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., Directory.Build.props))\Directory.Build.props" />
+  <Import Project="..\Directory.Build.props" />
```
dagood: Why is the standard import lookup changed?

tools-local/tasks/core-setup.tasks.sln
```diff
@@ -1,22 +0,0 @@
-﻿
```
dagood: Why is this deleted?

tools-local/tasks/installer.tasks/installer.tasks.csproj
```diff
@@ -16,8 +13,8 @@
 
   <ItemGroup>
     <PackageReference Include="NuGet.ProjectModel" Version="$(RefOnlyNugetProjectModelPackageVersion)" />
-    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.1.0" />
-    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="2.1.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="$(MicrosoftDotNetPlatformAbstractionsPackageVersion)" />
```
dagood: This changes the version from 2.1.0 to 3.0.0, which impacts source-build. Is it necessary in this PR?

eng/DisableSourceControlManagement.targets
ViktorHofer: I searched through the whole dotnet organization, this property is dead: https://github.com/search?q=org%3Adotnet+DisableSourceControlManagementTargetsImported&type=Code

src/Directory.Build.props
ViktorHofer: Yeah I know, I'm sorry for that. A lot of changes were connected and I wasn't sure if it will work out like I imagined. I will try to avoid that in future.

src/publish/Directory.Build.targets
ViktorHofer: no, I just saw that the root Directory.Build.targets is imported at the end here which doesn't make much sense as nested D.B.targets/props file should be incremental to what's already there in the root. But as this is completely unrelated I can undo the change in this PR.

tools-local/tasks/core-setup.tasks.sln
ViktorHofer: this folder will be migrated into the runtime root repo. To reduce noise with the sln (opening it in different VS versions will update its version) and as the solution only contains one project, I removed it.

tools-local/tasks/installer.tasks/installer.tasks.csproj
ViktorHofer: Yes, I was also concerned about this change. I will replace it back to the 2.1.0 version but we should consolidate the different versions soon.

src/signing/Directory.Build.props
ViktorHofer: I don't really understand what you mean here? I just simplified the syntax.

dagood: This pattern reduces maintainability:
* It makes it possible to accidentally skip over a dir.props if you add a new dir.props and forget to update all instances of this pattern in the child dirs.
* It increases the number of changes in the resulting PR even if you do manage to replace everything correctly.

src/publish/Directory.Build.targets
dagood: The standard pattern I'm aware of is to import the "above" dir.props at the beginning of the file, and the "above" dir.targets at the end of the file, unless you need to override some targets in the "above" dir.targets. This resembles the project file convention and how the default SDK imports work:

```
<Project>
  <Import dir.props>
  [project-specific properties and targets]
  <Import dir.targets>
</Project>
```

I don't have a particular list what kind of problems this prevents (what exactly is more predictable this way)--I don't mind changing, just weird to see it as a change in this PR.

src/Directory.Build.props
dagood: I tend to do rebasing and such to set up nice PRs for my larger sets of changes when the number of commits and backtracking gets too big to submit as-is.

src/signing/Directory.Build.props
ViktorHofer: I reverted all these changes as they are just noise. Thanks for sharing your opinion.

src/publish/Directory.Build.targets
ViktorHofer: I have a different opinion here but we can discuss that at a later point, probably after the consolidation when we do a more overall cleanup.

ViktorHofer:
I assume this is now ready to go in? :)

ViktorHofer:
Of course. Thanks for reviewing! 👍

PR #8724, opened 2019-11-01
Use Version suffix and clean versions

ViktorHofer:
Requires https://github.com/dotnet/arcade/pull/4266

I'm consolidating all the version properties to use the `Version` suffix instead of `PackageVersion`. This aligns to Arcade conventions and extension point which also use the `Version` suffix.

- Cleaning the unused dependencies in Version.Details.xml and the sourcelink one as that one is pushed down by arcade.
- Removing the `MicrosoftDotNetMaestroTasksVersion` which is set in Arcade as well.
- Removing unused package names.
- Use package names which are defined in Versions.props in the code.
- Consolidate on using Version suffix everywhere (corefx, coreclr, core-setup, arcade)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -19,72 +19,41 @@
     <NETCoreAppFrameworkVersion>$(MajorVersion).$(MinorVersion)</NETCoreAppFrameworkVersion>
     <NETCoreAppFramework>netcoreapp$(NETCoreAppFrameworkVersion)</NETCoreAppFramework>
   </PropertyGroup>
-  <PropertyGroup>
-    <MicrosoftDotNetMaestroTasksVersion>1.0.0-beta.18619.4</MicrosoftDotNetMaestroTasksVersion>
-  </PropertyGroup>
   <!--Package versions-->
   <PropertyGroup>
     <!-- arcade -->
-    <MicrosoftDotNetBuildTasksFeedPackageVersion>5.0.0-beta.19530.15</MicrosoftDotNetBuildTasksFeedPackageVersion>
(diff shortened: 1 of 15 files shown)


eng/Versions.props
```diff
@@ -19,83 +19,50 @@
...
+    <RefOnlyNugetPackagingVersion>4.9.4</RefOnlyNugetPackagingVersion>
   </PropertyGroup>
   <!--Package names-->
   <PropertyGroup>
     <MicrosoftDotNetBuildTasksFeedPackage>Microsoft.DotNet.Build.Tasks.Feed</MicrosoftDotNetBuildTasksFeedPackage>
     <MicrosoftPrivateCoreFxNETCoreAppPackage>Microsoft.Private.CoreFx.NETCoreApp</MicrosoftPrivateCoreFxNETCoreAppPackage>
     <MicrosoftNETCorePlatformsPackage>Microsoft.NETCore.Platforms</MicrosoftNETCorePlatformsPackage>
     <MicrosoftNETCoreTargetsPackage>Microsoft.NETCore.Targets</MicrosoftNETCoreTargetsPackage>
-    <NETStandardLibraryPackage>NETStandard.Library</NETStandardLibraryPackage>
+    <NETStandardLibraryPackage>netstandard.library</NETStandardLibraryPackage>
```
dagood: Why is this lowecased? The package name is spelled with uppercase chars.

ViktorHofer: `NETStandardLibraryPackage` is used in libraries to build a path to the local package cache package. On Unix having the upper-case one fails because of case-sensitivity. When we are using properties for that we are usually just using their lowercase name as that works in PackageReferences fine as well.

dagood: Yeah, suppose there's no need to keep the proper casing anywhere. Sounds good, I assume we'll clean up the rest at some point after consolidation and this one was just because it happened to be on the same line as some other change.

PR #8767, opened 2019-11-04
Allow ManPagesDir override, handle error better (consolidation)

dagood:
In the consolidated build, the docs have moved, so `<ManPagesDir>$(RepoRoot)Documentation/manpages</ManPagesDir>` is no longer accurate. This PR lets the property get overridden, and handles the error better when it the manpages aren't found. Details in code comment.

Requires the property to be set in the consolidated repo. (To `$(RepoRoot)/docs/installer/manpages` at the moment, I think.) Once we're merged, it can be resolved and removed.


ViktorHofer:
Thanks.

PR #8780, opened 2019-11-05
 Backport pipeline changes

jkoritzinsky:
Backport changes from dotnet/runtime to core-setup for CI in a manner that will automatically switch over when we consolidate.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,12 +1,18 @@
 trigger:
-  batch: true
   branches:
     include:
     - master
     - release/3.*
   paths:
+    include:
+    - '*'
+    - docs/installer/manpages/*
     exclude:
-    - Documentation/*
+    - src/coreclr/*
+    - src/libraries/*
+    - eng/pipelines/coreclr/*
+    - eng/pipelines/libraries/*
+    - docs/*
     - README.md
     - CONTRIBUTING.md
     - LICENSE.TXT
@@ -18,8 +24,15 @@ pr:
     - master
     - release/3.*
   paths:
+    include:
+    - '*'
+    - docs/installer/manpages/*
     exclude:
-    - Documentation/*
+    - src/coreclr/*
+    - src/libraries/*
+    - eng/pipelines/coreclr/*
+    - eng/pipelines/libraries/*
+    - docs/*
     - README.md
     - CONTRIBUTING.md
     - LICENSE.TXT
@@ -53,11 +66,22 @@ variables:
     - name: SignType
       value: $[ coalesce(variables.OfficialSignType, 'real') ]
 
+  - ${{ if contains(variables['Build.DefinitionName'], 'runtime') }}:
+    - name: pipelinesPath
+      value: /eng/pipelines/installer
+    - name: buildScriptFileName
+      value: installer
+  - ${{ if not(contains(variables['Build.DefinitionName'], 'runtime')) }}:
+    - name: pipelinesPath
+      value: /eng/pipelines
+    - name: buildScriptFileName
+      value: build
+
 stages:
 - stage: Build
   jobs:
   # -------- Build Bash legs (Linux and FreeBSD) --------
-  - template: /eng/pipelines/jobs/bash-build.yml
+  - template: $(pipelinesPath)/jobs/bash-build.yml
     parameters:
       crossBuild: true
       name: Linux_Arm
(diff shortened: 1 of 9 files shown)


eng/pipelines/jobs/bash-build.yml
```diff
@@ -40,8 +40,8 @@ jobs:
       ${{ if ne(parameters.name, 'FreeBSD_x64')}}:
         RunArguments: >-
           docker run --privileged --rm
-          -v "$(Build.SourcesDirectory):/root/coresetup"
-          -w="/root/coresetup"
+          -v "$(Build.SourcesDirectory):/root/runtime"
```
ViktorHofer: don't we need a condition here?

eng/pipelines/jobs/bash-build.yml
```diff
@@ -89,11 +92,16 @@ jobs:
 
       DockerRunMSBuild: >-
         docker run
-        -v $(Build.SourcesDirectory):/root/coresetup
+        -v $(Build.SourcesDirectory):/root/runtime
```
ViktorHofer: again: don't we need a condition here?

eng/pipelines/jobs/steps/upload-job-artifacts.yml
```diff
@@ -35,6 +35,13 @@ steps:
   continueOnError: true
   condition: always()
 
+- task: DeleteFiles@1
```
ViktorHofer: can you please explain this change?

azure-pipelines.yml
```diff
@@ -1,12 +1,20 @@
 trigger:
-  batch: true
```
ViktorHofer: So we don't want batching anymore because of different competing build definitions?

jkoritzinsky: I don't think we want batching in the runtime repository and it makes diagnosing issues in this repository difficult at times.
Since core-setup doesn't run a public CI job, the only post-merge validation is a batched official build. When that breaks, you have to figure out which commit caused the break, which takes a significant amount of time. Since we don't run tests on limited-hardware configurations anyway, I don't think the minor increase in number of official builds will cause any issues.

eng/pipelines/jobs/steps/upload-job-artifacts.yml
jkoritzinsky: This file is generated as part of creating the RPM packages for RHEL. It's a symlink to a file by a path that only exists in the Docker container. When we try to copy files by glob matching below, it tries to stat this path and fails. We delete this file in MSBuild, but it doesn't get deleted there unless the RPM package build succeeds. Without this change, binlogs fail to get uploaded when RPM packages aren't built.

eng/pipelines/jobs/bash-build.yml
jkoritzinsky: We don't need a condition here. This change (as well as the other similar changes) are mainly cosmetic. This argument to docker says "map the $(Build.SourcesDirectory) to the directory /root/runtime in the docker container". We could delay this particular change until after consolidation, but I felt that it fits better in this PR.

eng/pipelines/jobs/steps/upload-job-artifacts.yml
ViktorHofer: thanks.

azure-pipelines.yml
ViktorHofer: ok, thanks.

eng/pipelines/jobs/steps/upload-job-artifacts.yml
dagood: This is a suggestion I made to try to make the build debuggable even if we run into this situation:

https://github.com/dotnet/core-setup/blob/bff42d61cf44185dfb90dc4b344e11aa136d9f46/src/pkg/packaging/rpm/package.targets#L65-L72

Now that https://github.com/dotnet/core-setup/pull/8767 is in, the direct cause is gone, and I think it'd be nice to remove this workaround step from `upload-job-artifacts.yml` so it doesn't cause confusion when it shows up in all the jobs using this template.

dagood: Whoops, didn't refresh before commenting. Fine either way.

jkoritzinsky: I just added a new PR to remove it: https://github.com/dotnet/core-setup/pull/8791

PR #8785, opened 2019-11-06
Import with absolute paths

ViktorHofer:
This seems to could cause troubles, not sure if it fixes what @trylek is experiencing.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <IsRuntimeRepository Condition="Exists('$(MSBuildThisFileDirectory)..\..\.dotnet-runtime-placeholder')">true</IsRuntimeRepository>
   </PropertyGroup>
-  <Import Project="$(MSBuildThisFileDirectory)..\..\Directory.Build.props" Condition="'$(IsRuntimeRepository)' == 'true'" />
+  <Import Project="..\..\Directory.Build.props" Condition="'$(IsRuntimeRepository)' == 'true'" />
 
   <!--
     Import the arcade sdk with these requirements.
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,9 +2,9 @@
 
   <!-- TODO: (Consolidation) Remove when consolidated. -->
   <PropertyGroup>
-    <IsRuntimeRepository Condition="Exists('..\..\.dotnet-runtime-placeholder')">true</IsRuntimeRepository>
+    <IsRuntimeRepository Condition="Exists('$(MSBuildThisFileDirectory)..\..\.dotnet-runtime-placeholder')">true</IsRuntimeRepository>
   </PropertyGroup>
-  <Import Project="..\..\Directory.Build.props" Condition="'$(IsRuntimeRepository)' == 'true'" />
+  <Import Project="$(MSBuildThisFileDirectory)..\..\Directory.Build.props" Condition="'$(IsRuntimeRepository)' == 'true'" />
 
   <!--
     Import the arcade sdk with these requirements.
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -7,7 +7,7 @@
   -->
   <Import Project="packaging.stubs.targets" Condition="'$(MSBuildProjectExtension)' != '.pkgproj'" />
 
-  <Import Project="..\..\Directory.Build.targets" Condition="'$(IsRuntimeRepository)' == 'true'" />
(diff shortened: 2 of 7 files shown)


dagood:
What's the pattern here? Why these particular imports and not others that rely on the import being relative to the file doing the import? From the PR description, this sounds like a shot in the dark.

ViktorHofer:
Those are all the imports in core-setup in msbuild files that use a relative pattern. @trylek tried the fix out in the runtime repo and verified that it fixes wrong imports which cause the installer.tasks output to be placed in the wrong directory.

src/Directory.Build.targets
```diff
@@ -7,7 +7,7 @@
   -->
   <Import Project="packaging.stubs.targets" Condition="'$(MSBuildProjectExtension)' != '.pkgproj'" />
```
dagood: Why isn't this a "relative pattern"? Are you talking about `..` in particular?

This PR is striking me as extremely odd, since we rely on this working as-is in so much... why now?

ViktorHofer: > Relative paths in imported projects are interpreted relative to the directory of the importing project. Therefore, if a project file is imported into several project files in different locations, the relative paths in the imported project file will be interpreted differently for each imported project.

from https://docs.microsoft.com/en-us/visualstudio/msbuild/import-element-msbuild?view=vs-2019#remarks.
This is what is causing troubles for installer.tasks. Using absolute paths prevents that. For the sake of fixing the runtime repo, the change could be scoped down to only the two Directory.Build.props/targets files under tools-local but I thought it might a good idea to harden this everywhere.

dagood: For example, the simplest thing, `<Import Project="..\Directory.Build.props" />`, should not work if this PR is correct.

dagood: That doc seems incorrect. If it were only referring to properties and items, then it would be correct by my experience.

ViktorHofer: What part do you believe is incorrect?

dagood: The term `relative paths` is too broad in

> Relative paths in imported projects are interpreted relative to the directory of the importing project.

It's true that paths in properties aren't evaluated at all, so their meaning is defined based on where they're used, and item includes are expanded in the context of the importing project. However, imports must be relative to the file they appear in, or else it's in contradiction with even the most basic dir.props chains in practice:

```
src/foo/foo.csproj
src/foo/Directory.Build.props
src/Directory.Build.props
Directory.Build.props
```

Each `Directory.Build.props` contains `<Import Project="..\Directory.Build.props" />`. When `src/Directory.Build.props` is imported, if the above doc is true, then the file would try to import itself.

src/Directory.Build.props
```diff
@@ -2,9 +2,9 @@
 
   <!-- TODO: (Consolidation) Remove when consolidated. -->
   <PropertyGroup>
-    <IsRuntimeRepository Condition="Exists('..\..\.dotnet-runtime-placeholder')">true</IsRuntimeRepository>
+    <IsRuntimeRepository Condition="Exists('$(MSBuildThisFileDirectory)..\..\.dotnet-runtime-placeholder')">true</IsRuntimeRepository>
```
dagood: Are you sure this change alone isn't the fix? It would make sense to me if the condition is evaluated in context of the importing project.

src/Directory.Build.targets
ViktorHofer: @rainersigwald can you help out here?

dagood: An even simpler example is the line I left this comment on... `<Import Project="packaging.stubs.targets"` is against the doc, because `packaging.stubs.targets` exists in `src/`, not adjacent to every single project.

jkoritzinsky: I'm almost certain that `<Import>` paths are evaluated relative to the current file. In properties and items they're evaluated relative to the entry project file.

rainersigwald: Yeah, I think @dagood was right with https://github.com/dotnet/core-setup/pull/8785/files#r343228968. Since that condition isn't in the context of an import it gets "the usual" CWD rules, instead of the import-relative ones.

dagood: I'll file the doc bug.

dagood: Filed https://github.com/MicrosoftDocs/visualstudio-docs/issues/4287.

ViktorHofer: Thanks Davis and Rainer. Any suggestion what to do with this? Should I remove changes in the Import statements or leave them to be absolute?

rainersigwald: I lean toward backing them out, but don't feel strongly about it.

ViktorHofer: And thanks Jeremy as well of course.

Issue #8735, opened 2019-11-01
Disable producing targeting packs in servicing builds (unless patching)


dagood:
https://github.com/dotnet/core-setup/issues/8507 tracks disabling the NETStandard targeting pack for 3.1 in particular. However, we should disable NETCoreApp and WindowsDesktop's as well to match this part of [the framework pack design](https://github.com/dotnet/designs/pull/50):

> Targeting pack package versions will generally not increase past major.minor.0 where major.minor matches the the corresponding two-part TFM. For example, when there is a 3.0.1 .NET Core runtime, the targeting pack will likely remain at 3.0.0. This follows from the fact that targeting packs represent public API surface, which must not change in a patch version of the runtime. (With that said, we can reserve the right to modify the targeting pack in a patch release to fix a severe bug. In the rare event that this occurs, the targeting pack patch version could be incremented past 0.)

In theory, there isn't a problem with Core-Setup building the targeting packs as long as they aren't published *and* downstream repos don't point to the bad new version. However, this requires manual dependency PR fixup/pinning by the downstream repos. It would be friendlier for Core-Setup to not produce the bad versions in the first place. This also means that to turn on the build, we only need to change Core-Setup, rather than undo dependency pins in any downstream repos that have them.

/cc @nguerrera @mmitche @wtgodbe

dagood:
This is tied up with a general source-build problem... we need the old targeting pack to assemble the current SDK installers, so how will the old targeting pack source-built?

E.g. the 3.0.108 SDK installer could need the 3.0.0 targeting pack and 3.0.7 runtime.

This is the same core problem as https://github.com/dotnet/source-build/issues/923. I'm asking offline what the source-build team's current thoughts are on this.

My current understanding is that source-build generally assumes that the current sources are *close enough* to end up producing a working build when put together. In this case, I think that would mean forcing the targeting pack to always build when doing a source-build, and forcing the targeting pack's version back to the last-shipped version so that it can target the same runtimes as the Microsoft build. (Keeping source-build parity.)

/cc @MichaelSimons @leecow @dseefeld @crummel @adaggarwal

jzabroski:
@dagood You're pretty close to the heart of the problem, but it actually goes deeper and simpler than that:

[...]

Issue #8507, opened 2019-10-09
Disable producing NETStandard targeting pack in 3.1 branch


dagood:
Right now, Core-Setup `release/3.1` still produces 2.1-versioned .NET Standard targeting packs with prerelease versions, which is *below* the 2.1.0 version we produced for 3.0.0 (`release/3.0`). This is confusing.

We should disable this targeting pack from building. Repos should keep pointing at the released 2.1 targeting pack.

The .NET Standard targeting pack is odd because it doesn't really belong in Core-Setup. That's why the usual approach of bumping up the version isn't an obvious solution.

dleeapho:
When we do need to rev the targeting pack we'll need to manually enable it.

dagood:
This specific work is done. Followup porting to master is already tracked by https://github.com/dotnet/core-setup/issues/8507, other bugs are filed as followup servicing infra bugs.

PR #8828, opened 2019-11-11
[release/3.1] Add servicing config: project skip infrastructure

dagood:
#### Description

For https://github.com/dotnet/core-setup/issues/8735, *Disable producing targeting packs in servicing builds (unless patching)* and specifically https://github.com/dotnet/core-setup/issues/8507, *Disable producing NETStandard targeting pack in 3.1 branch*.

Ports https://github.com/dotnet/core-setup/pull/8827 from `release/3.0` to `release/3.1`.

This isn't necessary for the 3.1.0 GA build because the `NETStandard.Library.Ref` targeting pack dependency in Core-SDK is already pinned. This should be merged before the 3.1.1 servicing build so the NETCoreApp and WindowsDesktop targeting packs don't also need to be pinned.

#### Customer Impact

The way the build currently works, downstream repos like core-sdk have to manually pin the targeting pack versions so the auto-update PRs don't insert bad targeting packs that we don't plan to ship.

After this PR, dependencies on Core-Setup's targeting packs will no longer need to be pinned because the packs will only be produced when they should be used. This reduces the chance of creating a bad build, reducing maintenance cost of servicing. 

#### Regression?

No.

#### Risk

This may cause build breaks in downstream repos. This would be a good thing from a correctness standpoint, because it reveals bad assumptions that targeting packs are always produced.

To keep this PR simple, targeting pack tests now pass rather than fail if the nupkg isn't found on disk. This allows the tests to pass when the targeting pack isn't being patched. It adds a risk that the tests won't catch a situation where a targeting pack wasn't built but should have been. I believe this risk isn't significant: I've only observed missing nupkgs along with a build failure (so it would get noticed anyway) or major refactoring (which I don't see happening in `release/3.0`). We can backport improvements to these tests in the future.

---

/cc @wtgodbe @mmitche @Anipik @nguerrera

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -39,6 +39,7 @@
   -->
   <ItemGroup Condition="'$(StabilizePackageVersion)' == 'true'">
     <ProjectServicingConfiguration Include="Microsoft.NETCore.App.Ref" PatchVersion="0" />
+    <ProjectServicingConfiguration Include="Microsoft.WindowsDesktop.App.Ref" PatchVersion="0" />
     <ProjectServicingConfiguration Include="NETStandard.Library.Ref" PatchVersion="0" />
   </ItemGroup>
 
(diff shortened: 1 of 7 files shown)


dagood:
Closing, this has been merged by automatic branch port: https://github.com/dotnet/core-setup/pull/8881.

PR #8842, opened 2019-11-12
Update CoreCLR override to use the sharedFramework output folder instead of comparing with packages

jkoritzinsky:
Currently, the core-setup scripting to use a local CoreCLR build depends on the CoreCLR packages. With dotnet/coreclr#27841, we will be able to determine what files are part of the shared framework based on if they are in the sharedFramework folder.

This PR updates the CoreCLR override mechanism to use the sharedFramework folder instead of comparing files against the packages.

### DIFF
--- a/src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
+++ b/src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
@@ -4,40 +4,43 @@
   <Target Name="OverrideRuntimeFilesFromPackageResolve" Condition="'$(CoreCLROverridePath)' != ''" BeforeTargets="GetFilesFromPackageResolve">
     <Error Condition="!Exists('$(CoreCLROverridePath)')" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not exist." />
     <PropertyGroup>
-      <CoreCLRCrossTargetComponentFolderName Condition="'$(TargetArchitecture)' == 'arm64' and '$(BuildArchitecture)' != 'arm64'">x64</CoreCLRCrossTargetComponentFolderName>
-      <CoreCLRCrossTargetComponentFolderName Condition="'$(TargetArchitecture)' == 'arm' and '$(BuildArchitecture)' != 'arm' and '$(TargetsWindows)' == 'true'">x86</CoreCLRCrossTargetComponentFolderName>
-      <CoreCLRCrossTargetComponentFolderName Condition="'$(TargetArchitecture)' == 'arm' and '$(BuildArchitecture)' != 'arm' and '$(TargetsLinux)' == 'true'">x64</CoreCLRCrossTargetComponentFolderName>
-      <CoreCLRSharedFrameworkFolder>$([MSBuild]::NormalizePath($(CoreCLROverridePath),'sharedFramework'))</CoreCLRSharedFrameworkFolder>
-      <CoreCLRCrossTargetComponentFolder
-        Condition="'$(CoreCLRCrossTargetComponentFolderName)' != ''">$([MSBuild]::NormalizePath($(CoreCLROverridePath),$(CoreCLRCrossTargetComponentFolder),'sharedFramework'))</CoreCLRCrossTargetComponentFolder>
+      <CoreCLROverridePath>$([MSBuild]::NormalizeDirectory('$(CoreCLROverridePath)'))</CoreCLROverridePath>
+    </PropertyGroup>
+    <PropertyGroup>
+      <CoreCLRCrossTargetComponentDirName Condition="'$(TargetArchitecture)' == 'arm64' and '$(BuildArchitecture)' != 'arm64'">x64</CoreCLRCrossTargetComponentDirName>
+      <CoreCLRCrossTargetComponentDirName Condition="'$(TargetArchitecture)' == 'arm' and '$(BuildArchitecture)' != 'arm' and '$(TargetsWindows)' == 'true'">x86</CoreCLRCrossTargetComponentDirName>
+      <CoreCLRCrossTargetComponentDirName Condition="'$(TargetArchitecture)' == 'arm' and '$(BuildArchitecture)' != 'arm' and '$(TargetsLinux)' == 'true'">x64</CoreCLRCrossTargetComponentDirName>
+      <CoreCLRSharedFrameworkDir>$([MSBuild]::NormalizeDirectory('$(CoreCLROverridePath)','sharedFramework'))</CoreCLRSharedFrameworkDir>
+      <CoreCLRCrossTargetComponentDir
+        Condition="'$(CoreCLRCrossTargetComponentDirName)' != ''">$([MSBuild]::NormalizeDirectory('$(CoreCLROverridePath)','$(CoreCLRCrossTargetComponentDir)','sharedFramework'))</CoreCLRCrossTargetComponentDir>
     </PropertyGroup>
     <ItemGroup>
-      <CoreCLRFiles Include="$(CoreCLRSharedFrameworkFolder)/*.*" />
-      <CoreCLRCrossTargetFiles Condition="'$(CoreCLRCrossTargetComponentFolder)' != ''" Include="$(CoreCLRCrossTargetComponentFolder)/*.*" />
-      <CoreCLRFiles Include="$(CoreCLROverridePath)/Redist/**/*.dll" />
+      <CoreCLRFiles Include="$(CoreCLRSharedFrameworkDir)*.*" />
+      <CoreCLRCrossTargetFiles Condition="'$(CoreCLRCrossTargetComponentDir)' != ''" Include="$(CoreCLRCrossTargetComponentDir)*.*" />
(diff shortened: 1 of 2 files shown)


src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
```diff
@@ -4,39 +4,45 @@
...
     <PropertyGroup>
-      <CoreCLRPDBOverridePath Condition="'$(CoreCLRPDBOverridePath)' == '' and Exists('$(CoreCLROverridePath)/PDB')">$(CoreCLROverridePath)/PDB</CoreCLRPDBOverridePath>
+      <CoreCLRCrossTargetComponentFolder Condition="'$(TargetArchitecture)' == 'arm64'">x64</CoreCLRCrossTargetComponentFolder>
+      <CoreCLRCrossTargetComponentFolder Condition="'$(TargetArchitecture)' == 'arm' and '$(TargetsWindows)' == 'true'">x86</CoreCLRCrossTargetComponentFolder>
+      <CoreCLRCrossTargetComponentFolder Condition="'$(TargetArchitecture)' == 'arm' and '$(TargetsLinux)' == 'true'">x64</CoreCLRCrossTargetComponentFolder>
     </PropertyGroup>
     <ItemGroup>
-      <CoreCLRFiles Include="$(CoreCLROverridePath)/*.*" />
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/sharedFramework/*.*" />
+      <CoreCLRCrossTargetFiles Condition="'$(CoreCLRCrossTargetComponentFolder)' != ''" Include="$(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/*.*" />
```
dagood: Use normalization property functions to make less sensitive to `/`/`\` in various places?

src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
```diff
@@ -4,39 +4,45 @@
...
+      </CoreCLRFiles>
+      <CoreCLRFiles Include="$(CoreCLROverridePath)/System.Private.CoreLib.dll" />
+      <CoreCLRFiles
+        Include="$(CoreCLROverridePath)/sharedFramework/PDB/*.pdb;$(CoreCLROverridePath)/sharedFramework/PDB/*.dbg;$(CoreCLROverridePath)/sharedFramework/PDB/*.dwarf" />
+      <CoreCLRFiles
+        Include="$(CoreCLROverridePath)/PDB/System.Private.CoreLib.pdb;$(CoreCLROverridePath)/PDB/System.Private.CoreLib.ni.pdb" />
+      <CoreCLRCrossTargetFiles Condition="'$(CoreCLRCrossTargetComponentFolder)' != ''"
+        Include="$(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/PDB/*.pdb;
+        $(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/PDB/*.dbg;
+        $(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/PDB/*.dwarf" />
```
dagood: IMO preemptive chopping is easier to read:

```xml
<CoreCLRCrossTargetFiles
  Condition="'$(CoreCLRCrossTargetComponentFolder)' != ''"
  Include="
    $(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/PDB/*.pdb;
    $(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/PDB/*.dbg;
    $(CoreCLROverridePath)/$(CoreCLRCrossTargetComponentFolder)/sharedFramework/PDB/*.dwarf" />
```

src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
```diff
@@ -4,39 +4,45 @@
...
+
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="'@(CoreCLRFiles->'%(FileName)%(Extension)')' == '%(FileName)%(Extension)'"/>
+      <ReferenceCopyLocalPaths Remove="@(ReferenceCopyLocalPaths)" Condition="$([System.String]::Copy('%(FileName)').StartsWith('mscordaccore_'))" />
+      <ReferenceCopyLocalPaths Include="@(CoreCLRFiles)" />
     </ItemGroup>
 
     <Error Condition="'@(CoreCLRFiles)' == ''" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not contain any files." />
   </Target>
-
+
```
dagood: Trailing space added.

src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
```diff
@@ -4,39 +4,45 @@
   <Target Name="OverrideRuntimeFilesFromPackageResolve" Condition="'$(CoreCLROverridePath)' != ''" BeforeTargets="GetFilesFromPackageResolve">
     <Error Condition="!Exists('$(CoreCLROverridePath)')" Text="The path provided to CoreCLROverridePath ($(CoreCLROverridePath)) does not exist." />
     <PropertyGroup>
-      <CoreCLRPDBOverridePath Condition="'$(CoreCLRPDBOverridePath)' == '' and Exists('$(CoreCLROverridePath)/PDB')">$(CoreCLROverridePath)/PDB</CoreCLRPDBOverridePath>
+      <CoreCLRCrossTargetComponentFolder Condition="'$(TargetArchitecture)' == 'arm64'">x64</CoreCLRCrossTargetComponentFolder>
```
dagood: Maybe I'm not quite following the logic here, but won't this cause the build to try to use nonexistent x64 bits when building targeting arm64 on a real arm64 machine? Does this just need a `AND '$(BuildArchitecture)' != 'arm64'` like https://github.com/dotnet/arcade/blob/993af9410c505680b9a260f3bfd79515c936de12/src/Microsoft.DotNet.Build.Tasks.SharedFramework.Sdk/targets/framework.dependency.targets#L260 ?

dagood:
Test failure may be this known flakiness: https://github.com/dotnet/core-setup/issues/8643

src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
jkoritzinsky: Yep you're correct. Good catch!

src/pkg/projects/netcoreapp/src/localnetcoreapp.override.targets
```diff
@@ -4,34 +4,47 @@
...
     <PropertyGroup>
-      <CoreCLRPDBOverridePath Condition="'$(CoreCLRPDBOverridePath)' == '' and Exists('$(CoreCLROverridePath)/PDB')">$(CoreCLROverridePath)/PDB</CoreCLRPDBOverridePath>
+      <CoreCLRCrossTargetComponentFolderName Condition="'$(TargetArchitecture)' == 'arm64' and '$(BuildArchitecture)' != 'arm64'">x64</CoreCLRCrossTargetComponentFolderName>
+      <CoreCLRCrossTargetComponentFolderName Condition="'$(TargetArchitecture)' == 'arm' and '$(BuildArchitecture)' != 'arm' and '$(TargetsWindows)' == 'true'">x86</CoreCLRCrossTargetComponentFolderName>
+      <CoreCLRCrossTargetComponentFolderName Condition="'$(TargetArchitecture)' == 'arm' and '$(BuildArchitecture)' != 'arm' and '$(TargetsLinux)' == 'true'">x64</CoreCLRCrossTargetComponentFolderName>
+      <CoreCLRSharedFrameworkFolder>$([MSBuild]::NormalizePath($(CoreCLROverridePath),'sharedFramework'))</CoreCLRSharedFrameworkFolder>
+      <CoreCLRCrossTargetComponentFolder
+        Condition="'$(CoreCLRCrossTargetComponentFolderName)' != ''">$([MSBuild]::NormalizePath($(CoreCLROverridePath),$(CoreCLRCrossTargetComponentFolder),'sharedFramework'))</CoreCLRCrossTargetComponentFolder>
```
dagood: It's odd that these two dirs are treated as file paths (normalized without trailing `/` rather than using `NormalizeDirectory`) and the names don't have any of the conventional `Path`, `File`, `Dir`, or `Directory` suffixes. (At least... those are the conventions I'm aware of.) Is there a reason for this?

dagood: I think the function params need to be `'$()'` to make sure parsing doesn't trip up if the property contents look like MSBuild syntax.

jkoritzinsky: I'll change the naming.

jkoritzinsky: There wasn't any particular reason.

jkoritzinsky:
Failure is due to a lost connection in AzDO. Rerunning.

PR #8947, opened 2020-01-14
Making the 3.0 branch ready for next release 

Anipik:


mmitche (approved):
Please make sure to merge rather than squash

eng/Versions.props
```diff
@@ -3,10 +3,10 @@
   <PropertyGroup>
     <MajorVersion>3</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>2</PatchVersion>
+    <PatchVersion>3</PatchVersion>
     <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
     <!-- Enable to remove prerelease label. -->
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
+    <StabilizePackageVersion>true</StabilizePackageVersion>
```
dagood: This isn't my suggestion, https://github.com/dotnet/core-setup/pull/8945#discussion_r363990946:

> I believe the fix is pretty simple, just change `<ItemGroup Condition="'$(StabilizePackageVersion)' == 'true'">` to `<ItemGroup>` below.

I don't think this will damage the build, it's just a no-op.

dagood: Feel free to revert it to unblock this if it doesn't make sense. Since it turns out we aren't unstabilizing, it isn't necessary yet.

dagood (approved):

PR #8987, opened 2020-01-22
[release/3.1] Prompt dialog when running GUI app and hostfxr is < 3.*

mateoatr:
Issue: [dotnet/runtime#617](https://github.com/dotnet/runtime/issues/617)

Description
---
Since .NET Core 3.1, apphosts have the ability to prompt a dialog whenever a client doesn't have either a runtime or a required framework installed. Currently, clients with an installed version of .NET Core previous to 3.0 that try to run a GUI app will get no error message whatsoever, the application will just silently crash. This fixes the aforementioned scenario by prompting an error dialog that lets the client download a newer version of .NET Core. 

This is because versions of hostfxr previous to 3.0 have no error message propagation, causing the apphost to crash silently.

Customer impact
---

Customers trying to run a GUI apphost with a .NET Core version previous to 3.0 will not get any kind of error message, the apphost will silently crash, not giving any clue of what's wrong.
The error is written into Windows Event Log, but most end users won't go there to check.  
There's no workaround for either the user or the application developer as this happens too early in the host binaries for any application code to run.

Regression?
---

Not a regression - the dialog popup for errors in GUI apps is a new capability in 3.0 (as well as official support for GUI apps as a whole).

Risk
---

Low - the fix is targeted to the very specific case where the bug happens.

----

[Edit] The dialog that we show now:
image

cc @vitek-karas @jeffschwMSFT

### DIFF
--- a/src/corehost/cli/apphost/apphost.windows.cpp
+++ b/src/corehost/cli/apphost/apphost.windows.cpp
@@ -76,12 +76,19 @@ namespace
             while (std::getline(ss, line, _X('\n'))){
                 const pal::string_t prefix = _X("The framework '");
                 const pal::string_t suffix = _X("' was not found.");
+				const pal::string_t custom_prefix = _X("  _ ");
                 const pal::string_t url_prefix = _X("  - ") DOTNET_CORE_APPLAUNCH_URL _X("?");
                 if (starts_with(line, prefix, true) && ends_with(line, suffix, true))
                 {
                     dialogMsg.append(line);
                     dialogMsg.append(_X("\n\n"));
                 }
+				else if (starts_with(line, custom_prefix, true))
+				{
+					dialogMsg.erase();
+					dialogMsg.append(line.substr(custom_prefix.length()));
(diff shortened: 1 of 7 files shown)


src/corehost/cli/apphost/apphost.windows.cpp
```diff
@@ -76,12 +76,19 @@ namespace
             while (std::getline(ss, line, _X('\n'))){
                 const pal::string_t prefix = _X("The framework '");
                 const pal::string_t suffix = _X("' was not found.");
+				const pal::string_t custom_prefix = _X("  _ ");
```
vitek-karas: It seems you used Tabs for indentation - please make it consistent with the rest of the file (spaces).

vitek-karas: Seems to be the case for all .cpp/.h file changes in this PR.

vitek-karas (approved):
Other than fixing the whitespaces looks good.

MeiChin-Tsai:
approved. thx.

PR #8846, opened 2019-11-13
Add automatic SDL validation

dagood:
For https://github.com/dotnet/core-setup/issues/8669.

Validation build: https://dev.azure.com/dnceng/internal/_build/results?buildId=426399

Issue #8669, opened 2019-10-25
Add SDL tool automation tooling to official build


dagood:
The goal is to get `policheck` and `credscan` integrated for 3.1.

CoreFX impl: https://github.com/dotnet/corefx/pull/40294 (small fix https://github.com/dotnet/corefx/pull/40294).

dagood:
I (basically) copied what CoreFX has and ran a build, and the SDL step failed due to disk space while trying to pull down artifacts: https://dev.azure.com/dnceng/internal/_build/results?buildId=402553&view=results. Looking at how to change the agent pool it runs on.

I'm also looking at what values are needed for a few of the parameters to the tool. (Notification alias, TSA admin alias.) I haven't gotten to a point in the SDL runs where they're needed yet.

Number of unknowns seems fairly bounded, but could be peeling an onion after getting the disk space issue fixed.

dagood:
Filed https://github.com/dotnet/arcade/issues/4216 for the disk space issue. It doesn't appear to be configurable, so I'm trying builds with the `eng/common` job hacked up to see what fix we need.

dagood:
This is blocked on:

* An unexpected answer to https://github.com/dotnet/arcade/issues/4216
* OR both https://github.com/dotnet/core-eng/issues/8180 and https://github.com/dotnet/arcade/issues/3700 being resolved.

PR #8990, opened 2020-01-24
[release/3.1] Add automatic SDL validation

dagood:
#### Description

Ports SDL validation infrastructure from https://github.com/dotnet/core-setup/pull/8846 to `release/3.1`.

The goal is to get `policheck` and `credscan` integrated for 3.1 to reduce the need for manual runs. (There's still value: https://github.com/dotnet/arcade/issues/4216#issuecomment-553664380.)

For https://github.com/dotnet/core-setup/issues/8669.

I ran a mock build to validate, which passed: https://dev.azure.com/dnceng/internal/_build/results?buildId=496043&view=results

#### Customer Impact

None.

#### Regression?

No.

#### Risk

Very low. The mock build would have caught any reasonable problems. If there are issues once merged, the build will break and we will need to revert this change, but nothing worse than that.

---

/cc @vivmishra @mmitche

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,6 +39,8 @@ variables:
   - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
     - name: SignType
       value: $[ coalesce(variables.OfficialSignType, 'real') ]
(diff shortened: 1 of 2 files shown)


MichaelSimons (approved):

dagood:
The mock build succeeded. (Updated description.)

PR #9012, opened 2020-02-28
[release/3.1] HostModel: Retry ResourceUpdate on Win32 error

swaroop-sridhar:
## Issue
https://github.com/dotnet/runtime/issues/3832

## Customer Scenario
Building a WinExe with resources fails non-deterministically

Several customers have observed failure during resource update when the HostModel updates the AppHost (to transfer resources from the managed app).
The failure is not detereminisitc, not reproducible on our machines, and depends on specific computers/software running.
This indicates interference by other software while the HostWriter is updating the AppHost.

## Problem
The current implementation retries the resource update if an update because the device or drive is locked (say by an antivurus) HRESULT 0x21 and 0x6C. However, the failures reported have errors 0x5 (Access violation) and 0x6# (Open failed). Windows/Defender team said that file-locking with these error-codes is not expected.
However, different AVs work differently about examining files.

## Solution
We believe that the correct fix for this issue is to complete:

* To implement #3828 and #3829
* Ship AppHost with an extension/permissions not indicating an executable.

However the above is a fairly large change for servicing .net core 3.1.

So, this change implements a simpler fix intended for servicing 3.1 branch: Always retry the resource-update on Win32 error. While this may cause unnecessary retries on legitimate failures (ex: file missing) such cases are rare, because the SDK supplies the apphost, and the HostModel itself creates the file to update.

## Risk
Low

## Master Branch PR
https://github.com/dotnet/runtime/pull/32347

### DIFF
--- a/src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
+++ b/src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
@@ -41,16 +41,43 @@ namespace Microsoft.NET.HostModel
 
         public static void RetryOnWin32Error(Action func)
         {
-            bool IsWin32FileLockError(int hresult)
+            bool IsKnownIrrecoverableError(int hresult)
             {
                 // Error codes are defined in winerror.h
-                const int ErrorLockViolation = 33;
-                const int ErrorDriveLocked = 108;
-
                 // The error code is stored in the lowest 16 bits of the HResult
-                int errorCode = hresult & 0xffff;
 
-                return errorCode == ErrorLockViolation || errorCode == ErrorDriveLocked;
+                switch (hresult & 0xffff)
+                {
+                    case 0x00000001: // ERROR_INVALID_FUNCTION
+                    case 0x00000002: // ERROR_FILE_NOT_FOUND
+                    case 0x00000003: // ERROR_PATH_NOT_FOUND
+                    case 0x00000006: // ERROR_INVALID_HANDLE
+                    case 0x00000008: // ERROR_NOT_ENOUGH_MEMORY
+                    case 0x0000000B: // ERROR_BAD_FORMAT
+                    case 0x0000000E: // ERROR_OUTOFMEMORY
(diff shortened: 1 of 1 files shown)


src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
```diff
@@ -41,16 +41,43 @@ public static void RetryOnIOError(Action func)
...
-                return errorCode == ErrorLockViolation || errorCode == ErrorDriveLocked;
+                switch (hresult & 0xffff)
+                {
+                    case 0x00000001: // ERROR_INVALID_FUNCTION
+                    case 0x00000002: // ERROR_FILE_NOT_FOUND
+                    case 0x00000003: // ERROR_PATH_NOT_FOUND
+                    case 0x00000006: // ERROR_INVALID_HANDLE
+                    case 0x00000008: // ERROR_NOT_ENOUGH_MEMORY
+                    case 0x0000000B: // ERROR_BAD_FORMAT
+                    case 0x0000000E: // ERROR_OUTOFMEMORY
```
lpereira, line 57: Love the fact that there's an error code for both "not enough memory" and "out of memory"  :)

lpereira (approved):
I'm not familiar enough with Win32 to properly review this, but patch seems correct.

jeffschwMSFT (approved):

jeffschwMSFT:
Approved in tactics

avifatal:
Hi,
How do I update my dockers to avoid this problem?
Thanks

src/managed/Microsoft.NET.HostModel/AppHost/RetryUtil.cs
danmoseley: Possibly analogous to our distinction between InsufficientMemoryException and OutOfMemoryException.

swaroop-sridhar:
@avifatal the fix will be in the next servicing release of the .net core 3.1 SDK.
Once this is available, you can use the new SDK to avoid the problem.

PR #9049, opened 2020-07-16
Allow skipping package build during servicing

NikolaMilosavljevic:
Fixes https://github.com/dotnet/runtime/issues/37661

## Summary

This infra adds the ability to skip package build during servicing, for any package that should not be built and published to NuGet.org.

## Customer Impact

No. This is internal, build infra.

## Regression?

No.

## Testing

Repo build to ensure unchanged packages are not built.

## Risk

None.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -8,4 +8,31 @@
   <Import Project="packaging.stubs.targets" Condition="'$(MSBuildProjectExtension)' != '.pkgproj'" />
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., Directory.Build.targets))\Directory.Build.targets" />
+
+  <!--
+    Package build can be skipped by invoking PackIfServiced instead of Pack targets.
+
+    ProjectServicingConfiguration is used to determine if packaging should be skipped. See more in eng/Versions.props.
+  -->
+  <Target Name="GetCurrentProjectServicingConfiguration">
+    <ItemGroup>
+      <CurrentProjectServicingConfiguration
+        Include="@(ProjectServicingConfiguration)"
+        Condition="'%(Identity)' == '$(MSBuildProjectName)'" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="PackIfServiced"
+        DependsOnTargets="
+          SkipPackIfShouldNotBuild;
+          Pack" />
+
+  <Target Name="SkipPackIfShouldNotBuild"
+        DependsOnTargets="GetCurrentProjectServicingConfiguration">
+    <PropertyGroup>
+      <MostRecentProducedServicingPatchVersion>%(CurrentProjectServicingConfiguration.PatchVersion)</MostRecentProducedServicingPatchVersion>
+      <IsPackable Condition="'$(MostRecentProducedServicingPatchVersion)' != '$(PatchVersion)'">false</IsPackable>
+    </PropertyGroup>
+  </Target>
+
 </Project>
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -14,6 +14,11 @@
 
     ProjectServicingConfiguration is used to determine if packaging should be skipped. See more in eng/Versions.props.
   -->
+  <Target Name="PackIfServiced"
+          DependsOnTargets="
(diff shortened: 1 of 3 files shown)


src/Directory.Build.targets
```diff
@@ -8,4 +8,31 @@
...
+  <Target Name="PackIfServiced"
+        DependsOnTargets="
+          SkipPackIfShouldNotBuild;
+          Pack" />
+
+  <Target Name="SkipPackIfShouldNotBuild"
+        DependsOnTargets="GetCurrentProjectServicingConfiguration">
+    <PropertyGroup>
+      <MostRecentProducedServicingPatchVersion>%(CurrentProjectServicingConfiguration.PatchVersion)</MostRecentProducedServicingPatchVersion>
+      <IsPackable Condition="'$(MostRecentProducedServicingPatchVersion)' != '$(PatchVersion)'">false</IsPackable>
```
dagood, line 35: Hmmm, I think we might actually need to skip this for source-build, to avoid pulling in prebuilts. I *think* these binaries are actually included in the SDK (not just used as references to compile against) so we need to build nupkgs every time to propagate them.

I think this would be sufficient for us to tweak later if necessary (or not use if not necessary) while avoiding a patch:

```xml
<Target Name="SkipPackIfShouldNotBuild"
        Condition="'$(AlwaysPackEvenIfNotServicing)' != 'true'"
        DependsOnTargets="GetCurrentProjectServicingConfiguration">
```

It would be nice to figure out what our end goal is and make this more unified, but it's not worth it at the moment, just need to make sure it's configurable enough. I have an overview of the complex issue source-build is hitting, here, for the future: https://github.com/dotnet/source-build/tree/release/3.1/Documentation/planning/nongranular-servicing-readiness

src/Directory.Build.targets
```diff
@@ -8,4 +8,31 @@
...
+      <CurrentProjectServicingConfiguration
+        Include="@(ProjectServicingConfiguration)"
+        Condition="'%(Identity)' == '$(MSBuildProjectName)'" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="PackIfServiced"
+        DependsOnTargets="
+          SkipPackIfShouldNotBuild;
+          Pack" />
```
dagood: Misaligned `DependsOnTargets`

src/Directory.Build.targets
```diff
@@ -8,4 +8,31 @@
...
+
+  <Target Name="SkipPackIfShouldNotBuild"
+        DependsOnTargets="GetCurrentProjectServicingConfiguration">
```
dagood: Misaligned `DependsOnTargets`

src/Directory.Build.targets
```diff
@@ -8,4 +8,31 @@
   <Import Project="packaging.stubs.targets" Condition="'$(MSBuildProjectExtension)' != '.pkgproj'" />
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., Directory.Build.targets))\Directory.Build.targets" />
+
+  <!--
+    Package build can be skipped by invoking PackIfServiced instead of Pack targets.
```
dagood, line 13: Was this comment line meant to be on this target, or `PackIfServiced`? Maybe `PackIfServiced` should be defined first since it's a better introduction to what we're trying to accomplish with this set of targets?

NikolaMilosavljevic: Thanks - I'll reorder targets.

src/Directory.Build.targets
NikolaMilosavljevic: Will fix.

src/Directory.Build.targets
NikolaMilosavljevic: WIll fix.

src/Directory.Build.targets
NikolaMilosavljevic: Thanks - I'll add the condition.

jamshedd:
Approved for 3.1.7.

Anipik:
@dagood can you take another go at this one ?

src/Directory.Build.targets
dagood: Looks good. Filed https://github.com/dotnet/source-build/issues/1668 to track the source-build side of this.

dagood (approved):

PR #9055, opened 2020-08-03
Adds TLS to generating nuget packages

silvioguiso:


mmitche (approved):

dagood:
Do you think setting up the default error action could also prevent silent failures?

https://github.com/dotnet/core-setup/blob/965af060b4c893b50efbc7e4959fbb8b52879318/eng/common/tools.ps1#L53-L54

PR #9066, opened 2020-08-16
[release/3.1] Add dependencies for CPD strict in aspnetcore

dougbu:


eng/Version.Details.xml
```diff
@@ -106,6 +106,14 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>7ee84596d92e178bce54c986df31ccc52479e772 </Sha>
     </Dependency>
+    <Dependency Name="System.Net.Http.WinHttpHandler" Version="4.7.0">
```
dougbu: dotnet/aspnetcore is behind on this version. Should we move to `4.7.2` (released in 3.1.5)❔

wtgodbe: Yes

wtgodbe (changes requested):
We should also add System.IO.Pipelines & System.Runtime.CompilerServices.Unsafe, then unpin those in AspNetCore. They're only pinned _because_ of this strict CPD issue: https://github.com/dotnet/aspnetcore/pull/20695

dougbu:
> We should also add System.IO.Pipelines & System.Runtime.CompilerServices.Unsafe

Done. PR should now cover the core-setup portion of dotnet/aspnetcore#24937 and dotnet/aspnetcore#24944 as well as provide what dotnet/aspnetcore needs for strict coherency.

wtgodbe (approved):

PR #9113, opened 2021-01-13
Backport ARM64 IJW host fix to 3.1

jkoritzinsky:
Backport of dotnet/runtime#46933 to release/3.1

Today, activating the runtime through the IJW host on ARM64 crashes due to an invalid address and invalid instructions. This PR fixes the ARM64 bootstrapping stub so we can don't crash when the first call into the runtime is through a native export of a managed function via C++/CLI.

## Customer Impact

This fix allows users to call into managed code from native code in C++/CLI. This is one of the fundamental C++/CLI scenarios, which is not currently officially supported on ARM64 on .NET Core 3.1, but has been talked about for the future. Given that the most common C++/CLI scenario is the opposite, I would guess this would be at most 1 in 4 C++/CLI scenarios overall, but I don't have usage data to back that number up. The issue was reported internally by the C++ team.

## Testing

This fix has been verified offline with the people that reported the issue (C++ team).

## Risk

Minimal risk. Without this fix, this path crashes hard with an "unaligned address" failure. With the fix, the runtime successfully starts and the managed code is successfully called.

### DIFF
--- a/src/corehost/cli/ijwhost/arm64/bootstrap_thunk.cpp
+++ b/src/corehost/cli/ijwhost/arm64/bootstrap_thunk.cpp
@@ -10,21 +10,19 @@
 // Get thunk from the address that the thunk code provided
 bootstrap_thunk *bootstrap_thunk::get_thunk_from_cookie(std::uintptr_t cookie)
 {
-    
+
     // Cookie is generated via the first thunk instruction:
-    //  mov r12, pc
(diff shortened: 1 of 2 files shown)


jeffschwMSFT (approved):
Approved.  Please get a cr and I will take it for consideration this week.

AaronRobinsonMSFT (approved):

PR #9125, opened 2021-02-20
Port from 5.0: Fix relro, now, and PIE for installer

ivdiazsa:
# Port Description

This is a direct port from PR 685 in the runtime repo. This one was added during .NET Core 5.0 development but there is a customer need to also have it in the .NET Core 3.1 release. The other part is in the corefx repo, PR 43036.

This change adds the missing `-z,relro` and `-z,now` compile options, as well as the position independent related code and `-pie` linker option to the Unix installers.

# Customer Impact
Having these flags enabled provides the built binaries with an additional layer of security, which has become a necessary requirement for some compliance checks, as well as safer applications overall. This port was requested by teams in Azure.

# Regression
This was not a regression.

# Testing
The [checksec](https://github.com/slimm609/checksec.sh) tool was used to verify the executables and shared objects had been indeed built with `PIE` and `Full RELRO` enabled.

# Risk
The risk of this is pretty low, since it's been well tested in the current `main` branch, as well as .NET 5.0 releases.


src/settings.cmake
```diff
@@ -216,14 +223,12 @@ endif()
 # containing the reference instead of using definitions from other modules.
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Xlinker -Bsymbolic -Bsymbolic-functions")
-    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id=sha1")
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--build-id=sha1")
```
janvorli, line 220: The CMAKE_EXE_LINKER_FLAGS need to stay here and below, otherwise the host binaries would miss the build id sha1 in them. 
Also, the `-Wl,-z,relro,-z,now` should be added to this one too.

src/settings.cmake
```diff
@@ -4,6 +4,13 @@
 
 set (CMAKE_CXX_STANDARD 11)
 
+if(NOT MSVC)
```
janvorli: This doesn't look right as it would always print the warning on Unix (and PIE is Unix only concept). This whole 'if' block should go away.

janvorli (commented):
We also need to add the `-pie` linker option to the `CMAKE_EXE_LINKER_FLAGS`. IIRC, the `CMAKE_POSITION_INDEPENDENT_CODE` doesn't do that for some reason.

janvorli (approved):
LGTM, thank you!

aik-jahoda:
Is this PR targeting current release? If yes we need approve and Servicing-approved tag.

jeffschwMSFT (approved):
Approved.  I will take for consideration in 3.1.x

cc @blowdart and @GrabYourPitchforks

PR #9153, opened 2021-05-13
Add deps package for CBL Mariner

NikolaMilosavljevic:
Adding deps package for CBL Mariner.


azure-pipelines.yml
```diff
@@ -111,6 +111,7 @@ stages:
         imageRid: rhel.7
         rids:
         - centos.7
+        - cm1
```
jkoritzinsky: I think for the rid format, this should be cm.1

NikolaMilosavljevic: I tried to match existing CBL Mariner naming convention, for existing packages, i.e.: https://packages.microsoft.com/cbl-mariner/1.0/prod/update/x86_64/rpms/

@richlander any preference?

jkoritzinsky: Also cc: @ericstj and @safern

NikolaMilosavljevic: @richlander @ericstj @safern - any thoughts on the RID naming? We are trying to match CBL Mariner package naming. I don't believe these RIDs are used anywhere except package generation infra.

richlander: For the RID format, the `cm.1` suggestion looks good.

NikolaMilosavljevic: > For the RID format, the `cm.1` suggestion looks good.

That looks better to me, too. It would be slightly different than current CBL-Mariner package naming (`cm1`), but it's a good compromise that maintains both distro identity and .NET RID naming conventions.

NikolaMilosavljevic: Updated this PR with proposed change.

richlander (approved):

PR #9192, opened 2021-12-07
Update branding to 3.1.23

vseanreesermsft:


safern (approved):

wtgodbe:
@safern I think we need to update off of Ubuntu1604 here (probably also the case in corefx & coreclr)

safern:
> @safern I think we need to update off of Ubuntu1604 here (probably also the case in corefx & coreclr)

Yup. I can take care of that, will push to the branding prs.

PR #9195, opened 2021-12-17
[release/3.1] Fixes __rid_plat generation on alpine

ayakael:
./src/corehost/build.sh generates wrong plat_rid on Alpine. Generated RID is expected to be alpine.x.xx-xx, while the computed RID is linux-musl-xxx. This patches it by matching what's expected by rest of build system.

Made as part of Alpine Linux dotnet31 / dotnet5 packaging project, see dotnet/source-build#2695


carlossanlop:
Closing since 3.1 is going out of support in December.

PR #9196, opened 2022-01-13
Set min OS version on macos to 10.13

VSadov:
Fixes: https://github.com/dotnet/runtime/issues/61173

## Description

We did not set any particular min OS version in the build so it was using the SDK default and could change when toolset was updated and indeed changed from original 10.13 to 10.15 and then to 11.0

This change puts back 10.13 as min OS version on macos.

## Testing

I have verified locally via `otool -l `that both `apphost` and `dotnet` executables now have:

```
Load command 9
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.13
      sdk 11.0
```

## Risk

Low, compiler-flag-only change

### DIFF
--- a/eng/jobs/osx-build.yml
+++ b/eng/jobs/osx-build.yml
@@ -4,19 +4,16 @@ parameters:
 jobs:
 - job: ${{ parameters.name }}
   pool:
-    ${{ if eq(variables['System.TeamProject'], 'public') }}:
-      name: Hosted macOS
-    ${{ if ne(variables['System.TeamProject'], 'public') }}:
-      vmImage: macOS-latest
+      vmImage: 'macOS-latest'
   strategy:
-    matrix: 
+    matrix:
       debug:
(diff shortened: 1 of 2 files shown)


VSadov:
CC: @agocke

agocke (approved):
LGTM, we'll have to take this through servicing. I'll schedule it.

agocke:
I checked the support policy -- apparently we only support 10.15: https://github.com/dotnet/core/blob/main/release-notes/3.1/3.1-supported-os.md#macos

I think we should set that as the minimum.

VSadov:
OSX builds are failing with 
```
##[warning]An image label with the label MacOS does not exist.
,##[error]The remote provider was unable to process the request.
```

agocke:
Sorry, I thought about this more and I think you were right -- 10.13 is what we had before and is less risky. I'm going to bring this back to the earlier commit.

PR #9198, opened 2022-01-20
Add System.Text.Encodings.Web package reference

tommcdon:

PR #9201, opened 2022-03-24
Build DependencyModel in May

wtgodbe:
This project was updated to fix a Component Governance bug in January: https://github.com/dotnet/core-setup/pull/9198. However, it hasn't been serviced since then, which means downstream repos that depend on it (like dotnet/efcore) still have active CompGov bugs. Servicing it in May should fix that.

CC @dcwhittaker @Pilchie - the SLA for this is April 1, but the alert won't be fixed by this until mid-April - is that alright? Can we extend it? I could also add an explicit `PackageReference` to System.Text.Encodings.Web to the relevant project in EFCore, but again, we wouldn't be able to check that in until after April 1, so I think this is a better long-term fix.


dougbu (commented):
If this causes the updated package to flow, 👍 . But I'm not familiar with `@(ProjectServicingConfiguration)` and defer to other reviewers.

/btw the failures in CI looked transient. I reran the two failed jobs.

NikolaMilosavljevic (approved):

PR #9212, opened 2022-06-09
[release/3.1] Do not add x86 host to PATH on 64-bit machines

NikolaMilosavljevic:
Backport of https://github.com/dotnet/runtime/pull/69902 to dotnet/core-setup release/3.1

Fixes: https://github.com/dotnet/runtime/issues/70039

## Customer Impact

This fixes the remaining issue with non-native host paths in machine's environment (PATH).

The fix uses a different model than x64-on-arm64. It is a simpler change as it's easy to determine if machine is 64-bit, using WersionNT64 property. This solution avoids unnecessary changes in arcade infra and adding additional complexity.

## Testing

See main issue: https://github.com/dotnet/runtime/issues/70039#issuecomment-1151412442

## Risk

As with any breaking change, this will affect some users. In this case, anyone that wants 32-bit host (dotnet.exe) to be in PATH on 64-bit machines, will need to add host location to PATH manually.

### DIFF
--- a/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
+++ b/src/pkg/projects/Microsoft.NETCore.DotNetHost/host.wxs
@@ -42,6 +42,9 @@
         <?if $(var.Platform)~=x64 ?>
         <!-- For x64 installer, only add to PATH when actually on native architecture -->
         <Condition>NOT NON_NATIVE_ARCHITECTURE</Condition>
+        <?elseif $(var.Platform)~=x86 ?>
+        <!-- For x86 installer, only add to PATH when not on 64-bit platform -->
+        <Condition>NOT VersionNT64</Condition>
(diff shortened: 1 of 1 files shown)


carlossanlop:
Needs the `servicing-consider` label and a sign-off from area owners. Please also send an email to Tactics with a copy of the template so it is discussed for servicing.

NikolaMilosavljevic:
> Needs the `servicing-consider` label and a sign-off from area owners. Please also send an email to Tactics with a copy of the template so it is discussed for servicing.

Thanks. Yeah, I know. I was waiting on some testing to be completed before adding reviewers, tags and sending to Tactics.

NikolaMilosavljevic:
Approved offline - @joeloff @ericstj any concerns with this backport - can you approve it?

joeloff (approved):

